PR #1045, opened 2022-11-18
Add setValue and getValue for ValueVector

ray6080:
- Removed `values` in `ValueVector`, instead: (1) expose templated get/setValue interface with a given data type; (2) expose getData() for cases we need access to raw pointers, e.g., need to reference a value inside the vector; memcpy values into/out of the ValueVector, etc.

- Removed `using namespace std` from `types.h`.
- Removed the exception test for "MATCH", which crashes on MacOS due to error signals.

### DIFF
--- a/src/common/include/vector/value_vector.h
+++ b/src/common/include/vector/value_vector.h
@@ -23,14 +23,10 @@ public:
 
     ~ValueVector() = default;
 
-    inline void setState(shared_ptr<DataChunkState> state_) { state = move(state_); }
-    void addString(uint64_t pos, string value) const;
-    void addString(uint64_t pos, char* value, uint64_t len) const;
+    inline void setState(shared_ptr<DataChunkState> state_) { state = std::move(state_); }
 
     inline void setAllNull() { nullMask->setAllNull(); }
-
     inline void setAllNonNull() { nullMask->setAllNonNull(); }
-
     inline void setMayContainNulls() { nullMask->setMayContainNulls(); }
     // Note that if this function returns true, there are no null. However, if it returns false, it
     // doesn't mean there are nulls, i.e., there may or may not be nulls.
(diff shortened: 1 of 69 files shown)


src/common/include/vector/value_vector.h
```diff
@@ -41,30 +37,46 @@ class ValueVector {
...
 
-    inline node_offset_t readNodeOffset(uint64_t pos) const {
+    template<typename T>
+    T getValue(uint32_t pos) const {
+        return ((T*)valueBuffer.get())[pos];
+    }
+
+    // Note: this function doesn't handle overflow for STRING and LIST.
+    template<typename T>
+    void setValue(uint32_t pos, T val) {
```
andyfengHKU, line 61: I would special template for ku_list and ku_string 

```
template<>
void setValue(uint32_t pos, ku_string val) {
     addString(pos, val)
}

```

test/processor/physical_plan/operator/orderBy/order_by_key_encoder_test.cpp
```diff
@@ -52,9 +52,8 @@ class OrderByKeyEncoderTest : public Test {
         for (auto i = 0u; i < numOfOrderByCols; i++) {
             shared_ptr<ValueVector> valueVector =
                 make_shared<ValueVector>(INT64, memoryManager.get());
-            auto intValuePtr = (int64_t*)valueVector->values;
             for (auto j = 0u; j < numOfElementsPerCol; j++) {
-                intValuePtr[j] = 5;
+                valueVector->setValue(j, (int64_t)5);
```
andyfengHKU, line 56: No need to change. Just curious
is setValue<int64_t>(j, 5) not working in this case?

andyfengHKU (approved):

src/common/include/vector/value_vector.h
ray6080: I think you mean string here. I add a specialized template for it.
For ku_list and ku_string, we expect their overflows are already properly populated inside the vector, and if not, they actually should go through ValueVectorUtils, instead of this interface.

test/processor/physical_plan/operator/orderBy/order_by_key_encoder_test.cpp
ray6080: It won't. So this is the part that the change can introduce unsafe code with potential bugs.
Let's discuss this in person to go over the options here.

PR #1046, opened 2022-11-18
Fix arrow linking issue for M1-based mac

mewim:
In this PR, I fix the Apache Arrow linking issue on M1-based Mac for @weipang142857 and @printfCalvin. The arrow build I used in this PR is Homebrew-arrow at 10.0.0 (https://formulae.brew.sh/formula/apache-arrow).

**THIS PR AIMS TO ADDRESS THE IMMEDIATE REQUIREMENT FOR DEVELOPMENT, BUT THIS CONFIGURATION SHOULD NEVER BE MERGED INTO MASTER AS IT IS PLATFORM-SPECIFIC!!! WE NEED TO REVISIT IT AFTER TRANSITIONING TO CMAKE.**

The PR has been tested on my M1 MacBook Air and I can confirm that the build process completes successfully. It should also work for the same MacBook Air from the university. For linking on Linux, it should follow a similar principle, but the directory and search path may need to change. 

Note that I did NOT test the actual code in copy_csv_arrow_node.cpp, only compilation process has been tested.


weipang142857 (approved):
Tested, and it works on macOS.

Issue #968, opened 2022-11-11
Parser exception for multi-list extract/slicing syntax


aziz-mu:
Currently, slicing/extracting a list of list and then extracting or slicing the results is permitted, if done like so:
`MATCH (a:person) RETURN (a:courseScoresPerTerm[1])[1]`
however, the following syntax gives a parser error:
`MATCH (a:person) RETURN a:courseScoresPerTerm[1][1]`

Currently, openCypher supports this syntax so we should too.

PR #1054, opened 2022-11-21
fixed multi extract/slice

aziz-mu:
Fixed the issue outlined in #968 as well as added some tests

### DIFF
--- a/src/antlr4/Cypher.g4
+++ b/src/antlr4/Cypher.g4
@@ -334,8 +334,13 @@ oC_StringListNullOperatorExpression
     : oC_PropertyOrLabelsExpression ( oC_StringOperatorExpression | oC_ListOperatorExpression | oC_NullOperatorExpression )? ;
 
 oC_ListOperatorExpression
-    : ( SP ? '[' oC_Expression ']' )
-        | ( SP ? '[' oC_Expression? ':' oC_Expression? ']' );
+    : ( kU_ListExtractOperatorExpression | kU_ListSliceOperatorExpression ) oC_ListOperatorExpression ? ;
+
+kU_ListExtractOperatorExpression
+    : SP ? '[' oC_Expression ']' ;
+
+kU_ListSliceOperatorExpression
+    : SP ? '[' oC_Expression? ':' oC_Expression? ']' ;
 
 oC_StringOperatorExpression
     :  ( ( SP STARTS SP WITH ) | ( SP ENDS SP WITH ) | ( SP CONTAINS ) ) SP? oC_PropertyOrLabelsExpression ;
--- a/src/parser/transformer.cpp
+++ b/src/parser/transformer.cpp
@@ -498,36 +498,62 @@ unique_ptr<ParsedExpression> Transformer::transformListOperatorExpression(
     CypherParser::OC_ListOperatorExpressionContext& ctx,
     unique_ptr<ParsedExpression> propertyExpression) {
     auto rawExpression = propertyExpression->getRawName() + " " + ctx.getText();
-
-    if (ctx.children[1]->getText() == ":" || ctx.children[2]->getText() == ":") {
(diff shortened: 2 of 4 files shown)


acquamarin (approved):

src/antlr4/Cypher.g4
```diff
@@ -334,8 +334,8 @@ oC_StringListNullOperatorExpression
     : oC_PropertyOrLabelsExpression ( oC_StringOperatorExpression | oC_ListOperatorExpression | oC_NullOperatorExpression )? ;
 
 oC_ListOperatorExpression
-    : ( SP ? '[' oC_Expression ']' )
-        | ( SP ? '[' oC_Expression? ':' oC_Expression? ']' );
+    : ( ( SP ? '[' oC_Expression ']' )
```
andyfengHKU, line 337: this rule is getting complicates so I would try to rewrite as
```
oC_ListOperatorExpression
    : ( kU_ListExtractOperatorExpression | kU_ListSliceOperatorExpression ) oC_ListOperatorExpression ? ;

kU_ListExtractOperatorExpression
   :  SP ? '[' oC_Expression ']' ;

kU_ListSliceOperatorExpression
   :  SP ? '[' oC_Expression? ':' oC_Expression? ']' ;

```

Let me know if this makes sense or not

andyfengHKU (changes requested):

src/antlr4/Cypher.g4
aziz-mu: Done in [organized parser rules](https://github.com/kuzudb/kuzu/pull/1054/commits/9f3ba9ed3e3f304b4357a7353a8d26287a094dba)

src/parser/transformer.cpp
```diff
@@ -498,34 +498,48 @@ unique_ptr<ParsedExpression> Transformer::transformListOperatorExpression(
     CypherParser::OC_ListOperatorExpressionContext& ctx,
     unique_ptr<ParsedExpression> propertyExpression) {
     auto rawExpression = propertyExpression->getRawName() + " " + ctx.getText();
-
-    if (ctx.children[1]->getText() == ":" || ctx.children[2]->getText() == ":") {
+    if (ctx.kU_ListSliceOperatorExpression()) {
```
andyfengHKU, line 501: I'll approve this PR but you can make the following change before merging. The general principal of transformer is to follow grammar. Since you have split grammar into different rules, this part need to change accordingly.

```
unique_ptr<ParsedExpression> listOperator;
if (ctx.SliceOperator)
     listOperator = transformSlice ...
else if
     listOperator = transformExtract...
if ctx.oC_ListOperatorExpression 
        return transformListOperatorExpression(*ctx.oC_ListOperatorExpression(), move(listOperator));
return listOperator;
```

src/parser/transformer.cpp
```diff
@@ -498,34 +498,48 @@ unique_ptr<ParsedExpression> Transformer::transformListOperatorExpression(
...
         auto listSlice =
             make_unique<ParsedFunctionExpression>(LIST_SLICE_FUNC_NAME, move(rawExpression));
         listSlice->addChild(move(propertyExpression));
-        if (ctx.children[1]->getText() == ":") {
+        if (ctx.kU_ListSliceOperatorExpression()->children[1]->getText() == ":") {
```
andyfengHKU, line 505: I understand this code path but let's leave a comment here about parsing `[ : ]` `[ left : ]` `[ : right ]`. In future we might change grammar to make this more readable.

andyfengHKU (approved):

src/parser/transformer.cpp
aziz-mu: Done

src/parser/transformer.cpp
aziz-mu: Done

PR #1057, opened 2022-11-22
Multi labeled node scan

andyfengHKU:


src/processor/operator/scan_node_id.cpp
```diff
@@ -46,10 +28,42 @@ pair<uint64_t, uint64_t> ScanNodeIDSharedState::getNextRangeToRead() {
...
+        tableState->initialize(transaction);
+    }
+    initialized = true;
+}
+
+tuple<ScanTableNodeIDSharedState*, node_offset_t, node_offset_t>
+ScanNodeIDSharedState::getNextRangeToRead() {
+    unique_lock lck{mtx};
+    if (currentStateIdx == tableStates.size()) {
+        return make_tuple(nullptr, -1, -1);
```
ray6080, line 46: Change `-1` to `UINT64_MAX`.

src/processor/operator/scan_node_id.cpp
```diff
@@ -46,10 +28,42 @@ pair<uint64_t, uint64_t> ScanNodeIDSharedState::getNextRangeToRead() {
...
+    while (startOffset >= endOffset) {
+        currentStateIdx++;
+        if (currentStateIdx == tableStates.size()) {
+            return make_tuple(nullptr, -1, -1);
+        }
+        auto [_startOffset, _endOffset] = tableStates[currentStateIdx]->getNextRangeToRead();
+        startOffset = _startOffset;
+        endOffset = _endOffset;
+    }
+    auto tableState =
```
ray6080, line 58: I don't think you should check for `nullptr `here. Instead, you are not supposed to hit `currentStateIdx >= tableStates.size()`. I would put an assert here.

ray6080 (approved):

PR #1077, opened 2022-11-29
Add init global state interface

andyfengHKU:
This PR differentiate the initialization between global and local state. Global state should be only initialized once while local state is initialized for each thread.

Remove following code pattern in global state
```
if (!initialized)
    initialize
initialized = true
```

Other change
We used to rely on vectors to get input data type. This design is bad since it requires first generate vector (i.e. initialize local state) before accessing data type. This PR also changes this by giving data type during construction.

### DIFF
--- a/src/processor/processor_task.cpp
+++ b/src/processor/processor_task.cpp
@@ -7,14 +7,14 @@ void ProcessorTask::run() {
     // We need the lock when cloning because multiple threads can be accessing to clone,
     // which is not thread safe
     lock_t lck{mtx};
-    unique_ptr<PhysicalOperator> lastOp = sinkOp->clone();
+    auto clonedPipeline = sink->clone();
     lck.unlock();
-    auto& sink = (Sink&)*lastOp;
-    sink.execute(executionContext);
+    auto clonedSink = (Sink*)clonedPipeline.get();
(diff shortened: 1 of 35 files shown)


src/processor/processor_task.cpp
```diff
@@ -3,18 +3,21 @@
...
+void Pipeline::execute(ExecutionContext* context) {
+    sink->execute(context);
+}
+
 void ProcessorTask::run() {
     // We need the lock when cloning because multiple threads can be accessing to clone,
     // which is not thread safe
     lock_t lck{mtx};
-    unique_ptr<PhysicalOperator> lastOp = sinkOp->clone();
+    auto pipeline = make_unique<Pipeline>(sink);
```
ray6080, line 14: Actually I think constructing a pipeline here is confusing. I understand it is just a wrapper for now. but I would suggest not introduce the pipeline class in this pr.
Eventually, we should do the following refactoring, inside processor, a plan is decomposed as pipelines (instead of processor tasks. each pipeline is a sub-tree of the physical plan, which can contain sink, source, and vector of operators, as suggested by you), which have dependency between themselves, and processor tasks take in pipeline and schedule the execution.

ray6080 (approved):

PR #1087, opened 2022-12-01
Unlabeled rel

andyfengHKU:
Initial attempt to support unlabeled rel


src/processor/operator/generic_extend.cpp
```diff
@@ -0,0 +1,127 @@
...
+        listHandles.push_back(
+            make_shared<ListHandle>(*resultSet->getListSyncState(outVectorPos.dataChunkPos)));
+    }
+    nextListIdx = lists.size();
+    nextColumnIdx = columns.size();
+    currentList = nullptr;
+    return resultSet;
+}
+
+bool GenericExtend::getNextTuplesInternal() {
```
ray6080, line 23: One thing I would try is as follows:
a) for implementation of `findMoreOutput()`, merge `findInCurrentList` and `findInRemainingLists` into one function `scanFromList`, which will handle both cases.
b) move the block
```
    if (findMoreOutput()) {
         return true;
     }
```
into the while loop. So each time you check if you can read with current states (currentNodeOffset, currentList, nextList/ColumnIdx), if not, call getNextTuple, then reset states, and go back to the beginning to read with current states again.
3. thus, you don't need findOutput() inside the while condition.
I think, if possible, this will simplify the whole control logic.

ray6080 (approved):

ray6080 (commented):
The general comment is, as we discussed, in the next PR, we can try the idea of only having a single generic scan operator for all rel nodeIDs and properties. This can simplify sync states, and if it works fine, we shall propagate this to all scans (with filter pushdown) later.

src/include/processor/operator/generic_extend.h
```diff
@@ -0,0 +1,61 @@
...
+private:
+    DataPos inVectorPos;
+    DataPos outVectorPos;
+    vector<Column*> columns;
+    vector<Lists*> lists;
+    vector<shared_ptr<ListHandle>> listHandles;
+
+    shared_ptr<ValueVector> inVector;
+    shared_ptr<ValueVector> outVector;
+    uint32_t nextColumnIdx;
```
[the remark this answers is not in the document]
andyfengHKU: Decide not to change in this PR

src/processor/operator/generic_extend.cpp
andyfengHKU: Done

PR #1096, opened 2022-12-06
Unlabeled rel property

andyfengHKU:


src/processor/operator/generic_extend.cpp
```diff
@@ -5,34 +5,47 @@ namespace processor {
...
+    nextListIdx = adjColumnAndListCollection.lists.size();
+    nextColumnIdx = adjColumnAndListCollection.columns.size();
+    currentAdjList = nullptr;
+    currentPropertyLists.resize(outPropertyVectors.size());
+    currentPropertyListHandles.resize(outPropertyVectors.size());
 }
 
 bool GenericExtend::getNextTuplesInternal() {
     while (true) {
         if (findOutput()) {
```
ray6080, line 34: I should've commented this in the last PR, but I didn't add them.
I would replace `find` with `scan`. I don't think find is accurate here, because you are actually performing reading. Suggestions are: `findOutput` -> `scanForOutput`, `findInColumns` -> `scanFromColumns`, `findInLists` -> `scanFromLists`.
But up to you to make final decisions on these names.

ray6080 (approved):

PR #1101, opened 2022-12-07
Shortest Path Logical Operator and Query Planning

anuchak:


src/include/processor/operator/shortest_path.h
```diff
@@ -0,0 +1,62 @@
...
+#include "processor/operator/source_operator.h"
+#include "storage/storage_structure/column.h"
+#include "storage/storage_structure/lists/lists.h"
+
+using namespace kuzu::evaluator;
+
+namespace kuzu {
+namespace processor {
+
+class ShortestPathAdjCol : public BaseShortestPath {
```
ray6080, line 17: Separate these two classes to two files.

src/include/processor/operator/physical_operator.h
```diff
@@ -61,7 +62,7 @@ const string PhysicalOperatorTypeNames[] = {"AGGREGATE", "AGGREGATE_SCAN", "COLU
     "MULTIPLICITY_REDUCER", "PROJECTION", "SCAN_REL_PROPERTY", "RESULT_COLLECTOR", "SCAN_NODE_ID",
     "SCAN_COLUMN_PROPERTY", "SEMI_MASKER", "SET_STRUCTURED_NODE_PROPERTY", "SKIP", "ORDER_BY",
     "ORDER_BY_MERGE", "ORDER_BY_SCAN", "UNION_ALL_SCAN", "UNWIND", "VAR_LENGTH_ADJ_LIST_EXTEND",
-    "VAR_LENGTH_COLUMN_EXTEND"};
+    "VAR_LENGTH_COLUMN_EXTEND", "SHORTEST_PATH"};
```
ray6080, line 65: Let's be consistent with (var length) extend operators. Separate SHORTEST_PATH to COLUMN_SHORTEST_PATH, and ASJ_LIST_SHORTEST_PATH.

src/include/processor/operator/base_shortest_path.h
```diff
@@ -0,0 +1,52 @@
...
+
+#include "queue"
+
+#include "processor/operator/physical_operator.h"
+#include "storage/storage_structure/lists/lists.h"
+
+namespace kuzu {
+namespace processor {
+
+struct BFSState {
```
ray6080, line 11: Set initialized values for these variables.

src/processor/operator/shortest_path.cpp
```diff
@@ -0,0 +1,100 @@
...
+using namespace std;
+
+namespace kuzu {
+namespace processor {
+
+bool ShortestPathAdjList::getNextTuplesInternal() {
+    while (true) {
+        while (!bfsQueue.empty()) {
+            auto queueFrontNodeOffset = bfsQueue.front();
+            auto bfsFrontNodeState = bfsVisitedNodesMap[queueFrontNodeOffset];
```
ray6080, line 14: Is this a safe operation? 
1. You didn't initialize variables in BaseShortestPath, which is not a good habit. Because there can be unexpected initialized values on some platforms.
2. Why not first check if bfsVisitedNodesMap contains queueFrontNodeOffset, before you get the value (bfsFrontNodeState) from the map?

src/processor/operator/shortest_path.cpp
```diff
@@ -0,0 +1,100 @@
...
+            if (bfsFrontNodeState->bfsLevel >= lowerBound &&
+                bfsFrontNodeState->bfsLevel <= upperBound && bfsFrontNodeState->hasBeenOutput) {
+                for (auto i = destValueVector->state->getPositionOfCurrIdx();
```
ray6080, line 17: I don't understand the for loop here. Why are you looping over dest value vector?

src/processor/operator/shortest_path.cpp
```diff
@@ -0,0 +1,100 @@
+#include "processor/operator/shortest_path.h"
+
+#include "iostream"
+
+using namespace std;
+
+namespace kuzu {
+namespace processor {
+
+bool ShortestPathAdjList::getNextTuplesInternal() {
```
ray6080, line 10: This function is not well organized. Can you separate the logic of reading a pair of src and dest node ids, and the logic of computing shortest path between them?
So, you will have a computeShortestPath function, which takes in source and dest node id, and return a path (you can represent the path in any data type you want for now, e.g., string; vector of int64 with mixed node and rel ids; vector of steps, each contains a source node id, rel id, and dest node id)

Then, getNextTuplesInternal will be responsible for reading src and dest node ids from `srcValueVector` and `destValueVector`, and them call computeShortestPath, and finally print the result for it.

src/include/processor/operator/shortest_path.h
```diff
@@ -0,0 +1,64 @@
...
+        } else {
+            return make_unique<ShortestPath>(srcDataPos, destDataPos, lists, lowerBound, upperBound,
+                children[0]->clone(), id, paramsString);
+        }
+    }
+
+private:
+    DataPos srcDataPos;
+    DataPos destDataPos;
+    vector<Column*> columns;
```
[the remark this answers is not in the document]
anuchak: Yeah not required, had removed this.

src/include/processor/operator/shortest_path.h
```diff
@@ -0,0 +1,64 @@
...
+    vector<AdjLists*> lists;
```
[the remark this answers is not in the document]
anuchak: Separated

src/include/processor/operator/shortest_path.h
anuchak: Done

src/include/processor/operator/physical_operator.h
anuchak: Done

src/include/processor/operator/base_shortest_path.h
anuchak: added constructor to this struct for initializing the variables

src/processor/operator/shortest_path.cpp
anuchak: 1) I think the queue and map container always are initialized as empty, but anyway have added the initialization as part of the constructor now.

2) The map does not have any elements removed actually. And the only place it has elements added is the `addBfsLevel` function where the queue is also pushed with the element. So there is no case of the map missing any element. I've also added the map, queue proper initialization.

src/processor/operator/shortest_path.cpp
anuchak: Yeah I had done this for keeping it generic like if there are multiple destination nodes for same source.
But we kept `Flatten` so only 1 source and 1 destination should be present at any point.

I've removed the for loop and added code for only 1 destination node now.

src/processor/operator/shortest_path.cpp
anuchak: I've changed the format of the logic, currently it is like this:
<pre>
while(true):
  computeShortestPath;
  if(child->getNextTuples())
    getSrcNode and getDestNode;
</pre>

src/processor/mapper/plan_mapper.cpp
```diff
@@ -124,12 +127,43 @@ unique_ptr<PhysicalOperator> PlanMapper::mapLogicalOperatorToPhysical(
...
+    LogicalOperator* logicalOperator, MapperContext& mapperContext) {
+    auto* logicalShortestPath = (LogicalShortestPath*)logicalOperator;
+    auto prevOperator = mapLogicalOperatorToPhysical(logicalOperator->getChild(0), mapperContext);
+    auto sourceNode = logicalShortestPath->getSrcNodeExpression();
+    auto destNode = logicalShortestPath->getDestNodeExpression();
+    auto rel = logicalShortestPath->getRelExpression();
+    // assume this here for now
+    auto direction = REL_DIRECTIONS[0];
+    auto srcDataPos = mapperContext.getDataPos(sourceNode->getInternalIDPropertyName());
+    auto destDataPos = mapperContext.getDataPos(destNode->getInternalIDPropertyName());
```
ray6080, line 149: During the mapping phase, you would need to get data pos for relID, too. And pass it to the shortest path operator.

src/include/processor/operator/base_shortest_path.h
```diff
@@ -0,0 +1,58 @@
...
+          bfsVisitedNodesMap{map<uint64_t, shared_ptr<BFSState>>()} {}
+
+    virtual bool getNextTuplesInternal() = 0;
+
+    shared_ptr<ResultSet> init(ExecutionContext* context);
+
+    virtual unique_ptr<PhysicalOperator> clone() = 0;
+
+protected:
+    DataPos srcDataPos;
```
ray6080, line 45: I don't see the relIDVector? Don't you need also to read out rel IDs?

src/include/processor/operator/base_shortest_path.h
```diff
@@ -0,0 +1,58 @@
...
+    DataPos destDataPos;
+    uint64_t lowerBound;
+    uint64_t upperBound;
+    shared_ptr<ValueVector> srcValueVector;
+    shared_ptr<ValueVector> destValueVector;
+    map<uint64_t, shared_ptr<BFSState>> bfsVisitedNodesMap;
+    queue<uint64_t> bfsQueue = queue<uint64_t>();
+    BaseColumnOrList* storage;
+    MemoryManager* memoryManager;
```
ray6080, line 54: memoryManager is not needed.

src/include/processor/operator/base_shortest_path.h
```diff
@@ -0,0 +1,58 @@
...
+protected:
+    DataPos srcDataPos;
+    DataPos destDataPos;
+    uint64_t lowerBound;
+    uint64_t upperBound;
+    shared_ptr<ValueVector> srcValueVector;
+    shared_ptr<ValueVector> destValueVector;
+    map<uint64_t, shared_ptr<BFSState>> bfsVisitedNodesMap;
+    queue<uint64_t> bfsQueue = queue<uint64_t>();
+    BaseColumnOrList* storage;
```
ray6080, line 53: I think it's better to not keep BaseColumnOrList* in base class, instead, just keep lists or column separately in two child classes.

src/include/processor/operator/base_shortest_path.h
```diff
@@ -0,0 +1,58 @@
...
+          upperBound{upperBound}, bfsQueue{queue<uint64_t>()},
+          bfsVisitedNodesMap{map<uint64_t, shared_ptr<BFSState>>()} {}
+
+    virtual bool getNextTuplesInternal() = 0;
+
+    shared_ptr<ResultSet> init(ExecutionContext* context);
+
+    virtual unique_ptr<PhysicalOperator> clone() = 0;
+
+protected:
```
ray6080, line 44:
```
    DataPos srcDataPos;
    DataPos destDataPos;
    DataPos relIDDataPos;
    uint64_t lowerBound;   
    uint64_t upperBound; 
    uint64_t currentLayer;
    shared_ptr<ValueVector> srcValueVector;
    shared_ptr<ValueVector> destValueVector;
    shared_ptr<ValueVector> relIDVector;
    unique_ptr<Frontier> currentFrontier;
    unique_ptr<Frontier> nextFrontier;
    unordered_map<node_offset_t, unique_ptr<NodeState>> visitedNodes;
```

src/include/processor/operator/base_shortest_path.h
```diff
@@ -0,0 +1,58 @@
...
+
+#include "queue"
+
+#include "processor/operator/physical_operator.h"
+#include "storage/storage_structure/lists/lists.h"
+
+namespace kuzu {
+namespace processor {
+
+struct BFSState {
```
ray6080, line 11: I don't think you should keep children, listSyncState, listHandle, etc. in the state. For each node, its state can simply be as follows:
```
struct NodeState {
    uint64_t relID = UINT64_MAX;
    node_offset_t predecessor = UINT64_MAX; // should change to nodeID_t eventually.
}

```

src/processor/operator/shortest_path_adj_list.cpp
```diff
@@ -0,0 +1,92 @@
...
+                return false;
+            }
+            currSrcIdx = srcValueVector->state->getPositionOfCurrIdx();
+            currDestIdx = destValueVector->state->getPositionOfCurrIdx();
+        } while (srcValueVector->isNull(currSrcIdx) || destValueVector->isNull(currDestIdx) ||
+                 !addBFSLevel(srcValueVector->readNodeOffset(currSrcIdx), 1, UINT64_MAX));
+    }
+}
+
+void ShortestPathAdjList::computeShortestPath() {
```
ray6080, line 26: Let's simplify this function a bit by introducing more functions to decompose the computation logic.
```
// return false if no path can be found (within the lower and upper bound). otherwise return true.
bool ShortestPathAdjList::getShortestPath(nodeID_t srcNode, nodeID_t destNode) {
    // initialize the current frontier for layer 0.
    currentFrontier.addNode(UINT64_MAX, srcNode);
    // extend from the current frontier to the next one.
    while (currentLayer < upperBound) {
        destNode = currentLayer < lowerBound ? UINT64_MAX : destNode; // set destNodeID to invalid value if the layer is still less than the lower bound, so no matches with dest node can be found.
        if (extendToNextFrontier(destNode)) { return true; }
        // move vector in nextFrontier to currentFontier, and reset nextFrontier.
        resetFrontier();
        currentLayer++;
    }
    return false;
}
```

See the shared design doc for complete information.

src/include/processor/operator/base_shortest_path.h
```diff
@@ -0,0 +1,58 @@
...
+
+protected:
+    DataPos srcDataPos;
+    DataPos destDataPos;
+    uint64_t lowerBound;
+    uint64_t upperBound;
+    shared_ptr<ValueVector> srcValueVector;
+    shared_ptr<ValueVector> destValueVector;
+    map<uint64_t, shared_ptr<BFSState>> bfsVisitedNodesMap;
+    queue<uint64_t> bfsQueue = queue<uint64_t>();
```
ray6080, line 52: Instead of a queue, we can manually maintain two frontiers, one for the current layer, and one for the next layer.
```
    unique_ptr<Frontier> currentFrontier;
    unique_ptr<Frontier> nextFrontier;
```

src/include/processor/operator/shortest_path_adj_list.h
```diff
@@ -0,0 +1,44 @@
...
+    bool getNextBatchOfChildNodes(BFSState* bfsState);
+
+    PhysicalOperatorType getOperatorType() override {
+        return PhysicalOperatorType::SHORTEST_PATH_ADJ_LIST;
+    }
+
+    inline unique_ptr<PhysicalOperator> clone() override {
+        return make_unique<ShortestPathAdjList>(srcDataPos, destDataPos, storage, lowerBound,
+            upperBound, children[0]->clone(), id, paramsString);
+    }
```
ray6080, line 40: You would need `Lists* lists` here, and possibly `    shared_ptr<ListHandle> listHandle;
`

ray6080 (changes requested):
I created a doc for detailed designs, let's coordinate through that.
https://docs.google.com/document/d/1d-C9mVZ0NvscLB1ZiLTzwOyKvd5TRNJ5mzjKKxnZyCQ/edit?usp=sharing

ray6080 (changes requested):
See my new batch of comments, and let me know if you want to discuss something.

src/processor/mapper/plan_mapper.cpp
```diff
@@ -124,12 +127,47 @@ unique_ptr<PhysicalOperator> PlanMapper::mapLogicalOperatorToPhysical(
...
+        Column* column =
+            relStore.getAdjColumn(direction, sourceNode->getTableID(), rel->getTableID());
+        return make_unique<ShortestPathAdjCol>(srcDataPos, destDataPos, column,
+            rel->getLowerBound(), rel->getUpperBound(), move(prevOperator), getOperatorID(),
+            logicalShortestPath->getExpressionsForPrinting());
+    } else {
+        assert(relStore.hasAdjList(direction, sourceNode->getTableID(), rel->getTableID()));
+        auto relProperties =
+            (PropertyExpression*)logicalShortestPath->getRelPropertyExpression().get();
+        auto list = relStore.getRelPropertyLists(direction, sourceNode->getTableID(),
```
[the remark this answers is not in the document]
anuchak: done

src/processor/operator/shortest_path_adj_list.cpp
```diff
@@ -0,0 +1,129 @@
...
+}
+
+bool ShortestPathAdjList::computeShortestPath(uint64_t currIdx, uint64_t destIdx) {
+    bfsVisitedNodesMap = map<uint64_t, shared_ptr<NodeState>>();
+    currFrontier = make_shared<vector<node_offset_t>>();
+    nextFrontier = make_shared<vector<node_offset_t>>();
+    listHandle->reset();
+
+    uint64_t currLevel = 1, destNodeOffset;
+    if (!visitNextFrontierNode(srcValueVector->readNodeOffset(currIdx), UINT64_MAX, INT64_MAX)) {
```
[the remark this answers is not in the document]
anuchak: I've kept starting val of current level to 1, and changed loop condition to curr level < upper bound
This should ensure then that level is tracking number of edges between source and destination

src/include/processor/operator/base_shortest_path.h
```diff
@@ -0,0 +1,58 @@
...
+#include "processor/operator/physical_operator.h"
+#include "storage/storage_structure/lists/lists.h"
+
+namespace kuzu {
+namespace processor {
+
+struct NodeState {
+    uint64_t parentNodeID;
+    uint64_t relParentID;
+    shared_ptr<ValueVector> children;
```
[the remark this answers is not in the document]
anuchak: done

src/processor/operator/shortest_path_adj_list.cpp
```diff
@@ -0,0 +1,129 @@
...
+    }
+}
+
+bool ShortestPathAdjList::computeShortestPath(uint64_t currIdx, uint64_t destIdx) {
+    bfsVisitedNodesMap = map<uint64_t, shared_ptr<NodeState>>();
+    currFrontier = make_shared<vector<node_offset_t>>();
+    nextFrontier = make_shared<vector<node_offset_t>>();
+    listHandle->reset();
+
+    uint64_t currLevel = 1, destNodeOffset;
```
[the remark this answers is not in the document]
anuchak: I've changed the logic

src/processor/operator/shortest_path_adj_list.cpp
```diff
@@ -0,0 +1,129 @@
...
+            printShortestPath(destNodeOffset);
+            return true;
+        }
+        currLevel++;
+        resetFrontier();
+    }
+    return false;
+}
+
+bool ShortestPathAdjList::extendToNextFrontier(node_offset_t destNodeOffset) {
```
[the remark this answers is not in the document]
anuchak: done

src/processor/operator/shortest_path_adj_list.cpp
```diff
@@ -0,0 +1,129 @@
...
+                if (addToNextFrontier(nodeState, j, currFrontier->operator[](i), destNodeOffset)) {
+                    return true;
+                }
+            }
+        } while (getNextBatchOfChildNodes(nodeState));
+    }
+    return false;
+}
+
+bool ShortestPathAdjList::addToNextFrontier(NodeState*& nodeState, uint64_t index,
```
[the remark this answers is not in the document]
anuchak: done

src/processor/operator/shortest_path_adj_list.cpp
```diff
@@ -0,0 +1,129 @@
...
+            nodeState->relIDVector->getValue<int64_t>(index))) {
+        nextFrontier->push_back(nextFrontierNodeOffset);
+        if (nextFrontierNodeOffset == destNodeOffset) {
+            return true;
+        }
+    }
+    return false;
+}
+
+bool ShortestPathAdjList::visitNextFrontierNode(
```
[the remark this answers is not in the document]
anuchak: done

src/processor/operator/shortest_path_adj_list.cpp
```diff
@@ -0,0 +1,129 @@
...
+        relPropertyLists->readValues(relVector, *relListHandles);
+        auto newBFSState = make_shared<NodeState>(predecessor, parentRelID, children, relVector);
+        bfsVisitedNodesMap.emplace(nodeOffset, newBFSState);
+        return true;
+    } else {
+        return false;
+    }
+}
+
+bool ShortestPathAdjList::getNextBatchOfChildNodes(NodeState* bfsState) {
```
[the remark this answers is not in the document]
anuchak: done

src/include/processor/operator/base_shortest_path.h
```diff
@@ -0,0 +1,58 @@
...
+    virtual unique_ptr<PhysicalOperator> clone() = 0;
+
+protected:
+    DataPos srcDataPos;
+    DataPos destDataPos;
+    uint64_t lowerBound;
+    uint64_t upperBound;
+    shared_ptr<ValueVector> srcValueVector;
+    shared_ptr<ValueVector> destValueVector;
+    map<uint64_t, shared_ptr<NodeState>> bfsVisitedNodesMap;
```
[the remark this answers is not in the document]
anuchak: changed

src/include/processor/operator/shortest_path_adj_list.h
```diff
@@ -0,0 +1,63 @@
...
+        return PhysicalOperatorType::SHORTEST_PATH_ADJ_LIST;
+    }
+
+    inline unique_ptr<PhysicalOperator> clone() override {
+        return make_unique<ShortestPathAdjList>(srcDataPos, destDataPos, lists, relPropertyLists,
+            lowerBound, upperBound, children[0]->clone(), id, paramsString);
+    }
+
+private:
+    AdjLists* lists;
```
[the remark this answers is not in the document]
anuchak: added vector in main class

src/include/processor/operator/shortest_path_adj_list.h
```diff
@@ -0,0 +1,63 @@
...
+namespace processor {
+
+class ShortestPathAdjList : public BaseShortestPath {
+public:
+    ShortestPathAdjList(const DataPos& srcDataPos, const DataPos& destDataPos, AdjLists* adjList,
+        Lists* relPropertyLists, uint64_t lowerBound, uint64_t upperBound,
+        unique_ptr<PhysicalOperator> child, uint32_t id, const string& paramsString)
+        : BaseShortestPath{srcDataPos, destDataPos, lowerBound, upperBound, move(child), id,
+              paramsString},
+          lists{adjList}, relPropertyLists{move(relPropertyLists)} {}
```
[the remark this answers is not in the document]
anuchak: done

ray6080 (approved):

PR #1104, opened 2022-12-08
Multi labeled graph pattern

andyfengHKU:


src/include/processor/operator/generic_extend.h
```diff
@@ -35,44 +89,32 @@ class GenericExtend : public PhysicalOperator {
...
-    AdjLists* currentAdjList;
-    ListHandle* currentAdjListHandle;
-    vector<Lists*> currentPropertyLists;
-    vector<ListHandle*> currentPropertyListHandles;
-    node_offset_t currentNodeOffset;
+    // storage structures
+    unordered_map<table_id_t, unique_ptr<AdjAndPropertyCollection>>
+        adjAndPropertyCollectionPerNodeTable;
+    AdjAndPropertyCollection* currentAdjAndPropertyCollection = nullptr;
+    unique_ptr<ListSyncState> listSyncState = nullptr;
```
ray6080, line 117: You don't really need to keep a current list sync state here. I think it's more clear to move ListSyncState into AdjAndPropertyCollection, so each will have their own one.

ray6080 (approved):

PR #1108, opened 2022-12-09
Extend property reading

andyfengHKU:
This PR abstracts BaseExtend class to make sure all of our extends (i.e. ColumnExtend, ListExtend, GenericExtend) have the same behaviour. 

We read all rel properties together with extend to avoid appending listSyncState to resultSet. Also this simplifies our design of GenericExtend. The tricky part is that it's hard to elegantly sync list state between GenericExtend and GenericScanRelProperty. So we choose this simpler design until we have a better solution to read rel properties.

A hidden bug related to scan rel property after deletion can be found at adjList_extend.cpp


ray6080 (approved):
The general comment is we should rename our scan operators to be more accurate.
Suggestions:
```
AdjColumnExtend -> ScanRelTableColumns
AdjListExtend -> ScanRelTableLists
GenericExtend -> ScanGenericRelTable
ScanSingleTableProperties -> ScanSingleNodeTable
ScanMultiTableProperties -> ScanMultiNodeTable
```

PR #1114, opened 2022-12-12
Unlabeled query

andyfengHKU:


src/include/binder/expression/node_rel_expression.h
```diff
@@ -0,0 +1,35 @@
...
+        auto tableIDsMap = unordered_set<table_id_t>(tableIDs.begin(), tableIDs.end());
+        for (auto tableID : tableIDsToAdd) {
+            if (!tableIDsMap.contains(tableID)) {
+                tableIDs.push_back(tableID);
+            }
+        }
+    }
+    inline bool isMultiLabeled() const { return tableIDs.size() > 1; }
+    inline vector<table_id_t> getTableIDs() const { return tableIDs; }
+    // We don't support multi-labeled update
```
andyfengHKU, line 24: Remove this comment. This is incorrect, we will have this API eventually because we have fast-path code for single labeled query

src/include/binder/binder.h
```diff
@@ -112,14 +113,13 @@ class Binder {
     shared_ptr<NodeExpression> bindQueryNode(const NodePattern& nodePattern, QueryGraph& queryGraph,
         PropertyKeyValCollection& collection);
     shared_ptr<NodeExpression> createQueryNode(const NodePattern& nodePattern);
-    inline unordered_set<table_id_t> bindNodeTableIDs(const vector<string>& tableNames) {
+    inline vector<table_id_t> bindNodeTableIDs(const vector<string>& tableNames) {
         return bindTableIDs(tableNames, NODE);
     }
-    inline unordered_set<table_id_t> bindRelTableIDs(const vector<string>& tableNames) {
+    inline vector<table_id_t> bindRelTableIDs(const vector<string>& tableNames) {
```
ray6080, line 119: Why not just return `unordered_set`? same for `bindNodeTableIDs`.

src/include/binder/expression/node_rel_expression.h
```diff
@@ -0,0 +1,35 @@
...
+    inline table_id_t getSingleTableID() const {
+        assert(tableIDs.size() == 1);
+        return tableIDs[0];
+    }
+
+protected:
+    vector<table_id_t> tableIDs;
```
ray6080, line 31: Ditto, why not just use unordered_set for tableIDs?

ray6080 (approved):

src/include/binder/binder.h
andyfengHKU: We want to ensure the order of table_id to be the same across different platforms.

PR #1119, opened 2022-12-14
move schema to operator

andyfengHKU:
This is our initial step to split planner and optimizer

### DIFF
--- a/src/planner/operator/logical_hash_join.cpp
+++ b/src/planner/operator/logical_hash_join.cpp
@@ -1,0 +1,32 @@
+#include "planner/logical_plan/logical_operator/logical_hash_join.h"
+
+#include "planner/logical_plan/logical_operator/sink_util.h"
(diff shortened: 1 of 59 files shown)


src/planner/operator/logical_hash_join.cpp
```diff
@@ -0,0 +1,32 @@
...
+#include "planner/logical_plan/logical_operator/sink_util.h"
+
+namespace kuzu {
+namespace planner {
+
+void LogicalHashJoin::computeSchema() {
+    auto probeSchema = children[0]->getSchema();
+    auto buildSchema = children[1]->getSchema();
+    schema = probeSchema->copy();
+    if (mark == nullptr) {
```
ray6080, line 12: You can use joinType here, instead of checking null for mark.

ray6080 (approved):

PR #1123, opened 2022-12-18
Numeric ops

aziz-mu:
This PR Implements some of the functions from issue #942 . Specifically, it implements for integers:
 - bitwise and (&)
 - bitwise OR (|)
 - bit shift left (<<)
 - bit shift right (>>)
New tests are included as well. I'll do factorial in another PR as it requires additional parser support


test/test_files/tinysnb/function/arithmetic.test
```diff
@@ -332,3 +332,42 @@
...
+120
+180
+80
+80
+100
+160
+332
+
+-NAME BitshiftRightFunctionTest
+-QUERY MATCH (o:organisation) RETURN o.orgCode >> 3
```
andyfengHKU, line 368: I would also add tests to check if the numerical operator precedence follows the [cpp standard](https://en.cppreference.com/w/c/language/operator_precedence) (I don't think SQL has a standard on this).

e.g. a + 1 & 1 should be interpreted as (a + 1) & 1 since addition has a higher precedence

andyfengHKU (approved):

andyfengHKU (commented):
I have an additional comment on factorial. If factorial can be supported by simply modifying

```
oC_UnaryAddOrSubtractExpression 
    : ( MINUS SP? )? oC_StringListNullOperatorExpression ;
```
to 

```
oC_UnaryAddSubtractOrFactorialExpression
    : ( MINUS | FACTORIAL SP? )? oC_StringListNullOperatorExpression ;
```
then let's do it. Otherwise we don't need this syntax sugar for factorial given there is already a factorial function exists in the code base.

My concern is on the antlr4 side. Antlr4's performance will degrade as we add more and more rules.

test/test_files/tinysnb/function/arithmetic.test
aziz-mu: Done in latest commit

aziz-mu:
Turns out adding factorial syntax was a lot easier than I thought, so I added it as part of this PR too.

src/antlr4/Cypher.g4
```diff
@@ -329,13 +340,15 @@ oC_MultiplyDivideModuloExpression
 kU_MultiplyDivideModuloOperator : '*' | '/' | '%' ;
 
 oC_PowerOfExpression
-    : oC_UnaryAddOrSubtractExpression ( SP? '^' SP? oC_UnaryAddOrSubtractExpression )* ;
+    : oC_UnaryAddSubtractOrFactorialExpression ( SP? '^' SP? oC_UnaryAddSubtractOrFactorialExpression )* ;
 
-oC_UnaryAddOrSubtractExpression
-    : ( MINUS SP? )? oC_StringListNullOperatorExpression ;
+oC_UnaryAddSubtractOrFactorialExpression
+    : ( MINUS SP? )? oC_StringListNullOperatorExpression SP? FACTORIAL? ;
```
andyfengHKU, line 346: SP? FACTORIAL? -> ( SP? FACTORIAL )?

andyfengHKU (approved):

src/parser/transformer.cpp
```diff
@@ -470,12 +526,16 @@ unique_ptr<ParsedExpression> Transformer::transformPowerOfExpression(
...
 
-unique_ptr<ParsedExpression> Transformer::transformUnaryAddOrSubtractExpression(
-    CypherParser::OC_UnaryAddOrSubtractExpressionContext& ctx) {
+unique_ptr<ParsedExpression> Transformer::transformUnaryAddSubtractOrFactorialExpression(
+    CypherParser::OC_UnaryAddSubtractOrFactorialExpressionContext& ctx) {
     if (ctx.MINUS()) {
         return make_unique<ParsedFunctionExpression>(NEGATE_FUNC_NAME,
             transformStringListNullOperatorExpression(*ctx.oC_StringListNullOperatorExpression()),
             ctx.getText());
+    } else if (ctx.FACTORIAL()) {
```
andyfengHKU, line 535: The input might be `-2!` which contains both MINUS() and FACTORIAL().

Either define their order here (i.e. first create factorial function if necessary then create minus function is necessary) and leave a comment. Or u can define their order in antlr4.

I might do the ordering here because I'm a bit concerned with antlr4 performance as number of grammar rules increasing. But you can make the call here,

aziz-mu: Done

PR #1124, opened 2022-12-19
Wrap pybind11 API and and Fix #1106

### LINKED ISSUE #1106: Closing ipython session without closing conn causes BufferManagerException
#### How to reproduce:

1. open an iPython session. 
2. create a (kuzu) connection to a dataset 
3. close ipython (exit()) without closing the kuzu connection

#### Outcome:
"libc++abi: terminating with uncaught exception of type kuzu::common::BufferManagerException: Buffer manager exception: Deleting buffer that is still pinned. pinCount: 1 pageIdx: 12" -> the python process is killed

#### Environment:
python (3.7) running on a virtualenv on macosx

#### What I'd expect instead:
when ipython sends the kill signal to the C++ process, the buffer is unpinned

mewim:
This PR adds the following:
- A Python wrapper for the pybind11 Python API. This creates a layer around the APIs exposed by pybind11, which gives us the flexibility to override and extend Python APIs with native Python script instead of doing everything in C++. (Note that if an override is not created, the C++ endpoints are still exposed and can be called by the user, so we do not have to wrap everything in Python if not necessary). 
- Fix #1106. The fix is as follows:
  - In `query_result.py`, the `__del__` method is added, which will close the query result automatically.
  - In `query_result.py`, a reference to `connection` is added. This ensures that when the garbage collection is invoked for the `connection`, the corresponding query results have to be deleted first.
  - Similarly, a reference to `database` is added to the connection. Overall, the above measurements ensures the garbage collection order of query result -> connection -> database to prevent the memory leak caused by database being closed first.
  - To ensure #1106 is properly fixed, an automated test case is added (`test_query_result_close.py`). Note that the issue is not reproducible in `pytest`. Therefore, this script invokes a new Python process. It is verified that without the above measurements, this test script does fail.
 - Additionally, this PR overrides the APIs in `query_result.py` to use `sneak_case` in accordance with the Python standard.


tools/python_api/CMakeLists.txt
```diff
@@ -24,3 +28,7 @@ target_include_directories(
         PUBLIC
         ../../src/include
 )
+
+get_target_property(PYTHON_DEST _kuzu LIBRARY_OUTPUT_DIRECTORY)
+
+file(COPY ${SOURCE_PY} DESTINATION ${PYTHON_DEST})
```
andyfengHKU, line 34: new line

tools/python_api/src_py/connection.py
```diff
@@ -0,0 +1,14 @@
+from .query_result import QueryResult
+from . import _kuzu
+
+
+class connection:
```
andyfengHKU, line 5: should this be Connection?

tools/python_api/src_py/query_result.py
```diff
@@ -0,0 +1,29 @@
...
+        self._query_result.close()
+
+    def get_as_df(self):
+        return self._query_result.getAsDF()
+
+    def get_column_data_types(self):
+        return self._query_result.getColumnDataTypes()
+
+    def get_column_names(self):
+        return self._query_result.getColumnNames()
```
andyfengHKU, line 29: new line

andyfengHKU (approved):

.gitignore
```diff
@@ -6,6 +6,7 @@ bazel-*
 
 test_temp/
 build/
+cmake-build-debug/
```
ray6080, line 9: I don't think this is necessary. To avoid generation of this directory, you can configure build directory of CMake profile to `build` in CLion.

tools/python_api/CMakeLists.txt
```diff
@@ -3,17 +3,21 @@ project(_kuzu)
 
 set(CMAKE_CXX_STANDARD 20)
 
+file(GLOB SOURCE_PY
```
ray6080, line 6: I would rather not use GLOB, but list all necessary files here.

ray6080 (approved):

tools/python_api/CMakeLists.txt
mewim: I have changed the C++ target to use the explicit path instead of using GLOB. But I think we can keep the GLOB for Python files since CMake is not actually running any build script against the Python files. It is just copying the files to the build directory.

PR #1127, opened 2022-12-20
Source sink op interface

andyfengHKU:
This PR adds isSource and isSink interface to physical operator.

There two interfaces are used to
- calculate time metrics
- decompose processor task

### DIFF
--- a/src/include/processor/data_pos.h
+++ b/src/include/processor/data_pos.h
@@ -1,6 +1,7 @@
 #pragma once
 
 #include <cstdint>
+#include <utility>
 
 namespace kuzu {
 namespace processor {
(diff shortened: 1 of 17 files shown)


acquamarin (approved):

src/include/processor/data_pos.h
```diff
@@ -1,6 +1,7 @@
 #pragma once
 
 #include <cstdint>
+#include <utility>
```
ray6080, line 4: Why do we need this include?

ray6080 (approved):

src/include/processor/data_pos.h
andyfengHKU: std::pair

PR #1145, opened 2023-01-03
Throw exception if the database path directory is wrong

anuchak:
#1033 added a try catch block to handle incorrect database path directory

### DIFF
--- a/tools/python_api/test/test_exception.py
+++ b/tools/python_api/test/test_exception.py
@@ -1,4 +1,7 @@
 import pytest
+import sys
+sys.path.append('../build/')
+import kuzu
 
 
 def test_exception(establish_connection):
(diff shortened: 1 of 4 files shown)


andyfengHKU (changes requested):
Add tests on c++ and Python API

anuchak:
added python & cpp api tests

tools/python_api/test/test_exception.py
```diff
@@ -1,4 +1,7 @@
 import pytest
+import sys
+sys.path.append('../build/')
```
andyfengHKU, line 3: double check if this is needed.

andyfengHKU (approved):

tools/python_api/test/test_exception.py
anuchak: yes seems to be required, the tests failed after I removed it: https://github.com/kuzudb/kuzu/actions/runs/3838919604/jobs/6535990703

PR #1152, opened 2023-01-05
Refactor property variable expression

andyfengHKU:
This PR flips the dependency between property expression and node/rel expression.

Previously, node/rel expression is a child of property expression. In this PR, we ask node/rel expression to hold all properties associated with its table. This is needed because we will interpret `return a` as `returning all properties of a and aggregated in a map` which complies with openCypher semantic. So a node/rel expression need to have access to all of its properties.

As a side effect, this should also solve issue #1083


src/binder/bind/bind_graph_pattern.cpp
```diff
@@ -60,6 +60,49 @@ static void validateNodeRelConnectivity(const Catalog& catalog_, const RelExpres
                           " are not connected through rel " + rel.getRawName() + ".");
 }
 
+static vector<std::pair<std::string, vector<Property>>> getPropertyNameAndSchemasPairs(
```
ray6080, line 63: `getPropertyNameAndSchemasPairs` -> `getPropertyNameAndPropertiesPairs`

src/include/planner/join_order_enumerator_context.h
```diff
@@ -35,8 +35,8 @@ class JoinOrderEnumeratorContext {
     inline QueryGraph* getQueryGraph() { return queryGraph; }
 
     inline bool nodeNeedScanTwice(NodeExpression* node) {
```
ray6080, line 37: Change the function name `nodeNeedScanTwice` accordingly.

ray6080 (approved):

PR #1160, opened 2023-01-08
Node map agg

andyfengHKU:
This PR allows returning node/rel as all of their properties aggregated in json format in a single column. To do so, this PR

- Add BoundStatementResult which defines the schema of how to aggregate multiple columns into a single column in json format
- Move ownership of iterate tuple from iterator to query result. So iterator only writes to a vector of flat values.

### Other changes

- Move isReadOnly, isDDL, isCopyCSV check from planning stage to binding stage
- Alter preparedStatement to take a vector of logical plans and thus remove APIs exposed only for testing framework

### DIFF
--- a/src/include/binder/binder.h
+++ b/src/include/binder/binder.h
@@ -88,13 +88,12 @@ private:
 
     /*** bind projection clause ***/
     unique_ptr<BoundWithClause> bindWithClause(const WithClause& withClause);
-    unique_ptr<BoundReturnClause> bindReturnClause(
-        const ReturnClause& returnClause, unique_ptr<BoundSingleQuery>& boundSingleQuery);
+    unique_ptr<BoundReturnClause> bindReturnClause(const ReturnClause& returnClause);
 
     expression_vector bindProjectionExpressions(
         const vector<unique_ptr<ParsedExpression>>& projectionExpressions, bool containsStar);
-    // For RETURN clause, we write variable "v" as all properties of "v"
-    expression_vector rewriteProjectionExpressions(const expression_vector& expressions);
+    // Rewrite variable "v" as all properties of "v"
+    expression_vector rewriteNodeOrRelExpression(const Expression& expression);
 
     void bindOrderBySkipLimitIfNecessary(
         BoundProjectionBody& boundProjectionBody, const ProjectionBody& projectionBody);
--- a/src/include/binder/bound_statement_result.h
+++ b/src/include/binder/bound_statement_result.h
@@ -1,0 +1,57 @@
+#pragma once
+
+#include "binder/expression/expression.h"
+
+namespace kuzu {
+namespace binder {
+
+class BoundStatementResult {
+public:
+    BoundStatementResult() = default;
+    BoundStatementResult(
+        expression_vector columns, vector<expression_vector> expressionsToCollectPerColumn)
+        : columns{std::move(columns)}, expressionsToCollectPerColumn{
+                                           std::move(expressionsToCollectPerColumn)} {}
+
+    static unique_ptr<BoundStatementResult> createEmptyResult() {
+        return make_unique<BoundStatementResult>();
+    }
+
+    static unique_ptr<BoundStatementResult> createSingleStringColumnResult() {
+        auto result = make_unique<BoundStatementResult>();
+        auto stringColumn = make_shared<Expression>(LITERAL, DataType{STRING}, "outputMsg");
+        result->addColumn(stringColumn, expression_vector{stringColumn});
+        return result;
+    }
+
+    inline void addColumn(shared_ptr<Expression> column, expression_vector expressionToCollect) {
+        columns.push_back(std::move(column));
+        expressionsToCollectPerColumn.push_back(std::move(expressionToCollect));
+    }
+    inline expression_vector getColumns() const { return columns; }
+    inline vector<expression_vector> getExpressionsToCollectPerColumn() const {
+        return expressionsToCollectPerColumn;
+    }
+
+    inline expression_vector getExpressionsToCollect() {
+        expression_vector result;
+        for (auto& expressionsToCollect : expressionsToCollectPerColumn) {
+            for (auto& expression : expressionsToCollect) {
+                result.push_back(expression);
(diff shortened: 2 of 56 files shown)


src/include/main/query_result.h
```diff
@@ -10,78 +10,62 @@ using namespace kuzu::processor;
...
     std::string errMsg;
 
-    std::unique_ptr<QueryResultHeader> header;
+    // header information
+    std::vector<std::string> columnNames;
+    std::vector<common::DataType> columnDataTypes;
+    // data
     std::shared_ptr<processor::FactorizedTable> factorizedTable;
     std::unique_ptr<processor::FlatTupleIterator> iterator;
+    shared_ptr<FlatTuple> tuple;
```
ray6080, line 66: 1. There is an inconsistency here (shared_ptr) and in the `initResultTableAndIterator` function (make_unique).
2. Can we not use a `unique_ptr` here? And we return a `FlatTuple*` for the `getNext` function.

src/include/processor/mapper/plan_mapper.h
```diff
@@ -23,7 +23,8 @@ class PlanMapper {
         : storageManager{storageManager}, memoryManager{memoryManager},
           expressionMapper{}, catalog{catalog}, physicalOperatorID{0} {}
 
-    unique_ptr<PhysicalPlan> mapLogicalPlanToPhysical(LogicalPlan* logicalPlan);
+    unique_ptr<PhysicalPlan> mapLogicalPlanToPhysical(LogicalPlan* logicalPlan,
+        const expression_vector& expressionsToCollect, bool isDDLOrCopyCSV);
```
ray6080, line 27: I would suggest pass in `StatementType statementType` instead of `bool isDDLOrCopyCSV` here.

src/include/processor/physical_plan.h
```diff
@@ -8,29 +8,13 @@ namespace kuzu {
...
-                   PhysicalOperatorType::CREATE_REL_TABLE ||
-               lastOperator->getChild(0)->getOperatorType() == PhysicalOperatorType::DROP_TABLE;
-    }
+    explicit PhysicalPlan(unique_ptr<PhysicalOperator> lastOperator, bool isDDLOrCopyCSV)
+        : lastOperator{std::move(lastOperator)}, isDDLOrCopyCSV{isDDLOrCopyCSV} {}
 
 public:
     unique_ptr<PhysicalOperator> lastOperator;
-    bool readOnly;
+    bool isDDLOrCopyCSV;
```
ray6080, line 17: This can be changed to StatementType.

src/include/processor/result/factorized_table.h
```diff
@@ -325,14 +325,25 @@ class FactorizedTable {
...
+    explicit FlatTupleIterator(FactorizedTable& factorizedTable, vector<Value*> values);
 
     inline bool hasNextFlatTuple() {
         return nextTupleIdx < factorizedTable.getNumTuples() || nextFlatTupleIdx < numFlatTuples;
     }
 
-    shared_ptr<FlatTuple> getNextFlatTuple();
+    void getNextFlatTuple();
+
+    inline void resetState() {
```
ray6080, line 336: Will this function be inlined? It looks too complicated to be auto inlined.

src/include/processor/result/value.h
```diff
@@ -0,0 +1,92 @@
...
+        return val.intervalVal;
+    }
+
+    inline std::string getStringVal() const {
+        validateType(common::STRING);
+        return stringVal;
+    }
+
+    // TODO(Xiyang): Maybe we should do copy here.
+    inline const vector<unique_ptr<Value>>& getListVal() const {
```
ray6080, line 50: As discussed, it's better to expose both copy and reference interfaces.

ray6080 (approved):

src/include/binder/binder.h
```diff
@@ -88,13 +88,12 @@ class Binder {
 
     /*** bind projection clause ***/
     unique_ptr<BoundWithClause> bindWithClause(const WithClause& withClause);
-    unique_ptr<BoundReturnClause> bindReturnClause(
-        const ReturnClause& returnClause, unique_ptr<BoundSingleQuery>& boundSingleQuery);
+    unique_ptr<BoundReturnClause> bindReturnClause(const ReturnClause& returnClause);
 
     expression_vector bindProjectionExpressions(
         const vector<unique_ptr<ParsedExpression>>& projectionExpressions, bool containsStar);
     // For RETURN clause, we write variable "v" as all properties of "v"
```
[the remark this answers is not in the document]
andyfengHKU: This comment is still partially true, we still perform this rewrite but not in RETURN clause anymore. Updated the comment.

src/include/binder/bound_statement_result.h
```diff
@@ -0,0 +1,57 @@
...
+    BoundStatementResult(
+        expression_vector columns, vector<expression_vector> expressionsToCollectPerColumn)
+        : columns{std::move(columns)}, expressionsToCollectPerColumn{
+                                           std::move(expressionsToCollectPerColumn)} {}
+
+    static unique_ptr<BoundStatementResult> createEmptyResult() {
+        return make_unique<BoundStatementResult>();
+    }
+
+    static unique_ptr<BoundStatementResult> createSingleStringColumnResult() {
```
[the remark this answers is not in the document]
andyfengHKU: Opened an issue.

src/include/main/query_result.h
andyfengHKU: Using unique_ptr meaning we will need to expose getCopy and getReference interfaces. Ducked also uses a shared_ptr. So I wouldn't priority changing this.

src/include/processor/mapper/plan_mapper.h
andyfengHKU: Giving statementType to physicalPlan doesn't seem correct either. I will remove this field soon anyway.

src/include/processor/physical_plan.h
andyfengHKU: See comments above

Issue #1053, opened 2022-11-21
list slice/extract bugs


aziz-mu:
Some buggy queries found using list slice/extract:

- [ ] Gives an error and crashes (out of range list slice):  `MATCH (a:person) RETURN a.fName[7:]`
- [x] Gives an error but should just return blank output (out of range extract): `MATCH (a:person) RETURN a.courseScoresPerTerm[2]`
- [x] Also, extract gives segfault for index <= 0: `MATCH (a:person) RETURN a.courseScoresPerTerm[0]`
- [ ] Negative index on right extract: `MATCH (a:person) RETURN a.fName[:-2]`

PR #1172, opened 2023-01-11
list slice/extract bugs #1053

anuchak:
#1053

### DIFF
--- a/src/include/function/list/operations/list_extract_operation.h
+++ b/src/include/function/list/operations/list_extract_operation.h
@@ -30,16 +30,13 @@ public:
             return;
         }
         auto uint64Pos = (pos < 0) ? (pos + list.size + 1) : (uint64_t)pos;
-        if (uint64Pos == 0) {
-            uint64Pos = list.size;
-        }
         auto values = reinterpret_cast<T*>(list.overflowPtr);
         result = values[uint64Pos - 1];
         setValue(values[uint64Pos - 1], result, resultValueVector);
     }
 
     static inline void operation(ku_string_t& str, int64_t& idx, ku_string_t& result) {
-        if (str.len < idx) {
+        if (str.len < idx || abs(idx) > str.len) {
             result.set("", 0);
         } else {
             ArrayExtract::operation(str, idx, result);
--- a/src/include/function/list/operations/list_extract_operation.h
+++ b/src/include/function/list/operations/list_extract_operation.h
@@ -26,10 +26,12 @@ public:
     template<typename T>
     static inline void operation(
         ku_list_t& list, int64_t pos, T& result, ValueVector& resultValueVector) {
-        auto uint64Pos = (uint64_t)pos;
-        if (list.size < uint64Pos) {
-            throw RuntimeException("list_extract(list, index): index=" + TypeUtils::toString(pos) +
-                                   " is out of range.");
+        if (pos == 0 || abs(pos) > list.size) {
+            return;
+        }
+        auto uint64Pos = (pos < 0) ? (pos + list.size + 1) : (uint64_t)pos;
+        if (uint64Pos == 0) {
+            uint64Pos = list.size;
         }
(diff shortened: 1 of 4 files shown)


src/include/function/list/operations/list_extract_operation.h
```diff
@@ -26,10 +26,12 @@ struct ListExtract {
     template<typename T>
     static inline void operation(
         ku_list_t& list, int64_t pos, T& result, ValueVector& resultValueVector) {
-        auto uint64Pos = (uint64_t)pos;
-        if (list.size < uint64Pos) {
-            throw RuntimeException("list_extract(list, index): index=" + TypeUtils::toString(pos) +
-                                   " is out of range.");
+        if (pos == 0 || abs(pos) > list.size) {
```
andyfengHKU, line 29: This seems to be incorrect. You are returning without setting any value for result which leaves it into an undefined state. The right thing should be setting resultValueVector.setNull(pos or pos - 1) to make sure result is a null value.

Add test so see if you can reproduce the undefined state case.

src/include/function/list/operations/list_extract_operation.h
```diff
@@ -26,10 +26,12 @@ struct ListExtract {
...
+            return;
+        }
+        auto uint64Pos = (pos < 0) ? (pos + list.size + 1) : (uint64_t)pos;
+        if (uint64Pos == 0) {
```
andyfengHKU, line 33: what is this case stands for, will we ever go inside this if statement?

src/include/function/list/operations/list_extract_operation.h
```diff
@@ -26,10 +26,12 @@ struct ListExtract {
...
+            uint64Pos = list.size;
```
andyfengHKU, line 34: I don't see changes applied to string list extraction. Make sure their behaviour is consistent and add test.

src/include/function/list/operations/list_slice_operation.h
```diff
@@ -30,8 +30,12 @@ struct ListSlice {
 
     static inline void operation(ku_string_t& str, int64_t& begin, int64_t& end,
         ku_string_t& result, ValueVector& resultValueVector) {
-        int64_t startIdx = (begin == 0) ? 1 : begin;
-        int64_t endIdx = (end == 0) ? str.len : end;
+        int64_t startIdx = (begin == 0) ? 1 : ((begin < 0) ? (begin + str.len + 1) : begin);
```
andyfengHKU, line 33: the logic handling negative index seems to be same for list extract and slice. So refactor a list operation utils class/struct to hold this logic.

src/include/function/list/operations/list_slice_operation.h
```diff
@@ -30,8 +30,12 @@ struct ListSlice {
...
+        int64_t endIdx = (end == 0) ? str.len : ((end < 0) ? (end + str.len + 1) : end);
+        if (endIdx < startIdx || startIdx < 0 || endIdx < 0) {
```
andyfengHKU, line 35: Similar to list extraction, you seem to only modify one of the template functions. This one is for string slice but not general list slice. Make sure you modified both and test both cases. Also, don't forget to refactor/template common logic whenever possible.

A separate comment not relates to this PR (but let's change it in this PR), use substring operation to do string slice. It's always good to reuse code.

andyfengHKU (changes requested):

src/include/function/list/operations/list_extract_operation.h
anuchak: this case is not feasible, removed

src/include/function/list/operations/list_extract_operation.h
anuchak: string extract cases are handled in the array extract function, only 1 case was missing which was the -ve index out of range, I've added logic and test for it

anuchak:
This PR is being closed, Guodong will be taking over this ticket: #1053

PR #1182, opened 2023-01-16
add make debug with address sanitizer flag enabled

anuchak:
add make debug with address sanitizer flag enabled


.github/workflows/ci-workflow.yml
```diff
@@ -23,6 +23,20 @@ jobs:
       - name: test
         run: CC=gcc CXX=g++ make test
 
+  gcc-make-debug:
+    name: gcc make debug
+    needs: [ clang-formatting-check ]
```
mewim, line 28: We should make this depend on `gcc-build-test` instead. If the gcc build test does not success we should not run the debug build.

.github/workflows/ci-workflow.yml
```diff
@@ -23,6 +23,20 @@ jobs:
...
+    runs-on: kuzu-self-hosted-testing
+    steps:
+      - uses: actions/checkout@v2
+      - run: pip install -r tools/python_api/requirements_dev.txt
```
mewim, line 32: For all the `pip install` on CI, we should change to `pip install --user -r`

.github/workflows/ci-workflow.yml
anuchak: yes added that

.github/workflows/ci-workflow.yml
anuchak: done

PR #1183, opened 2023-01-16
Compile CI test in parallel

ray6080:


.github/workflows/ci-workflow.yml
```diff
@@ -21,7 +21,7 @@ jobs:
         run: CC=gcc CXX=g++ make release NUM_THREADS=32
 
       - name: test
-        run: CC=gcc CXX=g++ make test
+        run: CC=gcc CXX=g++ make test NUM_THREADS=16
```
andyfengHKU, line 24: let's use 32 instead of 16

andyfengHKU (approved):

.github/workflows/ci-workflow.yml
ray6080: Right. Changed.

PR #1192, opened 2023-01-18
Add Python binding for NODE & REL types; output query results to NetworkX

### LINKED ISSUE #1173: Python API `get_as_df()` returns incorrect result when being called for more than once
Steps to reproduce:
```
import kuzu
db = kuzu.Database("test")
conn = kuzu.Connection(db)
conn.execute("CREATE NODE TABLE person (ID INT64, fName STRING, gender INT64, isStudent BOOLEAN, isWorker BOOLEAN, "
             "age INT64, eyeSight DOUBLE, birthdate DATE, registerTime TIMESTAMP, lastJobDuration "
             "INTERVAL, workedHours INT64[], usedNames STRING[], courseScoresPerTerm INT64[][], PRIMARY "
             "KEY (ID))")
conn.execute('COPY person FROM "dataset/tinysnb/vPerson.csv" (HEADER=true)')
res = conn.execute("MATCH (p:person) RETURN p.fName")

res.get_as_df()
# returns correct result

res.get_as_df()
# returns all None or garbage
```

This is likely due to that there are no more tuples in `QueryResult`, but the `get_as_df` method still keeps reading. We should either throw an error in this case or allow jumping to the beginning of the `QueryResult` again.

mewim:
This PR does the following:
- Add Python binding for new NODE and REL type introduced in #1168. The NODE and REL types are returned as Python dicts.
- Add a new API `get_as_networkx()`, which converts the query results as NetworkX graph.
- Split C++ and Python tests.
- Fix #1173 and add a binding for the user to manually reset the iterators.

### DIFF
--- a/src/include/common/types/value.h
+++ b/src/include/common/types/value.h
@@ -109,6 +109,13 @@ public:
         properties.emplace_back(key, std::move(value));
     }
 
+    inline const vector<pair<std::string, unique_ptr<Value>>>& getProperties() const {
+        return properties;
+    }
+
+    nodeID_t getNodeID() const;
+    string getLabelName() const;
+
     inline unique_ptr<NodeVal> copy() const { return make_unique<NodeVal>(*this); }
 
     string toString() const;
@@ -129,7 +136,14 @@ public:
         properties.emplace_back(key, std::move(value));
     }
 
+    inline const vector<pair<std::string, unique_ptr<Value>>>& getProperties() const {
+        return properties;
+    }
+
     inline unique_ptr<RelVal> copy() const { return make_unique<RelVal>(*this); }
+
+    nodeID_t getSrcNodeID() const;
+    nodeID_t getDstNodeID() const;
 
     string toString() const;
 
--- a/tools/python_api/src_cpp/py_query_result.cpp
+++ b/tools/python_api/src_cpp/py_query_result.cpp
@@ -18,7 +18,8 @@ void PyQueryResult::initialize(py::handle& m) {
         .def("close", &PyQueryResult::close)
         .def("getAsDF", &PyQueryResult::getAsDF)
         .def("getColumnNames", &PyQueryResult::getColumnNames)
-        .def("getColumnDataTypes", &PyQueryResult::getColumnDataTypes);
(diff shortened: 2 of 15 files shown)


src/include/common/types/value.h
```diff
@@ -109,6 +109,13 @@ class NodeVal {
         properties.emplace_back(key, std::move(value));
     }
 
+    inline const vector<pair<std::string, unique_ptr<Value>>>& getProperties() const {
+        return properties;
+    }
+
+    inline const Value& getNodeID() const { return *idVal; }
```
andyfengHKU, line 116: We can probably directly do `inline const nodeID_t getNodeID() const { return idVal.getValue<nodeID_t>()}` since we know the data type. Ditto for getLabelName().

src/include/common/types/value.h
```diff
@@ -129,10 +136,17 @@ class RelVal {
...
 
+    inline const vector<pair<std::string, unique_ptr<Value>>>& getProperties() const {
+        return properties;
+    }
+
     inline unique_ptr<RelVal> copy() const { return make_unique<RelVal>(*this); }
 
     string toString() const;
 
+    inline const Value& getSrcNodeID() const { return *srcNodeIDVal; }
```
andyfengHKU, line 147: ditto return nodeID_t for getSrc/DstNodeID()

tools/python_api/src_cpp/py_query_result.cpp
```diff
@@ -132,3 +150,25 @@ py::list PyQueryResult::getColumnNames() {
...
+    const vector<pair<std::string, unique_ptr<Value>>>& properties) {
+    py::dict result;
+    for (auto i = 0u; i < properties.size(); ++i) {
+        auto& [name, value] = properties[i];
+        result[py::cast(name)] = convertValueToPyObject(*value);
+    }
+    return result;
+}
+
+py::dict PyQueryResult::convertNodeIdToPyDict(const Value& nodeIdVal) {
```
andyfengHKU, line 168: this function will probably go away if the API directly returns nodeID_t

andyfengHKU (approved):

ray6080 (approved):

PR #1193, opened 2023-01-20
Export query result (fixed sized values only) to arrow

ray6080:
- add `get_as_arrow()`, support BOOL, INT64, DOUBLE, DATE, TIMESTAMP, INTERVAL.
- rename `getResultValue` to `getValue`.

### DIFF
--- a/tools/python_api/requirements_dev.txt
+++ b/tools/python_api/requirements_dev.txt
@@ -3,3 +3,4 @@ pytest
 pandas
 networkx~=3.0.0
 numpy
+pyarrow~=10.0.0
--- a/tools/python_api/test/test_arrow.py
+++ b/tools/python_api/test/test_arrow.py
@@ -1,0 +1,60 @@
+import sys
+
+sys.path.append('../build/')
+import kuzu
+import pyarrow as pa
+import datetime
+
+
+def test_to_arrow(establish_connection):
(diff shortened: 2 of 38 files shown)


tools/python_api/test/test_arrow.py
```diff
@@ -0,0 +1,62 @@
+import sys
+
+sys.path.append('../build/')
+import kuzu
+import pyarrow as pa
+
+sys.path.append('../build/')
```
mewim, line 7: Duplicated path

tools/python_api/requirements_dev.txt
```diff
@@ -3,3 +3,4 @@ pytest
 pandas
 networkx~=3.0.0
 numpy
+pyarrow
```
mewim, line 6: We should probably specify a version here in case the compatibility breaks.

mewim (approved):
Overall looks good to me. I am not very familiar with the internals of Arrow format though.

tools/python_api/test/test_arrow.py
ray6080: Done.

tools/python_api/requirements_dev.txt
ray6080: Done.

Issue #1191, opened 2023-01-17
LDBC SNB benchmark queries


ray6080:
Besides our current customized benchmark queries under `benchmark/queries`, it's good to also add some LDBC SNB IC and IS queries (see here for cypher queries: https://github.com/ldbc/ldbc_snb_interactive_impls/tree/main/cypher/queries).

The first step is to identify queries we can support for now on a small dataset (ldbc0.1 or ldbc1), then we can gradually add them into our benchmark.

PR #1197, opened 2023-01-20
add ldbc snb simple and complex queries

anuchak:
LDBC SNB simple and complex queries, some had to be rewritten and there are some functions missing like collect, coalesce, shortest path.


ray6080 (approved):
LGTM. Thanks!

anuchak:
#1191 is partially completed, some functions missing, those queries will be added later

PR #1199, opened 2023-01-23
Export query result to arrow: string data type

ray6080:
- add support of string data type in arrow export
- fix csv copy mismatch in py test (leading to changes of ground truth in df and networkx tests)
- add more tests for arrow export


mewim (approved):
LGTM. I think we should probably add a test for UTF-8.

PR #1200, opened 2023-01-24
Implment `get_as_torch_geometric`

mewim:
This PR implements a function `get_as_torch_geometric()` which converts the query result into a `torch_geometric.data` object.

Principles: 
- Only full `NODE` and `REL`s are converted. Selected properties, such as `a.isWorker` is ignored.
- Only converts `INT64` and `BOOL` types. Other types are ignored. Multi-dimensional lists are supported as long as the dimension is consistent and there is no sub-list of different length.
- Edge attributes and labels are ignored for now.
- Best effort: it will try to convert every attribute that can be converted and ignore those that cannot be converted to `torch.Tensor`.
- No throw: it only shows warnings for incompatible types but never throws an error.


tools/python_api/src_py/query_result.py
```diff
@@ -65,16 +65,7 @@ def get_as_networkx(self, directed=True):
...
-        column_types = self.get_column_data_types()
-        column_to_extract = {}
-
-        # Iterate over columns and extract nodes and rels, ignoring other columns
-        for i in range(len(column_names)):
-            column_name = column_names[i]
-            column_type = column_types[i]
-            if column_type in [NODE_TYPE, REL_TYPE]:
-                column_to_extract[i] = (column_type, column_name)
+        column_to_extract = self._get_columns_to_extract()
```
ray6080, line 68: `column_to_extract` -> `columns_to_extract`

tools/python_api/src_py/query_result.py
```diff
@@ -115,3 +106,28 @@ def get_as_networkx(self, directed=True):
                 table_to_label_dict[_dst["table"]]) + "_" + str(_dst["offset"])
             nx_graph.add_edge(src_id, dst_id, **rel)
         return nx_graph
+
+    def _get_columns_to_extract(self):
+        column_names = self.get_column_names()
+        column_types = self.get_column_data_types()
+        column_to_extract = {}
```
ray6080, line 113: Ditto. `column_to_extract` -> `columns_to_extract`

tools/python_api/src_py/types.py
```diff
@@ -0,0 +1,16 @@
...
+    INT64 = "INT64"
+    DOUBLE = "DOUBLE"
+    STRING = "STRING"
+    DATE = "DATE"
+    TIMESTAMP = "TIMESTAMP"
+    INTERVAL = "INTERVAL"
+    LIST = "LIST"
+    NODE = "NODE"
+    REL = "REL"
+    NODE_ID = "NODE_ID"
```
ray6080, line 16: Update NODE_ID once Ziyi's PR is in.

tools/python_api/src_py/torch_geometric_result_converter.py
```diff
@@ -0,0 +1,237 @@
...
+        self.ignored_columns = set()
+        self.column_to_extract = self.query_result._get_columns_to_extract()
+
+    def __get_node_property_names(self, table_name):
+        if table_name in self.nodes_property_names_dict:
+            return self.nodes_property_names_dict[table_name]
+
+        PRIMARY_KEY_SYMBOL = "(PRIMARY KEY)"
+        LIST_SYMBOL = "[]"
+        result_str = self.query_result.connection._connection.get_node_property_names(
```
ray6080, line 25: This comment is beyond this PR, but let me document it here:

The parsing here makes me think that our cpp interface `string Connection::getNodePropertyNames(const string& tableName)` is incorrect. Instead of returning a `string`, it's better to return `vector<string>`, each inside which is a property name. The caller (shell) shall be responsible to organize them in a printable way.

tools/python_api/src_py/torch_geometric_result_converter.py
```diff
@@ -0,0 +1,237 @@
...
+            line_splited = line.split(" ")
+            if len(line_splited) < 2:
+                continue
+
+            prop_name = line_splited[0]
+            prop_type = " ".join(line_splited[1:])
+
+            is_primary_key = PRIMARY_KEY_SYMBOL in prop_type
+            prop_type = prop_type.replace(PRIMARY_KEY_SYMBOL, "")
+            depth = prop_type.count(LIST_SYMBOL)
```
ray6080, line 44: How about `dimension`?

tools/python_api/src_py/torch_geometric_result_converter.py
```diff
@@ -0,0 +1,237 @@
...
+
+    def __extract_properties_from_node(self, node, label, node_property_names):
+        import torch
+        for prop_name in node_property_names:
+            # Ignore properties that are marked as ignored
+            if (label, prop_name) in self.ignored_columns:
+                continue
+
+            # Ignore primary key
+            if node_property_names[prop_name]["is_primary_key"]:
```
ray6080, line 97: Why not put primary_key prop_name also into the ignored properties?

tools/python_api/src_py/torch_geometric_result_converter.py
```diff
@@ -0,0 +1,237 @@
...
+                    _src = row[i]["_src"]
+                    _dst = row[i]["_dst"]
+                    self.rels.add((_src["table"], _src["offset"],
+                                   _dst["table"], _dst["offset"]))
+
+    def __extract_properties_from_node(self, node, label, node_property_names):
+        import torch
+        for prop_name in node_property_names:
+            # Ignore properties that are marked as ignored
+            if (label, prop_name) in self.ignored_columns:
```
ray6080, line 93: `ignored_columns` -> `ignored_properties`

tools/python_api/src_py/types.py
```diff
@@ -0,0 +1,16 @@
+from enum import Enum
+
+
+class Type(Enum):
```
ray6080, line 4: Is there a way that we don't need to maintain a separate mapping in Python?

ray6080 (approved):

tools/python_api/src_py/query_result.py
```diff
@@ -115,3 +106,28 @@ def get_as_networkx(self, directed=True):
                 table_to_label_dict[_dst["table"]]) + "_" + str(_dst["offset"])
             nx_graph.add_edge(src_id, dst_id, **rel)
         return nx_graph
+
+    def _get_columns_to_extract(self):
```
ray6080, line 110: `_get_columns_to_extract` -> `_get_properties_to_extract`

ray6080 (approved):

tools/python_api/src_py/torch_geometric_result_converter.py
mewim: Ignored property is not read at all, but for primary key, it still needs to be read as we need to maintain the `pos_to_primary_key_dict`. I have changed the comment to `Read primary key but do not add it to the node properties`.

tools/python_api/src_py/types.py
mewim: As discussed in #1189, pybind11 has a memory leakage bug when binding enum class. We can change this `Enum` to binding after the bug is fixed.

PR #1201, opened 2023-01-25
Connection + Database Classes for Nodejs bindings

ynoza:
Followed the python binding structure to create a nodejs binding for kuzu focusing on the connection and database APIs.


tools/nodejs_api/src_cpp/node_connection.h
```diff
@@ -0,0 +1,14 @@
+#include <napi.h>
+#include <iostream>
+#include "main/kuzu.h"
+
+class NodeConnection : public Napi::ObjectWrap<NodeConnection> {
```
mewim, line 5: Connection is not taking `numThreads` argument for initialization and also missing method `setMaxNumThreadForExec`.

tools/nodejs_api/src_cpp/node_database.cpp
```diff
@@ -0,0 +1,69 @@
...
+    if (!info[1].IsNumber()) {
+        Napi::TypeError::New(env, "Database buffer manager size must be an int_64").ThrowAsJavaScriptException();
+    }
+
+    std::string databaseConfigString = info[0].ToString().Utf8Value().c_str();
+    std::cout << "Database config is " << databaseConfigString << std::endl;
+    DatabaseConfig databaseConfig(databaseConfigString);
+
+    std::int64_t bufferSize = info[1].As<Napi::Number>().DoubleValue();
+    std::cout  << "Database buffer size is " << bufferSize << std::endl;
```
mewim, line 43: Do not output anything in your methods.

tools/nodejs_api/src_cpp/node_database.h
```diff
@@ -0,0 +1,21 @@
+#ifndef KUZU_NODE_DATABASE_H
+#define KUZU_NODE_DATABASE_H
+
+#include <napi.h>
+#include <iostream>
+#include "main/kuzu.h"
+
+class NodeDatabase : public Napi::ObjectWrap<NodeDatabase> {
```
mewim, line 8: Database class is missing `setLoggingLevel ` but it is a minor feature, so it is OK not to have it for this iteration.

tools/nodejs_api/sample.js
```diff
@@ -0,0 +1,14 @@
+// Make sure the test directory is removed as it will be recreated
+const fs = require("fs");
+const {Database, Connection} = require("./src_nodejs/index.js");
```
mewim, line 3: You should probably use CMake to copy the built module and the js files into a proper directory structure (refer to https://github.com/kuzudb/kuzu/blob/89fba7c1328b543747fc4948e6bff0961263276b/tools/python_api/CMakeLists.txt#L32) so you can do something like `require("build/kuzu/")`

tools/nodejs_api/src_cpp/node_database.cpp
```diff
@@ -0,0 +1,69 @@
...
+    exports.Set("NodeDatabase", t);
+    return exports;
+}
+
+
+NodeDatabase::NodeDatabase(const Napi::CallbackInfo& info) : Napi::ObjectWrap<NodeDatabase>(info)  {
+    Napi::Env env = info.Env();
+    Napi::HandleScope scope(env);
+
+    if (info.Length()!=2) {
```
mewim, line 28: The buffer manager size should be optional. If it is not provided, setting it to 0 so that the system can automatically configure it: https://github.com/kuzudb/kuzu/blob/89fba7c1328b543747fc4948e6bff0961263276b/tools/python_api/src_cpp/py_database.cpp#L6

tools/nodejs_api/src_cpp/node_database.cpp
```diff
@@ -0,0 +1,69 @@
...
+        Napi::TypeError::New(env, "Need database config string and buffer manager size").ThrowAsJavaScriptException();
+    }
+    if (!info[0].IsString()) {
+        Napi::TypeError::New(env, "Database config parameter must be a string").ThrowAsJavaScriptException();
+    }
+    if (!info[1].IsNumber()) {
+        Napi::TypeError::New(env, "Database buffer manager size must be an int_64").ThrowAsJavaScriptException();
+    }
+
+    std::string databaseConfigString = info[0].ToString().Utf8Value().c_str();
```
mewim, line 38: You should not cast the value to `c_str`. The cpp class supports UTF-8 string (non-English character). Casting it to `c_str` will break this functionality.

mewim (changes requested):

tools/nodejs_api/src_cpp/node_database.cpp
ynoza: > ` constructor(databaseConfigString, bufferSize = 0) {
        this.database = new kuzu.NodeDatabase(databaseConfigString, bufferSize);
    }`
Is it fine if I just do this in the javascript side -> if a user doesn't specifiy an argument then it will pass 0 to the NodeDatabase class

mewim: You can do it on JS side

tools/nodejs_api/sample.js
ynoza: Is it ok if I do require("build/kuzu/index.js")?

ynoza: nvm

tools/nodejs_api/src_cpp/node_connection.cpp
```diff
@@ -0,0 +1,119 @@
...
+  exports.Set("NodeConnection", t);
+  return exports;
+}
+
+
+NodeConnection::NodeConnection(const Napi::CallbackInfo& info) : Napi::ObjectWrap<NodeConnection>(info)  {
+  Napi::Env env = info.Env();
+  Napi::HandleScope scope(env);
+
+  if (info.Length()!=2 || !info[0].IsObject() || !info[1].IsNumber()) {
```
andyfengHKU, line 31: the error msg is confusing given the three check in this if statement

tools/nodejs_api/src_cpp/node_database.h
```diff
@@ -0,0 +1,15 @@
...
+public:
+    static Napi::Object Init(Napi::Env env, Napi::Object exports);
+    NodeDatabase(const Napi::CallbackInfo& info);
+    ~NodeDatabase();
+    friend class NodeConnection;
```
andyfengHKU, line 10: why do u need friend class?

tools/nodejs_api/sample.js
```diff
@@ -0,0 +1,26 @@
+// Make sure the test directory is removed as it will be recreated
+const fs = require("fs");
+const {Database, Connection} = require("./build/kuzu");
+
+try {
+  fs.rmSync("./test", { recursive: true });
+} catch (e) {
+  // ignore
```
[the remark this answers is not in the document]
mewim: Actually I wrote this. Since this test script always recreate a test database, we remove the old one before running, but node throws an error when removing non-existent file, which can be ignored.

tools/nodejs_api/src_cpp/node_database.h
ynoza: Because I need to access a kuzu::main::database field in the NodeConnection class (since in the initialization of Connection it takes in the database) and rather than make that field public I think it makes sense to make NodeConnection a friend so that it can access private fields like that.

tools/nodejs_api/src_cpp/node_connection.cpp
ynoza: It only says` needs database class passed in` because on the javascript side numThreads (the second param) has a default value so the user doesn't need to specify it - hence an error will only occur if a valid database class isn't passed in.

mewim (approved):

PR #1209, opened 2023-01-27
Export query result to arrow: NodeID, Node, Rel, LIST data types

ray6080:


src/common/arrow/arrow_row_batch.cpp
```diff
@@ -106,82 +164,179 @@ void ArrowRowBatch::templateCopyNonNullValue<bool>(
...
+template<>
+void ArrowRowBatch::templateCopyNonNullValue<LIST>(
+    ArrowVector* vector, const main::DataTypeInfo& typeInfo, Value* value, std::int64_t pos) {
+    vector->data.resize((pos + 2) * sizeof(std::uint32_t));
+    auto offsets = (std::uint32_t*)vector->data.data();
+    auto numElements = value->listVal.size();
+    offsets[pos + 1] = offsets[pos] + numElements;
+    auto numChildElements = offsets[pos + 1] + 1;
+    auto currentNumBytesForChildValidity = vector->childData[0]->validity.size();
+    auto numBytesForChildValidity = getNumBytesForBits(numChildElements);
```
mewim, line 185: Maybe add a comment here about what is `ChildValidity` and the meaning of `0xFF`.

src/common/arrow/arrow_row_batch.cpp
```diff
@@ -106,82 +164,179 @@ void ArrowRowBatch::templateCopyNonNullValue<bool>(
...
-    switch (value->dataType.typeID) {
+template<>
+void ArrowRowBatch::templateCopyNonNullValue<LIST>(
+    ArrowVector* vector, const main::DataTypeInfo& typeInfo, Value* value, std::int64_t pos) {
+    vector->data.resize((pos + 2) * sizeof(std::uint32_t));
+    auto offsets = (std::uint32_t*)vector->data.data();
+    auto numElements = value->listVal.size();
+    offsets[pos + 1] = offsets[pos] + numElements;
+    auto numChildElements = offsets[pos + 1] + 1;
+    auto currentNumBytesForChildValidity = vector->childData[0]->validity.size();
```
mewim, line 184: Maybe add a comment here about what is `ChildValidity` and the meaning of `0xFF`.

src/common/arrow/arrow_row_batch.cpp
```diff
@@ -106,82 +164,179 @@ void ArrowRowBatch::templateCopyNonNullValue<bool>(
...
+    Value offsetVal((std::int64_t)nodeID.offset);
+    Value tableIDVal((std::int64_t)nodeID.tableID);
+    appendValue(vector->childData[0].get(), *typeInfo.childrenTypesInfo[0], &offsetVal);
+    appendValue(vector->childData[1].get(), *typeInfo.childrenTypesInfo[1], &tableIDVal);
+}
+
+template<>
+void ArrowRowBatch::templateCopyNonNullValue<NODE>(
+    ArrowVector* vector, const main::DataTypeInfo& typeInfo, Value* value, std::int64_t pos) {
+    appendValue(
```
mewim, line 213: Why do we need to add the id and label manually? Could it just be a for-loop with 0 to start?

src/common/arrow/arrow_row_batch.cpp
```diff
@@ -106,82 +164,179 @@ void ArrowRowBatch::templateCopyNonNullValue<bool>(
...
+    for (auto& [name, val] : value->nodeVal->getProperties()) {
+        appendValue(vector->childData[propertyId].get(), *typeInfo.childrenTypesInfo[propertyId],
+            val.get());
+        propertyId++;
+    }
+}
+
+template<>
+void ArrowRowBatch::templateCopyNonNullValue<REL>(
+    ArrowVector* vector, const main::DataTypeInfo& typeInfo, Value* value, std::int64_t pos) {
```
mewim, line 227: Same question here

src/common/arrow/arrow_row_batch.cpp
```diff
@@ -106,82 +164,179 @@ void ArrowRowBatch::templateCopyNonNullValue<bool>(
...
+    case NODE: {
+        templateCopyNonNullValue<NODE>(vector, typeInfo, value, pos);
+    } break;
+    case REL: {
+        templateCopyNonNullValue<REL>(vector, typeInfo, value, pos);
     } break;
     default: {
         throw RuntimeException("Data type " + Types::dataTypeToString(value->dataType) +
-                               " is not supported for arrow exportation.");
+                               " is not supported for arrow exportation. in copyNonNullValue");
```
mewim, line 278: `is not supported for arrow exportation in copyNonNullValue.`

src/main/query_result.cpp
```diff
@@ -5,12 +5,65 @@
...
+    }
+    }
+    return std::move(columnTypeInfo);
+}
+
+vector<unique_ptr<DataTypeInfo>> QueryResult::getColumnTypesInfo() {
+    vector<unique_ptr<DataTypeInfo>> result;
+    for (auto i = 0u; i < columnDataTypes.size(); i++) {
+        auto columnTypeInfo = DataTypeInfo::getInfoForDataType(columnDataTypes[i], columnNames[i]);
+        if (columnTypeInfo->typeID == common::NODE || columnTypeInfo->typeID == common::REL) {
```
mewim, line 39: Maybe this outer `if` can be removed and the `else` can be changed to `else if (columnTypeInfo->typeID == common::REL)`

mewim (approved):
Overall LGTM.

Two general comments:
1. In the Python class for query result (https://github.com/kuzudb/kuzu/blob/88a989cc2e094d264b1bebc83e2974816775a128/tools/python_api/src_py/query_result.py#L44), we should add `import arrow` before calling the CPP function, so that if `pyarrow` is not installed, the user gets a `ModuleNotFoundError` exception early without having to deal with the confusing exceptions in CPP code.
1. The function seems to throw error for un-processable data types. Should we consider not throwing error and do a best-effort with warning if we add a new type but the arrow conversion has not been implemented?

src/common/arrow/arrow_row_batch.cpp
ray6080: Currently in NodeVal, properties are stored in separate fields from id and label, that's why we need to read them out separately as `value->nodeVal->getNodeIDVal()`, `value->nodeVal->getLabelVal()`, `value->nodeVal->getProperties()`

src/common/arrow/arrow_row_batch.cpp
ray6080: Same as NodeVal.

src/common/arrow/arrow_row_batch.cpp
ray6080: Done.

src/common/arrow/arrow_row_batch.cpp
ray6080: Reworked this exception message.

src/main/query_result.cpp
ray6080: Right. Done.

ray6080:
> Overall LGTM.
> 
> Two general comments:
> 
> 1. In the Python class for query result (https://github.com/kuzudb/kuzu/blob/88a989cc2e094d264b1bebc83e2974816775a128/tools/python_api/src_py/query_result.py#L44
>    ), we should add `import pyarrow` before calling the CPP function, so that if `pyarrow` is not installed, the user gets a `ModuleNotFoundError` exception early without having to deal with the confusing exceptions in CPP code.
> 2. The function seems to throw error for un-processable data types. Should we consider not throwing error and do a best-effort with warning if we add a new type but the arrow conversion has not been implemented?

Thanks! Added import check for `get_as_df` and `get_as_arrow`. For exception, we've covered data types we have right now. for new data types, I think it's better to throw exception than trying to convert unknown type by ourself.

PR #1225, opened 2023-01-31
Fix PyG converter crash when no property is extracted at all

mewim:


tools/python_api/src_py/torch_geometric_result_converter.py
```diff
@@ -182,6 +183,8 @@ def __mark_property_unconverted(self, label, prop_name):
     def __populate_edges_dict(self):
         # Post-process edges, map internal ids to positions
         for r in self.rels:
+            if (r[0], r[1]) not in self.internal_id_to_pos_dict or (r[2], r[3]) not in self.internal_id_to_pos_dict:
```
andyfengHKU, line 186: Seems there should be a warning here because it only happens when user fail to return the src/dst node for a rel which is an undefined behaviour (when converting to pyg) in my opinion. Up to u to decide.

andyfengHKU (approved):

PR #1226, opened 2023-02-01
Index nested loop join optimizer

andyfengHKU:
- Add isPrimaryKey check to PropertyExpression
- Make IndexScan a non-source operator
- Add index-nested-loop-join optimizer as a temporary solution before generic hash join is implemented.


src/binder/bind/bind_graph_pattern.cpp
```diff
@@ -211,10 +211,16 @@ shared_ptr<NodeExpression> Binder::createQueryNode(const NodePattern& nodePatter
     for (auto tableID : tableIDs) {
         nodeTableSchemas.push_back(catalog.getReadOnlyVersion()->getNodeTableSchema(tableID));
     }
+    auto isSingleTable = nodeTableSchemas.size() == 1;
     for (auto& [propertyName, propertySchemas] :
         getNodePropertyNameAndPropertiesPairs(nodeTableSchemas)) {
+        auto isPrimaryKey = false;
```
ray6080, line 217: auto isPrimaryKey = isSingleTable && (nodeTableSchemas[0]->getPrimaryKey().propertyID == propertySchemas[0].propertyID)

src/include/binder/expression/property_expression.h
```diff
@@ -40,6 +44,7 @@ class PropertyExpression : public Expression {
     }
 
 private:
+    bool isPrimaryKey_ = false;
```
ray6080, line 47: Why do we need the `_` postfix?

src/include/optimizer/optimizer.h
```diff
@@ -0,0 +1,17 @@
+#pragma once
+
+#include "planner/logical_plan/logical_plan.h"
+
+namespace kuzu {
+namespace optimizer {
+
+class Optimizer {
+public:
+    Optimizer() = default;
```
ray6080, line 10: Do we need these default constructor and destructor?

src/include/optimizer/optimizer.h
```diff
@@ -0,0 +1,17 @@
...
+    ~Optimizer() = default;
+
+    void optimize(planner::LogicalPlan* plan);
```
ray6080, line 13: Can we make this function static?

src/include/optimizer/index_nested_loop_join_optimizer.h
```diff
@@ -0,0 +1,34 @@
...
+// This is a temporary optimizer that rewrites
+//      Filter(a.ID=b.ID)
+//      CrossProduct                   into           IndexNestedLoopJoin(b)
+//   S(a)           S(b)                              S(a)
+// In the absense of a generic hash join operator.
+// We should merge this operator to filter push down + ASP when the generic hash join is
+// implemented.
+class IndexNestedLoopJoinOptimizer {
+public:
+    IndexNestedLoopJoinOptimizer() = default;
```
ray6080, line 19: Ditto. Do we need these default constructor and destructor?

src/include/optimizer/index_nested_loop_join_optimizer.h
```diff
@@ -0,0 +1,34 @@
...
+    ~IndexNestedLoopJoinOptimizer() = default;
+
+    shared_ptr<planner::LogicalOperator> rewrite(shared_ptr<planner::LogicalOperator> op);
```
ray6080, line 22: Why not set this function as static?

ray6080 (approved):

src/include/binder/expression/property_expression.h
andyfengHKU: because it conflicts with get function name

Issue #756, opened 2022-08-19
Sequential read for adjList and properties


andyfengHKU:
Similar to how we handle overflow sequential reading. AdjList sequential read can be optimized by caching a pinned page and only unpin when the list to read is in a different page.

ray6080:
This is outdated now. We now switch to the node-group based storage layout and use optimistic reads.

PR #1254, opened 2023-02-07
Sequential read for StringPropertyLists

rfdavid:
As previously discussed with @andyfengHKU, this PR implements the method `scanSequentialStringOverflow` when reading the overflow part of a string in `StringPropertyList::readFromSmallList` and `StringPropertyList::readFromLargeList`.

Related to #756


andyfengHKU (approved):

andyfengHKU (changes requested):
Actually let's pause a bit more

PR #1257, opened 2023-02-08
Rework headers for APIs

ray6080:
This is reworked on top of @mewim 's #1237 .

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -12,8 +12,13 @@ add_subdirectory(storage)
 add_subdirectory(transaction)
 
 add_library(kuzu STATIC ${ALL_OBJECT_FILES})
-target_link_libraries(kuzu PUBLIC antlr4_cypher antlr4_runtime utf8proc re2
-        parquet_lib arrow_lib arrow_deps Threads::Threads)
-target_include_directories(
-        kuzu PUBLIC $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
-        $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>)
+target_link_libraries(kuzu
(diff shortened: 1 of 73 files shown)


src/main/client_context.cpp
```diff
@@ -0,0 +1,9 @@
+#include "main/client_context.h"
+
+namespace kuzu {
+namespace main {
+
+ClientContext::ClientContext() : numThreadsForExecution{1} {}
+
+} // namespace main
+} // namespace kuzu
```
mewim, line 9: New line here

src/CMakeLists.txt
```diff
@@ -12,8 +12,12 @@ add_subdirectory(storage)
...
-        parquet_lib arrow_lib arrow_deps Threads::Threads)
-target_include_directories(
-        kuzu PUBLIC $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
-        $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>)
+target_link_libraries(kuzu
+        PUBLIC antlr4_cypher antlr4_runtime utf8proc re2 parquet_lib arrow_lib arrow_deps Threads::Threads)
+target_include_directories(kuzu
+        PUBLIC $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include> $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>)
+add_library(kuzu_shared SHARED ${ALL_OBJECT_FILES})
+target_link_libraries(kuzu_shared
```
mewim, line 20: We should add `set_target_properties(kuzu_shared PROPERTIES OUTPUT_NAME kuzu)` so that the shared lib is also called `libkuzu` instead of `libkuzu_shared`

mewim (approved):

src/main/client_context.cpp
ray6080: Done.

src/CMakeLists.txt
ray6080: Done.

PR #1287, opened 2023-02-14
WIP: nodejs binding QueryResult class + async execute + callback all/each + parameterised queries

ynoza:
Note that in all_each_async_worker.cpp Execute and OnOk the initialisation of the returned Napi::Array was done in 2 passes because storing a Napi::Array as a field and modifying/adding to it during async operations gave the following error: 

Fatal error in HandleScope::HandleScope
Entering the V8 API without proper locking in place

Trace/breakpoint trap (core dumped)


tools/nodejs_api/src_cpp/all_each_async_worker.cpp
```diff
@@ -0,0 +1,83 @@
...
+
+#include <chrono>
+#include <thread>
+#include <vector>
+
+#include "node_query_result.h"
+
+using namespace std;
+
+AllEachAsyncWorker::AllEachAsyncWorker(Function& callback, shared_ptr<kuzu::main::QueryResult>& queryResult,
```
mewim, line 11: Change the name. Call it something like `QueryResultIterator`

tools/nodejs_api/src_cpp/all_each_async_worker.cpp
```diff
@@ -0,0 +1,83 @@
...
+            arr.Set(i, rowArray);
+        }
+    }
+    if (executionType == "all") {
+        Callback().Call({Env().Null(), arr});
+    }
+};
+
+
+Napi::Value AllEachAsyncWorker::ConvertToNapiObject(const kuzu::common::Value& value, Napi::Env env) {
```
mewim, line 53: Separate this from the class into a utility class.

tools/nodejs_api/src_nodejs/common.js
```diff
@@ -0,0 +1,10 @@
+function callbackWrapper(err, callback) {
```
mewim, line 1: Get rid of the `callbackWrapper`.

tools/nodejs_api/src_nodejs/connection.js
```diff
@@ -1,12 +1,29 @@
...
     constructor(database, numThreads = 0) {
         this.#connection = new kuzu.NodeConnection(database.database, numThreads);
     }
 
-    execute(query){
-        return this.#connection.execute(query);
+    execute(query, callback=null){
+        if (callback) {
+            this.#connection.execute(query, (err, queryResult) => {
+                callbackWrapper(err, () => {
```
mewim, line 14: Same here. You can just pass the callback directly.

tools/nodejs_api/src_nodejs/queryResult.js
```diff
@@ -0,0 +1,47 @@
...
+        if (this.#isClosed){
+            return;
+        }
+        this.#queryResult.close();
+        this.#isClosed = true;
+    }
+
+    each(eachCallback, doneCallback) {
+        this.checkForQueryResultClose();
+        this.#queryResult.each((err, result) => {
```
mewim, line 27: The semantic is not correct. You should pass both the `eachCallbak` and `doneCallback` to C++ side.

tools/nodejs_api/src_nodejs/queryResult.js
```diff
@@ -0,0 +1,63 @@
...
+        this.#isClosed = true;
+    }
+
+    async each(rowCallback, doneCallback, errorCallback) {
+        this.checkForQueryResultClose();
+        while (this.#queryResult.hasNext()) {
+            await this.getNext().then(row => {
+                rowCallback(row);
+            }).catch(err => {
+                errorCallback(err);
```
mewim, line 29: You should terminate on error instead of keep fetching next

tools/nodejs_api/src_cpp/util.cpp
```diff
@@ -0,0 +1,95 @@
...
+        if (floor(temp) == temp) {
+            return Value::createValue<int64_t>(temp);
+        } else {
+            return Value::createValue<double_t>(temp);
+        }
+    } else if (val.IsString()) {
+        string temp = val.ToString();
+        return Value::createValue<string>(temp);
+    } else if (val.IsDate()) {
+        double dateMilliseconds = val.As<Napi::Date>().ValueOf();
```
mewim, line 80: There is a method `FromEpochMs` for `timestamp_t`. Consider using it instead of parsing it manually.

tools/nodejs_api/src_cpp/util.cpp
```diff
@@ -0,0 +1,95 @@
...
+    return result;
+}
+
+kuzu::common::Value Util::transformNapiValue(Napi::Value val) {
+    if (val.IsBoolean()) {
+        bool temp = val.ToBoolean();
+        return Value::createValue<bool>(temp);
+    } else if (val.IsNumber()) {
+        double temp = val.ToNumber().DoubleValue();
+        if (floor(temp) == temp) {
```
mewim, line 71: Overall this is not a very robust way of checking integer. Maybe refer to this discussion: https://github.com/nodejs/node-addon-api/issues/265

tools/nodejs_api/src_cpp/util.cpp
```diff
@@ -0,0 +1,95 @@
...
+        }
+        return arr;
+    }
+    case kuzu::common::DATE: {
+        auto dateVal = value.getValue<date_t>();
+        int32_t year, month, day;
+        kuzu::common::Date::Convert(dateVal, year, month, day);
+        return Napi::Date::New(env, GetEpochFromDate(year, month, day));
+    }
+    default:
```
mewim, line 35: Please add support for `INTERVAL` and `TIMESTAMP`

tools/nodejs_api/src_cpp/node_query_result.cpp
```diff
@@ -0,0 +1,107 @@
...
+    }
+    return info.Env().Undefined();
+}
+
+Napi::Value NodeQueryResult::GetNext(const Napi::CallbackInfo& info) {
+    Napi::Env env = info.Env();
+    Napi::HandleScope scope(env);
+
+    if (info.Length()!=1 || !info[0].IsFunction()) {
+        Napi::TypeError::New(env, "Each needs a callback").ThrowAsJavaScriptException();
```
mewim, line 70: Consider remove these checks in C++ and do these checks in JS, so that if the input is not valid, just terminate early instead of propagating to C++.

tools/nodejs_api/src_cpp/node_query_result.cpp
```diff
@@ -0,0 +1,107 @@
...
+        Napi::TypeError::New(env, "Unsuccessful getNext callback: " + std::string(exc.what())).ThrowAsJavaScriptException();
+    }
+    return info.Env().Undefined();
+}
+
+Napi::Value NodeQueryResult::HasNext(const Napi::CallbackInfo& info) {
+    Napi::Env env = info.Env();
+    Napi::HandleScope scope(env);
+
+    if (info.Length()!=0) {
```
mewim, line 94: Consider remove these checks in C++ and do these checks in JS, so that if the input is not valid, just terminate early instead of propagating to C++.

tools/nodejs_api/src_cpp/node_query_result.cpp
```diff
@@ -0,0 +1,107 @@
...
+        Napi::TypeError::New(env, "Unsuccessful queryResult close: " + std::string(exc.what())).ThrowAsJavaScriptException();
+    }
+
+}
+
+Napi::Value NodeQueryResult::All(const Napi::CallbackInfo& info) {
+    Napi::Env env = info.Env();
+    Napi::HandleScope scope(env);
+
+    if (info.Length()!=1 || !info[0].IsFunction()) {
```
mewim, line 50: Consider remove these checks in C++ and do these checks in JS, so that if the input is not valid, just terminate early instead of propagating to C++.

tools/nodejs_api/src_cpp/node_query_result.cpp
```diff
@@ -0,0 +1,107 @@
...
+
+    Napi::Function t = DefineClass(env, "NodeQueryResult", {
+       InstanceMethod("close", &NodeQueryResult::Close),
+       InstanceMethod("all", &NodeQueryResult::All),
+       InstanceMethod("getNext", &NodeQueryResult::GetNext),
+       InstanceMethod("hasNext", &NodeQueryResult::HasNext),
+    });
+
+    constructor = Napi::Persistent(t);
+    constructor.SuppressDestruct();
```
mewim, line 22: Please double check if you should use `SuppressDestruct`. As per https://github.com/nodejs/node-addon-api/blob/main/doc/reference.md:
> `Napi::Reference` objects allocated in static space, such as a global static instance, must call the `SuppressDestruct` method to prevent its destructor, running at program shutdown time, from attempting to reset the reference when the environment is no longer valid. Avoid using this if at all possible.

I don't think this is the correct use case.

tools/nodejs_api/src_cpp/node_query_result.cpp
```diff
@@ -0,0 +1,107 @@
...
+    Napi::HandleScope scope(env);
+
+    Napi::Function t = DefineClass(env, "NodeQueryResult", {
+       InstanceMethod("close", &NodeQueryResult::Close),
+       InstanceMethod("all", &NodeQueryResult::All),
+       InstanceMethod("getNext", &NodeQueryResult::GetNext),
+       InstanceMethod("hasNext", &NodeQueryResult::HasNext),
+    });
+
+    constructor = Napi::Persistent(t);
```
mewim, line 21: Please double check if you need a persistent reference: https://github.com/nodejs/node-addon-api/blob/main/doc/object_reference.md

tools/nodejs_api/src_cpp/include/node_connection.h
```diff
@@ -13,5 +13,5 @@ class NodeConnection : public Napi::ObjectWrap<NodeConnection> {
   Napi::Value Execute(const Napi::CallbackInfo& info);
   void SetMaxNumThreadForExec(const Napi::CallbackInfo& info);
   Napi::Value GetNodePropertyNames(const Napi::CallbackInfo& info);
-  unique_ptr<kuzu::main::Connection> connection;
+  shared_ptr<kuzu::main::Connection> connection;
```
mewim, line 16: Why do you need a `shared_ptr` here? Why `unique_ptr` would not work?

tools/nodejs_api/src_cpp/include/each_single_async_worker.h
```diff
@@ -0,0 +1,21 @@
+#pragma once
+#include <napi.h>
+#include "main/kuzu.h"
+
+using namespace Napi;
+
+class EachSingleAsyncWorker : public AsyncWorker {
```
mewim, line 7: Change name to `EachAsyncWorker`

tools/nodejs_api/CMakeLists.txt
```diff
@@ -43,3 +43,12 @@ file(COPY ${CMAKE_JS_SRC} DESTINATION ./kuzu)
 add_library(${PROJECT_NAME} SHARED ${SOURCE_FILES} ${CMAKE_JS_SRC})
 set_target_properties(${PROJECT_NAME} PROPERTIES PREFIX "" SUFFIX ".node")
 target_link_libraries(${PROJECT_NAME} PRIVATE ${CMAKE_JS_LIB} kuzu)
+
+# Include Node-API wrappers
+execute_process(COMMAND node -p "require('node-addon-api').include"
+        WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}
+        OUTPUT_VARIABLE NODE_ADDON_API_DIR
+        )
+string(REPLACE "\n" "" NODE_ADDON_API_DIR ${NODE_ADDON_API_DIR})
```
mewim, line 52: Please do not try to minify the scripts in this way. If we need to do this, probably we should use webpack instead of doing this manually.

tools/nodejs_api/src_cpp/all_async_worker.cpp
```diff
@@ -0,0 +1,43 @@
...
+            i++;
+        }
+    } catch(const std::exception &exc) {
+        SetError("Unsuccessful async All/Each callback: " + std::string(exc.what()));
+    }
+};
+
+void AllAsyncWorker::OnOK() {
+    Napi::Array arr = Napi::Array::New(Env(), allResult.size());
+    for (size_t i=0; i < allResult.size(); i++) {
```
mewim, line 34: I think it is actually possible to move this work from the main thread with `ThreadSafeFunction` (https://github.com/nodejs/node-addon-api/blob/main/doc/threadsafe_function.md). Please consider using it instead of performing the conversions in `OnOK`.

tools/nodejs_api/src_cpp/each_single_async_worker.cpp
```diff
@@ -0,0 +1,35 @@
...
+            rowResult.emplace_back(
+                make_unique<kuzu::common::Value>(kuzu::common::Value(*row->getValue(j))));
+        }
+    } catch(const std::exception &exc) {
+        SetError("Unsuccessful async Each callback: " + std::string(exc.what()));
+    }
+}
+
+void EachSingleAsyncWorker::OnOK() {
+    Napi::Array rowArray = Napi::Array::New(Env(), rowResult.size() + 1);
```
mewim, line 28: I think it is actually possible to move this work from the main thread with `ThreadSafeFunction` (https://github.com/nodejs/node-addon-api/blob/main/doc/threadsafe_function.md). Please consider using it instead of performing the conversions in `OnOK`.

tools/nodejs_api/src_cpp/util.cpp
mewim: Also, add support for `NODE` and `REL`

mewim (changes requested):

tools/nodejs_api/src_cpp/include/node_connection.h
ynoza: Because I'm passing the connection to the ExecuteWorker and still need the connection class itself to have access to connection for future executions.

tools/nodejs_api/src_cpp/node_query_result.cpp
ynoza: I think it's better to have it both places so I added it to the js side and still have it here - lemme know if this is a bad approach (i think it improves fault tolerance).

mewim (approved):

PR #1290, opened 2023-02-15
Decouple group pos between operator and planner

andyfengHKU:


src/include/planner/logical_plan/logical_operator/schema.h
```diff
@@ -130,6 +131,16 @@ class SchemaUtils {
 public:
     static std::vector<binder::expression_vector> getExpressionsPerGroup(
         const binder::expression_vector& expressions, const Schema& schema);
+    // Given a set of factorization group, a leading group is selected as the unFlat group (caller
+    // should ensure at most one unFlat group which is our general assumption of factorization). If
+    // all groups are flat, we select any (the first) group as leading group.
+    static f_group_pos getLeadingGroupPos(
```
ray6080, line 137: I remember we used the term "PrimaryGroup" previously? (or that was actually never used?)

ray6080 (approved):

src/include/planner/logical_plan/logical_operator/schema.h
andyfengHKU: I think it was electing a primary expression within a group. Though we can change this function name easily in future.

PR #1299, opened 2023-02-17
Single buffer manager with multiple size classes

ray6080:
This is the first PR towards the goal that our BM can allocate pages with multiple size classes, e.g., 4KB, 8KB, ..., 1GB, etc, that are backed by a single memory pool.

This PR changes internal implementation of BM, while still keeps its interfaces unchanged, and also page sizes.
**Only 4KB and 256KB are two supported size classes in this PR**, which will be extended to more size classes in the following PR.

**MemoryManager**
Inside MM, there are multiple MemoryAllocator, each is responsible for allocating/reclaiming memory blocks of its size class (backed by a temp in-mem file) from the BM. Same page reuse logic is implemented to avoid allocating new pages when possible.

**BufferManager**
Inside BM, there are multiple BufferPools, each corresponds to a collection of frames with a fixed page size class.
The buffer pool is backed by one or more mmap regions (virtual memory space). There can be more than one region due to that `mmap()` has length limit of `SIZE_T_MAX`. 

During construction, the buffer pool maps its frames to its own allocated regions, and leaves the responsibility of handling memory fragmentation to the OS.
Pin and Unpins will still go through FileHandle. The locking and mapping from page id to frame id logic remains unchanged, but the eviction strategy is now changed to queue-based so it can adapt to variable-sized frames easily.
See comments above `claimAFrame()` for details.

The physical memory usage is tracked by an atomic `usedMemory`, which guarantees the memory usage is always below what user specifies `maxSize`. usedMemory is incremented when we fill an empty frame and decremented when we actually remove a page out of a frame.

Thanks to Fadhil Abubaker (@fabubaker) who helped us with [the initial prototype](https://github.com/fabubaker/kuzu) on the mmap idea in his course project with Semih.


benchmark/queries/ldbc-sf100/aggregation/q28.benchmark
```diff
@@ -1,9 +0,0 @@
--NAME q28
```
ray6080, line 1: This will be reverted in this pr.

semihsalihoglu-uw (changes requested):
As we discussed, there is a bug in this code and we should just take our time on this design. We might want to implement two different ways to manage the evictionQueue.

1. Global Eviction Queue for all buffer pools of all sizes: and unpinned frames move forward in the queue and pinned pages go to the beginning of the queue.  This would require accessing a global lock but we would have a very homogeneous way of managing pages.
2. Do Gclock on different bufferpools. That is pick a buffer pool (so fix a size) that has enough available memory and try to evict from that level.  This looks simpler because we have this code already and we trust the code. And it is probably better in terms of synchronization. In both cases, we would try to find enough evictable frames from the layer. Frankly, this is what I would start with and my strategy would be to order the buffer pools in terms of their evictable memory. For each level that has enough memory that what we need, pick it with probability proportional to its available memory. So for example, suppose we need to evict 128KB of memory and there were two buffer pools with evictable memory > 128KB. Suppose one stores 4KB-size pages and has 10MB evictable memory, and the other stores 256KB pages and has 5MB evictable memory. Then you can pick 256KB pages with 0.33 probability and 4KB-size pages with 66% probability. Let me know if this makes sense.    

I will try to read the LeanStore paper to better understand what they are doing because I'm curious.

ray6080:
The design has shifted, I should open a new PR instead of continuing on this one.

PR #1305, opened 2023-02-21
Add Python API binding for prepared statement

mewim:
This PR adds binding for prepared statement to Python API. It exposes the `prepare` function to Python API and adds `PreparedStatement` class to Python. The execute is modified accordingly to support both query string and prepared statement as a parameter. If a query string is provided, it will be prepared automatically.

Additionally, this PR adds support for interval data type (`timedelta` in Python) as a query parameter for Python execute.

### DIFF
--- a/tools/python_api/src_cpp/py_prepared_statement.cpp
+++ b/tools/python_api/src_cpp/py_prepared_statement.cpp
@@ -1,0 +1,18 @@
+#include "include/py_prepared_statement.h"
+
+#include "binder/binder.h"
+#include "planner/logical_plan/logical_plan.h"
+
+void PyPreparedStatement::initialize(py::handle& m) {
+    py::class_<PyPreparedStatement>(m, "PreparedStatement")
+        .def("get_error_message", &PyPreparedStatement::getErrorMessage)
+        .def("is_success", &PyPreparedStatement::isSuccess);
+}
+
+py::str PyPreparedStatement::getErrorMessage() const {
+    return preparedStatement->getErrorMessage();
+}
+
+bool PyPreparedStatement::isSuccess() const {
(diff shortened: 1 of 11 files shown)


tools/python_api/src_cpp/py_prepared_statement.cpp
```diff
@@ -0,0 +1,23 @@
...
+
+py::str PyPreparedStatement::getErrorMessage() const {
+    return preparedStatement->getErrorMessage();
+}
+
+bool PyPreparedStatement::isSuccess() const {
+    return preparedStatement->isSuccess();
+}
+
+bool PyPreparedStatement::isReadOnly() const {
```
andyfengHKU, line 21: let's not expose this

andyfengHKU (approved):
LGTM

tools/python_api/src_py/prepared_statement.py
```diff
@@ -0,0 +1,68 @@
...
+        connection : Connection
+            Connection to a database.
+        query : str
+            Query to prepare.
+        """
+
+        self._prepared_statement = connection._connection.prepare(query)
+        # Keep a reference to the connection to prevent make sure the connection
+        # is not garbage collected before the prepared statement.
+        self.connection = connection
```
andyfengHKU, line 32: As we discussed, this may not be needed.

tools/python_api/src_cpp/py_prepared_statement.cpp
mewim: Removed

tools/python_api/src_py/prepared_statement.py
mewim: I removed this and the tests still pass, so this is not needed

PR #1315, opened 2023-02-23
Fix #998

### LINKED ISSUE #998: Handle large lists in Intersect
Large lists are not handled yet in our Intersect.
IntersectBuild operator should correctly append and sort large lists altogether.

As you add dataset with large list, also test multi-labeled extend

ray6080:
This PR adds the cartesian product logic to handle chunked large lists in intersect.
Notice: partially caching intermediate intersect results can further improve the performance when there are multiple large lists to be intersected at the same time, which will be added later.

### DIFF
--- a/src/processor/operator/intersect/intersect.cpp
+++ b/src/processor/operator/intersect/intersect.cpp
@@ -30,20 +30,23 @@ void Intersect::initLocalStateInternal(ResultSet* resultSet, ExecutionContext* c
     }
 }
 
-std::vector<uint8_t*> Intersect::probeHTs(const std::vector<nodeID_t>& keys) {
-    std::vector<uint8_t*> tuples(keys.size());
-    hash_t tmpHash;
-    for (auto i = 0u; i < keys.size(); i++) {
-        Hash::operation<nodeID_t>(keys[i], false, tmpHash);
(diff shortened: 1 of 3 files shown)


src/processor/operator/intersect/intersect.cpp
```diff
@@ -160,17 +150,60 @@ void Intersect::populatePayloads(
...
+            std::fill(tupleIdxPerBuildSide.begin(), tupleIdxPerBuildSide.end(), 0);
         }
-        auto tuples = probeHTs(getProbeKeys());
-        auto listsToIntersect = fetchListsToIntersectFromTuples(tuples, isIntersectListAFlatValue);
+        // Cartesian product of all flat tuples probed from all build sides.
+        // Notice: when there are large adjacency lists in the build side, which means the list is
+        // too large to fit in a single ValueVector, we end up chunking the list as multiple tuples
+        // in FTable. Thus, when performing the intersection, we need to perform cartesian product
+        // between all flat tuples probed from all build sides.
+        auto numBuildSides = sharedHTs.size();
```
andyfengHKU, line 193: you are using sharedHTs.size() in multiple places. Better to create an inline function

andyfengHKU (approved):

src/processor/operator/intersect/intersect.cpp
ray6080: Done.

PR #1327, opened 2023-02-28
Sequential read for StringPropertyList

rfdavid:
This PR implements overflow cache in the `DiskOverflowFile::readStringToVector` function in the same way that was implemented in `scanSequentialStringOverflow.`, so it can be reused in different parts of the code.

## Benchmark

- Node table with two fields STRING and INT64 `CREATE NODE TABLE City(name STRING, population INT64, PRIMARY KEY (name))`
- total 986125 rows
- `name` string has ~ 25kb for each node property
- Processor: M2 Max, 32GB ram

Results of `SCAN_NODE_PROPERTY` after running `PROFILE MATCH (c:City) RETURN c.name;` 5 times on master branch and on this branch:

| Run  |  1 | 2  | 3  | 4  | 5 |
|---|---|---|---|---|---|
| Master | 37181.326 | 1571.216 | 1405.098 | 1409.182 | 1417.050
| Branch | 37127.368 | 1562.977 | 1395.112 | 1518.187 | 1394.889

Related to #756 (note: this only addresses string overflow, not lists)

I have read and agree to the terms under [CLA.md](https://github.com/kuzudb/kuzu/blob/master/CONTRIBUTING.md)

### DIFF
--- a/src/include/storage/storage_structure/column.h
+++ b/src/include/storage/storage_structure/column.h
@@ -123,12 +123,12 @@ private:
     inline void scan(transaction::Transaction* transaction, common::ValueVector* resultVector,
         PageElementCursor& cursor) override {
         Column::scan(transaction, resultVector, cursor);
-        diskOverflowFile.scanSequentialStringOverflow(transaction->getType(), *resultVector);
+        diskOverflowFile.scanStrings(transaction->getType(), *resultVector);
     }
     void scanWithSelState(transaction::Transaction* transaction, common::ValueVector* resultVector,
         PageElementCursor& cursor) override {
         Column::scanWithSelState(transaction, resultVector, cursor);
-        diskOverflowFile.scanSequentialStringOverflow(transaction->getType(), *resultVector);
+        diskOverflowFile.scanStrings(transaction->getType(), *resultVector);
     }
 };
 
--- a/src/include/storage/storage_structure/disk_overflow_file.h
+++ b/src/include/storage/storage_structure/disk_overflow_file.h
@@ -36,19 +36,14 @@ public:
         return copy;
     }
 
-    void readStringsToVector(
-        transaction::TransactionType trxType, common::ValueVector& valueVector);
-    void readStringToVector(transaction::TransactionType trxType, common::ku_string_t& kuStr,
-        common::InMemOverflowBuffer& inMemOverflowBuffer);
+    void scanStrings(transaction::TransactionType trxType, common::ValueVector& valueVector);
 
     inline void scanSingleStringOverflow(
         transaction::TransactionType trxType, common::ValueVector& vector, uint64_t vectorPos) {
         assert(vector.dataType.typeID == common::STRING && !vector.isNull(vectorPos));
         auto& kuString = ((common::ku_string_t*)vector.getData())[vectorPos];
-        readStringToVector(trxType, kuString, vector.getOverflowBuffer());
+        lookupString(trxType, kuString, vector.getOverflowBuffer());
     }
-    void scanSequentialStringOverflow(
-        transaction::TransactionType trxType, common::ValueVector& vector);
     inline void scanSingleListOverflow(
         transaction::TransactionType trxType, common::ValueVector& vector, uint64_t vectorPos) {
         assert(vector.dataType.typeID == common::VAR_LIST && !vector.isNull(vectorPos));
@@ -77,6 +72,15 @@ public:
     }
 
 private:
+    struct OverflowPageCache {
+        BufferManagedFileHandle* bufferManagedFileHandle = nullptr;
+        common::page_idx_t pageIdx = UINT32_MAX;
+        uint8_t* frame = nullptr;
+    };
+    void lookupString(transaction::TransactionType trxType, common::ku_string_t& kuStr,
+        common::InMemOverflowBuffer& inMemOverflowBuffer);
+    void lookupString(transaction::TransactionType trxType, common::ku_string_t& kuStr,
+        common::InMemOverflowBuffer& inMemOverflowBuffer, OverflowPageCache& overflowPageCache);
     void addNewPageIfNecessaryWithoutLock(uint32_t numBytesToAppend);
     void readListToVector(transaction::TransactionType trxType, common::ku_list_t& kuList,
         const common::DataType& dataType, common::InMemOverflowBuffer& inMemOverflowBuffer);
@@ -85,6 +89,9 @@ private:
     void setListRecursiveIfNestedWithoutLock(const common::ku_list_t& inMemSrcList,
         common::ku_list_t& diskDstList, const common::DataType& dataType);
     void logNewOverflowFileNextBytePosRecordIfNecessaryWithoutLock();
+    void pinOverflowPageCache(BufferManagedFileHandle* bufferManagedFileHandleToPin,
+        common::page_idx_t pageIdxToPin, OverflowPageCache& overflowPageCache);
+    void unpinOverflowPageCache(OverflowPageCache& overflowPageCache);
 
 private:
     // This is the index of the last free byte to which we can write.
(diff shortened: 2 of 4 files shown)


src/include/storage/storage_structure/column.h
```diff
@@ -134,13 +134,13 @@ class StringPropertyColumn : public PropertyColumnWithOverflow {
         const std::shared_ptr<common::ValueVector>& resultVector,
         PageElementCursor& cursor) override {
         Column::scan(transaction, resultVector, cursor);
-        diskOverflowFile.scanSequentialStringOverflow(transaction->getType(), *resultVector);
+        diskOverflowFile.readStringsToVector(transaction->getType(), *resultVector);
```
andyfengHKU, line 137: let's rename the function to scanStrings()

src/include/storage/storage_structure/disk_overflow_file.h
```diff
@@ -37,19 +37,25 @@ class DiskOverflowFile : public StorageStructure {
         return copy;
     }
 
+    struct OverflowCache {
```
andyfengHKU, line 40: Rename to OverflowPageCache

src/include/storage/storage_structure/disk_overflow_file.h
```diff
@@ -37,19 +37,25 @@ class DiskOverflowFile : public StorageStructure {
...
+
     void readStringsToVector(
         transaction::TransactionType trxType, common::ValueVector& valueVector);
     void readStringToVector(transaction::TransactionType trxType, common::ku_string_t& kuStr,
-        common::InMemOverflowBuffer& inMemOverflowBuffer);
+        common::InMemOverflowBuffer& inMemOverflowBuffer, OverflowCache& overflowCache);
 
     inline void scanSingleStringOverflow(
         transaction::TransactionType trxType, common::ValueVector& vector, uint64_t vectorPos) {
+        OverflowCache overflowCache;
```
andyfengHKU, line 53: for `scanSingleStringOverflow` I think the general assumption is it's pure random so we should not try to apply cache.

src/storage/storage_structure/disk_overflow_file.cpp
```diff
@@ -11,20 +11,28 @@ using namespace kuzu::common;
...
 
+void DiskOverflowFile::unpinOverflowCache(OverflowCache& overflowCache) {
+    if (overflowCache.pageIdx != UINT32_MAX) {
+        bufferManager.unpin(*overflowCache.fileHandle, overflowCache.pageIdx);
+    }
+}
+
 void DiskOverflowFile::readStringsToVector(TransactionType trxType, ValueVector& valueVector) {
     assert(!valueVector.state->isFlat());
+    OverflowCache overflowCache;
```
andyfengHKU, line 22: I'm slight concerned whether this object creation will cause overhead. Usually the compiler optimize these things so it shouldn't matter. 

But can u try to benchmark a bit to see if we become slower if we wrap things as a struct? If we do become slower I would use three primitive types instead.

src/storage/storage_structure/disk_overflow_file.cpp
```diff
@@ -33,51 +41,16 @@ void DiskOverflowFile::readStringToVector(
...
-void DiskOverflowFile::scanSequentialStringOverflow(TransactionType trxType, ValueVector& vector) {
-    FileHandle* cachedFileHandle = nullptr;
-    page_idx_t cachedPageIdx = UINT32_MAX;
-    uint8_t* cachedFrame = nullptr;
-    for (auto i = 0u; i < vector.state->selVector->selectedSize; ++i) {
-        auto pos = vector.state->selVector->selectedPositions[i];
-        if (vector.isNull(pos)) {
-            continue;
+    if (pageIdxToPin != overflowCache.pageIdx) { // cache miss
+        if (overflowCache.pageIdx != UINT32_MAX) {
```
andyfengHKU, line 45: call unpinOverflowCache instead

andyfengHKU (changes requested):

src/include/storage/storage_structure/disk_overflow_file.h
```diff
@@ -37,19 +37,25 @@ class DiskOverflowFile : public StorageStructure {
...
+        FileHandle* fileHandle = nullptr;
+        common::page_idx_t pageIdx = UINT32_MAX;
+        uint8_t* frame = nullptr;
+    };
+
     void readStringsToVector(
         transaction::TransactionType trxType, common::ValueVector& valueVector);
     void readStringToVector(transaction::TransactionType trxType, common::ku_string_t& kuStr,
-        common::InMemOverflowBuffer& inMemOverflowBuffer);
+        common::InMemOverflowBuffer& inMemOverflowBuffer, OverflowCache& overflowCache);
```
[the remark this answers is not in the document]
rfdavid: that's a very good point! overloading `lookupString` so we can use it in `scanStringOverflow` without trying to cache (which is only additional overhead since it won't be triggered).

src/storage/storage_structure/disk_overflow_file.cpp
rfdavid: great catch

src/include/storage/storage_structure/disk_overflow_file.h
```diff
@@ -37,19 +37,24 @@ class DiskOverflowFile : public StorageStructure {
         return copy;
     }
 
-    void readStringsToVector(
-        transaction::TransactionType trxType, common::ValueVector& valueVector);
-    void readStringToVector(transaction::TransactionType trxType, common::ku_string_t& kuStr,
+    struct OverflowPageCache {
```
ray6080, line 40: Can you move this struct to private? I guess it doesn't need to be exposed as public for now.

src/storage/storage_structure/disk_overflow_file.cpp
```diff
@@ -39,45 +47,24 @@ void DiskOverflowFile::readStringToVector(
...
+    PageByteCursor cursor;
+    TypeUtils::decodeOverflowPtr(kuStr.overflowPtr, cursor.pageIdx, cursor.offsetInPage);
+    auto [fileHandleToPin, pageIdxToPin] =
+        StorageStructureUtils::getFileHandleAndPhysicalPageIdxToPin(
+            fileHandle, cursor.pageIdx, *wal, trxType);
+    if (pageIdxToPin != overflowPageCache.pageIdx) { // cache miss
+        unpinOverflowPageCache(overflowPageCache);
+        overflowPageCache.frame = bufferManager.pin(*fileHandleToPin, pageIdxToPin);
+        overflowPageCache.fileHandle = fileHandleToPin;
+        overflowPageCache.pageIdx = pageIdxToPin;
```
ray6080, line 64: I would consider wrapping these three lines as `pinOverflowPageCache`, but i don't have a strong opinion on this, up to you.
```
        overflowPageCache.frame = bufferManager.pin(*fileHandleToPin, pageIdxToPin);
        overflowPageCache.fileHandle = fileHandleToPin;
        overflowPageCache.pageIdx = pageIdxToPin;
```

ray6080 (commented):
Thanks! Looks good to me.
Just two minor comments:
1. you wrote in the description `name string has ~ 25kb for each node property`, did you mean 25bytes here? I don't think we can store strings whose length are 25KB right now.
2. can you add one more performance numbers comparison on lists string reading? my understanding is that we should see some performance gains in that case.

We can take another quick look, then we can get this in. Thanks!

src/include/storage/storage_structure/disk_overflow_file.h
rfdavid: Hi! Thank you so much for your comments.

1. You are right, I actually created a csv with a very long string, but after querying, I realized it was stored only 4096 bytes.
2. I'll benchmark it soon. I believe the numbers should be better, considering the cache is implemented for the child type STRING (unless the object creating is causing some overhead, as mentioned by @andyfengHKU)

src/storage/storage_structure/disk_overflow_file.cpp
rfdavid: I think it makes sense since I already have `unpinOverflowPageCache`, which definitely increases readability. thanks

andyfengHKU (approved):

Issue #1338, opened 2023-03-02
Remove CSV reader


andyfengHKU:
Since we are using the CSV reader from arrow. We should no longer maintain our own CSV reader.

mewim:
In #1341, I have removed the `CSVReader` class and renamed all `csv_reader` to `csv_reader_config`. @ray6080 Please double check if `NodeStructuredStringPropertyTest` is still needed and also consider if we should move the `CSVReaderConfig` to another place.

PR #1341, opened 2023-03-03
Remove csv reader

mewim:
#1338


andyfengHKU (approved):
I think the copy test was meant to test CSVReader class but I might be wrong. Since we are removing CSVReader, can u open an issue and assign to @ray6080 to go back and see if we can remove this test.

In the issue also mention we should either rename cdv_reader.h and cpp or move the CSVReaderConfig somewhere else and delete the file.

mewim:
I did not close #1338 and added a note in it.

PR #1357, opened 2023-03-07
Flat schema

andyfengHKU:
- Rename `computeSchema` to `computeFactorizedSchema`
- Add `computeFlatSchema` so that operators resolve its input schema from child. As a result, we no longer maintain `expressionsToMaterialize` as a field in sink operators.


ray6080 (approved):
Looks good. Let's get this in!

PR #1365, opened 2023-03-10
Separate FileHandle and BufferManagedFileHandle

ray6080:
This PR separates `FileHandle` and `BufferManagedFileHandle`.
The former one is used for files that are **not** managed by BM, and the latter one is for ones managed by BM.

- `BufferManagedFileHandle` supports both versioned and non-versioned file handles.
- Page resize logic now are all unified to grow one page group (1024 pages) at one time (grow means page capacity will increase 1024 at one time, not actual num of pages).
- `StorageStructureID` is moved from FileHandle to `StorageStructure`. `StorageStructureID` is mostly used to logically identify the semantic of a storage structure, for example, a node table column. FileHandle shouldn't be aware of those information.


src/include/storage/storage_structure/lists/list_headers.h
```diff
@@ -124,7 +123,7 @@ class ListHeaders : public BaseListHeaders {
         return (*headersDiskArray)[offset];
     };
 
-    inline VersionedFileHandle* getFileHandle() { return versionedFileHandle.get(); }
+    inline FileHandle* getFileHandle() { return fileHandle.get(); }
```
andyfengHKU, line 126: remove if not needed

src/include/storage/wal_replayer.h
```diff
@@ -15,15 +14,16 @@ namespace storage {
...
     // This interface is used for recovery only. We always recover the disk files before
     // constructing the storageManager and catalog. So this specialized recovery constructor
     // doesn't take in storageManager and bufferManager.
-    WALReplayer(WAL* wal);
+    explicit WALReplayer(WAL* wal);
 
-    WALReplayer(WAL* wal, StorageManager* storageManager, BufferManager* bufferManager,
-        MemoryManager* memoryManager, catalog::Catalog* catalog, bool isCheckpoint);
+    WALReplayer(WAL* wal, StorageManager* storageManager, MemoryManager* memoryManager,
+        catalog::Catalog* catalog, bool isCheckpoint, bool isRecovering = false);
```
andyfengHKU, line 26: Unless you are setting isRecovering to True in any case, don't expose isRecovering as a default argument.

andyfengHKU (approved):

src/include/storage/storage_structure/lists/list_headers.h
ray6080: Done.

src/include/storage/wal_replayer.h
ray6080: Done.

Issue #1098, opened 2022-12-07
Multi-label cyclic query bug


andyfengHKU:
- [ ] `test/test_files/tinySNB/cyclic/multi_label.test`
- [ ] `MATCH (a:person)-[e1:knows]->(b:person)-[e2:knows]->(c:person), (a)-[e3:knows]->(c) WHERE a.fName='Alice' AND e1.meetTime=timestamp('1986-10-21 21:08:31.521') RETURN b.fName, c.fName`

Issue #1371, opened 2023-03-13
Incorrect handling of multiple keys in the same chunk when inserting to join hash table.


andyfengHKU:
The following query will output incorrect result
```
MATCH (a:person)-[e1:knows]->(b:person), (a)-[e2:meets]->(b) WHERE a.age > 0 RETURN a.fName, b.fName
```
This is because e2 is a column extend and thus both keys "a" and "b" are in the same data chunk. However, our join hash table seems to unable to differentiate whether "a" and "b" are in the same chunk or not and thus always taking a Cartesian product.

PR #1384, opened 2023-03-17
Factorization aware dp

andyfengHKU:
This PR changes the set of plans that kept in DP table during enumeration. Previous, we only keep plans with top K (K=50) lowest cost. This PR keeps plans with top 1 lowest cost for each different factorization structure.

The rationale is 
```
Consider a triangle + an edge
MATCH (a)->(b)->(c), (a)->(c), (c)->(d)
First it make sense to close the cycle as early as possible, so (a)->(b)->(c), (a)->(c)  will be handled first with WCOJ. The output factorization structures
[a, b] * [c] , [b, c] * [a] , [a, c] * [b]  (where the second group is unFlat) are considered as homogeneous. So any one could be picked in the DP level 3 for subgraph `(a)->(b)->(c), (a)->(c)`
But then at the next level, there is a clear disadvantage for [a, b] * [c] because further extension undermines the factorization.
```

This PR also fixes issue #1098 and #1371

### DIFF
--- a/src/include/binder/query/reading_clause/query_graph.h
+++ b/src/include/binder/query/reading_clause/query_graph.h
@@ -12,6 +12,10 @@ const uint8_t MAX_NUM_VARIABLES = 64;
 
 class QueryGraph;
 struct SubqueryGraph;
+struct SubqueryGraphHasher;
+using subquery_graph_set_t = std::unordered_set<SubqueryGraph, SubqueryGraphHasher>;
+template<typename T>
+using subquery_graph_V_map_t = std::unordered_map<SubqueryGraph, T, SubqueryGraphHasher>;
 
 // hash on node bitset if subgraph has no rel
 struct SubqueryGraphHasher {
@@ -51,7 +55,7 @@ struct SubqueryGraph {
 
     std::unordered_set<uint32_t> getNodeNbrPositions() const;
     std::unordered_set<uint32_t> getRelNbrPositions() const;
-    std::unordered_set<SubqueryGraph, SubqueryGraphHasher> getNbrSubgraphs(uint32_t size) const;
+    subquery_graph_set_t getNbrSubgraphs(uint32_t size) const;
     std::vector<uint32_t> getConnectedNodePos(const SubqueryGraph& nbr) const;
 
     // E.g. query graph (a)-[e1]->(b) and subgraph (a)-[e1], although (b) is not in subgraph, we
@@ -65,9 +69,8 @@ struct SubqueryGraph {
     }
 
 private:
-    std::unordered_set<SubqueryGraph, SubqueryGraphHasher> getBaseNbrSubgraph() const;
-    std::unordered_set<SubqueryGraph, SubqueryGraphHasher> getNextNbrSubgraphs(
-        const SubqueryGraph& prevNbr) const;
+    subquery_graph_set_t getBaseNbrSubgraph() const;
+    subquery_graph_set_t getNextNbrSubgraphs(const SubqueryGraph& prevNbr) const;
 };
 
 // QueryGraph represents a connected pattern specified in MATCH clause.
--- a/src/include/planner/logical_plan/logical_operator/schema.h
+++ b/src/include/planner/logical_plan/logical_operator/schema.h
@@ -7,9 +7,15 @@
 namespace kuzu {
 namespace planner {
 
-typedef uint32_t f_group_pos;
-typedef std::unordered_set<f_group_pos> f_group_pos_set;
+using f_group_pos = uint32_t;
+using f_group_pos_set = std::unordered_set<f_group_pos>;
 constexpr f_group_pos INVALID_F_GROUP_POS = UINT32_MAX;
+
+class Schema;
+struct SchemaHasher;
+struct SchemaApproximateEquality;
+template<typename T>
+using schema_map_t = std::unordered_map<Schema*, T, SchemaHasher, SchemaApproximateEquality>;
 
 class FactorizationGroup {
     friend class Schema;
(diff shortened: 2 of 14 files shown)


src/include/planner/logical_plan/logical_operator/schema.h
```diff
@@ -57,7 +63,9 @@ class FactorizationGroup {
 
 class Schema {
 public:
-    inline f_group_pos getNumGroups() const { return groups.size(); }
+    inline size_t getNumGroups() const { return groups.size(); }
+    inline size_t getNumFlatGroups() const { return getNumGroups(true /* isFlat */); }
+    size_t getNumUnFlatGroups() const { return getNumGroups(false /* isFlat */); }
```
andyfengHKU, line 68: inline

src/processor/operator/hash_join/join_hash_table.cpp
```diff
@@ -42,9 +42,12 @@ void JoinHashTable::append(const std::vector<ValueVector*>& vectorsToAppend) {
     // TODO(Guodong): use compiling information to remove the for loop.
     auto numTuplesToAppend = 1;
     for (auto i = 0u; i < numKeyColumns; i++) {
-        numTuplesToAppend *= (vectorsToAppend[i]->state->isFlat() ?
-                                  1 :
-                                  vectorsToAppend[i]->state->selVector->selectedSize);
+        // At most one unFlat key data chunk. If there are multiple unFlat key vectors, they must
+        // share the same state.
+        if (!vectorsToAppend[i]->state->isFlat()) {
```
andyfengHKU, line 47: This logic is correct but still garbage in the sense that this should be a shared logic implemented in factorized table. Go back and update comment and issue.

src/processor/operator/intersect/intersect.cpp
```diff
@@ -17,9 +17,7 @@ void Intersect::initLocalStateInternal(ResultSet* resultSet, ExecutionContext* c
         for (auto i = 0u; i < dataInfo.payloadsDataPos.size(); i++) {
             auto vector = resultSet->getValueVector(dataInfo.payloadsDataPos[i]);
             // Always skip the first two columns in the fTable: build key and intersect key.
-            // TODO(Guodong): this is a potential bug because you cannot guarantee intersect key is
-            // the second column. Once this is solved, go back and refactor projection push down for
-            // intersect.
+            // TODO(Guodong): this is a bad practice. This assumption requires
```
andyfengHKU, line 20: Update comment

src/include/binder/query/reading_clause/query_graph.h
```diff
@@ -12,6 +12,10 @@ const uint8_t MAX_NUM_VARIABLES = 64;
 
 class QueryGraph;
 struct SubqueryGraph;
+struct SubqueryGraphHasher;
+using sub_query_graph_set_t = std::unordered_set<SubqueryGraph, SubqueryGraphHasher>;
```
ray6080, line 16: `sub_query_graph_set_t` -> `subquery_graph_set_t`

src/include/binder/query/reading_clause/query_graph.h
```diff
@@ -12,6 +12,10 @@ const uint8_t MAX_NUM_VARIABLES = 64;
...
+template<typename T>
+using sub_query_graph_map_t = std::unordered_map<SubqueryGraph, T, SubqueryGraphHasher>;
```
ray6080, line 18: `sub_query_graph_map_t` -> `subquery_graph_map_t`

src/planner/operator/schema.cpp
```diff
@@ -99,6 +99,43 @@ void Schema::clear() {
...
+}
+
+std::size_t SchemaHasher::operator()(const Schema* const& schema) const {
+    return std::hash<size_t>{}(schema->getNumFlatGroups()) ^
+           std::hash<size_t>{}(schema->getNumUnFlatGroups());
+}
+
+// We use this equality in join order enumeration to make sure at each DP level, we don't just keep
+// the best plan, but keep best plan for each unique factorization schema.
+// In order to balance enumeration time, we use an approximate equality check to reduce computation.
```
ray6080, line 117: - Should we also check the equality of num of expressions in the left and right schemas?
- Can you add comments on the approximate equal logic?
```
Two schemas (left and right) are considered approximately equal, if:
1) The left schema and the right schema have same number of flat and unflat groups.
2) Any expression in the left schema also exists in the right schema, and its corresponding factorization groups in the left and right schemas are both flat or both unflat.
```

src/include/planner/subplans_table.h
```diff
@@ -13,30 +13,43 @@ namespace kuzu {
...
     void addPlan(const SubqueryGraph& subqueryGraph, std::unique_ptr<LogicalPlan> plan);
-    void finalizeLevel(uint32_t level);
 
     void clear();
 
 private:
-    std::vector<std::unique_ptr<SubqueryGraphPlansMap>> subPlans;
+    struct PlanSet {
+        logical_plan_vector_t plans;
+        schema_map_t<common::vector_idx_t> schemaToPlanIdx;
```
ray6080, line 40: This is the part i'm a bit confused. 
```
        logical_plan_vector_t plans;
        schema_map_t<common::vector_idx_t> schemaToPlanIdx;
```
Why not just keep a map from schema to logical plan? so you don't need to keep a vector separately.

ray6080 (approved):

src/include/planner/logical_plan/logical_operator/schema.h
```diff
@@ -7,10 +7,16 @@
 namespace kuzu {
 namespace planner {
 
-typedef uint32_t f_group_pos;
-typedef std::unordered_set<f_group_pos> f_group_pos_set;
+using f_group_pos = uint32_t;
+using f_group_pos_set = std::unordered_set<f_group_pos>;
```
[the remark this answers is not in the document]
andyfengHKU: As we discussed, we will not change this.

src/include/planner/logical_plan/logical_plan.h
```diff
@@ -5,6 +5,9 @@
 namespace kuzu {
 namespace planner {
 
+class LogicalPlan;
+using logical_plan_vector_t = std::vector<std::unique_ptr<LogicalPlan>>;
```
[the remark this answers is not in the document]
andyfengHKU: removed

src/planner/operator/schema.cpp
andyfengHKU: done

src/include/planner/subplans_table.h
andyfengHKU: because I need to return plans as a vector

PR #1398, opened 2023-03-21
Separate WAL page idx lock from page lock in BMFileHandle

ray6080:
This PR introduces `WALPageIdxGroup` in BMFileHandle, the WALPageIdxGroup keeps a set of WAL page indexes for a page group with updated pages in WAL file.
This also changes how BMFileHandle stores wal page indexes, switching from vector-based to map-based. We assume each write transaction will only touch a small number of page groups, so map-based method can be efficient.

After the change, we can now remove two interfaces, `pinWithoutAcquiringPageLock ` and `unpinWithoutAcquiringPageLock`, in BufferManager. PageLock is only used specifically by BufferManager to correctly manage pages and frames.
Any operations that require sync on accessing wal page idx of an original page should acquire the WAL page idx lock.

Additionally, this PR fixes a potential bug in `BMFileHandle::clearWALPageIdxIfNecessary`.
There was a missing lock acquire (only `releasePageLock(pageIdx);` was called) in the function, but adding lock acquiring logic can lead to dead locks in the test `SetNodeStructuredPropTransactionTest::SetVeryLongStringErrorsTest`.
The reason is writing long string overflow triggers exception thrown, which didn't correctly release locks.
See changes in `StringPropertyColumn::writeValueForSingleNodeIDPosition`.
The change is also propagated to `ListPropertyColumn::writeValueForSingleNodeIDPosition` though it has nothing to do with the failing test.


src/include/storage/buffer_manager/bm_file_handle.h
```diff
@@ -46,13 +46,42 @@ class PageState {
...
 
+// This class is used to keep the WAL page idxes of a page group in the original file handle.
+// For each original page in the page group, it has a corresponding WAL page idx. The WAL page idx
+// is initialized as INVALID_PAGE_IDX to indicate that the page does not have any updates in WAL
+// file. To synchronize accesses to the WAL version page idxes, we use a lock for each WAL page idx.
+class WALPageIdxGroup {
+public:
+    WALPageIdxGroup();
+
+    inline void acquireWALPageIdxLock(common::page_idx_t originalPageIdxInGroup) {
```
semihsalihoglu-uw, line 57: Might be good to clarify what originalPageIdxInGroup means with an example. As I understand it is the offset of the page in the pageIdxGroup, right?

src/include/storage/buffer_manager/bm_file_handle.h
```diff
@@ -114,8 +151,10 @@ class BMFileHandle : public FileHandle {
     std::vector<std::unique_ptr<PageState>> pageStates;
     // Each file page group corresponds to a frame group in the VMRegion.
     std::vector<common::page_group_idx_t> frameGroupIdxes;
-    std::vector<std::vector<common::page_idx_t>> pageVersions;
-    std::vector<std::unique_ptr<std::atomic_flag>> pageGroupLocks;
+    // For each page group, if it has any WAL page version, we keep a `WALPageIdxGroup` in this map.
+    // `WALPageIdxGroup` records the WAL page idx for each page in the page group.
+    // Accesses to this map is synchronized by `fhSharedMutex`.
+    std::unordered_map<common::page_group_idx_t, std::unique_ptr<WALPageIdxGroup>> walPageIdxGroups;
```
semihsalihoglu-uw, line 157: Not sure if this change is necessary. 
```
    std::vector<std::vector<common::page_idx_t>> pageVersions;
    std::vector<std::unique_ptr<std::atomic_flag>> pageGroupLocks;
```
These data structures are a very tiny (1/1024th only even less) fraction of the total memory footprint compared to all PageSrates, so we don't need to optimize this. Your call but overall make sure you think about these ballpark numbers before you optimize such details in code.

src/storage/buffer_manager/bm_file_handle.cpp
```diff
@@ -62,124 +67,124 @@ void BMFileHandle::addNewPageGroupWithoutLock() {
...
     if (numPageGroups == frameGroupIdxes.size()) {
         return;
     }
     assert(numPageGroups < frameGroupIdxes.size());
     frameGroupIdxes.resize(numPageGroups);
-    pageGroupLocks.resize(numPageGroups);
-    pageVersions.resize(numPageGroups);
+    for (auto& [pageGroupIdx, _] : walPageIdxGroups) {
+        if (pageGroupIdx >= numPageGroups) {
+            walPageIdxGroups.erase(pageGroupIdx);
```
semihsalihoglu-uw, line 107: I wonder if you should also check if walPageIdxGroups has become empty() and if so clear() or reconstruct it. My worry is that this map is getting inserted things and erased things continuously but we don't run any code to clear() it at some point. Not sure how badly maps can behave if they are continuously inserted into and emptied.

src/storage/buffer_manager/bm_file_handle.cpp
```diff
@@ -62,124 +67,124 @@ void BMFileHandle::addNewPageGroupWithoutLock() {
...
+        }
+    }
+    pageCapacity = numPageGroups * StorageConstants::PAGE_GROUP_SIZE;
+    // TODO(Reviewer): Should the physical file also be truncated here?
```
semihsalihoglu-uw, line 111: I dont think so. `removePageIdxAndTruncateIfNecessary` should be called only when a transaction aborted but in that case we never actually write anything to the original file (we only may have written to the wal, which gets deleted anyways).

src/storage/storage_structure/storage_structure.cpp
```diff
@@ -16,12 +16,14 @@ static inline bool isInRange(uint64_t val, uint64_t start, uint64_t end) {
...
     auto pageIdxInOriginalFile = fileHandle->addNewPage();
     auto pageIdxInWAL = wal->logPageInsertRecord(storageStructureID, pageIdxInOriginalFile);
-    bufferManager.pinWithoutAcquiringPageLock(
+    bufferManager.pin(
         *wal->fileHandle, pageIdxInWAL, BufferManager::PageReadPolicy::DONT_READ_PAGE);
-    fileHandle->createPageVersionGroupIfNecessary(pageIdxInOriginalFile);
-    fileHandle->setWALPageVersion(pageIdxInOriginalFile, pageIdxInWAL);
+    fileHandle->addWALPageIdxGroupIfNecessary(pageIdxInOriginalFile);
+    fileHandle->setWALPageIdx(pageIdxInOriginalFile, pageIdxInWAL);
+    // TODO(Reviewer): Do we need to set the page dirty here? Instead, can we only set the page
```
semihsalihoglu-uw, line 23: If you have added a brand new page to a file, that is a write operation, so that page needs to be flushed so that it persists if it gets evicted. So we should mark the page as dirty.

src/storage/storage_structure/column.cpp
```diff
@@ -189,14 +201,20 @@ void ListPropertyColumn::writeValueForSingleNodeIDPosition(
             ((ku_list_t*)(updatedPageInfoAndWALPageFrame.frame +
                           mapElementPosToByteOffset(updatedPageInfoAndWALPageFrame.posInPage)));
         auto kuListToWriteFrom = vectorToWriteFrom->getValue<ku_list_t>(posInVectorToWriteFrom);
-        diskOverflowFile.writeListOverflowAndUpdateOverflowPtr(
-            kuListToWriteFrom, *kuListToWriteTo, vectorToWriteFrom->dataType);
+        try {
```
semihsalihoglu-uw, line 204: Can u write a comment explaining what's going on here in general and why you're releasing originalPageLock if there is an exception (when was the lock acquired (I believe in beginUpdatingPage but it's not very clear)).

semihsalihoglu-uw (approved):

src/storage/storage_structure/column.cpp
ray6080: Done.

src/storage/buffer_manager/bm_file_handle.cpp
ray6080: Changed the logic to not erase while iterating, and directly clear when numPageGroups is 0.

src/include/storage/buffer_manager/bm_file_handle.h
ray6080: I agree the memory footprint is small, which is not worth optimizing. But I think using map here simplifies the logic, thus we don't need to grow the vector of WALPageIdxGroup each time we add a new page group, even without any versioned wal pages.

src/include/storage/buffer_manager/bm_file_handle.h
ray6080: Right. Added a comment.

PR #1406, opened 2023-03-23
Encode plan when inserting to dp table

andyfengHKU:
This PR removes the custom hasher for schema because they are not computational efficient. Instead, we now encode a logical plan's factorization schema into a bitset by considering the flat information of certain expressions. As a consequence

- We improve the computational time of enumeration.
- We reduce the number of plans kept in each subgraph.
- Fewer plans are being generated.


src/include/planner/subplans_table.h
```diff
@@ -13,14 +13,62 @@ namespace kuzu {
...
+    std::vector<std::unique_ptr<LogicalPlan>>& getPlans() { return plans; }
 
+private:
+    // To balance computation time, we encode plan by only considering the flat information of the
+    // nodes that are involved in current subgraph.
+    std::bitset<MAX_NUM_QUERY_VARIABLES> encodePlan(const LogicalPlan& plan);
+
+private:
+    binder::expression_vector nodeIDsToEncode;
+    constexpr static uint32_t MAX_NUM_PLANS = 10;
```
ray6080, line 43: I would separate static const values from variables.
```
private:
    constexpr static uint32_t MAX_NUM_PLANS = 10;

private:
    binder::expression_vector nodeIDsToEncode;
    std::vector<std::unique_ptr<LogicalPlan>> plans;
    std::unordered_map<std::bitset<MAX_NUM_QUERY_VARIABLES>, common::vector_idx_t> planCode2planIdx;
```

src/include/planner/subplans_table.h
```diff
@@ -13,14 +13,62 @@ namespace kuzu {
...
+    }
+
+    std::vector<SubqueryGraph> getSubqueryGraphs();
+
+    void addPlan(const SubqueryGraph& subqueryGraph, std::unique_ptr<LogicalPlan> plan);
+
+    inline void clear() { subgraph2Plans.clear(); }
+
+private:
+    constexpr static uint32_t MAX_NUM_SUBGRAPH = 50;
```
ray6080, line 67: Ditto.Separate the const value and variables.

src/include/planner/subplans_table.h
ray6080: planCode2planIdx -> encodedPlan2PlanIdx

ray6080 (approved):

PR #1409, opened 2023-03-24
BM optimistic reads

ray6080:
This PR adds `optimisticRead` interface to BM following the [VMCache design](https://www.cs.cit.tum.de/fileadmin/w00cfj/dis/_my_direct_uploads/vmcache.pdf).  Any writes to file pages should go through `pin`/`unpin`, and reads can go through `optimisticRead`, which lifts the read-read lock contentions.

### Page states and versions
Pages managed by BM now have four states: a) Locked, b) Unlocked, c) Marked, d) Evicted.
Every page is initialized as in the Evicted state. When a page is pinned, it transits to the Locked state.
When the caller finishes changes to the page, it calls `unpin`, which releases the lock on the page, thus, the page transits to the Unlocked state.
Any optimistic reads on Unlocked pages should not make any changes to the page, and reads on Marked pages will first set their states to Unlocked, and then read the page optimistically. For Evicted pages, optimistic reads will trigger pin and unpin to read pages from disk into frames.

Besides the page state, each page also has a version number. The version number is used to identify any potential writes on the page. Each time a page transits from Locked to Unlocked state, it will increment its version. When a page is pinned, then unpinned, it will transit from Locked to the Unlocked state, and its version will be incremented. During the pin and unpin, we assume the page's content in its corresponding frame might have changed, thus, we increment the version number to forbid stale reads on it.

### Eviction
We still use queue-based eviction strategy.
To track pages that can potentially be evicted, we push unlocked pages into the eviction queue, and set their states to Marked, which means they are ready to be evicted.
After the candidate was enqueued:
1) if the page's version has changed, which indicates that the page was pinned and unpinned again, the candidate should be discarded. Due to that there should be another candidate pointing to the same page in the queue.
2) if the page's state is Unlocked, but its version hasn't changed, which indicates that the page was optimistically read, the page should be set to Marked and the candidate should be moved back to the queue.

### DIFF
--- a/src/include/storage/buffer_manager/bm_file_handle.h
+++ b/src/include/storage/buffer_manager/bm_file_handle.h
@@ -6,44 +6,87 @@
 namespace kuzu {
 namespace storage {
 
-static constexpr uint64_t IS_IN_FRAME_MASK = 0x8000000000000000;
-static constexpr uint64_t DIRTY_MASK = 0x4000000000000000;
-static constexpr uint64_t PAGE_IDX_MASK = 0x3FFFFFFFFFFFFFFF;
-
-enum class LockMode : uint8_t { SPIN = 0, NON_BLOCKING = 1 };
-
 class BMFileHandle;
 class BufferManager;
 
 // Keeps the state information of a page in a file.
 class PageState {
-public:
-    inline bool isInFrame() const { return pageIdx & IS_IN_FRAME_MASK; }
-    inline void setDirty() { pageIdx |= DIRTY_MASK; }
-    inline void clearDirty() { pageIdx &= ~DIRTY_MASK; }
-    inline bool isDirty() const { return pageIdx & DIRTY_MASK; }
-    inline common::page_idx_t getPageIdx() const {
-        return (common::page_idx_t)(pageIdx & PAGE_IDX_MASK);
-    }
-    inline uint64_t incrementPinCount() { return pinCount.fetch_add(1); }
-    inline uint64_t decrementPinCount() { return pinCount.fetch_sub(1); }
-    inline void setPinCount(uint64_t newPinCount) { pinCount.store(newPinCount); }
-    inline uint64_t getPinCount() const { return pinCount.load(); }
-    inline uint64_t getEvictionTimestamp() const { return evictionTimestamp.load(); }
-    inline uint64_t incrementEvictionTimestamp() { return evictionTimestamp.fetch_add(1); }
-    inline void releaseLock() { lock.clear(); }
+    static constexpr uint64_t DIRTY_MASK = 0x0080000000000000;
+    static constexpr uint64_t STATE_MASK = 0xFF00000000000000;
+    static constexpr uint64_t VERSION_MASK = 0x00FFFFFFFFFFFFFF;
+    static constexpr uint64_t NUM_BITS_TO_SHIFT_FOR_STATE = 56;
 
-    bool acquireLock(LockMode lockMode);
-    void setInFrame(common::page_idx_t pageIdx);
-    void resetState();
+public:
+    static constexpr uint64_t UNLOCKED = 0;
+    static constexpr uint64_t LOCKED = 1;
+    static constexpr uint64_t MARKED = 2;
+    static constexpr uint64_t EVICTED = 3;
+
+    PageState() {
+        stateAndVersion.store(EVICTED << NUM_BITS_TO_SHIFT_FOR_STATE, std::memory_order_release);
+    }
+
+    inline uint64_t getState() { return getState(stateAndVersion.load()); }
+    inline static uint64_t getState(uint64_t stateAndVersion) {
+        return (stateAndVersion & STATE_MASK) >> NUM_BITS_TO_SHIFT_FOR_STATE;
+    }
+    inline static uint64_t getVersion(uint64_t stateAndVersion) {
+        return stateAndVersion & VERSION_MASK;
+    }
+    inline static uint64_t updateStateWithSameVersion(
+        uint64_t oldStateAndVersion, uint64_t newState) {
+        return ((oldStateAndVersion << 8) >> 8) | (newState << NUM_BITS_TO_SHIFT_FOR_STATE);
+    }
+    inline static uint64_t updateStateAndIncrementVersion(
+        uint64_t oldStateAndVersion, uint64_t newState) {
+        return (((oldStateAndVersion << 8) >> 8) + 1) | (newState << NUM_BITS_TO_SHIFT_FOR_STATE);
+    }
+    inline void spinLock(uint64_t oldStateAndVersion) {
+        while (true) {
+            if (tryLock(oldStateAndVersion)) {
+                return;
+            }
+        }
+    }
+    inline bool tryLock(uint64_t oldStateAndVersion) {
+        return stateAndVersion.compare_exchange_strong(
+            oldStateAndVersion, updateStateWithSameVersion(oldStateAndVersion, LOCKED));
+    }
+    inline void unlock() {
+        assert(getState(stateAndVersion.load()) == LOCKED);
+        stateAndVersion.store(updateStateAndIncrementVersion(stateAndVersion.load(), UNLOCKED),
+            std::memory_order_release);
+    }
+    // Change page state from Mark to Unlocked.
+    inline bool tryClearMark(uint64_t oldStateAndVersion) {
+        assert(getState(oldStateAndVersion) == MARKED);
+        return stateAndVersion.compare_exchange_strong(
+            oldStateAndVersion, updateStateWithSameVersion(oldStateAndVersion, UNLOCKED));
+    }
+    inline bool tryMark(uint64_t oldStateAndVersion) {
+        return stateAndVersion.compare_exchange_strong(
+            oldStateAndVersion, updateStateWithSameVersion(oldStateAndVersion, MARKED));
+    }
+
+    inline void setDirty() {
+        assert(getState(stateAndVersion.load()) == LOCKED);
+        stateAndVersion |= DIRTY_MASK;
+    }
+    inline void clearDirty() {
+        assert(getState(stateAndVersion.load()) == LOCKED);
+        stateAndVersion &= ~DIRTY_MASK;
+    }
+    inline bool isDirty() const { return stateAndVersion & DIRTY_MASK; }
+    uint64_t getStateAndVersion() const { return stateAndVersion.load(); }
+
+    inline void resetToEvicted() {
+        stateAndVersion.store(EVICTED << NUM_BITS_TO_SHIFT_FOR_STATE, std::memory_order_release);
+    }
 
 private:
-    std::atomic_flag lock = ATOMIC_FLAG_INIT;
-    // Highest 1st bit indicates if this page is loaded or not, 2nd bit indicates if this
-    // page is dirty or not. The rest 62 bits records the page idx inside the file.
-    uint64_t pageIdx = 0;
-    std::atomic<uint32_t> pinCount = 0;
-    std::atomic<uint64_t> evictionTimestamp = 0;
+    // Highest 1 bit is dirty bit, and the rest are page state and version bits.
+    // In the rest bits, the lowest 1 byte is state, and the rest are version.
+    std::atomic<uint64_t> stateAndVersion;
 };
 
 // This class is used to keep the WAL page idxes of a page group in the original file handle.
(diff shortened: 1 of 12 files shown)


src/storage/buffer_manager/buffer_manager.cpp
```diff
@@ -13,35 +11,32 @@ namespace storage {
...
-            // Remove the candidate from the eviction queue if it is still pinned or if it has
-            // been recently visited.
+        auto pageStateAndVersion = evictionCandidate.pageState->getStateAndVersion();
+        if (evictionCandidate.isSecondChanceEvictable(pageStateAndVersion)) {
+            // The page was optimistically read, mark it as MARKED, and enqueue to be evicted later.
+            evictionCandidate.pageState->tryMark(pageStateAndVersion);
+            queue->enqueue(evictionCandidate);
             continue;
-        } else {
+        } else if (evictionCandidate.isEvictable(pageStateAndVersion)) {
```
semihsalihoglu-uw, line 31: I think making isEvictable the if and sSecondChanceEvictable else if will make it easier to read this code.

src/storage/buffer_manager/buffer_manager.cpp
```diff
@@ -135,13 +168,17 @@ bool BufferManager::claimAFrame(
...
+                evictionCandidate.pageState->tryMark(pageStateAndVersion);
+                evictionQueue->enqueue(evictionCandidate);
+            }
             continue;
         }
-        // We found a page whose pin count can be 0, and potentially haven't been accessed since
-        // enqueued. We try to evict the page from its frame by calling `tryEvictPage`, which will
-        // check if the page's pin count is actually 0 and the page has not been accessed recently,
-        // if so, we evict the page from its frame.
+        // We found a page potentially haven't been accessed since enqueued. We try to evict the
```
semihsalihoglu-uw, line 179: `We found a page that potentially hasn't been accessed ...`.

src/storage/buffer_manager/buffer_manager.cpp
```diff
@@ -13,35 +11,32 @@ namespace storage {
...
+            evictionCandidate.pageState->tryMark(pageStateAndVersion);
+            queue->enqueue(evictionCandidate);
             continue;
-        } else {
+        } else if (evictionCandidate.isEvictable(pageStateAndVersion)) {
             queue->enqueue(evictionCandidate);
             break;
+        } else {
+            // Cases to remove the candidate from the queue:
+            // 1) The page is currently pinned, remove the candidate from the queue.
```
semihsalihoglu-uw, line 36: I would clarify the comment to say talk about the state first instead of the action that lead to the state. So something like: 
`The page is currently LOCKED (because it was pinned), remove ...`
`The page's version number was different (because it was pinned and unpinned ...`.

src/storage/buffer_manager/buffer_manager.cpp
```diff
@@ -70,47 +65,85 @@ BufferManager::~BufferManager() = default;
...
+            continue;
+        }
+        }
+    }
+}
+
+void BufferManager::optimisticRead(BMFileHandle& fileHandle, common::page_idx_t pageIdx,
+    const std::function<void(uint8_t*)>& func) {
+    auto pageState = fileHandle.getPageState(pageIdx);
+    while (true) {
```
semihsalihoglu-uw, line 97: This is a piece of code that makes me appreciate papers (in this case the VMCache paper) describe their core algorithms in sufficient detail so we can actually use them as reference. This is very similar to Listing 2 of the paper.

src/storage/buffer_manager/buffer_manager.cpp
```diff
@@ -70,47 +65,85 @@ BufferManager::~BufferManager() = default;
...
-    if (pageState && pageState->getPinCount() >= 1) {
-        pageState->setDirty();
-        fileHandle.releasePageLock(pageIdx);
-    } else {
-        fileHandle.releasePageLock(pageIdx);
-        throw BufferManagerException("If a page is not in memory or is not pinned, cannot set "
-                                     "it to isDirty = true. filePath: " +
-                                     fileHandle.getFileInfo()->path +
-                                     " pageIdx: " + std::to_string(pageIdx) + ".");
+    while (true) {
```
semihsalihoglu-uw, line 128: I think this while loop is not necessary. Similarly you should just set the state to dirty since a pinned page is guaranteed to be LOCKED. In fact, you might be able to just remove this function if there is a single line call you can make to change the state of a page to dirty. That is I see that you have an `->unlock()` call that does not take the existing state as an argument. You can have a `setLockedPageDirty()` function.

src/storage/buffer_manager/buffer_manager.cpp
```diff
@@ -135,13 +168,17 @@ bool BufferManager::claimAFrame(
             freeUsedMemory(pageSizeToClaim);
             return false;
         }
-        if (!evictionCandidate.isEvictable()) {
+        auto pageStateAndVersion = evictionCandidate.pageState->getStateAndVersion();
+        if (!evictionCandidate.isEvictable(pageStateAndVersion)) {
+            if (evictionCandidate.isSecondChanceEvictable(pageStateAndVersion)) {
+                evictionCandidate.pageState->tryMark(pageStateAndVersion);
```
semihsalihoglu-uw, line 174: Ditto about not checking tryMark's output.

src/storage/buffer_manager/buffer_manager.cpp
```diff
@@ -156,47 +193,43 @@ bool BufferManager::claimAFrame(
...
-    // evictionTimestamp.
-    if (!candidate.isEvictable()) {
-        pageState.releaseLock();
+    auto currStateAndVersion = pageState.getStateAndVersion();
+    // We check if the page is evictable again. Note that if the page's state or version has
+    // changed after the check, `tryLock` will fail, and we will abort the eviction of this page.
+    if (!candidate.isEvictable(currStateAndVersion) || !pageState.tryLock(currStateAndVersion)) {
         return 0;
     }
     // Else, flush out the frame into the file page if the frame is dirty. Then remove the page
```
semihsalihoglu-uw, line 216: I would say:
```
// At this point, the page is LOCKED. Next, flush out the frame ...
```

semihsalihoglu-uw (approved):

src/storage/buffer_manager/buffer_manager.cpp
ray6080: Done. Also moved this function to bm_file_handle.

src/storage/buffer_manager/buffer_manager.cpp
ray6080: Done.

src/storage/buffer_manager/buffer_manager.cpp
ray6080: Done.

src/storage/buffer_manager/buffer_manager.cpp
```diff
@@ -13,35 +11,32 @@ namespace storage {
...
         }
-        if (evictionCandidate.pageState->getPinCount() != 0 ||
-            evictionCandidate.pageState->getEvictionTimestamp() !=
-                evictionCandidate.evictionTimestamp) {
-            // Remove the candidate from the eviction queue if it is still pinned or if it has
-            // been recently visited.
+        auto pageStateAndVersion = evictionCandidate.pageState->getStateAndVersion();
+        if (evictionCandidate.isSecondChanceEvictable(pageStateAndVersion)) {
+            // The page was optimistically read, mark it as MARKED, and enqueue to be evicted later.
+            evictionCandidate.pageState->tryMark(pageStateAndVersion);
```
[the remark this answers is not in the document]
ray6080: The candidate is still in the queue, so next time when the candidate is dequeued, we will try mark again.

src/storage/buffer_manager/buffer_manager.cpp
```diff
@@ -13,35 +11,32 @@ namespace storage {
...
-// pushed. c0 will consist of (ptr to p5, 1), where the latter is the eviction timestamp at the time
-// c0 is put into the queue. Suppose p5 is put into the eviction queue again (e.g., because it was
-// pinned and unpinned). At this point we create another EvictionCandidate object c1 (ptr to p5, 2)
-// where the latter eviction timestamp is now incremented by 1, which makes c1 now not evictable.
-// This idea is inspired by DuckDB's queue-based eviction implementation.
+// 1) If the candidate page's version has changed, which means the page was pinned and unpinned, we
+// remove the candidate.
+// 2) If the candidate page's state is UNLOCKED, and its page version hasn't changed, which means
+// the page was optimistically read, we mark it as MARKED, and move it to the back of the queue.
+// 3) If the candidate page's state is LOCKED, we remove the candidate.
```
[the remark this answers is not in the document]
ray6080: Done.

src/storage/buffer_manager/buffer_manager.cpp
```diff
@@ -13,35 +11,32 @@ namespace storage {
...
-// the eviction queue, it will end up with a timestamp 1. When we push a page into the queue, we
-// create an EvictionCandidate object for the page. Let's call this object c0 when p5 is first
-// pushed. c0 will consist of (ptr to p5, 1), where the latter is the eviction timestamp at the time
-// c0 is put into the queue. Suppose p5 is put into the eviction queue again (e.g., because it was
-// pinned and unpinned). At this point we create another EvictionCandidate object c1 (ptr to p5, 2)
-// where the latter eviction timestamp is now incremented by 1, which makes c1 now not evictable.
-// This idea is inspired by DuckDB's queue-based eviction implementation.
+// 1) If the candidate page's version has changed, which means the page was pinned and unpinned, we
+// remove the candidate.
+// 2) If the candidate page's state is UNLOCKED, and its page version hasn't changed, which means
```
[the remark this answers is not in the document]
ray6080: Done.

src/storage/buffer_manager/buffer_manager.cpp
```diff
@@ -13,35 +11,32 @@ namespace storage {
...
-// time the page is pushed into the eviction queue. For example, the first time p5 is pushed into
-// the eviction queue, it will end up with a timestamp 1. When we push a page into the queue, we
-// create an EvictionCandidate object for the page. Let's call this object c0 when p5 is first
-// pushed. c0 will consist of (ptr to p5, 1), where the latter is the eviction timestamp at the time
-// c0 is put into the queue. Suppose p5 is put into the eviction queue again (e.g., because it was
-// pinned and unpinned). At this point we create another EvictionCandidate object c1 (ptr to p5, 2)
-// where the latter eviction timestamp is now incremented by 1, which makes c1 now not evictable.
-// This idea is inspired by DuckDB's queue-based eviction implementation.
+// 1) If the candidate page's version has changed, which means the page was pinned and unpinned, we
+// remove the candidate.
```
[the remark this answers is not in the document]
ray6080: Done.

src/storage/buffer_manager/buffer_manager.cpp
```diff
@@ -13,35 +11,32 @@ namespace storage {
 
 // In this function, we try to remove as more as possible candidates that are not evictable from the
```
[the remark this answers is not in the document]
ray6080: Done.

src/include/storage/buffer_manager/bm_file_handle.h
```diff
@@ -6,44 +6,86 @@
...
+        return (stateAndVersion & STATE_MASK) >> NUM_BITS_TO_SHIFT_FOR_STATE;
+    }
+    inline static uint64_t getVersion(uint64_t stateAndVersion) {
+        return stateAndVersion & VERSION_MASK;
+    }
+    inline static uint64_t updateStateWithSameVersion(
+        uint64_t oldStateAndVersion, uint64_t newState) {
+        return ((oldStateAndVersion << 8) >> 8) | (newState << NUM_BITS_TO_SHIFT_FOR_STATE);
+    }
+    inline static uint64_t updateStateAndIncrementVersion(
```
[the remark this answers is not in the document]
ray6080: I keep it to make it more clear what the bit shifting and mask operations do.

src/storage/buffer_manager/buffer_manager.cpp
ray6080: Ditto. The candidate is still in the queue, so next time when the candidate is dequeued, we will try mark again.

src/storage/buffer_manager/buffer_manager.cpp
ray6080: Yep! me too.

src/include/storage/buffer_manager/buffer_manager.h
```diff
@@ -12,25 +12,46 @@ class logger;
 namespace kuzu {
 namespace storage {
 
+// This class keeps state info for pages potentially can be evicted. Candidates are organized in the
```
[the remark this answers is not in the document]
ray6080: Done. Simplified comments here to refer to the state transition diagram.

src/include/storage/buffer_manager/buffer_manager.h
```diff
@@ -86,6 +109,34 @@ class EvictionQueue {
  * queue based replacement policy and the MADV_DONTNEED hint to explicitly control evictions. See
  * comments above `claimAFrame()` for more details.
  *
+ * Page states in BM:
+ * A page can be in one of the four states: a) Locked, b) Unlocked, c) Marked, d) Evicted.
```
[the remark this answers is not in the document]
ray6080: Done.

PR #1415, opened 2023-03-28
Cardinality est

andyfengHKU:
This PR introduces basic cardinality estimator, for primary-foreign key join we simply use the following equation
```
|LEFT| * |RIGHT| / domain(primary key)
```
This equation should be widely applicable to joins between node and rel.

For predicate cardinality estimation we rely on magic numbers. All cardinality estimations are based on factorized tuples.

For cost model, we simply use the sum of all intermediate result size.

### DIFF
--- a/src/include/planner/join_order/cardinality_estimator.h
+++ b/src/include/planner/join_order/cardinality_estimator.h
@@ -1,0 +1,48 @@
+#include "binder/query/reading_clause/query_graph.h"
+#include "planner/logical_plan/logical_plan.h"
+#include "storage/store/nodes_statistics_and_deleted_ids.h"
+#include "storage/store/rels_statistics.h"
+
+namespace kuzu {
+namespace planner {
+
+class CardinalityEstimator {
+public:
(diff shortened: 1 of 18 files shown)


src/include/planner/join_order/cardinality_estimator.h
```diff
@@ -0,0 +1,48 @@
...
+    uint64_t estimateFlatten(const LogicalPlan& childPlan, f_group_pos groupPosToFlatten);
+    uint64_t estimateFilter(const LogicalPlan& childPlan, const binder::Expression& predicate);
+
+    double getExtensionRate(
+        const binder::RelExpression& rel, const binder::NodeExpression& boundNode);
+
+private:
+    inline uint64_t atLeastOne(uint64_t x) { return x == 0 ? 1 : x; }
+
+    uint64_t getNumNodes(const std::string& nodeIDName) {
```
ray6080, line 32: getNumNodes -> getDomain / getDomCardinality?

ray6080 (approved):

PR #1417, opened 2023-03-28
Rework asp optimizer

andyfengHKU:
- rename ASP to AccHashJoin
- disable SIP from probe to build when probe is much greater than build
- enable SIP from build to probe when probe is much greater than build
- optimize enumeration to avoid generating hash join whose cost is already known to be greater than the max cost for a given subgraph


src/optimizer/acc_hash_join_optimizer.cpp
```diff
@@ -54,26 +101,22 @@ void ASPOptimizer::visitIntersect(planner::LogicalOperator* op) {
...
-        return false;
-    }
-    return true;
+    // Probe side is not selective. So we don't apply acc hash join.
+    return subPlanContainsFilter(probeRoot);
 }
 
 binder::expression_map<std::vector<planner::LogicalOperator*>>
-ASPOptimizer::resolveScanNodesToApplySemiMask(const binder::expression_vector& nodeIDCandidates,
+AccHashJoinOptimizer::resolveScanNodesToApplySemiMask(
```
andyfengHKU, line 118: rename signature

src/include/common/constants.h
```diff
@@ -130,10 +130,13 @@ struct LoggerConstants {
...
 };
 
-struct EnumeratorKnobs {
+struct PlannerKnobs {
     static constexpr double NON_EQUALITY_PREDICATE_SELECTIVITY = 0.1;
     static constexpr double EQUALITY_PREDICATE_SELECTIVITY = 0.01;
     static constexpr uint64_t BUILD_PENALTY = 2;
+    // Avoid doing probe to build SIP if we have to accumulate a probe side that is much bigger than
+    // build side. Also avoid doing build to probe SIP if probe side is not much bigger than build.
+    static constexpr uint64_t ACC_HASH_JOIN_RATIO = 5;
```
ray6080, line 139: ACC_HJ_PROBE_BUILD_RATIO?

src/processor/mapper/map_hash_join.cpp
```diff
@@ -40,8 +40,15 @@ std::unique_ptr<PhysicalOperator> PlanMapper::mapLogicalHashJoinToPhysical(
     auto hashJoin = (LogicalHashJoin*)logicalOperator;
     auto outSchema = hashJoin->getSchema();
     auto buildSchema = hashJoin->getChild(1)->getSchema();
-    auto buildSidePrevOperator = mapLogicalOperatorToPhysical(hashJoin->getChild(1));
-    auto probeSidePrevOperator = mapLogicalOperatorToPhysical(hashJoin->getChild(0));
+    std::unique_ptr<PhysicalOperator> probeSidePrevOperator;
+    std::unique_ptr<PhysicalOperator> buildSidePrevOperator;
```
ray6080, line 44: Leave a comment here that we need to map the side into which the SIP is passed first.

src/optimizer/acc_hash_join_optimizer.cpp
```diff
@@ -12,36 +12,83 @@ using namespace kuzu::planner;
...
+void AccHashJoinOptimizer::visitOperator(planner::LogicalOperator* op) {
     // bottom up traversal
     for (auto i = 0u; i < op->getNumChildren(); ++i) {
         visitOperator(op->getChild(i).get());
     }
     visitOperatorSwitch(op);
 }
 
-void ASPOptimizer::visitHashJoin(planner::LogicalOperator* op) {
+void AccHashJoinOptimizer::visitHashJoin(planner::LogicalOperator* op) {
```
ray6080, line 27: Leave a comment here that we always prioritize left to right sip, compared to right to left one.

src/include/optimizer/acc_hash_join_optimizer.h
```diff
@@ -4,16 +4,20 @@
...
+class AccHashJoinOptimizer : public LogicalOperatorVisitor {
 public:
     void rewrite(planner::LogicalPlan* plan);
 
 private:
     void visitOperator(planner::LogicalOperator* op);
 
     void visitHashJoin(planner::LogicalOperator* op) override;
+
+    bool tryLeftToRightHashJoinSIP(planner::LogicalOperator* op);
```
ray6080, line 18: tryApplyLeftToRightSIP and tryApplyRightToLeftSIP?

src/include/planner/logical_plan/logical_operator/side_way_info_passing.h
```diff
@@ -8,6 +8,9 @@ namespace planner {
 enum class SidewaysInfoPassing : uint8_t {
     NONE = 0,
     LEFT_TO_RIGHT = 1,
```
ray6080, line 10: I wonder if we should always directly say PROBE_TO_BUILD, BUILD_TO_PROBE, instead of LEFT_TO_RIGHT, and RIGHT_TO_LEFT? The former one seems more clear, when there are cases we kind of mix the terms of build, probe and left, right altogether.

ray6080 (approved):

PR #1430, opened 2023-03-30
Fix issue 1428, 1408

andyfengHKU:


src/expression_evaluator/function_evaluator.cpp
```diff
@@ -31,6 +31,19 @@ bool FunctionExpressionEvaluator::select(SelectionVector& selVector) {
     for (auto& child : children) {
         child->evaluate();
     }
+    if (selectFunc == nullptr) {
```
ray6080, line 34: Can you comment or add a todo here to explain that this is a temp path for functions (e.g., list contains) that didn't implement `select` interface? We will remove this once we've done changes to those functions.

ray6080 (approved):

PR #1455, opened 2023-04-06
Lazy ctor

ynoza:


src/include/common/types/date_t.h
```diff
@@ -105,7 +105,7 @@ class Date {
 
     KUZU_API static int64_t getEpochNanoSeconds(const date_t& date);
 
-    KUZU_API static int64_t getEpochMicroSeconds(const date_t& date);
+    KUZU_API static int64_t getEpochMilSeconds(const date_t& date);
```
mewim, line 108: Same here

tools/nodejs_api/src_cpp/node_connection.cpp
```diff
@@ -21,22 +23,80 @@ Napi::Object NodeConnection::Init(Napi::Env env, Napi::Object exports) {
...
+}
+
+
+Napi::Value NodeConnection::GetConnection(const Napi::CallbackInfo& info) {
+  Napi::Env env = info.Env();
+  Napi::HandleScope scope(env);
+  // Construct context data
+  context = new ThreadSafeConnectionContext(env, connection, database, numThreads);
+
+  // Create a new ThreadSafeFunction.
```
mewim, line 75: These comments seems to come from the example. You should remove them.

tools/nodejs_api/src_nodejs/connection.js
```diff
@@ -4,17 +4,44 @@ const QueryResult = require("./queryResult.js");
 class Connection {
     #connection;
     #database;
+    #initPromise;
+    #initialized;
     constructor(database, numThreads = 0) {
-        if (typeof database !== "object" || database.constructor.name !==  "Database" || typeof numThreads !== "number" || !Number.isInteger(numThreads)){
+        if (typeof database !== "object" || database.constructor.name !== "Database" || typeof numThreads !== "number" || !Number.isInteger(numThreads)) {
```
mewim, line 10: You should split the error messages to make it easier for the user to understand what is wrong. For example:
```
if (typeof database !== "object" || database.constructor.name !== "Database"){
  throw new Error("The database object is not valid.");
}
if (typeof numThreads !== "number" || !Number.isInteger(numThreads)) {
  throw new Error("numThreads is not a valid integer.");
}

```

tools/nodejs_api/src_nodejs/connection.js
```diff
@@ -4,17 +4,44 @@ const QueryResult = require("./queryResult.js");
...
             throw new Error("Connection constructor requires a database object and optional numThreads integer as argument(s)");
         }
         this.#database = database;
         this.#connection = new kuzu.NodeConnection(database.database, numThreads);
+        this.#initPromise = this.#connection.getConnection();
```
mewim, line 15: You should not start initializing upon calling constructor. It will make it hard for the user to catch the initialization error. Do the initialization upon the user issues the first query.

tools/nodejs_api/src_nodejs/connection.js
```diff
@@ -4,17 +4,44 @@ const QueryResult = require("./queryResult.js");
...
+        }
+        return this.#connection;
+    }
+
+    async execute(query, opts = {}) {
+        const connection = await this.getConnection();
+        const optsKeys = Object.keys(opts);
         if (typeof opts !== "object") {
             throw new Error("optional opts in execute must be an object");
+        } else if (optsKeys.length > 2) {
```
mewim, line 38: No need to do this type of check in Javascript. The idea of [Duck-typing](https://medium.com/@eamonocallaghan/what-is-duck-typing-in-javascript-f3eb10853361) is common in Javascript. So you should proceed as long as you get what you need. If there is additional vars / functions, you can just ignore them instead of raising an error.

tools/nodejs_api/src_nodejs/connection.js
```diff
@@ -29,16 +56,18 @@ class Connection {
         const queryResult = new QueryResult(this, nodeQueryResult);
         if ('callback' in opts) {
             const callback = opts['callback'];
-            if (typeof callback !== "function" || callback.length !== 2){
+            if (typeof callback !== "function" || callback.length !== 2) {
```
mewim, line 59: Similar here. There is absolutely no need to check if the callback takes 2 params. The user may not be interested in error or result at all.

tools/nodejs_api/src_nodejs/queryResult.js
```diff
@@ -2,47 +2,60 @@ class QueryResult {
     #connection
     #database;
     #queryResult
-    #isClosed
     constructor(connection, queryResult) {
-        if (typeof queryResult !== "object" || queryResult.constructor.name !==  "NodeQueryResult"){
-            throw new Error("QueryResult constructor requires a NodeQueryResult object as an argument");
+        if (typeof connection !== "object" || connection.constructor.name !==  "Connection"){
```
mewim, line 6: Since the user does not call the constructor of `QueryResult` (i.e. the only way to get the object is through the return value of one of our function), there is no need to do this type of check. You should change it to assertion or just don't do a check at all.

tools/nodejs_api/src_nodejs/queryResult.js
```diff
@@ -2,47 +2,60 @@ class QueryResult {
...
-    }
-
      async each(rowCallback, doneCallback) {
-        this.checkForQueryResultClose();
-         if (typeof rowCallback !== 'function' || rowCallback.length !== 2 || typeof doneCallback !== 'function' || doneCallback.length !== 0) {
-             throw new Error("all must have 2 callbacks: rowCallback takes 2 arguments: (err, result), doneCallback takes none");
-         }
-        await this.#queryResult.each(rowCallback, doneCallback);
+         return new Promise(async (resolve, reject) => {
+             if (typeof rowCallback !== 'function' || rowCallback.length !== 2 || typeof doneCallback !== 'function' || doneCallback.length !== 0) {
```
mewim, line 17: Same here. Don't check the callback function

mewim (changes requested):
See my comments

src/main/connection.cpp
```diff
@@ -294,7 +294,10 @@ void Connection::bindParametersNoLock(PreparedStatement* preparedStatement,
             throw Exception("Parameter " + name + " not found.");
         }
         auto expectParam = parameterMap.at(name);
-        if (expectParam->dataType != value->getDataType()) {
+        if (value->getDataType().getTypeID() == DataTypeID::INT64
```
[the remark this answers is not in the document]
ynoza: Without this I wasn't able to distinguish between an integer and double with NAPI. So if I remove this I will have to remove that functionality for the datatypes as well.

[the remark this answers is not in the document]
mewim: Well, this is definitely not not the correct place to handle this (i.e., you should not change the general database implementation to handle Node.js specific problem. As you are working on the Node.js driver only, the general principle is that you should not change any existing API implementation. If you think certain features cannot be implemented, you should remove or disable or comment out the features. But in your final PR, you need to make sure not to modify any existing implementation, as specified in my message.

PR #1468, opened 2023-04-16
Shortest Path changes

anuchak:
This PR has all the current changes of Shortest Path including frontend and physical operators.


andyfengHKU (changes requested):
Let's merge the physical operator part and well the test suite. So can u do the following

- Remove code that is not under processor/
- Describe how you generated the test dataset, i.e. what are connectivity pattern.

To pass CI you will need to comment out test code as well.

anuchak:
All front end related changes for shortest path have been removed.

andyfengHKU (approved):

PR #1469, opened 2023-04-16
Rework arrow array copy in node table copier

ray6080:
Get rid of the `toString()` conversion in copying from arrow to column chunk in node copier.
This can boost the node copy performance a lot.
On `uw5`, the copy of ldbc-100 comment csv file can be reduced from 1031s to 286s.


andyfengHKU (approved):
Don't forget to update description

PR #1485, opened 2023-04-23
Add c, cpp examples

andyfengHKU:


examples/c/main.c
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    // Print query result.
+    printf("%s", kuzu_query_result_to_string(result));
+
+    kuzu_query_result_destroy(result);
+    kuzu_connection_destroy(conn);
+    kuzu_database_destroy(db);
+    return 0;
+}
```
mewim, line 48: New line here

examples/c/main.c
```diff
@@ -0,0 +1,48 @@
...
+        int64_t age = kuzu_value_get_int64(value);
+        kuzu_value_destroy(value);
+
+        printf("name: %s, age: %lld \n", name, age);
+        free(name);
+        kuzu_flat_tuple_destroy(tuple);
+    }
+
+    // Print query result.
+    printf("%s", kuzu_query_result_to_string(result));
```
mewim, line 42: This causes memory leak as `kuzu_query_result_to_string` allocates a string buffer. The `kuzu_query_result_to_string` result should be saved into a variable and freed manually.

test/c_api/flat_tuple_test.cpp
```diff
@@ -54,8 +54,8 @@ TEST_F(CApiFlatTupleTest, ToString) {
     columnWidths[0] = 10;
     columnWidths[1] = 5;
     columnWidths[2] = 10;
-    auto str = kuzu_flat_tuple_to_string(flatTuple, columnWidths, 3, "\t", 1024);
-    ASSERT_EQ(std::string(str), " Alice    \t 35  \t 1.731000 ");
+    auto str = kuzu_flat_tuple_to_string(flatTuple);
```
mewim, line 57: Since the function no longer requires passing a `columnWidths` array. It should also be removed from this test case.

ray6080 (approved):

test/c_api/flat_tuple_test.cpp
andyfengHKU: This test is meant to replace the original to_string API with a new one that doesn't require columnWidths. to_string() is still needed.

mewim: I mean that the lines:
```
    auto columnWidths = (uint32_t*)malloc(3 * sizeof(uint32_t));
    columnWidths[0] = 10;
    columnWidths[1] = 5;
    columnWidths[2] = 10;
```
and
```
    free(columnWidths);
```
should be removed as the function is no longer called with `columnWidths`.

PR #1489, opened 2023-04-25
SSP pipeline

andyfengHKU:
This PR implements the non-blocking version of shortest path.

In the front end, SHORTEST_PATH edge is treated in the same way as a regular query edge except that we don't allow property reading on SHORTEST_PATH edge.

In the back end, we add a RecursiveJoin operator that computes the SHORTEST_PATH for a single src node. Parallelism is achieved by having different threads working on different source node. Note this approach will suffer from skewness.


src/include/processor/operator/var_length_extend/bfs_state.h
```diff
@@ -0,0 +1,59 @@
...
+struct SSPMorsel {
+    // Level state
+    uint8_t currentLevelNumber;
+    uint64_t currentLevelNodeIdx; // controls scan of current BFS level
+    std::unique_ptr<BFSLevel> currentBFSLevel;
+    std::unique_ptr<BFSLevel> nextBFSLevel;
+
+    // Visited state
+    uint64_t numVisitedNodes;
+    uint8_t* visitedNodes;
```
semihsalihoglu-uw, line 29: Is this necessary? Can you not always get this pointer from `visitedNodesBuffer`? Similarly for distanceBuffer? There are a lot of fields in this class and it is hard to keep track of them.

src/include/processor/operator/var_length_extend/bfs_state.h
```diff
@@ -0,0 +1,59 @@
...
+    std::unique_ptr<uint8_t[]> visitedNodesBuffer;
+    uint8_t* distance;
+    std::unique_ptr<uint8_t[]> distanceBuffer;
+
+    // Start offset of src node. We need to exclude start node from result.
```
semihsalihoglu-uw, line 34: Can you say some more? Why do we need to exclude start from result? Is that a semantics specific to shortest paths? Let's explain because we will forget.

src/processor/operator/var_length_extend/bfs_state.cpp
```diff
@@ -0,0 +1,79 @@
...
+    if (currentLevelNumber == upperBound) { // upper limit reached.
+        return true;
+    }
+    if (numVisitedNodes == maxOffset) { // all dst has been reached.
+        return true;
+    }
+    return false;
+}
+
+void SSPMorsel::markSrc(common::offset_t offset) {
```
semihsalihoglu-uw, line 40: Do you want to make `markSrc`, `unMarkSrc` and `markVisited` inlined functions?

src/include/processor/operator/var_length_extend/bfs_state.h
```diff
@@ -0,0 +1,59 @@
...
+    SSPMorsel(common::offset_t maxOffset_, uint8_t upperBound_);
+
+    // Reset state for a new src node.
+    void resetState();
+    // If BFS has completed.
+    bool isComplete();
+    // Mark src as visited.
+    void markSrc(common::offset_t offset);
+    // UnMark src as NOT visited to avoid output src.
+    void unMarkSrc();
```
semihsalihoglu-uw, line 49: unMark -> unmark.

src/include/processor/operator/var_length_extend/bfs_state.h
```diff
@@ -0,0 +1,59 @@
...
+
+    SSPMorsel(common::offset_t maxOffset_, uint8_t upperBound_);
+
+    // Reset state for a new src node.
+    void resetState();
+    // If BFS has completed.
+    bool isComplete();
+    // Mark src as visited.
+    void markSrc(common::offset_t offset);
+    // UnMark src as NOT visited to avoid output src.
```
semihsalihoglu-uw, line 48: to avoid output src -> to avoid outputting src.

src/processor/operator/var_length_extend/recursive_join.cpp
```diff
@@ -0,0 +1,119 @@
...
+        op = op->getChild(0);
+    }
+    scanBFSLevel = (ScanBFSLevel*)op;
+
+    // Create two datachunks each with 1 nodeID value vector.
+    // DataChunk 1 has flat state and contains the temporary src node ID vector for recursive join.
+    // DataChunk 2 has unFlat state and contains the temporary dst node ID vector for recursive
+    // join.
+    localResultSet = std::make_unique<ResultSet>(2);
+    localResultSet->insert(0, std::make_shared<common::DataChunk>(1));
```
semihsalihoglu-uw, line 92: I think some of the 0's and 1's here is connected to the:

```
        DataPos tmpSrcNodePos{0, 0};
        DataPos tmpDstNodePos{1, 0};
```
in map_extend. And I think some of the other 1's are just the num vectors in DataChunks. Let's make what is connected and what is not explicit. One way is to have a constants for those that are connected somewhere appropriate and refer to them, so we can search easily where they are used. For example, you can make `tmpSrcNodePos` as a constant `srcNodeIDsPos` in a RecursiveJoin and use it. Similarly for dstNodeIDPos.

src/processor/operator/var_length_extend/recursive_join.cpp
```diff
@@ -0,0 +1,119 @@
...
+    dstNodeIDVector = resultSet->getValueVector(dstNodeIDVectorPos);
+    distanceVector = resultSet->getValueVector(distanceVectorPos);
+    sspMorsel = std::make_unique<SSPMorsel>(maxNodeOffset, upperBound);
+    initLocalRecursivePlan(context);
+    bfsScanState.resetState();
+}
+
+bool RecursiveJoin::getNextTuplesInternal(ExecutionContext* context) {
+    while (true) {
+        if (bfsScanState.sizeScanned < sspMorsel->numVisitedNodes) {
```
semihsalihoglu-uw, line 26: I think this control flow is clearer. I would still add a relatively long comment explaining the flow and the assumptions. The way the first ` if (bfsScanState.sizeScanned < sspMorsel->numVisitedNodes)` condition reads is still a bit confusing because it reads as if we output immediately as long as we visit some nodes. I'd write this above of immediately below `while(true)`.

`
There are two high level steps to RecursiveJoin. (1) BFS Computation phase: Grab a new source to do a BFS and compute an entire BFS starting from a single source; (2) Outputting phase: Once a computation finishes, we output the results in pieces of vectors to the parent operator.  These 2 steps are repeated iteratively until all sources to do a recursive computation are exhausted. The first if clause checks if we are in the outputting phase and if so, scans a vector to output and returns true. Otherwise, we compute a new BFS. 
`

src/processor/operator/var_length_extend/recursive_join.cpp
```diff
@@ -0,0 +1,119 @@
...
+    }
+    sspMorsel->resetState();
+    assert(srcNodeIDVector->state->isFlat());
+    auto nodeID = srcNodeIDVector->getValue<common::nodeID_t>(
+        srcNodeIDVector->state->selVector->selectedPositions[0]);
+    sspMorsel->markSrc(nodeID.offset);
+    while (!sspMorsel->isComplete()) {
+        auto nodeOffset = sspMorsel->getNextNodeOffset();
+        if (nodeOffset != common::INVALID_NODE_OFFSET) {
+            // Found a starting node from current BFS level.
```
semihsalihoglu-uw, line 53: Let's change BFS Levels to frontiers.

src/processor/operator/var_length_extend/bfs_state.cpp
```diff
@@ -0,0 +1,79 @@
...
+}
+
+bool SSPMorsel::isComplete() {
+    if (currentBFSLevel->size() == 0) { // no more to extend.
+        return true;
+    }
+    if (currentLevelNumber == upperBound) { // upper limit reached.
+        return true;
+    }
+    if (numVisitedNodes == maxOffset) { // all dst has been reached.
```
semihsalihoglu-uw, line 34: all destinations have been reached.

semihsalihoglu-uw (approved):

PR #1502, opened 2023-04-29
Add sideway information passing for shortest path extend

andyfengHKU:


src/include/processor/operator/mask.h
```diff
@@ -0,0 +1,141 @@
+#pragma once
+
+#include <mutex>
+
+#include "storage/store/node_table.h"
+
+namespace kuzu {
+namespace processor {
+
+// Note: All classes in this file are NOT thread-safe.
```
semihsalihoglu-uw, line 10: Note: Classes in this file are NOT thread-safe.

semihsalihoglu-uw (approved):

src/include/processor/operator/mask.h
```diff
@@ -0,0 +1,141 @@
...
+            mask->setMask(offset, currentMaskValue + 1);
+        }
+    }
+
+    inline uint8_t getNumMasks() const { return numMasks; }
+    inline void incrementNumMasks() { numMasks++; }
+
+private:
+    std::unique_ptr<Mask> mask;
+    uint8_t numMasks;
```
ray6080, line 57: The term `Mask` here is a bit confusing now, as the `Mask` struct is actually physical representation, while in `numMasks`, "Mask" is a logical concept, we are trying to indicate how many logical masks are passed here and physically stored inside the `std::unique_ptr<Mask> mask;`.

I suggest my changes in the comment under `Mask` struct.

src/include/processor/operator/mask.h
```diff
@@ -0,0 +1,141 @@
...
+namespace processor {
+
+// Note: All classes in this file are NOT thread-safe.
+struct MaskUtil {
+    static inline common::offset_t getMorselIdx(common::offset_t offset) {
+        return offset >> common::DEFAULT_VECTOR_CAPACITY_LOG_2;
+    }
+};
+
+struct Mask {
```
ray6080, line 17: I think it's better to rename Mask to MaskData or MaskBuffer here, and move `setMask` / `isMasked` out of this struct (to `MaskCollection` as utility functions maybe). This struct will be further simplified to only contain a byte array initialized to 0 used for mask.

ray6080 (approved):

PR #1505, opened 2023-05-02
Vlq rework

andyfengHKU:
This PR changes variable length join evaluation strategy from DFS to BFS and unify with shortest path join


src/include/processor/operator/mask.h
```diff
@@ -36,6 +36,7 @@ class MaskCollection {
     MaskCollection() : numMasks{0} {}
 
     inline void init(common::offset_t maxOffset) {
+        std::unique_lock lck{mtx};
```
semihsalihoglu-uw, line 39: Was this a bug?

src/include/processor/operator/recursive_extend/bfs_state.h
```diff
@@ -0,0 +1,153 @@
...
+    NOT_VISITED_DST = 0,
+    VISITED_DST = 1,
+    NOT_VISITED = 2,
+    VISITED = 3,
+};
+
+struct Frontier {
+    std::vector<common::offset_t> nodeOffsets;
+    // Multiplicity stands for number of paths that can reach an offset.
+    std::unordered_map<common::offset_t, uint64_t> offset2Multiplicity;
```
semihsalihoglu-uw, line 18: 2 things: 
1) I'd call this nodeMultiplicity. Also, I'd rather not adopt the style of using 2 (for "to") and 4 (for "for).
2) This is only needed by VariableLengthJoin and not by ShortestPathRecursiveJoin. We should not inject it into ShortestPathRecursiveJoin. Maybe you can template BaseBFSMorsel with the type of Frontier it has and have a Frontier and FrontierWithMultiplicity. Something along those lines?

src/processor/operator/recursive_extend/variable_length_recursive_join.cpp
```diff
@@ -0,0 +1,71 @@
...
+            vectorSize++;
+        }
+        numPath -= numPathFitInCurrentVector;
+        outputCursor += numPath == 0; // No more path to scan under current cursor
+    }
+    dstNodeIDVector->state->initOriginalAndSelectedSize(vectorSize);
+    return true;
+}
+
+void VariableLengthRecursiveJoin::computeBFS(ExecutionContext* context) {
```
semihsalihoglu-uw, line 39: You could really write 1 function for computeBFS. It's almost the same code across VariableLengthRecursiveJoin and ShortestPathRecursiveJoin. The code would look like:
```
void BaseRecursiveJoin::computeBFS(ExecutionContext* context) {
    auto morsel = bfsMorsel.get();
    auto nodeID = srcNodeIDVector->getValue<common::nodeID_t>(
        srcNodeIDVector->state->selVector->selectedPositions[0]);
    morsel->markSrc(nodeID.offset);
    while (!morsel->isComplete()) {
        auto nodeOffset = morsel->getNextNodeOffset();
        if (nodeOffset != common::INVALID_NODE_OFFSET) {
            // Found a starting node from current frontier.
            scanBFSLevel->setNodeID(common::nodeID_t{nodeOffset, nodeTable->getTableID()});
            while (recursiveRoot->getNextTuple(context)) { // Exhaust recursive plan.
                updateVisitedNode(morsel, nodeOffset);
            }
        } else {
            // Otherwise move to the next frontier.
            morsel->finishBFSMorsel();
        }
    }
}
```
The only difference is that VariableLengthRecursiveJoin does some work related to multiplicity. One happens in the updateVisitedNode (or updateVisitedState in ShortestPathRecursiveJoin). You can pass in nodeOffset to a common virtual function updateVisitedNode and VariableLengthRecursiveJoin::updateVisitedNode can read the multiplicity there instead of outside the while loop.

The other difference is that VariableLengthRecursiveJoin has to call the `updateNumPathFromCurrentFrontier` function. You can call a virtual morsel->finishBFSMorsel(). In ShortestPathRecursiveJoin this can just be        `morsel->moveNextLevelAsCurrentLevel();` and in VariableLengthRecursiveJoin it can be two lines `morsel->moveNextLevelAsCurrentLevel();  morsel->updateNumPathFromCurrentFrontierIfNecessary();`. 

I think it's probably worth it, at least for now, to factor out the common algorithm across these functions. We can diverge later once we add different versions of these recursive joins.

src/processor/operator/recursive_extend/shortest_path_recursive_join.cpp
```diff
@@ -0,0 +1,69 @@
...
+                updateVisitedState(morsel);
+            }
+        } else {
+            // Otherwise move to the next frontier.
+            morsel->moveNextLevelAsCurrentLevel();
+        }
+    }
+}
+
+void ShortestPathRecursiveJoin::updateVisitedState(ShortestPathBFSMorsel* morsel) {
```
semihsalihoglu-uw, line 55: Any reason this is called updateVisitedState vs updateVisitedNode in, respectively, ShortestPathRecursiveJoin and VariableLengthJoin?

semihsalihoglu-uw (approved):

src/include/processor/operator/recursive_extend/shortest_path_recursive_join.h
```diff
@@ -0,0 +1,58 @@
+#pragma once
+
+#include "recursive_join.h"
+
+namespace kuzu {
+namespace processor {
+
+class ShortestPathRecursiveJoin : public BaseRecursiveJoin {
+public:
+    ShortestPathRecursiveJoin(uint8_t lowerBound, uint8_t upperBound, storage::NodeTable* nodeTable,
```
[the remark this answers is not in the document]
andyfengHKU: No we made a decision to purposely use 2 constructors to avoid having NULL child for source of a pipeline. If we change this we should change every single source operator.

src/include/processor/operator/mask.h
andyfengHKU: yes

PR #1507, opened 2023-05-03
Auto-parse end2end tests

rfdavid:
An initial proposal to scan end-to-end files inside the e2e directory and dynamically run the tests.

This PR is related to the first task in the proposal TODO list:
- Set up the framework to auto-parse and run end2end tests without manually writing cpp code.
- Rewrite all read-only tests with new auto-parsing and running framework.

### How to add new end-to-end tests

For each e2e test, create a file named "test.group" containing the following fields:
```
-GROUP [group name]
-TEST [test name]
-DATASET [dataset]
-CHECK_ORDER true - use runTestAndCheckOrder() instead of runTest().
```

### Implementation details

- `e2e_read_test` recursively scan for `.test` files inside `test/test_files`
- If the directory contains the configuration file `test.group`, parse the file and extract the necessary information to create the tests dynamically
- Although `READ_ONLY` is in the test_files, it's been used only for informative purposes


test/runner/e2e_read_test.cpp
```diff
@@ -3,159 +3,69 @@
...
-        return TestHelper::appendKuzuRootPath("dataset/tinysnb-serial/");
+std::vector<TestConfig> scanEndToEndTestFiles(const std::string& path) {
+    TestConfig testConfig;
+    std::vector<TestConfig> tests;
+
+    for (const auto& entry : std::filesystem::directory_iterator(path)) {
+        std::string dirName = entry.path().filename().string();
+        testConfig.testSuiteName = TestHelper::convertSnakeCaseToCamelCase(dirName + "Test");
+        std::replace(dirName.begin(), dirName.end(), '_', '-');
+        testConfig.dataset = "dataset/" + dirName + "/";
```
ray6080, line 40: Can you take a look at `FileUtils::joinPath`? I think we can reuse that function to join path parts.

test/runner/e2e_read_test.cpp
```diff
@@ -3,159 +3,69 @@
...
-public:
-    std::string getInputDir() override {
-        return TestHelper::appendKuzuRootPath("dataset/tinysnb-serial/");
+std::vector<TestConfig> scanEndToEndTestFiles(const std::string& path) {
+    TestConfig testConfig;
+    std::vector<TestConfig> tests;
+
+    for (const auto& entry : std::filesystem::directory_iterator(path)) {
+        std::string dirName = entry.path().filename().string();
+        testConfig.testSuiteName = TestHelper::convertSnakeCaseToCamelCase(dirName + "Test");
```
ray6080, line 38: I think you can leave "Test" postfix out of the `convertSnakeCaseToCamelCase` call, so it saves a bit processing, though doesn't matter much. 😄

ray6080 (changes requested):
Nice start!

I have one design change that I hope to discuss with you.
Currently in this PR, we rely on the directory naming to figure out the test suite name and dataset path, which is fine, but limits the organization of test files, and coupling the directory name with test semantic is not a clean design in my mind. I think it's better to just have one file under the root level directory (`tinysnb`, `npy_1d`, etc.) that documents test suite name and dataset path.
You can see the first example in my proposal document. Let me know what do you think, we can discuss what's a better way.

ray6080 (commented):
Also, I noticed one thing, when I actually run tests, I see message like this "TinysnbTest.Agg  # GetParam() = Agg". Do you have any ideas why we have `# GetParam() = Agg`?

rfdavid:
> 

Thank you @ray6080 for the valuable feedback!
I totally agree, I'll work on parsing the `.group` file to extract the test suite properties (group, dataset, dbpath, others). I already have an idea in mind, I'll share it once I figure out some details.

rfdavid:
> Also, I noticed one thing, when I actually run tests, I see message like this "TinysnbTest.Agg # GetParam() = Agg". Do you have any ideas why we have `# GetParam() = Agg`?

I realize I'm setting `value_param` inside `testing::RegisterTest` even though we're not using params. I'll remove it.

test/runner/e2e_read_test.cpp
```diff
@@ -3,159 +3,69 @@
...
-    runTest(TestHelper::appendKuzuRootPath("test/test_files/tinysnb/function/interval.test"));
-    runTest(TestHelper::appendKuzuRootPath("test/test_files/tinysnb/function/list.test"));
-    runTest(TestHelper::appendKuzuRootPath("test/test_files/tinysnb/function/arithmetic.test"));
-    runTest(TestHelper::appendKuzuRootPath("test/test_files/tinysnb/function/boolean.test"));
-    runTest(TestHelper::appendKuzuRootPath("test/test_files/tinysnb/function/string.test"));
-    runTest(TestHelper::appendKuzuRootPath("test/test_files/tinysnb/function/cast.test"));
-    runTest(TestHelper::appendKuzuRootPath("test/test_files/tinysnb/function/case.test"));
-}
+    explicit RunDynamicTest(std::vector<std::string> testFiles, std::string dataset)
+        : testFiles_(testFiles), dataset_(dataset) {}
```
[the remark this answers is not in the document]
rfdavid: wdym by `move` both `testFiles` and `dataset`?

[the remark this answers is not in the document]
ray6080: I mean you should use `std::move` here to avoid copy:
```
: testFiles(std::move(testFiles)), dataset(std::move(dataset)) {}
```

You don't need to add the "\_" postfix to class fields. We usually don't do that.
When there is shadowing of class variable names in function params, we usually add "_" postfix to function params.

test/runner/e2e_read_test.cpp
```diff
@@ -2,160 +2,65 @@
...
 
-class TinySnbSerialReadTest : public DBTest {
-public:
-    std::string getInputDir() override {
-        return TestHelper::appendKuzuRootPath("dataset/tinysnb-serial/");
+std::vector<TestConfig> scanTestFiles(const std::string& path) {
+    TestConfig testConfig;
+    std::vector<TestConfig> tests;
+    std::string previousDirectory;
+    for (const auto& entry : std::filesystem::recursive_directory_iterator(path)) {
```
ray6080, line 39: I'm not familiar with `recursive_directory_iterator`. Just one question here: I guess `recursive_directory_iterator` have the iteration order guarantee? It seems you rely on that as you are checking `FileUtils::getParentPathStem(entry) != previousDirectory`. Specifically, I just want to make sure that the following case is covered in this function.

```
test_files/test.group
test_files/t1.test
test_files/group1/test.group
test_files/group1/g1.test
test_files/group2/case1/test.group
test_files/group2/case1/g2c1.test
```

Alternatively, I think something like the following also makes sense, if you don't want to rely on the `recursive_directory_iterator`:
```
void scanTestFiles(const std::string& path, std::vector<TestConfig>& configs) {
    for (entry in path) {
        if (test group file exists) { 
            auto config = parseTestGroup(path); 
            if (config.isValid()) {
                configs.push_back(config);
            }
        }
        if (entry is directory) { scanTestFiles(entry.path, configs); }
    }
}

TestConfig parseTestGroup(const std::string& path) {
    auto testConfig = TestHelper::parseGroupFile(testGroupFile);
    for (entry in path) {
        if (entry is a file && entry's extension is ".test") {
            testConfig.files.push_back(entry.path().string());
        }
    }
    return testConfig;
}
```

Your call here. I don't have preferences over these two solutions.

ray6080 (approved):
LGTM. Please squash before you merge. Thanks!

test/runner/e2e_read_test.cpp
rfdavid: Thanks, that's a good point.
From my tests, I haven't seen something like this (which will result in failure):
```
test_files/group2/case1/test.group
test_files/group1/g1.test
test_files/group2/case1/g2c1.test
```
However, looking at the documentation, it seems the order is not guaranteed. 
Your solution is more reliable and clear, I'll implement that.

rfdavid: I initially implemented your code, the only necessary change was to add a bool to check if the current directory was processed and not create multiple configs. It worked, but I slightly changed it to check the directories first and then parse them. I believe it's cleaner, the functions are very small. I hope you're ok with that.

test/test_helper/test_helper.cpp
```diff
@@ -12,6 +12,30 @@ using namespace kuzu::main;
...
+TestConfig TestHelper::parseGroupFile(const std::string& path) {
+    TestConfig result;
+    if (!FileUtils::fileOrPathExists(path)) {
+        return result;
+    }
+    std::ifstream ifs(path);
+    std::string line;
+    while (getline(ifs, line)) {
+        if (line.starts_with("-GROUP")) {
+            result.testGroup = line.substr(7, line.length());
```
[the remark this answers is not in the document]
rfdavid: Added `setConfigValue` to solve this.
Once I start working on the test file parser, I was thinking of improving this parser by separating it into other classes and having better validation. I'll share more specifics soon.

test/runner/e2e_read_test.cpp
ray6080: Sure! LGTM!

PR #1517, opened 2023-05-06
Add aggreate key dependency optimizer

andyfengHKU:
This PR adds AggregateKeyDependencyOptimizer which analyzes user input group by keys' dependency and split into `keys` and `dependentKeys`. A `dependentKey` has functional dependency on a `key`. `keys` will be hashed during hash aggregate while `dependentKeys` will be directly materialized into f-table. 

Consider the following example
```
RETURN a.ID, a.age, COUNT(*)
```

Based on user input, both `a.ID` and `a.age` are stated as hash keys. However, since `a.ID` is the primary key, `a.age` has functional dependency on `a.ID` so we only need to hash `a.ID` in hash aggregate.

This optimizer is more important given Cypher grammar because the grammar itself doesn't differentiate `SELECT` and `GROUP BY` which means user cannot group by a subset of selection list. Another case is Cypher allows group by `Node` or `Rel` which by definition is to group by all properties of `Node` or `Rel`. A more reasonable solution is to only group by their internal IDs and treat properties as dependentKeys.

## Performance benchmark
Dataset: LDBC100
Machine: M1 Pro, 16GB memory, maximum 4 threads
Query: `MATCH (a:Person)-[:knows]->(b:Person) RETURN keys, COUNT(*) LIMIT 1;` (Note that Aggregate is a Sink operator that is not affected by LIMIT, LIMIT 1 is to reduce print size).

| Keys | Optimizer ON    | Optimizer OFF |
| ----- | -------- | ------- |
| ID | 207 ms  | 198 ms   |
| ID, firstName| 204 ms  | 292 ms   |
| ID, firstName, lastName | 214 ms  | 356 ms   |

### DIFF
--- a/src/include/optimizer/agg_key_dependency_optimizer.h
+++ b/src/include/optimizer/agg_key_dependency_optimizer.h
@@ -1,0 +1,26 @@
+#pragma once
+
+#include "logical_operator_visitor.h"
+#include "planner/logical_plan/logical_plan.h"
+
+namespace kuzu {
+namespace optimizer {
+
+// This optimizer analyzes the dependency between group by keys. If key2 depends on key1 (e.g. key1
+// is a primary key column) we only hash on key1 and saves key2 as a payload.
+class AggKeyDependencyOptimizer : public LogicalOperatorVisitor {
+public:
+    void rewrite(planner::LogicalPlan* plan);
+
+private:
+    void visitOperator(planner::LogicalOperator* op);
+
+    void visitAggregate(planner::LogicalOperator* op) override;
+    void visitDistinct(planner::LogicalOperator* op) override;
+
+    std::pair<binder::expression_vector, binder::expression_vector> resolveKeysAndDependentKeys(
+        const binder::expression_vector& keys);
+};
+
+} // namespace optimizer
+} // namespace kuzu
--- a/src/include/planner/logical_plan/logical_operator/logical_aggregate.h
+++ b/src/include/planner/logical_plan/logical_operator/logical_aggregate.h
@@ -7,11 +7,18 @@ namespace planner {
 
 class LogicalAggregate : public LogicalOperator {
 public:
-    LogicalAggregate(binder::expression_vector expressionsToGroupBy,
-        binder::expression_vector expressionsToAggregate, std::shared_ptr<LogicalOperator> child)
+    LogicalAggregate(binder::expression_vector keyExpressions,
+        binder::expression_vector aggregateExpressions, std::shared_ptr<LogicalOperator> child)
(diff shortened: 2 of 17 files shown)


src/include/optimizer/agg_key_dependency_optimizer.h
```diff
@@ -0,0 +1,26 @@
+#pragma once
+
+#include "logical_operator_visitor.h"
+#include "planner/logical_plan/logical_plan.h"
+
+namespace kuzu {
+namespace optimizer {
+
+// This optimizer analyze the dependency between group by keys. If key2 depends on key1 (e.g. key1
```
semihsalihoglu-uw, line 9: analyzes.

src/include/planner/logical_plan/logical_operator/logical_aggregate.h
```diff
@@ -21,30 +28,40 @@ class LogicalAggregate : public LogicalOperator {
...
     bool hasDistinctAggregate();
+    void insertAllExpressionsToGroupAndScope(f_group_pos groupPos);
 
 private:
-    binder::expression_vector expressionsToGroupBy;
-    binder::expression_vector expressionsToAggregate;
+    binder::expression_vector keyExpressions;
+    // A payload expression depend on a keyExpression (e.g. a.age depends on a.ID) and will not be
+    // treated as a hash key during hash aggregation.
+    binder::expression_vector payloadExpressions;
```
semihsalihoglu-uw, line 63: Your call but I'd call this dependentKeyExpressions to be more specific. `payload` refers generally to values that correspond to keys in maps (e.g., `aggregateExpressions` can also be referred to as payloads according to this terminology.  If you do the change, I'd change the PR comments and in some other places throughout the code as well.

src/optimizer/agg_key_dependency_optimizer.cpp
```diff
@@ -0,0 +1,76 @@
...
+    for (auto& expression : keys) {
+        if (expression->expressionType == common::PROPERTY) {
+            auto propertyExpression = (binder::PropertyExpression*)expression.get();
+            if (propertyExpression->isPrimaryKey() || propertyExpression->isInternalID()) {
+                primaryKeys.push_back(propertyExpression);
+            }
+        }
+    }
+    // Collect variable names whose primary key is part of group keys
+    std::unordered_set<std::string> primaryVarNames;
```
semihsalihoglu-uw, line 52: I think if you gave an example, as we discussed this could be a bit more readable. preimaryKeys vs primaryVarNames and in general varNames vs expressions was not directly clear (though the code is clear enough so I could predict that's what was happening).

semihsalihoglu-uw (approved):

PR #1537, opened 2023-05-12
Fix return star order

andyfengHKU:
This PR fix return star order to follow paring order.


andyfengHKU:
This commit will be picked in Windows PR.

PR #1538, opened 2023-05-14
Recursive path

andyfengHKU:
This PR keeps track of path (currently implemented as `LIST[INTERNAL_ID]`) for recursive join

### DIFF
--- a/src/include/processor/operator/recursive_extend/path_scanner.h
+++ b/src/include/processor/operator/recursive_extend/path_scanner.h
@@ -1,0 +1,95 @@
+#pragma once
+
+#include <stack>
+
+#include "bfs_state.h"
+
+namespace kuzu {
+namespace processor {
+
+/*
+ * FixedLengPathScanner scans all paths of a fixed length k. This is done by starting a backward
+ * traversals from only the destination nodes in the k'th frontier (assuming the first frontier has
+ * index 0) over the backwards edges stored between the frontiers stored in the BFSMorsel that
+ * was used to store the data related to the BFS that was computed in the RecursiveJoin operator.
+ * To identify the destination nodes in the k'th frontier, we use a semi mask that marks the
+ * destination nodes (or the BFSMorsel::isAllDstTarget() function which indicates that every node is
+ * a possible destination).
+ */
+struct FixedLengthPathScanner {
+    using nbrs_t = std::vector<common::offset_t>*;
+    // FixedLengthPathScanner scans from bfsMorsel frontiers and semi mask.
+    const BaseBFSMorsel& bfsMorsel;
+    // Number of extension performed during recursive join
+    size_t numRels;
+    std::vector<common::offset_t> path;
+
+    // DFS states
+    std::stack<nbrs_t> nbrsStack;
+    std::stack<int64_t> cursorStack;
+
+    size_t lastFrontierCursor;
+
+    FixedLengthPathScanner(const BaseBFSMorsel& bfsMorsel, size_t numRels)
+        : bfsMorsel{bfsMorsel}, numRels{numRels} {
+        path.resize(numRels + 1);
+        resetState();
+    }
+
+    size_t scanPaths(common::table_id_t tableID, common::ValueVector* pathOffsetVector,
+        common::ValueVector* pathDataVector, common::ValueVector* dstNodeIDVector,
+        common::sel_t& offsetVectorPos, common::sel_t& dataVectorPos);
+
+    inline void resetState() { lastFrontierCursor = 0; }
+
+private:
+    // Scan current stacks until exhausted or vector is filled up.
+    void continueDfs(common::table_id_t tableID, common::ValueVector* pathOffsetVector,
+        common::ValueVector* pathDataVector, common::ValueVector* dstNodeIDVector,
+        common::sel_t& offsetVectorPos, common::sel_t& dataVectorPos);
+
+    // Initialize stacks for given offset.
+    void initDfs(common::offset_t offset, size_t currentDepth);
+
+    void writePathToVector(common::table_id_t tableID, common::ValueVector* pathOffsetVector,
+        common::ValueVector* pathDataVector, common::ValueVector* dstNodeIDVector,
+        common::sel_t& offsetVectorPos, common::sel_t& dataVectorPos);
+};
+
+/*
+ * Variable-length joins return union of paths with different length (e.g. *2..3). Note that we only
+ * keep track of the backward edges between the frontier in the RecursiveJoin operator (these
+ * frontiers are stored in the BaseBFSMorsel that was used to keep the data related to the BFS that
+ * was computed in the RecursiveJoin). Therefore we cannot start from the src and traverse to find
+ * all paths of all lengths. We can only start from nodes in a particular frontier and traverse
+ * backwards to the source. But whenever we start from a particular frontier, say the k'th frontier,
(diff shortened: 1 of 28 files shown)


src/processor/operator/recursive_extend/path_scanner.cpp
```diff
@@ -0,0 +1,77 @@
+#include "processor/operator/recursive_extend/path_scanner.h"
+
+namespace kuzu {
+namespace processor {
+
+bool FixedSizePathScanner::getNext() {
```
semihsalihoglu-uw, line 6: I have a few comments:
1) You seem to have different DFSScanner for different depths. This is not really necessary and would traverse the smaller depth paths multiple times. I am guessing you are doing this because you want to output all 3-length paths, and then all 4-length paths, etc.. But why do you have to do this (I vaguely recall you saying Neo4j does this but I don't think we have to do it). 

2) You are using a recursive implementation of DFS, which is destined to be very inefficient. You should try to have stack-based iterative implementation. Something based on the core iterative DFS algorithm, if you really have to do DFS. 

3) I also see that in BaseRecursiveJoin::scanOutput() function you are outputting a  vector of paths. But you are not doing anything in the DFS to output a vector of paths. You still do one step of DFS at a time. You can modify your DFS, so that for a prefix of nodes, you can do the "DFS traversal" one adjacency list at a time and output.

These are common wisdom implementation tricks for fast dfs. You might want to benchmark to measure the effects.

semihsalihoglu-uw (changes requested):

src/include/processor/operator/recursive_extend/path_scanner.h
```diff
@@ -0,0 +1,68 @@
...
+    size_t depth;
+    // Current path. TODO(Xiyang): we can remove this and directly read from cursor and nbrs.
+    std::vector<common::offset_t> path;
+
+    // States
+    std::vector<size_t> cursorPerDepth;
+    std::vector<std::vector<common::offset_t>*> nbrsPerDepth;
+    common::offset_t currentDstOffset;
+
+    FixedSizePathScanner(const BaseBFSMorsel& bfsMorsel, size_t depth)
```
semihsalihoglu-uw, line 21: It's confusing that FixedSizePathScanner takes a BFSMorsel. Can you change this so it's more clear what's really neeed?

semihsalihoglu-uw (commented):
As per our discussion, I would still improve on these:
1) Try to implement a iterative version that also can write multiple paths into ValueVectors inside the FixedLengthPathScanner. You might want to benchmark these optimizations to see if they matter but they're common wisdom I think.
2) I would add more comments to clarify how the entire algorithm will work. You would need to clarify that: (i) Why do you need multiple FixedLengthPathScanners? (ii) where you go over multiple FixedLengthPathScanners? and (iii) More comments inside the FixedSizePathScanner::getNext() function to clarify the high-level logic.

Also, it would be good to have @anuchak review the PR too. It would help him familiarize himself and see our discussions.

src/include/processor/operator/recursive_extend/path_scanner.h
andyfengHKU: Leave a comment but decide not to change this yet.

src/include/processor/operator/recursive_extend/path_scanner.h
```diff
@@ -0,0 +1,88 @@
...
+
+#include <stack>
+
+#include "bfs_state.h"
+
+namespace kuzu {
+namespace processor {
+
+/*
+ * PathScanner Design
```
semihsalihoglu-uw, line 11: Try to work harder on these comments. You are not giving enough details and context on some of your comments, e.g., "We scan from dst to src because semi mask might be available and we only need to scan a subset of all dst nodes." You are expecting the reader to infer a lot much from little detail.

I'd remove it all. First you can put this comment on top of PathScanner (not FixedLengthPathScanner).
```
 * Variable-length joins return union of paths with different length (e.g. *2..3). Note that we only keep 
 * track of the backward edges between the frontier in the RecursiveJoin operator (these frontiers are stored in the BaseBFSMorsel that was used to keep the data related to the BFS that was computed in the RecursiveJoin). Therefore we cannot start 
 * from the src and traverse to find all paths of all lengths. We can only start from nodes in a particular 
 * frontier and traverse backwards to the source. But whenever we start from a particular frontier, say 
 * the k'th frontier, we can only traverse paths of length k. Therefore, PathScanner scans 
 * these paths length by length, i.e. we first scan all length-2 paths, then scan all length-3 paths. 
```

Then here, you can put this comment:

```
 * FixedLengPathScanner scans all paths of a fixed length k. This is done by starting a backward
 * traversals from only the destination nodes in the k'th frontier (assuming the first frontier has
 * index 0) over the backwards edges stored between the frontiers stored in the BFSMorsel that
 * was used to store the data related to the BFS that was computed in the RecursiveJoin operator.
 * To identify the destination nodes in the k'th frontier, we use a semi mask that marks the destination 
 * nodes (or the BFSMorsel::isAllDstTarget() function which indicates that every node is a possible
 * destination).
```

src/processor/operator/recursive_extend/path_scanner.cpp
```diff
@@ -0,0 +1,102 @@
...
+bool FixedLengthPathScanner::continueDfs() {
+    auto level = 0;
+    while (!nbrsStack.empty()) {
+        auto& cursor = cursorStack.top();
+        cursor++;
+        if (cursor < nbrsStack.top()->size()) { // Found a new nbr
+            auto offset = nbrsStack.top()->at(cursor);
+            path[level] = offset;
+            if (level == 0) { // Found a new nbr at level 0. Found a new path.
+                return true;
```
semihsalihoglu-uw, line 47: I would do this:

```
void FixedLengthPathScanner::continueDfs(common::table_id_t tableID,
    common::ValueVector* pathOffsetVector, common::ValueVector* pathDataVector,
    common::ValueVector* dstNodeIDVector, common::sel_t& offsetVectorPos,
    common::sel_t& dataVectorPos) {
    auto level = 0;
    while (!nbrsStack.empty()) {
        auto& cursor = cursorStack.top();
        cursor++;
        if (cursor < nbrsStack.top()->size()) { // Found a new nbr
            auto offset = nbrsStack.top()->at(cursor);
            path[level] = offset;
            if (level == 0) { // Found a new nbr at level 0. Found a new path.
               writePathToVector(tableID, pathOffsetVector, pathDataVector, dstNodeIDVector,
                    offsetVectorPos, dataVectorPos);
               if (offsetVectorPos == common::DEFAULT_VECTOR_CAPACITY) {
                    return;
               } 
           }
            // Push new stack.
            cursorStack.push(-1);
            nbrsStack.push(&bfsMorsel.frontiers[level]->bwdEdges.at(offset));
            level--;
        } else { // Failed to find a nbr. Pop stack.
            cursorStack.pop();
            nbrsStack.pop();
            level++;
        }
    }
}
```

The number of continueDFS() function calls you are making right now is commensurate with the number and lengths of the paths. You can avoid these. You can also make `writePathToVector` an inlined function. The modified function I wrote about writes as many paths as it can, up to common::DEFAULT_VECTOR_CAPACITY many paths. Then the caller can check if it needs to start a new DFS  or not. If you want you can make this function return true to indicate you have written as many as common::DEFAULT_VECTOR_CAPACITY and false if there are more paths to write, which the caller can use to decide whether to start a new DFS or not.

semihsalihoglu-uw (approved):

PR #1547, opened 2023-05-17
Recursive join no path tracking optimizer

andyfengHKU:
This PR implements the optimization where path tracking is not needed in recursive join.

### DIFF
--- a/src/include/processor/operator/recursive_extend/path_scanner.h
+++ b/src/include/processor/operator/recursive_extend/path_scanner.h
@@ -8,85 +8,129 @@ namespace kuzu {
 namespace processor {
 
 /*
- * FixedLengPathScanner scans all paths of a fixed length k. This is done by starting a backward
- * traversals from only the destination nodes in the k'th frontier (assuming the first frontier has
- * index 0) over the backwards edges stored between the frontiers stored in the BFSMorsel that
- * was used to store the data related to the BFS that was computed in the RecursiveJoin operator.
- * To identify the destination nodes in the k'th frontier, we use a semi mask that marks the
- * destination nodes (or the BFSMorsel::isAllDstTarget() function which indicates that every node is
- * a possible destination).
+ * BaseFrontierScanner scans all dst nodes from k'th frontier. To identify the
+ * destination nodes in the k'th frontier, we use a semi mask that marks the destination nodes (or
+ * targetDstOffsets.empty() which indicates that every node is a possible destination).
  */
-struct FixedLengthPathScanner {
-    using nbrs_t = std::vector<common::offset_t>*;
-    // FixedLengthPathScanner scans from bfsMorsel frontiers and semi mask.
-    const BaseBFSMorsel& bfsMorsel;
+struct BaseFrontierScanner {
+    std::vector<Frontier*> frontiers;
+    const std::unordered_set<common::offset_t>& targetDstOffsets;
     // Number of extension performed during recursive join
-    size_t numRels;
-    std::vector<common::offset_t> path;
+    size_t k;
 
+    size_t lastFrontierCursor;
+    common::offset_t currentDstOffset;
+
+    BaseFrontierScanner(const std::unordered_set<common::offset_t>& targetDstOffsets, size_t k)
+        : targetDstOffsets{targetDstOffsets}, k{k}, lastFrontierCursor{0},
+          currentDstOffset{common::INVALID_OFFSET} {}
+    virtual ~BaseFrontierScanner() = default;
+
+    size_t scan(common::table_id_t tableID, common::ValueVector* pathVector,
+        common::ValueVector* dstNodeIDVector, common::sel_t& offsetVectorPos,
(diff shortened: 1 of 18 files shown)


src/include/processor/operator/recursive_extend/path_scanner.h
```diff
@@ -8,85 +8,116 @@ namespace kuzu {
...
-        common::ValueVector* pathDataVector, common::ValueVector* dstNodeIDVector,
-        common::sel_t& offsetVectorPos, common::sel_t& dataVectorPos);
+/*
+ * FrontierWithEdgeScanner scans all paths of a fixed length k. This is done by starting
+ * a backward traversals from only the destination nodes in the k'th frontier (assuming the first
+ * frontier has index 0) over the backwards edges stored between the frontiers stored in the
+ * BFSMorsel that was used to store the data related to the BFS that was computed in the
+ * RecursiveJoin operator.
+ */
+struct FrontierWithEdgeScanner : public FrontierScanner {
```
semihsalihoglu-uw, line 53: My main comment is that I see some inconsistencies in the coding style when building different versions of different classes you have: FrontierScanners, Morsels, Frontiers. Sometimes you name the base class as "BaseX" (e.g., in Morsel) and sometimes just "X" (e.g., Frontier). Sometimes you store some data in a common field in the base class and then reinterpret it (e.g., FrontierScanners reinterpret the frontiers stored in the base class to FrontierWithMultiplicity or FrontierWithEdge) and sometimes you store these fields in the extending classes. 

Also at least for the Morsels, it seemed like templating might be better (and probably more efficient during query processing). I would clean this up to ensure you have the same style in every class. It would things more parallel and easier to read. I didn't think hard about what to do here, but I'm happy to think this through quickly with you.

src/include/processor/operator/recursive_extend/path_scanner.h
```diff
@@ -8,85 +8,116 @@ namespace kuzu {
...
 
 /*
- * FixedLengPathScanner scans all paths of a fixed length k. This is done by starting a backward
- * traversals from only the destination nodes in the k'th frontier (assuming the first frontier has
- * index 0) over the backwards edges stored between the frontiers stored in the BFSMorsel that
- * was used to store the data related to the BFS that was computed in the RecursiveJoin operator.
- * To identify the destination nodes in the k'th frontier, we use a semi mask that marks the
- * destination nodes (or the BFSMorsel::isAllDstTarget() function which indicates that every node is
- * a possible destination).
+ * FrontierScanner scans all dst nodes from k'th frontier. To identify the
```
semihsalihoglu-uw, line 11: I think you should reorganize some of these comments. You repeat the same point in multiple places, e.g., that some Sanner scans dst nodes from the k'th frontier. For example, think through what should be the class comments for this base class. Maybe here you should only say that this class is a base class for some other classes or something. Or here you mention what this base class does and in the extending classes just mention what extra things they do instead of repeating. So try to reorganize the comments to avoid duplication yet covering everything you want to say about these classes.

src/processor/operator/recursive_extend/bfs_state.cpp
```diff
@@ -63,5 +44,32 @@ void ShortestPathBFSMorsel::resetVisitedState() {
...
+        visitedNodes[nbrOffset] = VISITED_DST;
+        numVisitedDstNodes++;
+        nextFrontierWithEdge->addEdge(boundOffset, nbrOffset);
+    } else if (visitedNodes[nbrOffset] == NOT_VISITED) {
+        visitedNodes[nbrOffset] = VISITED;
+        nextFrontierWithEdge->addEdge(boundOffset, nbrOffset);
+    }
+}
+
+void ShortestPathNoTrackPathMorsel::markVisited(
```
semihsalihoglu-uw, line 61: Looking at this function and the above also makes me thinking templating could be a better solution here.

semihsalihoglu-uw (approved):

src/processor/mapper/map_extend.cpp
```diff
@@ -160,19 +159,27 @@ std::unique_ptr<PhysicalOperator> PlanMapper::mapLogicalRecursiveExtendToPhysica
             tmpSrcNodePos, std::vector<DataPos>{tmpDstNodePos}, std::move(scanFrontier),
             getOperatorID(), emptyParamString);
     }
+    std::unique_ptr<RecursiveJoinDataInfo> dataInfo;
+    if (extend->trackPath()) {
+        auto pathVectorPos = DataPos(outSchema->getExpressionPos(*rel));
+        dataInfo = std::make_unique<RecursiveJoinDataInfo>(outDataPoses, colIndicesToScan,
+            inNodeIDVectorPos, outNodeIDVectorPos, tmpDstNodePos, true /* trackPath */,
```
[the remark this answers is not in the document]
andyfengHKU: I choose to not change this to enum yet because for a very long time we will just need to check between track/not track. So we probably don't need an ENUM with 2 cases.

Issue #1521, opened 2023-05-08
Testing Framework V2


ray6080:
## Goal
The goal of this proposal is to improve our existing testing framework, especially end-to-end testings:
- re-organize existing tests in a better way according to functionalities of our database (see our [documentation website](https://kuzudb.com/docs/data-import)), and remove unnecessary legacy tests. For example, to test COPY, we have tests distributed in several places, such as `test/copy/copy_*.cpp`, `test/runner/e2d_copy_transaction_test.cpp`, and `test/runner/e2e_read_test.cpp`, we should re-organize them altogether, so it's easier to navigate and make changes. 
- **enhance our end-to-end tests**. Currently, our end-to-end tests are quite simple. They cannot cover lots of complicated testing cases, such as exceptions, multiple statements with transactions, file-based result verification, etc. We need to make our test files more expressible and also easy to write. This will allow us to convert more inline tests to end-to-end tests.
- allow partial and parallel runs of specific test groups. This is to speed up our test runs as we add more tests in the future and make it easy to run tests locally only relevant to one's interests.
- find and fix critical code paths that are not covered by our tests. As we iterate, we should pay some attention to codecov reports.

The general principle for our testing is to avoid testing components individually, instead we should route all tests, when possible, in the end-to-end way through Cypher statements.
The main reason is that, through our experience, inline tests are not easy to maintain and do not bring much extra benefits compared to end-to-end ones.

Fuzzy tests is something we're interested but requires more thoughts, we are not yet sure when and how to add it. I'm looking into mutation-based fuzzy tests, and keep it as a TODO item for now.

### Overview of Our Existing Tests
We use GTest as the underlying testing framework.
Logically, there are two testing modes: inline and end-to-end.
Inline tests require manually writing cpp code with gTests. Usually these serve as unit tests for an individual module. Examples are tests on parse, binder, optimizer, etc. Besides, most update tests are also inline, due to that they often require more complicated logic for testing, which is beyond our current end-to-end tests.
End2end tests test the system as a black-box. This is done through Cypher statements written in test files, see `test/test_files`.

Generally, we have two categories of tests in terms of their functionality: read-only and updates.
The former one is mainly read-only queries, including tests on data types, functions, expressions, and query clauses.
The latter one covers DDL, COPY, and DML. Tests on updates are usually coupled with tests on transactions and recovery.

Additionally, we have specialized tests in our language bindings, such as py tests and c api tests.

### Basic End-to-end Tests
[...]

PR #1548, opened 2023-05-17
Testing framework v2

rfdavid:
### Testing Framework v2

The new framework uses the following logic:

- `e2e_read_test` scans a directory or opens a specific test file if it's passed as an argument. After we support `write` operation, we should change this name.
- A `TestCase` is created for each `.test` file. `TestParser` is responsible for reading and tokenizing the test file
- `TestCase` uses `TestParser` to check the tokens (STATEMENT, CHECK_ORDER, GROUP etc)
- If the test file is correct, a `TestCase` object is created containing a vector of `TestStatement`. `TestStatement` is a struct (derived from `TestQueryConfig`) containing the information about that statement.
- There is no need for a `.group` file anymore. The .test file must contain a header with the following information: `-GROUP`, `-DATASET` and `-TEST`.
- After a `TestCase` is created, `e2e_read_test` will register to GTest.
- `TestRunner` will be responsible to run the test. It reads the `TestCase`, prepare the logical plans and executes. `TestRunner` will check if the result corresponds to the expected in the statement.

This is still a work in progress. It currently supports multiple statements, exceptions and statement blocks.

#### TODO

- [x] Convert the current `.test` files to the current format
- [ ] Reorganize the test files (for example, merge test files `four_hop.test` and `five_hop.test`)

Related to #1521


test/include/test_runner/test_parser.h
```diff
@@ -0,0 +1,94 @@
+#include <cstring>
+
+#include "common/file_utils.h"
+#include "main/kuzu.h"
```
ray6080, line 4: Do we need to include `"main/kuzu.h"` ? Also `using namespace kuzu::main;`. Can you double check?

test/test_runner/test_parser.cpp
```diff
@@ -0,0 +1,227 @@
...
+            testCase->dataset = currentToken.params[1];
+            break;
+        }
+        case TokenType::SKIP: {
+            testCase->skipTest = true;
+            return;
+        }
+        case TokenType::SEPARATOR: {
+            return;
+            break;
```
ray6080, line 54: Remove `break` here.

test/test_runner/test_parser.cpp
```diff
@@ -0,0 +1,227 @@
...
+            extractStatementBlock();
+            break;
+        }
+        case TokenType::STATEMENT_BLOCK: {
+            checkMinimumParams(1);
+            for (auto& statement : testCase->variableStatements) {
+                if (statement->blockName == currentToken.params[1]) {
+                    statement->name = name;
+                    testCase->statements.push_back(std::make_unique<TestStatement>(*statement));
+                }
```
ray6080, line 164: Should we error here if the statement block is not found (i.e., not defined)?

test/test_runner/test_parser.cpp
```diff
@@ -0,0 +1,227 @@
...
+    case TokenType::STATEMENT:
+    case TokenType::QUERY: {
+        checkMinimumParams(1);
+        statement->query = paramsToString();
+        break;
+    }
+    case TokenType::RESULT: {
+        extractExpectedResult(statement);
+        return statement;
+        break;
```
ray6080, line 99: Remove `break` here.

test/include/test_runner/test_case.h
```diff
@@ -0,0 +1,40 @@
...
+    bool enumerate = false;
+    bool checkOutputOrder = false;
+};
+
+struct TestCase {
+    std::string group;
+    std::string name;
+    std::string dataset;
+    std::vector<std::unique_ptr<TestStatement>> statements;
+    std::vector<std::unique_ptr<TestStatement>> variableStatements;
```
ray6080, line 30: My suggestion for `variableStatements` is use `unordered_map<string, std::vector<std::unique_ptr<TestStatement>>>`, so when you need to extract a statement block, you can access based on the variable name, instead of looping all statements.

ray6080 (approved):
One general comment is that we'd better differentiate test case names for different test files. For example, tests under `tinysnb/function` can be named as `ArithmeticFunction`, `BooleanFunction`, etc., instead of just as `Function` for all of them.

ray6080 (approved):
One general comment is that we'd better differentiate test case names for different test files. For example, tests under `tinysnb/function` can be named as `ArithmeticFunction`, `BooleanFunction`, etc., instead of just as `Function` for all of them.

ray6080 (approved):
One general comment is that we'd better differentiate test case names for different test files. For example, tests under `tinysnb/function` can be named as `ArithmeticFunction`, `BooleanFunction`, etc., instead of just as `Function` for all of them.

rfdavid:
> One general comment is that we'd better differentiate test case names for different test files. For example, tests under `tinysnb/function` can be named as `ArithmeticFunction`, `BooleanFunction`, etc., instead of just as `Function` for all of them.

Sounds good! I added to the issue TODO list and I'll change it in the next PR.

src/common/string_utils.cpp
```diff
@@ -19,5 +19,14 @@ std::vector<std::string> StringUtils::split(
     return result;
 }
 
+std::vector<std::string> StringUtils::splitByAnySpace(const std::string& input) {
```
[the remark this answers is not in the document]
rfdavid: This function split single/double spaces, tabs etc. I don't think it's necessary use "anySpace", I'll change to `splitBySpace`.

test/include/test_runner/test_case.h
```diff
@@ -0,0 +1,39 @@
+#pragma once
+
+#include "main/kuzu.h"
+
+using namespace kuzu::main;
```
ray6080, line 5: Remove `using namespace kuzu::main;`. Our convention is that we never use `using namespace` in header files, but you can use them in cpp files. In header files, we add the namespace prefix as needed.

test/include/test_runner/test_runner.h
```diff
@@ -0,0 +1,31 @@
+#include "common/file_utils.h"
+#include "gtest/gtest.h"
+#include "parser/parser.h"
+#include "planner/logical_plan/logical_plan_util.h"
+#include "planner/planner.h"
+#include "test_runner/test_case.h"
+
+using namespace kuzu::planner;
```
ray6080, line 8: Ditto. Remove `using namespace kuzu::planner;` here.

test/include/test_runner/test_case.h
rfdavid: Should I do the same with `test_helper.h` as well?

ray6080: Up to you. Just keep a note is fine.

Issue #1522, opened 2023-05-08
Add undirected query


andyfengHKU:
## Goal
The goal of this issue is to support undirected query in a similar fashion as Neo4j. 

Neo4j supports 3 types of relationship

- Left-to-right, e.g.`(a)-[e]->(b)`
- Right-to-left, e.g. `(a)<-[e]-(b)`
- Undirected, e.g. `(a)-[e]-(b)`

Currently we support left-to-right and right-to-left in the system and missing the undirected one. 

## Design Details
The semantic of undirected is simply the union of left-to-right and right-to-left and should be implemented by solely changing the extend operator.

In query compilation, we should change query edge to be able to take `UNDIRECTED` as a direction and propagate this change to logic extend. 

In query processing, recall that we have implemented a GenericScanRelTable which takes one adjList from each rel table and scan them one by one. Undirected extend should further modify GenericScanRelTable such that it takes both fwd and bwd adjList from each rel table. From a high level view, GenericScanRelTable still scans multiple adjList. Undirected extend simply  adds more adjList to scan.  

## TODOs
- [x] Add undirected query edge and propagate this change to logical extend. Changes include but not limit to parser, binder, logical extend and cardinality estimator.
- [x] Modify GenericScanRelTable to take fwd and bwd adjList from a single rel table. (Example test query `(a:Person)-[:knows]->(b:Person)`)
- [x] Modify GenericScanRelTable to take fwd and bwd adjColumn from a single rel table. (Example test query `(a:Person)-[:meets]->(b:Person)`)
- [x]  Modify GenericScanRelTable to take fwd and bwd adjList/Column from multiple rel tables. (Example test query `(a:Person)-[:meets|:knows]->(b:Person)`)

PR #1552, opened 2023-05-19
Undirected query for multi-labelled and unlabelled nodes. 

aziz-mu:
This commit finishes up implementing undirected query, and includes support for queries in the form: 
`MATCH (A)-[N]-(B)`
`MATCH (A|B|C|..)-[N|M|..]-(E|F|G|..)`
`MATCH ()-[]-()`

This pertains to issue  #1522

### DIFF
--- a/src/binder/bind/bind_graph_pattern.cpp
+++ b/src/binder/bind/bind_graph_pattern.cpp
@@ -27,9 +27,35 @@ Binder::bindGraphPattern(const std::vector<std::unique_ptr<PatternElement>>& gra
     return std::make_pair(std::move(queryGraphCollection), std::move(propertyCollection));
 }
 
+// For undirected pattern (A)-[R]-(B), we need to match R in both FWD and BWD direction.
+// Since computation always starts from one node, we need to rewrite the node table names to be
+// the union of both node table names, i.e. (A|B)-[R]-(A|B)
+void static rewriteNodeTableNameForUndirectedRel(const PatternElement& patternElement) {
+    auto leftNode = patternElement.getFirstNodePattern();
+    for (auto i = 0u; i < patternElement.getNumPatternElementChains(); ++i) {
+        auto patternElementChain = patternElement.getPatternElementChain(i);
+        auto rightNode = patternElementChain->getNodePattern();
+        if (patternElementChain->getRelPattern()->getDirection() == ArrowDirection::BOTH) {
+            std::vector<std::string> tableNameUnion = {};
+            auto leftTableNames = leftNode->getTableNames();
+            auto rightTableNames = rightNode->getTableNames();
+            if (!leftTableNames.empty() && !rightTableNames.empty()) {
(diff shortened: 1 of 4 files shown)


src/binder/bind/bind_graph_pattern.cpp
```diff
@@ -27,12 +28,52 @@ Binder::bindGraphPattern(const std::vector<std::unique_ptr<PatternElement>>& gra
...
+    } else {
+        thisNode->updateTableNames(rightNodeTableNames);
+    }
+}
+
 // Grammar ensures pattern element is always connected and thus can be bound as a query graph.
 std::unique_ptr<QueryGraph> Binder::bindPatternElement(
     const PatternElement& patternElement, PropertyKeyValCollection& collection) {
     auto queryGraph = std::make_unique<QueryGraph>();
+    addUndirectedTableNames(patternElement, -1);
```
andyfengHKU, line 73: Can we perform this rewriting/union logic by traversing `patternElement` twice?

If you look at the binding logic, at a high level it's looping through `patternElement` and keeping track of `leftNode` and `rightNode`
```
boundLeftNode = bind(first node pattern)
for (chain : patternElementChains) {
     boundRightNode = bind(chain.node)
     boundRel = bind(chain.rel)
     boundLeftNode = boundRightNode
}
```

We can do a similar rewriting/union logic and put it before binding. Something along the line of
```
leftNode = first node pattern
for (chain : patternElementChains) {
     rightNode = chain.node
     if (chain.rel.isUndirected) {
         tableNameUnion = unionTableName(leftNode, rightNode)
         leftNode.setTableNames(tableNameUnion);
         leftNode.setTableNames(tableNameUnion);
     }
}
```
To me this is more consistent and easier to understand. Also don't forget to add comments explaining why are we taking the union

src/binder/bind/bind_graph_pattern.cpp
```diff
@@ -1,3 +1,4 @@
+
```
andyfengHKU, line 1: remove

src/binder/bind/bind_graph_pattern.cpp
```diff
@@ -27,9 +27,34 @@ Binder::bindGraphPattern(const std::vector<std::unique_ptr<PatternElement>>& gra
     return std::make_pair(std::move(queryGraphCollection), std::move(propertyCollection));
 }
 
+void static addUndirectedTableNames(const PatternElement& patternElement) {
```
andyfengHKU, line 30: rename to `rewriteNodeTableNameForUndirectedRel` 

also put the comment above the function signature. Change the comment to


"For undirected pattern (A)-[R]-(B), we need to match R in both FWD and BWD direction. Since computation always start from one node, we need to rewrite node table name to be the union of both node table names, i.e. (A|B)-[R]-(A|B)"

andyfengHKU (approved):

PR #1555, opened 2023-05-20
Push recursive path length into operator

andyfengHKU:


src/binder/bind_expression/bind_function_expression.cpp
```diff
@@ -17,11 +17,9 @@ std::shared_ptr<Expression> ExpressionBinder::bindFunctionExpression(
     auto& parsedFunctionExpression = (ParsedFunctionExpression&)parsedExpression;
     auto functionName = parsedFunctionExpression.getFunctionName();
     StringUtils::toUpper(functionName);
-    // check for special function binding
-    if (functionName == ID_FUNC_NAME) {
-        return bindInternalIDExpression(parsedExpression);
-    } else if (functionName == LABEL_FUNC_NAME) {
-        return bindLabelFunction(parsedExpression);
+    auto result = rewriteFunctionExpression(parsedExpression, functionName);
```
andyfengHKU, line 20: Comment when do we need to rewrite a function expression

semihsalihoglu-uw: Yes, it would be good to give an example query in the comment to explain this. 1 example goes a long way.

src/binder/bind_expression/bind_function_expression.cpp
```diff
@@ -207,28 +226,22 @@ std::shared_ptr<Expression> ExpressionBinder::bindNodeLabelFunction(const Expres
...
-                                        std::make_unique<LogicalType>(LogicalTypeID::STRING))),
-            populateLabelValues(relTableIDs, *catalogContent));
-    children.push_back(createLiteralExpression(std::move(labelsValue)));
-    auto execFunc = function::LabelVectorOperation::execFunction;
-    auto bindData =
-        std::make_unique<function::FunctionBindData>(LogicalType(LogicalTypeID::STRING));
-    auto uniqueExpressionName = ScalarFunctionExpression::getUniqueName(LABEL_FUNC_NAME, children);
-    return make_shared<ScalarFunctionExpression>(LABEL_FUNC_NAME, FUNCTION, std::move(bindData),
-        std::move(children), execFunc, nullptr, uniqueExpressionName);
+    auto& rel = (RelExpression&)expression;
```
semihsalihoglu-uw, line 243: space before expression?

src/include/processor/operator/recursive_extend/path_scanner.h
```diff
@@ -27,16 +27,24 @@ struct BaseFrontierScanner {
...
+    inline void writeDstNodeOffsetAndLength(common::ValueVector* dstNodeIDVector,
+        common::ValueVector* pathLengthVector, common::sel_t& offsetVectorPos,
+        common::table_id_t tableID) {
+        dstNodeIDVector->setValue<common::nodeID_t>(
+            offsetVectorPos, common::nodeID_t{currentDstOffset, tableID});
+        pathLengthVector->setValue<int64_t>(offsetVectorPos, (int64_t)k);
+    }
 };
 
 /*
```
semihsalihoglu-uw, line 50: In this comment should you say `DstNodeScanner scans dst node offsets and lengths of paths only."?

src/optimizer/projection_push_down_optimizer.cpp
```diff
@@ -214,8 +214,18 @@ binder::expression_vector ProjectionPushDownOptimizer::pruneExpressions(
     const binder::expression_vector& expressions) {
     expression_set expressionsAfterPruning;
     for (auto& expression : expressions) {
-        if (expression->expressionType != common::PROPERTY ||
-            propertiesInUse.contains(expression)) {
+        switch (expression->expressionType) {
+        case common::VARIABLE: {
```
semihsalihoglu-uw, line 218: Can you explain why this had to change and what exactly is happening here? Maybe add some comments too.

src/processor/operator/recursive_extend/path_scanner.cpp
```diff
@@ -86,14 +87,14 @@ void PathScanner::initDfs(common::offset_t offset, size_t currentDepth) {
...
+    common::sel_t& offsetVectorPos, common::sel_t& dataVectorPos) {
+    assert(offsetVectorPos < common::DEFAULT_VECTOR_CAPACITY);
     auto pathLength = path.size();
     auto listEntry = common::ListVector::addList(pathVector, pathLength);
     pathVector->setValue(offsetVectorPos, listEntry);
-    dstNodeIDVector->setValue<common::nodeID_t>(
-        offsetVectorPos, common::nodeID_t{path[pathLength - 1], tableID});
-    assert(offsetVectorPos < common::DEFAULT_VECTOR_CAPACITY);
+    assert(path[pathLength - 1] == currentDstOffset);
+    writeDstNodeOffsetAndLength(dstNodeIDVector, pathLengthVector, offsetVectorPos, tableID);
```
semihsalihoglu-uw, line 97: I'm a bit confused about the PR. so it looks like regardless of whether there is or there is not len(p) type expression, you always write path lengths to a value vector. Is that correct?

I was expecting to see a flag that you insert into these operators that will decide whether to write the lengths or not.

src/processor/mapper/map_extend.cpp
```diff
@@ -211,11 +213,12 @@ std::unique_ptr<PhysicalOperator> PlanMapper::mapLogicalRecursiveExtendToPhysica
     if (extend->trackPath()) {
         auto pathVectorPos = DataPos(outSchema->getExpressionPos(*rel));
         dataInfo = std::make_unique<RecursiveJoinDataInfo>(outDataPoses, colIndicesToScan,
-            inNodeIDVectorPos, outNodeIDVectorPos, tmpDstNodePos, true /* trackPath */,
-            pathVectorPos);
+            inNodeIDVectorPos, outNodeIDVectorPos, lengthVectorPos, tmpDstNodePos,
+            true /* trackPath */, pathVectorPos);
```
semihsalihoglu-uw, line 217: Ditto (from my previous PR comments) to replace bools with enums to increase readability.

semihsalihoglu-uw (approved):

semihsalihoglu-uw (approved):

src/processor/operator/recursive_extend/path_scanner.cpp
andyfengHKU: Yes I always write path length to a vector because it's cost is same as writing dst node offsets.

Trying to be smart about writing length is doable but codes becomes a lot complicated after I made the change because we need to switch for track `path & length, path & !length, !path & length, !path & !length` in multiple places. I decide to go for readability and always write length for now.

Using enum is fine. I've moved from bool to enum.

PR #1560, opened 2023-05-22
Refactor extend direction & rel data direction

andyfengHKU:


src/include/storage/wal/wal_record.h
```diff
@@ -1,5 +1,6 @@
 #pragma once
 
+#include "common/rel_direction.h"
```
ray6080, line 3: Should this include be added here?

ray6080 (approved):

PR #1561, opened 2023-05-22
Rework ResultSetDescriptor

andyfengHKU:


src/include/processor/result/result_set.h
```diff
@@ -4,13 +4,18 @@
...
 #include "common/data_chunk/data_chunk.h"
 #include "processor/data_pos.h"
+#include "result_set_descriptor.h"
 
 namespace kuzu {
 namespace processor {
 
-class ResultSet {
-public:
+struct ResultSet {
```
ray6080, line 12: Why change this to `struct`?

src/include/processor/result/result_set.h
```diff
@@ -4,13 +4,18 @@
...
+    uint64_t multiplicity;
+    std::vector<std::shared_ptr<common::DataChunk>> dataChunks;
+
     explicit ResultSet(uint32_t numDataChunks) : multiplicity{1}, dataChunks(numDataChunks) {}
+    static std::unique_ptr<ResultSet> getResultSet(
```
ray6080, line 17: I think you can also make this as a constructor, which is more clear. `ResultSet::getResultSet` looks a bit confusing.

src/processor/result/result_set_descriptor.cpp
```diff
@@ -1,31 +1,31 @@
...
 namespace processor {
 
-ResultSetDescriptor::ResultSetDescriptor(const Schema& schema) {
-    for (auto i = 0u; i < schema.getNumGroups(); ++i) {
-        auto group = schema.getGroup(i);
-        auto dataChunkDescriptor = std::make_unique<DataChunkDescriptor>();
-        if (group->isSingleState()) {
-            dataChunkDescriptor->setSingleState();
-        }
+std::unique_ptr<ResultSetDescriptor> ResultSetDescriptor::getResultSetDescriptor(
```
ray6080, line 8: Same as `ResultSet`, it's better to implement this as a constructor.

ray6080 (approved):

PR #1566, opened 2023-05-24
add LSQB queries as test

aziz-mu:
- Added LSQB queries q1 - q9 as a test. Queries q8, q9 are translated to use `WHERE NOT EXISTS {}` syntax
Queries' descriptions can be found at: https://github.com/ldbc/lsqb
- Added the dataset sf-0.1 from https://repository.surfsara.nl/datasets/cwi/lsqb . This is the smallest dataset to run the LSQB queries (~50MB).

Expected output from tests sourced from https://github.com/ldbc/lsqb/blob/main/expected-output/expected-output.csv. All tests pass.


test/lsqb_test/CMakeLists.txt
```diff
@@ -0,0 +1 @@
+add_kuzu_test(lsqb_test lsqb_test.cpp)
```
ray6080, line 1: I think we can just simply call this test as `lsqb` without the `_test` suffix. Also the directory name.

test/lsqb_test/lsqb_test.cpp
```diff
@@ -0,0 +1,21 @@
...
+        createDBAndConn();
+        initGraph();
+    }
+
+    std::string getInputDir() override { return TestHelper::appendKuzuRootPath("dataset/sf-0.1/"); }
+};
+
+TEST_F(LSQBTest, LSQBTest) {
+    runTest(TestHelper::appendKuzuRootPath("test/test_files/lsqb/lsqb_queries.test"));
+}
```
ray6080, line 21: New line here.

ray6080 (approved):

andyfengHKU (approved):

PR #1567, opened 2023-05-24
Enable multiple test cases inside one test file

rfdavid:
This enables using multiple `-CASE` inside the same test file, so GoogleTest registers multiple test cases based on this parameter.


test/include/test_runner/test_group.h
```diff
@@ -19,18 +19,22 @@ struct TestStatement {
     bool checkOutputOrder = false;
 };
 
-struct TestCase {
+// Test group is a collection of test cases. Each test case can
```
ray6080, line 22: Test group is a collection of test cases in a single file.

test/test_runner/test_parser.cpp
```diff
@@ -1,6 +1,7 @@
 #include "test_runner/test_parser.h"
 
 #include <fstream>
+#include <iostream>
```
ray6080, line 4: Do we need this include? `#include <iostream>`

test/test_runner/test_parser.cpp
```diff
@@ -133,19 +134,19 @@ void TestParser::extractStatementBlock() {
...
             auto statement = std::make_unique<TestStatement>();
             extractStatement(statement.get());
-            testCase->variableStatements[blockName].push_back(std::move(statement));
+            testGroup->testCasesStatementBlocks[blockName].push_back(std::move(statement));
         }
     }
 }
 
 void TestParser::parseBody() {
+    std::string testGroupName;
```
ray6080, line 143: I think you mean `testCaseName`, instead of `testGroupName`, here, right?

ray6080 (approved):

PR #1581, opened 2023-05-28
Multi label recursive join

andyfengHKU:


src/include/processor/operator/recursive_extend/path_scanner.h
```diff
@@ -13,23 +13,17 @@ namespace processor {
  * targetDstOffsets.empty() which indicates that every node is a possible destination).
```
ray6080, line 13: Can you change the file name to `frontier_scanner.h`?

ray6080 (approved):
One general comment is that we should clean up the use of `struct` and use `class` when necessary. e.g., BaseFrontierScanner, Frontier, etc.

PR #1582, opened 2023-05-28
Cloud storage

yuchenZhangTG:
1. Support COPY from different filesystem (AWS S3, GCP GCS Storage)
`copy Comment from "gs://kuzubi/example.csv";`
2. Support COPY from compressed CSV files
`copy Comment from "gs://kuzubi/example.csv.gz";`

I have read and agree to the terms under CLA.md.


yuchenZhangTG:
Emmm, my changes require system installed curl.

I see your test is on self-host machines. Is it possible to install CURL on that.

mewim:
Hi @yuchenZhangTG we would like to ensure that Kùzu can be built without installing any third party dependencies manually. If your change requires `libcurl`, could you please bring it `third_party`?

yuchenZhangTG:
> Hi @yuchenZhangTG we would like to ensure that Kùzu can be built without installing any third party dependencies manually. If your change requires `libcurl`, could you please bring it to `third_party`?

As mentioned in https://arrow.apache.org/docs/dev/developers/cpp/building.html, 

```
google_cloud_cpp_storage: for Google Cloud Storage support, requires system cURL and can use the BUNDLED method described below
```
System CURL is required by AWSSDK and google_cloud_cpp_storage. It seems custom build CURL cannot be detected in the dependency build of Apache arrow.

The dependency build generate its own cmake commands and environment variables. I tried to pass in different -D and environemnt variables, but it does not work.

Also I tried to add CURL in the external folder next to Apache arrow. Since third_party is built after apache arrow.

@mewim  How do you think.

ray6080 (commented):
Hi @yuchenZhangTG , this feature looks awesome! We do hope to integrate with cloud storage well and have some ideas in our roadmap (though not publicly written down yet). 
As for this PR, can you suggest a good way for testing? Should we have a private bucket for some of our testing dataset on S3/GCS, and deploy secrets on our CI machine?

Btw, feel free to join our slack workspace here (https://join.slack.com/t/kuzudb/shared_invite/zt-1w0thj6s7-0bLaU8Sb~4fDMKJ~oejG_g) :smile We'd like to learn more from your use cases with this feature!

yuchenZhangTG:
@ray6080 
The current implementation uses Filesystem module in Apache Arrow. But it requires system installed libcurl-dev, and it seems that custom built CURL not work. 

For testing, yeah, it would be great if you set up your own AWS and GCP account, and have private buckets for testing. The cost for cloud storage is not expensive and you should be able to cover the cost. It is better to have secrets hosted on your CI test machines (Sometimes, people put secrets in the Github code, it is convenient, but it is not safe). As far as I know, GCP does not charge for downloading, and AWS only charge for cross-region download. It is also OK to make buckets public if you want to share the use case on the Kuzu cloud (the one on your website demo).

For cloud import
I loaded 180MB snappy parquet using 11s from local SSD and 14s from GCS bucket. So I would say copying from GCS is pretty fast.
For compressed CSV
I tested on Google Cloud Storage (GCS) and loaded SF30 LDBC BI dataset and import comment.**csv.gz** files. it worked.

ray6080:
> @ray6080 The current implementation uses Filesystem module in Apache Arrow. But it requires system installed libcurl-dev, and it seems that custom built CURL not work.
> 
> For testing, yeah, it would be great if you set up your own AWS and GCP account, and have private buckets for testing. The cost for cloud storage is not expensive and you should be able to cover the cost. It is better to have secrets hosted on your CI test machines (Sometimes, people put secrets in the Github code, it is convenient, but it is not safe). As far as I know, GCP does not charge for downloading, and AWS only charge for cross-region download. It is also OK to make buckets public if you want to share the use case on the Kuzu cloud (the one on your website demo).
> 
> For cloud import I loaded 180MB snappy parquet using 11s from local SSD and 14s from GCS bucket. So I would say copying from GCS is pretty fast. For compressed CSV I tested on Google Cloud Storage (GCS) and loaded SF30 LDBC BI dataset and import comment.**csv.gz** files. it worked.

Thanks! @yuchenZhangTG . This looks promising. We will set up a testing pipeline for this asap after our coming release next week.
To have a work around for the third party dependency problem for now, is it possible to set up a new cmake option `ENABLE_HTTPS`, and pass it into the arrow cmake file? So we don't introduce extra system installed dependency when the option is not enabled.
In the longer term, we will set up an extension mechanism similar to what [DuckDB](https://duckdb.org/docs/extensions/httpfs) does, and offload cloud part to extensions. 
Let us know if you have better ideas!

yuchenZhangTG:
> 

The option looks good to me. 

The option in DuckDB is HTTPFS, so it is HTTP (support http) + FS (support AWS S3). My feature does not support HTTP but S3 and GCS. So it is better call it `ENABLE_FILESYSTEM`.

ray6080:
> The option in DuckDB is HTTPFS, so it is HTTP (support http) + FS (support AWS S3). My feature does not support HTTP but S3 and GCS. So it is better call it `ENABLE_FILESYSTEM`.
> 
You're right. How about we be more explicit and call this `ENABLE_CLOUD_FS`? so we can differential it from the local fs.

> DuckDB write their own HTTPFS extension. The extension is separatable from their package, and user install it using `load httpfs` command. However, for Kuzu, where we can simply utilize file system in Apache Arrow, we need to reload the whole arrow package.
> 
Yep. I agree that it's not a good idea to do the fs extension with arrow. We have a battle plan to remove arrow inside the team, that's why i mentioned the extension idea.

> Another solution For building from source, we turn on FileSystem in Arrow by default. If CURL package is not available we turn off FileSystem and give a warning message. If `ENABLE_FILESYSTEM` is true and CURL is not available, we throw an error, and ask for sudo installation of CURL lib.

This sounds good to me. @mewim what do you think?
> For pre-compiled package, I suggested the build with filesystem.

Agreed!

mewim:
@ray6080 @yuchenZhangTG This sounds good to me. I double checked and it seems that we have installed `curl` on the CI runners: https://github.com/kuzudb/kuzu/blob/5b10a0f47b27027d5d23c802c353fb739652ffda/scripts/dockerized-ci-tests-runner/Dockerfile#L6. I am not sure why the previous build did not go through. Is there another library required to make this work?

yuchenZhangTG:
@mewim 
I mean curllib not curl.
```
sudo apt-get install libcurl4-openssl-dev
```
add libcurl4-openssl-dev will resolve the issue.

src/binder/bind/bind_copy.cpp
```diff
@@ -40,23 +40,56 @@ std::unique_ptr<BoundStatement> Binder::bindCopyClause(const Statement& statemen
...
-        throw BinderException{StringUtils::string_format("Invalid file path: {}.", filePaths[0])};
+        throw BinderException{
+            StringUtils::string_format("Invalid file path: {}.", fileUriOrPaths[0])};
     }
     return boundFilePaths;
 }
 
+void Binder::throwExceptionIfNotOK(const arrow::Status& status) {
+    if (!status.ok()) {
+        throw CopyException(status.ToString());
```
ray6080, line 89: Change the exception to `BinderException`.

src/include/binder/binder.h
```diff
@@ -1,10 +1,11 @@
 #pragma once
-
 #include "binder/query/bound_regular_query.h"
 #include "common/copier_config/copier_config.h"
 #include "expression_binder.h"
 #include "parser/query/regular_query.h"
 #include "query_normalizer.h"
+#include <arrow/api.h>
+#include <arrow/filesystem/api.h>
```
ray6080, line 8: I hope we don't introduce arrow headers in the header file, as this will be propagated to `main` module, too.
Can we add a forward declaration of `arrow::Status` instead? Or move the `throwExceptionIfNotOK` function to cpp file.

src/include/storage/copier/table_copy_executor.h
ray6080, line 1: Hmm, can you double check if we need this file? I think we've refactored `TableCopyExecutor` to `TableCopyUtils`.

ray6080 (commented):
Thanks @yuchenZhangTG ! The PR looks good to me. I hope to take another quick look before we merge.
Btw, can you take a look at our CLA file? And please add "I have read and agree to the terms under CLA.md." to the PR description if you're fine with it. Thanks!

external/arrow/apache_arrow.cmake
```diff
@@ -33,3 +47,9 @@ ExternalProject_Add(apache_arrow
         -DARROW_BUILD_UTILITIES=OFF -DARROW_BUILD_TESTS=OFF -DARROW_ENABLE_TIMING_TESTS=OFF -DARROW_FUZZING=OFF
         <SOURCE_DIR>/cpp
         UPDATE_COMMAND "")
+
+
+
+if(NOT ${CURL_FOUND})
+        message(WARNNING ": AWS S3 and Google Cloud Storage are turned off. To enable it, install system curllib and rebuild.")
+endif()
```
mewim, line 55: Add a newline here

src/common/copier_config/copier_config.cpp
```diff
@@ -39,5 +39,21 @@ std::string CopyDescription::getFileTypeName(FileType fileType) {
...
 }
 
+CopyDescription::FileType CopyDescription::getFileType(std::string& fileName) {
+    if (fileName.ends_with("parquet"))
+        return FileType::PARQUET;
+    if (fileName.ends_with("npy"))
+        return FileType::NPY;
+    if (fileName.ends_with("csv"))
+        return FileType::CSV;
+    // for "filename.csv.gz"
```
mewim, line 49: This does not seem to be a very robust way of checking `.csv.gz`. It seems to match all of `.csv.xxx`. I think we should explicitly check for extension.

mewim:
I have redeployed all the CI runners with `libcurl`

Makefile
```diff
@@ -5,6 +5,7 @@ FORCE_COLOR=
 NUM_THREADS=
 SANITIZER_FLAG=
 ROOT_DIR=$(CURDIR)
+ENABLE_FS=
```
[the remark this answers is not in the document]
yuchenZhangTG: changed to `ENABLE_REMOTE_FS`, because HDFS can be easily added in the future.

src/include/binder/binder.h
yuchenZhangTG: #include <arrow/api.h>
#include <arrow/filesystem/api.h>

are moved to the CPP files.

yuchenZhangTG:
> Thanks @yuchenZhangTG . Can you manually squash your commits into one? That's a convention we're following. Then I think it's ready to merge!

Sounds great.

ray6080 (approved):

ray6080 (commented):
@yuchenZhangTG It seems there is something wrong with the asan test run, which lasts unusually long. Better let's figure out what happens before merging. 

Meanwhile, I've set up a public GCS bucket here (https://storage.googleapis.com/kuzu-test/tinysnb) with csv files under [here](https://github.com/kuzudb/kuzu/tree/master/dataset/tinysnb). Can you verify if this is good for use and help us to set up a test case on this?
My idea is we create a new dataset directory `tinysnb-cloud` under our [dataset](https://github.com/kuzudb/kuzu/tree/master/dataset) directory with the `schema.cypher` and `copy.cypher` file, but replace COPY statements in `copy.cypher`. Then I will add a new test file under our [copy tests](https://github.com/kuzudb/kuzu/tree/master/test/test_files/copy), which will automatically trigger copying the dataset on cloud into our database and run a bunch of queries.
Does this sound good to you?

yuchenZhangTG:
> @yuchenZhangTG It seems there is something wrong with the asan test run, which lasts unusually long. Better let's figure out what happens before merging.
> 
> Meanwhile, I've set up a public GCS bucket here (https://storage.googleapis.com/kuzu-test/tinysnb) with csv files under [here](https://github.com/kuzudb/kuzu/tree/master/dataset/tinysnb). Can you verify if this is good for use and help us to set up a test case on this? My idea is we create a new dataset directory `tinysnb-cloud` under our [dataset](https://github.com/kuzudb/kuzu/tree/master/dataset) directory with the `schema.cypher` and `copy.cypher` file, but replace COPY statements in `copy.cypher`. Then I will add a new test file under our [copy tests](https://github.com/kuzudb/kuzu/tree/master/test/test_files/copy), which will automatically trigger copying the dataset on cloud into our database and run a bunch of queries. Does this sound good to you?

When I work with nodejs test, the following `npx cmake-js print-cmakejs-include `
gives error `Unexpected token .`

Do you have idea why it happens ?

yuchenZhangTG:
@ray6080 
I see the copy tests are moved to end2end. Is ASAN still checked here.

I added a GCS test case in the end2end copy test. It seems that GCS anonymous access need to add a username `anonymous` before the bucket name
e.g.:
`COPY person FROM "gs://anonymous@kuzu-test/tinysnb/vPerson.csv"(HEADER=true);`

test/test_files/copy/copy_gcs.test
```diff
@@ -0,0 +1,30 @@
+-GROUP CopyGcsTest
+-DATASET CSV tinysnb
+
+--
+
+# The purpose of this test is to make sure that we correctly initialize the rel table for the
```
ray6080, line 6: Can you remove the comment here? I don't think it's relevant.

test/test_files/copy/copy_gcs.test
```diff
@@ -0,0 +1,30 @@
...
+                                                          comments STRING[], MANY_MANY)
+---- ok
+-STATEMENT create rel table meets (FROM person TO person, MANY_ONE);
+---- ok
+-STATEMENT COPY person FROM "gs://anonymous@kuzu-test/tinysnb/vPerson.csv"(HEADER=true);
+---- ok
+-STATEMENT COPY knows FROM "gs://anonymous@kuzu-test/tinysnb/eKnows.csv"(HEADER=true);
+---- ok
+-QUERY MATCH (:person)-[:knows]->(:person) return count(*)
+---- ok
```
ray6080, line 30: Can you rewrite the last test query as following?
```
-QUERY MATCH (:person)-[:knows]->(:person) return count(*)
---- 1
14
```

test/test_files/copy/copy_gcs.test
```diff
@@ -0,0 +1,30 @@
+-GROUP CopyGcsTest
+-DATASET CSV tinysnb
```
ray6080, line 2: Change "tinysnb" to "empty", since we want to start from an empty db from scratch here.

test/test_files/copy/copy_gcs.test
```diff
@@ -0,0 +1,30 @@
...
+
+-STATEMENT create rel table knows (FROM person TO person, date DATE, meetTime
+                                                          TIMESTAMP, validInterval INTERVAL,
+                                                          comments STRING[], MANY_MANY)
+---- ok
+-STATEMENT create rel table meets (FROM person TO person, MANY_ONE);
+---- ok
+-STATEMENT COPY person FROM "gs://anonymous@kuzu-test/tinysnb/vPerson.csv"(HEADER=true);
+---- ok
+-STATEMENT COPY knows FROM "gs://anonymous@kuzu-test/tinysnb/eKnows.csv"(HEADER=true);
```
ray6080, line 27: Remove "(HEADER=true)" in copy knows statement as the file doesn't contain a header.

test/test_files/copy/copy_gcs.test
```diff
@@ -0,0 +1,30 @@
...
+              registerTime TIMESTAMP, lastJobDuration interval, workedHours INT64[], 
+              usedNames STRING[], courseScoresPerTerm INT64[][], grades INT64[4], height float, 
+              PRIMARY KEY (ID));
+---- ok
+
+-STATEMENT create rel table knows (FROM person TO person, date DATE, meetTime
+                                                          TIMESTAMP, validInterval INTERVAL,
+                                                          comments STRING[], MANY_MANY)
+---- ok
+-STATEMENT create rel table meets (FROM person TO person, MANY_ONE);
```
ray6080, line 23: Remove this statement?

ray6080:
> @ray6080 I see the copy tests are moved to end2end. Is ASAN still checked here.
> 
> I added a GCS test case in the end2end copy test. It seems that GCS anonymous access need to add a username `anonymous` before the bucket name e.g.: `COPY person FROM "gs://anonymous@kuzu-test/tinysnb/vPerson.csv"(HEADER=true);`

Thanks @yuchenZhangTG . I left some comments to fix the test. Also I found two small issues:
1. without adding "anonymous", the COPY (`COPY person FROM "gs://kuzu-test/tinysnb/vPerson.csv"(HEADER=true);`) seems to be in stuck for a very long time, is this expected? Is it possible to do checks and timeouts here through arrow?
2. besides find package check on `CURL`, should we also check on `crypto (openssl)`? On my local M1 machine, the curl package is found, but the build still fails, i guess it's because of the lack of `crypto` (find package for openssl fails on my local).
```
Undefined symbols for architecture arm64:
  "_CFAllocatorCreate", referenced from:
      _aws_wrapped_cf_allocator_new in libarrow_bundled_dependencies.a(allocator.c.o)
  "_CFRelease", referenced from:
      absl::lts_20211102::time_internal::cctz::local_time_zone() in libarrow_bundled_dependencies.a(time_zone_lookup.cc.o)
      _aws_wrapped_cf_allocator_destroy in libarrow_bundled_dependencies.a(allocator.c.o)
  "_CFStringGetCString", referenced from:
      absl::lts_20211102::time_internal::cctz::local_time_zone() in libarrow_bundled_dependencies.a(time_zone_lookup.cc.o)
  "_CFStringGetLength", referenced from:
      absl::lts_20211102::time_internal::cctz::local_time_zone() in libarrow_bundled_dependencies.a(time_zone_lookup.cc.o)
  "_CFStringGetMaximumSizeForEncoding", referenced from:
      absl::lts_20211102::time_internal::cctz::local_time_zone() in libarrow_bundled_dependencies.a(time_zone_lookup.cc.o)
  "_CFTimeZoneCopyDefault", referenced from:
      absl::lts_20211102::time_internal::cctz::local_time_zone() in libarrow_bundled_dependencies.a(time_zone_lookup.cc.o)
  "_CFTimeZoneGetName", referenced from:
      absl::lts_20211102::time_internal::cctz::local_time_zone() in libarrow_bundled_dependencies.a(time_zone_lookup.cc.o)
  "_EVP_MD_size", referenced from:
      google::cloud::storage::v1_42_0::internal::MD5HashFunction::Finish() && in libarrow_bundled_dependencies.a(hash_function_impl.cc.o)
  "___CFConstantStringClassReference", referenced from:
      CFString in libarrow_bundled_dependencies.a(allocator.c.o)
ld: symbol(s) not found for architecture arm64
```

yuchenZhangTG:
1. 

> > @ray6080 I see the copy tests are moved to end2end. Is ASAN still checked here.
> > I added a GCS test case in the end2end copy test. It seems that GCS anonymous access need to add a username `anonymous` before the bucket name e.g.: `COPY person FROM "gs://anonymous@kuzu-test/tinysnb/vPerson.csv"(HEADER=true);`
> 
> Thanks @yuchenZhangTG . I left some comments to fix the test. Also I found two small issues:
> 
> 1. without adding "anonymous", the COPY (`COPY person FROM "gs://kuzu-test/tinysnb/vPerson.csv"(HEADER=true);`) seems to be in stuck for a very long time, is this expected? Is it possible to do checks and timeouts here through arrow?
> 2. besides find package check on `CURL`, should we also check on `crypto (openssl)`? On my local M1 machine, the curl package is found, but the build still fails, i guess it's because of the lack of `crypto` (find package for openssl fails on my local).
> 
> ```
> Undefined symbols for architecture arm64:
>   "_CFAllocatorCreate", referenced from:
>       _aws_wrapped_cf_allocator_new in libarrow_bundled_dependencies.a(allocator.c.o)
>   "_CFRelease", referenced from:
>       absl::lts_20211102::time_internal::cctz::local_time_zone() in libarrow_bundled_dependencies.a(time_zone_lookup.cc.o)
>       _aws_wrapped_cf_allocator_destroy in libarrow_bundled_dependencies.a(allocator.c.o)
>   "_CFStringGetCString", referenced from:
>       absl::lts_20211102::time_internal::cctz::local_time_zone() in libarrow_bundled_dependencies.a(time_zone_lookup.cc.o)
>   "_CFStringGetLength", referenced from:
>       absl::lts_20211102::time_internal::cctz::local_time_zone() in libarrow_bundled_dependencies.a(time_zone_lookup.cc.o)
>   "_CFStringGetMaximumSizeForEncoding", referenced from:
>       absl::lts_20211102::time_internal::cctz::local_time_zone() in libarrow_bundled_dependencies.a(time_zone_lookup.cc.o)
>   "_CFTimeZoneCopyDefault", referenced from:
>       absl::lts_20211102::time_internal::cctz::local_time_zone() in libarrow_bundled_dependencies.a(time_zone_lookup.cc.o)
>   "_CFTimeZoneGetName", referenced from:
>       absl::lts_20211102::time_internal::cctz::local_time_zone() in libarrow_bundled_dependencies.a(time_zone_lookup.cc.o)
>   "_EVP_MD_size", referenced from:
>       google::cloud::storage::v1_42_0::internal::MD5HashFunction::Finish() && in libarrow_bundled_dependencies.a(hash_function_impl.cc.o)
>   "___CFConstantStringClassReference", referenced from:
>       CFString in libarrow_bundled_dependencies.a(allocator.c.o)
> ld: symbol(s) not found for architecture arm64
> ```

yuchenZhangTG:
> > @ray6080 I see the copy tests are moved to end2end. Is ASAN still checked here.
> > I added a GCS test case in the end2end copy test. It seems that GCS anonymous access need to add a username `anonymous` before the bucket name e.g.: `COPY person FROM "gs://anonymous@kuzu-test/tinysnb/vPerson.csv"(HEADER=true);`
> 
> Thanks @yuchenZhangTG . I left some comments to fix the test. Also I found two small issues:
> 
> 1. without adding "anonymous", the COPY (`COPY person FROM "gs://kuzu-test/tinysnb/vPerson.csv"(HEADER=true);`) seems to be in stuck for a very long time, is this expected? Is it possible to do checks and timeouts here through arrow?
> 2. besides find package check on `CURL`, should we also check on `crypto (openssl)`? On my local M1 machine, the curl package is found, but the build still fails, i guess it's because of the lack of `crypto` (find package for openssl fails on my local).
> 
> ```
> Undefined symbols for architecture arm64:
>   "_CFAllocatorCreate", referenced from:
>       _aws_wrapped_cf_allocator_new in libarrow_bundled_dependencies.a(allocator.c.o)
>   "_CFRelease", referenced from:
>       absl::lts_20211102::time_internal::cctz::local_time_zone() in libarrow_bundled_dependencies.a(time_zone_lookup.cc.o)
>       _aws_wrapped_cf_allocator_destroy in libarrow_bundled_dependencies.a(allocator.c.o)
>   "_CFStringGetCString", referenced from:
>       absl::lts_20211102::time_internal::cctz::local_time_zone() in libarrow_bundled_dependencies.a(time_zone_lookup.cc.o)
>   "_CFStringGetLength", referenced from:
>       absl::lts_20211102::time_internal::cctz::local_time_zone() in libarrow_bundled_dependencies.a(time_zone_lookup.cc.o)
>   "_CFStringGetMaximumSizeForEncoding", referenced from:
>       absl::lts_20211102::time_internal::cctz::local_time_zone() in libarrow_bundled_dependencies.a(time_zone_lookup.cc.o)
>   "_CFTimeZoneCopyDefault", referenced from:
>       absl::lts_20211102::time_internal::cctz::local_time_zone() in libarrow_bundled_dependencies.a(time_zone_lookup.cc.o)
>   "_CFTimeZoneGetName", referenced from:
>       absl::lts_20211102::time_internal::cctz::local_time_zone() in libarrow_bundled_dependencies.a(time_zone_lookup.cc.o)
>   "_EVP_MD_size", referenced from:
>       google::cloud::storage::v1_42_0::internal::MD5HashFunction::Finish() && in libarrow_bundled_dependencies.a(hash_function_impl.cc.o)
>   "___CFConstantStringClassReference", referenced from:
>       CFString in libarrow_bundled_dependencies.a(allocator.c.o)
> ld: symbol(s) not found for architecture arm64
> ```

@ray6080 

1. I found the same issue. GCS anonymouse will get data from an `http://` URI, while the authorized is from a `gs://` URI. Because difference in the host name, it requires `anonymous` for a anonymous access. I got the same behavior for without anonymous. Apache arrow default `retry_limit_seconds` is 15min. It means that it retry for 15 mins if it fails. Cloud applications usually have large timeout values because they somtimes allow network glitch.

2. I have not try build on arm or M1 chip. You can add the check to this PR.

ray6080:
Thanks @yuchenZhangTG , I think we can merge after you add the openssl check for arrow builds and squashed commits. The win test fails due to not able to find `OpenSSL` and `cURL`, which we will try to fix soon.

mewim:
@ray6080 I installed curl and openssl on Windows runner. Now the code compiles on Windows but the test seems to fail. Also, the asan test seems to timeout.

yuchenZhangTG:
> Thanks @yuchenZhangTG , I think we can merge after you add the openssl check for arrow builds and squashed commits. The win test fails due to not able to find `OpenSSL` and `cURL`, which we will try to fix soon.

Great I already add the openssl check. I am not sure how to fix asan.

benjaminwinger:
I took a look at the rust test failure and was able to get it to work with this change:
```
diff --git a/tools/rust_api/build.rs b/tools/rust_api/build.rs
index 268a9e5b..01fd8af7 100644
--- a/tools/rust_api/build.rs
+++ b/tools/rust_api/build.rs
@@ -67,8 +67,9 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
         // openssl-sys has better cross-platform logic, but just using that doesn't work.
         if env::var("KUZU_TESTING").is_ok() {
             println!("cargo:rustc-link-lib=dylib=ssl");
-            println!("cargo:rustc-link-lib=dylib=crypto");
         }
+        println!("cargo:rustc-link-lib=dylib=crypto");
+        println!("cargo:rustc-link-lib=dylib=curl");

         println!("cargo:rustc-link-lib=static=parquet");
         println!("cargo:rustc-link-lib=static=arrow"); 
```

However applying that now will just give you a merge conflict since there were some changes to get the rust api tests running on windows since this branch was last updated (and presumably there will be similar issues on windows).

Plus, it really should be conditional on whether or not `ENABLE_REMOTE_FS` is set, as well as if it's enabled automatically when openssl/curl are detected, but perfecting mimicking the cmake code in the rust build file might be difficult. 
I'm going to see if there's a way we can have cmake export information about which libraries were linked so that the rust build can use that instead.

benjaminwinger:
Google cloud cpp is at least getting to the build stage, but it's failing for some reason still (it was working for me locally). I also still need to finish making the test be run only when the feature is enabled.

On windows, the only way it seems that I could get it to detect curl in all of the kuzu build, the arrow build, and the google-cloud-cpp build, was defining `CURL_INCLUDE_DIR` and `CURL_LIBRARY` and passing those values to the arrow build, which then had to be patched because, while it's passing the corresponding values for OpenSSL to google cloud cpp, it's not for CURL. This patch should probably be upstreamed (patch is written as a python file so that it can fail if cmake applies it multiple times without breaking the build).
For some reason, backslashes in the `CURL_*` paths aren't being normalized to forward slashes and are causing build errors. It works fine if you just use forward slashes in those paths.

benjaminwinger:
This also seems to be cause the arrow build to approach the maximum path length on windows. I suspect that the rust remote-fs feature will be difficult to build on windows as a result. 

But shortening the build directory paths a little helped, and it's succeeding now with the exception of coverage.

ray6080:
@yuchenZhangTG @benjaminwinger I'm proposing to close this PR as this has been a while, and we've now shifted to a new csv reader, instead of arrow. So this PR is not aligned with it anymore. Happy to re-open it if any of you are interested in continuing on this with the new csv reader, though a new series of PR for that sounds more reasonable.

ray6080:
Close this for now as it is outdated for a while.

PR #1584, opened 2023-05-29
Convert exception to end to end test

rfdavid:
This PR implements the following:

- Adds support for `-BEGIN_WRITE_TRANSACTION` within the test file
- Splits `exception.test` into multiple files
- Raises an exception if the test file is invalid
- Deletes `e2e_exception_test.cpp` file


test/runner/e2e_read_test.cpp
```diff
@@ -45,6 +45,8 @@ void parseAndRegisterTestGroup(const std::string& path) {
                     return new EndToEndReadTest(dataset, bufferPoolSize, std::move(testStatements));
                 });
         }
+    } else {
+        throw Exception("Invalid test file");
```
ray6080, line 49: Can you add the file path into the exception here?
Also, I would add a `TestException` class and throw `TestException` here. See `common/exception.h`.

ray6080 (approved):

PR #1587, opened 2023-05-29
Use GetFileSizeEX instead of GetFileSize on windows

benjaminwinger:
Following up https://github.com/kuzudb/kuzu/pull/1573#discussion_r1207504782. I also added error handling.

### DIFF
--- a/src/common/file_utils.cpp
+++ b/src/common/file_utils.cpp
@@ -26,7 +26,14 @@ FileInfo::~FileInfo() {
 
 int64_t FileInfo::getFileSize() {
 #ifdef _WIN32
-    return GetFileSize((HANDLE)handle, nullptr);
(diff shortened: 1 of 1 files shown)


acquamarin (approved):

benjaminwinger:
[LARGE_INTEGER](https://learn.microsoft.com/en-us/windows/win32/api/winnt/ns-winnt-large_integer-r1) with `QuadPart` seems to be windows 10+ only. I think the LowPart/HighPart version was used on older versions of windows ([OLD_LARGE_INTEGER](https://learn.microsoft.com/en-us/windows/win32/api/subauth/ns-subauth-old_large_integer)). 

I think everything prior to this works on versions as old as windows XP (at least, the windows APIs anyway).

PR #1589, opened 2023-05-29
Add LDBC interactive short queries to test

aziz-mu:
Adding the LDBC sf0.1 dataset, as well as 5/7 tests of the SNB Interactive short tests.

- Dataset is about 65 MB
- Interactive short queries 2 and 6 were removed due to having `[REPLY_OF*0..]` , which isn't possible to do currently.
- In query 4, the `coalesce()` function available in Neo4J was replaced with a `CASE WHEN` statement, since it only has 2 arguments the expected behaviour stays the same
- Arbitrary IDs that existed in the database were chosen for the queries. Query ref can be found at https://github.com/ldbc/ldbc_snb_interactive_impls/tree/main/cypher/queries
- All results were validated using Neo4J


test/test_files/ldbc/ldbc-interactive/interactive-short.test
```diff
@@ -0,0 +1,46 @@
...
+32985348834375|Alfred|Hoffmann|20120803220915016
+30786325578932|Alexander|Hleb|20120529180010845
+28587302323722|Joao|Oliveira|20120415042428397
+26388279067534|Emperor of Brazil|Dom Pedro II|20120119192154433
+19791209301592|Michal|Kaminski|20110917214017976
+10995116278468|Anders|Andersson|20110102223928664
+2199023256437|Rudolf|Engel|20100413164422450
+
+-NAME IS4
+-QUERY MATCH (m:Post:Comment {id: 481036337190}) RETURN m.creationDate as messageCreationDate, CASE WHEN m.content is NULL THEN m.imageFile ELSE m.content END AS messageContent;
```
andyfengHKU, line 26: Can u add a comment saying we should support this as `coalesce()`

test/test_files/ldbc/ldbc-interactive/interactive-short.test
```diff
@@ -0,0 +1,46 @@
+-GROUP LDBCTest
+-DATASET ldbc-sf01
+-BUFFER_POOL_SIZE 1073741824
+
+--
+
+-CASE LDBCInteractiveShort
```
andyfengHKU, line 7: For IS2 & 6,

can u replace `replyOf*` with a bounded recursive pattern e.g. `replyOf*1..2`. Add document how changed the query. I think we won't be able to support Kleene star for a long time. Meanwhile it will be good to have some similar query running.

andyfengHKU (approved):

PR #1595, opened 2023-05-30
Build Python wheels on Windows

benjaminwinger:
Tested here: https://github.com/kuzudb/kuzu/actions/runs/5126793944/jobs/9221725746 (I overrode the workflow settings to have it run on that branch, which, along with faking the version, is the differences from this PR).

cibuildwheel is run manually because the quoted arguments passed to it in the windows/powershell version of the cibuildwheel action don't work properly for options which are left empty (on kuzu's windows runner at least).

~~TODO: double-check that the wheels work.~~ Done. All the wheels contain the extension and I installed and tested the python3.11 wheel.

I rewrote the `package_tar.sh` in python so that it would run more easily on Windows. 

It was also necessary to fix `make clean` on windows, as it is called in `setup.py` and would return an error if the directories to clean don't exist.

And while I'm fairly sure that `ARCHIVE_OUTPUT_DIRECTORY` worked to control the output directory of the python extension when I was just testing building the extension before, the extension wasn't being stored in the correct directory again, which setting `RUNTIME_OUTPUT_DIRECTORY` fixed (setting all three should be exhaustive).


mewim:
Could you please trigger `Build-Linux-Wheels` and `Build-Mac-Wheels` on CI to make sure it does not break any existing workflow? As long as it still works for both Mac and Linux we can merge it.

mewim:
It seems this breaks both Mac and Linux wheel build:  https://github.com/kuzudb/kuzu/actions/runs/5150769886 and https://github.com/kuzudb/kuzu/actions/runs/5150768166. Could you please look into it?

mewim (approved):

PR #1597, opened 2023-05-31
Convert e2e_update_node.cpp to e2e test files

rfdavid:
This PR performs the following:

- Converts `e2e_update_node.cpp` into `tinysnb/update_node/create.test`, `tinysnb/update_node/delete.test` and `tinysnb/update_node/set.test`.
- Adds `-DEFINE` to create variables
- Adds `ARANGE start end` to create lists
Example:
```
-DEFINE MY_LIST ARANGE 0 10
# MY_LIST = [0,1,2,3,4,5,6,7,8,9,10]
-STATEMENT MATCH (a:person) WHERE a.ID=0 SET a.Pref ="${MY_LIST}"
...
# Not really a useful example but just to illustrate
-DEFINE R_EXCEPTION "Runtime exception: "
-CASE DivideBy0Error
-QUERY RETURN 1 / 0
---- error
${R_EXCEPTION} Divide by zero.

-CASE ModuloBy0Error
-QUERY RETURN 1 % 0
---- error
${R_EXCEPTION}  Modulo by zero.
```


ray6080 (approved):
1. Can you add commented tests to the test file (skipped or commented) too? I'm afraid we will forget these tests later when we turn them on.
2. Let's discuss the `DEFINE` syntax a bit more, I hope it can be more formally defined and we can extend it later. for example, `-DEFINE R_EXCEPTION Runtime exception: ` is not a clear syntax to me. It should be something along `-DEFINE R_EXCEPTION "Runtime exception: "`, or `-DEFINE R_EXCEPTION STRING("Runtime exception: ")`, so its data type is clear. I feel like this is important and we should think a bit more on this.

ray6080:
One feedback, though not related to this PR, is that when a test cases contain multiple test units (each identified by a NAME), is there a way that we can print which test unit fails in the log? Currently I believe we only report which test case fails, but it's unclear where the test case breaks.

PR #1611, opened 2023-06-02
Test Framework: Support CSV to Parquet conversion

rfdavid:
## Introduction 

This commit adds the capability of converting CSV datasets to parquet datasets on the fly for the tests by simply using the following command in the test header: 
```
-DATASET PARQUET CSV_TO_PARQUET(dataset)
```

To load dataset without any conversion:

```
-DATASET CSV tinysnb
-DATASET NPY npy-1d
-DATASET PARQUET demo-db/parquet
```

## How it works

1. Create a directory `dataset/parquet_temp/tinysnb`
2. Copy `schema.cypher` to the created directory
3.  Read and parse `copy.cypher`, extract the csv file names, header/no header information, csv delimiter
4. Create a new copy.cypher with the new COPY commands and paths
5. Convert .csv files to .parquet files to the parquet temp directory
6. Set dataset path to parquet temp directory
7. Remove parquet temp directory after all tests run. 

Related to #1521


src/include/common/file_utils.h
```diff
@@ -63,6 +63,12 @@ class FileUtils {
         return std::filesystem::exists(path);
     }
 
+    static inline void createDirIfNotExists(const std::string& path) {
```
ray6080, line 66: Better to move this function close to `static void createDir(const std::string& dir);`.

test/include/test_helper/test_helper.h
```diff
@@ -40,9 +40,14 @@ class TestHelper {
     static constexpr char E2E_TEST_FILES_DIRECTORY[] = "test/test_files";
     static constexpr char SCHEMA_FILE_NAME[] = "schema.cypher";
     static constexpr char COPY_FILE_NAME[] = "copy.cypher";
+    static constexpr char PARQUET_TEMP_DATASET_PATH[] = "dataset/parquet_temp/";
 
     static std::string getTmpTestDir() { return appendKuzuRootPath("test/unittest_temp/"); }
 
+    static std::string appendParquetDatasetTempDir(const std::string& dataset) {
+        return TestHelper::appendKuzuRootPath(TestHelper::PARQUET_TEMP_DATASET_PATH + dataset);
```
ray6080, line 48: Better to call `FileUtils::joinPath` to concatenation file paths?
Also in `appendKuzuRootPath`.

test/include/test_runner/csv_to_parquet_converter.h
```diff
@@ -0,0 +1,47 @@
...
+    static std::vector<CopyCommandInfo> readCopyCommandsFromCopyCypherFile(
+        const std::string& dataset);
+
+    static void convertCSVFilesToParquet(
+        const std::vector<CSVToParquetConverter::CopyCommandInfo>& copyCommands);
+
+    static CopyCommandInfo createCopyCommandInfo(
+        const std::string& dataset, const std::string& copyStatement);
+
+    static arrow::Status RunCSVToParquetConversion(const std::string& inputFile,
```
ray6080, line 31: Is there a special reason this function name starts with upper case?

test/test_runner/csv_to_parquet_converter.cpp
```diff
@@ -0,0 +1,126 @@
...
+    ARROW_ASSIGN_OR_RAISE(csvTable, csvReader->Read());
+    ARROW_ASSIGN_OR_RAISE(outFileStream, arrow::io::FileOutputStream::Open(outputFile));
+    PARQUET_THROW_NOT_OK(parquet::arrow::WriteTable(
+        *csvTable, arrow::default_memory_pool(), outFileStream, csvTable->num_rows()));
+    return arrow::Status::OK();
+}
+
+void CSVToParquetConverter::copySchema(
+    const std::string& csvDatasetPath, const std::string& parquetDatasetPath) {
+    auto parquetSchemaFile = parquetDatasetPath + "/" + std::string(TestHelper::SCHEMA_FILE_NAME);
```
ray6080, line 42: Ditto here. Use `FileUtils::joinPath`.

test/test_runner/csv_to_parquet_converter.cpp
```diff
@@ -0,0 +1,126 @@
...
+    if (!FileUtils::fileOrPathExists(parquetSchemaFile)) {
+        try {
+            std::filesystem::copy(csvDatasetPath + "/schema.cypher", parquetDatasetPath);
```
ray6080, line 45: Better add this into `FileUtils` and call `FileUtils::copyFile()`.

src/common/string_utils.cpp
```diff
@@ -31,6 +31,12 @@ std::vector<std::string> StringUtils::splitBySpace(const std::string& input) {
     return result;
 }
 
+std::string StringUtils::extractSubstring(std::string& str, char delimiter) {
```
ray6080, line 34: I don't think the function name is accurate here. As a general substring extractor (in `StringUtils`), it seems to assume that there are exactly two delimiters, and extract the substring between them, which only covers a subset of general substring extracting.

For now, this is only for the special usage that you are extracting path from the COPY statement. 
I would just put it under `CSVToParquetConverter`, and call it `extractPath(std::string& str)` maybe.

test/test_runner/csv_to_parquet_converter.cpp
```diff
@@ -0,0 +1,126 @@
...
+namespace testing {
+
+arrow::Status CSVToParquetConverter::RunCSVToParquetConversion(
+    const std::string& inputFile, const std::string& outputFile, char delimiter, bool hasHeader) {
+    std::shared_ptr<arrow::io::FileOutputStream> outFileStream;
+    std::shared_ptr<arrow::io::ReadableFile> infile;
+    std::shared_ptr<arrow::Table> csvTable;
+    ARROW_ASSIGN_OR_RAISE(infile, arrow::io::ReadableFile::Open(inputFile));
+    auto readOptions = arrow::csv::ReadOptions::Defaults();
+    auto parseOptions = arrow::csv::ParseOptions::Defaults();
```
ray6080, line 27: My understanding is that you are relying on arrow's auto csv reader to figure out data types for each csv column. I'm not sure if it's always as expected, especially when it comes to dates/timestamps/nested data types. I believe arrow converts some nested data types into strings. Can you double check the metadata of generated parquet files?
Ideally, we need to figure out data types for each column, and let the arrow reader be aware of that.
I'm fine with this auto reader for now if it works mostly as expected.

Actually I think the best way is we should add native support of exporting tables to parquet files in Cypher.
Then the conversion would be much easier.

Besides this, I'm curious what's the default row group size when we dump data into parquet?

ray6080 (approved):

scripts/parquet/csv_to_parquet.py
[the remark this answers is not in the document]
rfdavid: It might be useful if you want to generate a dataset to store in our codebase. Other than that, I can't think of anything. We can also remove `demo-db/parquet` too.

test/test_runner/csv_to_parquet_converter.cpp
rfdavid: That's right, I'm relying on arrow's auto csv reader and it is not much assertive. Dates and timestamps are being converted into strings.

PR #1634, opened 2023-06-05
Add [skip-benchmark] to skip benchmark on CI

rfdavid:
Skip benchmark tests when [skip-benchmark] is present in the commit message.
Sometimes we need to update files that don't necessarily need to run benchmarks but still need to run tests, for instance, the test framework.


rfdavid:
It doesn't make much difference, I will archive for now.

PR #1639, opened 2023-06-06
Remove glob cmake file

ray6080:
The use of INTERFACE here in glob requires a pretty high cmake version 3.19.
This PR simplified glob lib to not create an interface library, same as `concurrentqueue`, `nlohmann_json`, etc.

### DIFF
--- a/third_party/CMakeLists.txt
+++ b/third_party/CMakeLists.txt
@@ -1,6 +1,6 @@
 add_subdirectory(antlr4_runtime)
 add_subdirectory(antlr4_cypher)
+add_subdirectory(glob)
 add_subdirectory(utf8proc)
 add_subdirectory(pybind11)
 add_subdirectory(re2)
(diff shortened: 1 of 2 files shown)


ray6080:
> [This line](https://github.com/kuzudb/kuzu/blob/0248f056b494d60a6fc0a7bf8ce12c5a3dac5556/src/common/CMakeLists.txt#L24) should also be removed if the Glob interface library is being removed.
> 
> I think just replacing `add_library(Glob INTERFACE glob/glob.hpp)` with `add_library(Glob INTERFACE)` would also be sufficient; I don't think including the headers in an interface library actually provides any benefit other than metadata for certain IDEs (and even then, I think using `target_sources(INTERFACE ...)` achieves the same thing, and that's been supported since cmake 3.1).

Sg. I will replace it with `add_library(Glob INTERFACE)`.

PR #1645, opened 2023-06-06
Sanitize escape char in Transformer::transformSymbolicName

gaurav8297:
Escaped symbol name coming from the parser is of form "\`Some.Value\`". Therefore, we need to sanitize it such that we don't store the symbol with escape characters.

### DIFF
--- a/src/parser/transformer.cpp
+++ b/src/parser/transformer.cpp
@@ -914,7 +914,10 @@ std::string Transformer::transformSymbolicName(CypherParser::OC_SymbolicNameCont
     if (ctx.UnescapedSymbolicName()) {
         return ctx.UnescapedSymbolicName()->getText();
     } else if (ctx.EscapedSymbolicName()) {
-        return ctx.EscapedSymbolicName()->getText();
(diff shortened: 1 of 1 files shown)


src/parser/transformer.cpp
```diff
@@ -914,7 +914,14 @@ std::string Transformer::transformSymbolicName(CypherParser::OC_SymbolicNameCont
     if (ctx.UnescapedSymbolicName()) {
         return ctx.UnescapedSymbolicName()->getText();
     } else if (ctx.EscapedSymbolicName()) {
-        return ctx.EscapedSymbolicName()->getText();
+        std::string escapedSymbolName = ctx.EscapedSymbolicName()->getText();
+        // escapedSymbolName symbol will be of form "`Some.Value`". Therefore, we need to sanitize
+        // it such that we don't store the symbol with escape character.
+        if (escapedSymbolName.starts_with(ESCAPE_SYMBOL_CHAR) &
```
andyfengHKU, line 920: I guess we don't need to do this `if` check but should prune first and last char blindly?

andyfengHKU (approved):

PR #1647, opened 2023-06-07
Convert copy tests to end to end tests

rfdavid:
This PR converts `copy/` tests to end-to-end tests, and introduces the commands `REPEAT` and `<FILE>` support to the test framework:

Using `REPEAT` command:

```
-CASE LongStringError
-DEFINE AliceStr REPEAT 819 "Alice"

-NAME LongStringAlice
-QUERY MATCH (p:person) RETURN p.fName;
---- 1
${AliceStr}A
```

If your query output is too big and you desire to store the results in a file, save it inside `test/test_files/outputs` and use the following command to compare the results:

```
-QUERY MATCH (p0:person)-[r:knows]->(p1:person) WHERE p0.ID = 0 RETURN p0.ID, ID(r), p1.ID
---- 5001
<FILE>:copy_large_list_fwd_output.csv
```


test/test_files/copy/copy_long_string.test
```diff
@@ -0,0 +1,19 @@
+-GROUP CopyLongStringTest
+-DATASET CSV copy-fault-tests/long-string
+
+--
+
+-CASE LongStringError
+-DEFINE AliceStr REPEAT 819 "Alice"
```
ray6080, line 7: I assume you're following Postgres's `repeat` function? If so, let's switch the order of these two params. `REPEAT "Alice" 819` is more consistent with existing functions.
If not, I suggest the principle that we always follow existing syntaxes, instead of coin new ones.

test/test_runner/test_parser.cpp
```diff
@@ -204,6 +219,15 @@ std::string TestParser::parseCommand() {
     return params.substr(1, params.size() - 2);
 }
 
+std::string TestParser::parseCommandRepeat() {
```
ray6080, line 222: My understanding on `repeat` is that it should be limited to strings, so can you add checks here on the param that it is string literal (i.e., wrapped with `""`)?

ray6080 (approved):

src/common/string_utils.cpp
```diff
@@ -40,5 +40,17 @@ void StringUtils::replaceAll(
     }
 }
 
+std::string StringUtils::extractStringBetween(
+    const std::string& input, char delimiterStart, char delimiterEnd, bool includeDelimiter) {
```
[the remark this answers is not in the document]
rfdavid: In this case, it returns everything. However, I added a validation to return an empty string. I think it's better. I also added a quick test.

test/include/test_helper/test_helper.h
```diff
@@ -42,6 +42,7 @@ class TestHelper {
     static constexpr char COPY_FILE_NAME[] = "copy.cypher";
     static constexpr char PARQUET_TEMP_DATASET_PATH[] = "dataset/parquet_temp_";
     static constexpr char TMP_TEST_DIR[] = "test/unittest_temp_";
+    static constexpr char OUTPUTS_PATH[] = "/test/test_files/outputs/";
```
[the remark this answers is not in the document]
rfdavid: 1. fixed
2. no, fixed
3. fixed

test/test_runner/test_parser.cpp
rfdavid: added validation

PR #1654, opened 2023-06-09
Scan and parse test files only once

### LINKED ISSUE #1651: Test performance drops quickly after running LSQB test
I notice that before LSQB test, our average test cases takes around 0.15s. While after running LSQB test, all subsequent test cases take longer, average 0.6s on my machine. I wonder if this is because LSQB test bump BM size up and we don't reset BM size to a smaller size afterwards?

rfdavid:
This PR implements scanning and parsing the test files only once instead of every `e2e_test` run.

closes #1651

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@ bazel-*
 .clwb
 .cache
 .bazelrc
+test/test_files/test_list
 
 test_temp/
 build/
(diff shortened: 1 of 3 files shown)


ray6080 (approved):
I wonder if we should just not rely on `ctest`, but directly run `e2e_test` binary?

rfdavid:
> I wonder if we should just not rely on `ctest`, but directly run `e2e_test` binary?

Yes, we can directly run `e2e_test` binary by passing the test file or directory. We use `ctest` utility to organize what tests to run and provide a nice output. If we are going to run `e2e_test` directly, we will have to create a minimal utility to organize what tests to run, perform `e2e_test` passing the file, capture the output, summarize and print the results. We can have more control. It might be a good alternative.

PR #1686, opened 2023-06-16
Add windows precompiled binary workflow and shared library to pre-compiled binary archive

benjaminwinger:
I also modified the script which produces the pre-compiled binaries so that it includes both the shared library `.dll` and the import library `.lib` (needed for linking stuff against the `dll`; the `dll` is only needed at runtime as far as I know). The last release just included a libkuzu.lib (presumably just the static library? I didn't check; the pre-compiled-bin script wasn't working, so I'm guessing that was created manually).

One question though: should the libraries be `kuzu_shared.dll` and `kuzu_shared.lib` (to distinguish them from kuzu.lib, which is the name of the static library), or just be `kuzu.dll` and `kuzu.lib`?

### DIFF
--- a/.github/workflows/windows-precompiled-bin-workflow.yml
+++ b/.github/workflows/windows-precompiled-bin-workflow.yml
@@ -1,0 +1,42 @@
+name: Build-Windows-Precompiled-Binaries
+
+on:
+  workflow_dispatch:
+    inputs:
+      packageVersion:
+        description: "Version of output package"
+        required: true
+        default: "0.0.1"
+
+env:
+  PACKAGE_VERSION: ${{ github.event.inputs.packageVersion }}
+
+jobs:
+  build-precompiled-bin:
+    runs-on: self-hosted-windows
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Install networkx
+        run: python.exe -m pip install networkx --user
+
+      - name: Build precompiled binaries
+        run: python.exe build.py
+        working-directory: ./scripts/pre-compiled-bins/
+
+      - uses: actions/upload-artifact@v3
+        with:
+          name: libkuzu-${{ github.event.inputs.packageVersion }}-windows-x86_64
+          path: |
+            ./scripts/pre-compiled-bins/kuzu.h
+            ./scripts/pre-compiled-bins/kuzu.hpp
+            ./scripts/pre-compiled-bins/kuzu_shared.dll
+            ./scripts/pre-compiled-bins/kuzu_shared.lib
+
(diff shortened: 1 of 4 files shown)


mewim:
1. Could you please teat that all of the pre-compiled workflows are still working correctly by running `Build-Linux-Precompiled-Binaries` and `Build-Mac-Precompiled-Binaries`. You may not be able to test the Windows workflow itself before the PR has been merged but let's ensure it does not break the existing ones.
2. Before releasing 0.0.4, @acquamarin tested the `lib` and `dll` we built manually but it seems he cannot get the `dll` work. Could you please check if the `dll` itself is working by building [Introduction Examples](https://github.com/kuzudb/kuzudb.github.io/blob/main/docs/getting-started/introduction-examples.md). If the `dll` is working for you, could you please also update the documentation on how to make the `dll` work on Windows?
3. Since we did not get the `dll` working, we did not include it in the last release. I think DuckDB releases both the `lib` and `dll` for Windows, so we can follow that. The name `kuzu.dll` and `kuzu.lib` are fine.

mewim (commented):
1. Could you please teat that all of the pre-compiled workflows are still working correctly by running `Build-Linux-Precompiled-Binaries` and `Build-Mac-Precompiled-Binaries`. You may not be able to test the Windows workflow itself before the PR has been merged but let's ensure it does not break the existing ones.
2. Before releasing 0.0.4, @acquamarin tested the `lib` and `dll` we built manually but it seems he cannot get the `dll` work. Could you please check if the `dll` itself is working by building and running the [Introduction Examples](https://github.com/kuzudb/kuzudb.github.io/blob/main/docs/getting-started/introduction-examples.md). If the `dll` is working for you, could you please also update the documentation on how to make the `dll` work on Windows?
3. Since we did not get the `dll` working, we did not include it in the last release. I think DuckDB releases both the `lib` and `dll` for Windows, so we can follow that. The name `kuzu.dll` and `kuzu.lib` are fine.

benjaminwinger:
> Before releasing 0.0.4, @acquamarin tested the lib and dll we built manually but it seems he cannot get the dll work. Could you please check if the dll itself is working by building and running the [Introduction Examples](https://github.com/kuzudb/kuzudb.github.io/blob/main/docs/getting-started/introduction-examples.md). If the dll is working for you, could you please also update the documentation on how to make the dll work on Windows?

I tested it prior to making this PR, though not with those examples. I'll test it again and update those docs.

mewim:
OK. If renaming the DLL is not easy, let’s just leave it as is.

mewim (approved):

Issue #1670, opened 2023-06-13
Integrate NPY copy into the existing COPY pipeline


ray6080:
We support npy as a file format to be copied into node tables, specifically into fixed size array columns.
Currently, npy copy goes through `NodeCopyExecutor`, which is separate from our query processor. While copying of csv and parquet files goes through the query processor (in the issue title, I call it "COPY pipeline").
For simplicity of code, we should move npy copy to go through the query processor too.
Specifically, we need to remove `NodeCopyExecutor`, and move the logic of copying npy files to `CopyNode` and `ReadFile` operators.

Two main issues need to be addressed for this change:
1. Change `NpyReader` to read a DataChunk at a time, so it can fit into `ReadFile` opeartor. The use of `mmap` can be removed since we read sequentially, which should be fine from the perspective of IO performance.
2. [NPY copy](https://kuzudb.com/docs/data-import/npy-import.html) follows the syntax of `COPY BY COLUMN`, thus one file corresponds to one column, which is quite different from CSV and Parquet files. CSV and Parquet files cover all columns in a single file. So `CopyNode` should be aware of that npy is copying a DataChunk into one column at a time, instead of all columns.

PR #1697, opened 2023-06-20
Read NPY chunk at a time

aziz-mu:
This PR solves (1) from the issue https://github.com/kuzudb/kuzu/issues/1670, although some design changes were made from the issue: 
(i) We are keeping `mmap` for now as it simplifies reading the header
(ii) We are using `arrow::RecordBatch` to match the design of `ReadFile`

### DIFF
--- a/src/storage/copier/npy_reader.cpp
+++ b/src/storage/copier/npy_reader.cpp
@@ -205,5 +205,37 @@ void NpyReader::validate(LogicalType& type_, offset_t numRows, const std::string
             filePath, tableName));
     }
 }
+
+std::shared_ptr<arrow::DataType> NpyReader::getArrowType() const {
+    auto thisType = getType();
+    if (thisType == LogicalTypeID::DOUBLE) {
+        return arrow::float64();
+    } else if (thisType == LogicalTypeID::FLOAT) {
+        return arrow::float32();
+    } else if (thisType == LogicalTypeID::INT64) {
+        return arrow::int64();
+    } else if (thisType == LogicalTypeID::INT32) {
+        return arrow::int32();
+    } else if (thisType == LogicalTypeID::INT16) {
+        return arrow::int16();
+    }
+}
+
+std::shared_ptr<arrow::RecordBatch> NpyReader::readBlock(common::block_idx_t blockIdx) const {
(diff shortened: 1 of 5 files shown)


src/storage/in_mem_storage_structure/in_mem_column_chunk.cpp
```diff
@@ -48,6 +48,10 @@ uint32_t InMemColumnChunk::getDataTypeSizeInColumn(common::LogicalType& dataType
     }
 }
 
+void InMemColumnChunk::copyChunk(std::shared_ptr<arrow::RecordBatch> batch) {
```
ray6080, line 51: Add an assertion here batch only contains one column. Also, you can ignore passing `nullptr` here, otherwise, add a inline comment here `copyArrowArray(*batch->column(0), nullptr /* nodeOffsets */);`.

How about call this function `copyArrowBatch` instead of `copyChunk`? I think the former one is more accurate here.

src/storage/copier/npy_reader.cpp
```diff
@@ -205,5 +205,39 @@ void NpyReader::validate(LogicalType& type_, offset_t numRows, const std::string
...
+    uint64_t rowNumber = CopyConstants::NUM_ROWS_PER_BLOCK_FOR_NPY * blockIdx;
+    auto rowPointer = getPointerToRow(rowNumber);
+    auto buffer =
+        std::make_shared<arrow::Buffer>(rowPointer, CopyConstants::NUM_ROWS_PER_BLOCK_FOR_NPY);
+    auto arrowType = getArrowType();
+    int64_t length = CopyConstants::NUM_ROWS_PER_BLOCK_FOR_NPY < getNumRows() - rowNumber ?
+                         CopyConstants::NUM_ROWS_PER_BLOCK_FOR_NPY :
+                         getNumRows() - rowNumber;
+    auto arr = std::make_shared<arrow::PrimitiveArray>(arrowType, length, buffer);
+    auto field = std::make_shared<arrow::Field>("name", arrowType);
```
ray6080, line 234: Maybe give a more meaningful name than "name"? How about "NPY_FIELD"? And make it a const in NpyReader.

ray6080 (approved):

PR #1703, opened 2023-06-21
Rust API on Windows

benjaminwinger:
I've managed to get rust working on Windows, with the exception that compiling against the debug version doesn't work very well, so I've set it to link against the release version in both debug and release mode on Windows (it only works if you override the `CFLAGS`/`CXXFLAGS` environment variables, but it's not sufficient to do that in `build.rs` since it has to also affect one of the dependencies).

### DIFF
--- a/.github/workflows/ci-workflow.yml
+++ b/.github/workflows/ci-workflow.yml
@@ -122,6 +122,13 @@ jobs:
           call "C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Auxiliary\Build\vcvars64.bat"
           make test NUM_THREADS=18
 
+      - name: Rust test
+        shell: cmd
+        run: |
+          call "C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Auxiliary\Build\vcvars64.bat"
+          set OPENSSL_DIR=C:\Program Files\OpenSSL-Win64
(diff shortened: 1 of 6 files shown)


benjaminwinger:
I ran into the same issue with arrow and openssl that I did with the Linux tests. It seems that the `arrow_bundled_dependencies` library will use openssl if it finds it on your system, regardless of whether or not `ARROW_USE_OPENSSL` is enabled (I'm not sure why; as far as I can tell parquet is compiling the [nossl](https://github.com/apache/arrow/blob/0d1f7234124e67b98b3aa0aa3e521d42c2f3165d/cpp/src/parquet/encryption/encryption_internal_nossl.cc) version of their `encryption_internal` code). 

I've set it up to link against openssl when running the tests on windows, but it does have more difficulty finding it than on Linux. Adding the `openssl-sys` crate as a dependency on Windows helps with finding the library, but you either need vcpkg to detect it (something the openssl-sys error message mentions), or to manually set `OPENSSL_DIR` (which I've included in the github workflow). I might have to look into how using vcpkg works, but if we're relying on `OPENSSL_DIR` it might make more sense to just add `${OPENSSL_DIR}/lib` to the linker search paths ourselves since the openssl rust bindings from `openssl-sys` aren't needed.

andyfengHKU (approved):

PR #1706, opened 2023-06-21
fix WITH NULL segfault

aziz-mu:
fixes the second query in issue #1643 . The first query seems to be an issue with the CASE WHEN function itself, so I'll open a new issue for that one.

### DIFF
--- a/src/include/processor/operator/scan_node_id.h
+++ b/src/include/processor/operator/scan_node_id.h
@@ -44,10 +44,12 @@ public:
     inline NodeTableScanState* getTableState(uint32_t idx) const { return tableStates[idx].get(); }
 
     inline void initialize(transaction::Transaction* transaction) {
-        auto numMask = tableStates[0]->getSemiMask()->getNumMasks();
-        for (auto& tableState : tableStates) {
-            assert(tableState->getSemiMask()->getNumMasks() == numMask);
-            tableState->initializeMaxOffset(transaction);
+        if (!tableStates.empty()) {
+            auto numMask = tableStates[0]->getSemiMask()->getNumMasks();
+            for (auto& tableState : tableStates) {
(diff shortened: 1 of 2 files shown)


src/include/processor/operator/scan_node_id.h
```diff
@@ -44,10 +44,12 @@ class ScanNodeIDSharedState {
     inline NodeTableScanState* getTableState(uint32_t idx) const { return tableStates[idx].get(); }
 
     inline void initialize(transaction::Transaction* transaction) {
-        auto numMask = tableStates[0]->getSemiMask()->getNumMasks();
-        for (auto& tableState : tableStates) {
-            assert(tableState->getSemiMask()->getNumMasks() == numMask);
-            tableState->initializeMaxOffset(transaction);
+        if (!tableStates.empty()) {
```
andyfengHKU, line 47: Can u tell why this need to change? TableStates should never be empty otherwise we shouldn't generate a ScanNodeID operator.

The query below works fine if I revert this change.

andyfengHKU (changes requested):

src/include/processor/operator/scan_node_id.h
aziz-mu: The query gives a segfault if running it on an empty database. I (wrongly) assumed that it didn't work on any db, I can remove the test on demo_db.

It seems then that there is an issue somewhere upstream with generating a ScanNodeID operator when the db is empty, so maybe a different fix is necessary? Let me know if so, then I'll close this PR for now since I am mainly working on something else right now

This is quite a small bug, the reason I was interested in it was because I thought it was causing the LDBC IC1 to fail, but after some more investigating it seems that it's failing due to a separate issue with the `CASE WHEN` binder.

Issue #1692, opened 2023-06-19
Support COPY TO statement


rfdavid:
### Proposal

This feature will enable users to export query results to either CSV or PARQUET files using our current `COPY` command. In the same way, we already perform `COPY FROM`, the idea is to use the same statement using `COPY TO`, similar to [DuckDB Copy](https://duckdb.org/docs/sql/statements/copy.html).

### Examples

Write node table Person to `person.csv` 
`COPY Person TO "person.csv"`

Add HEADER and DELIMITER properties
`COPY Person TO "person.csv" (HEADER=true, DELIM='|');`

Write a specific query result to `result.parquet`
`COPY "MATCH (a:person) RETURN a.name" TO "result.parquet"`

**Todo**

- [ ] Implement `writeToParquet` API
- [ ] Implement `COPY TO` without query and additional properties
- [ ] Support additional properties / queries (tbd)

PR #1716, opened 2023-06-23
Support COPY TO statement

rfdavid:
This PR introduces the command `COPY ([QUERY]) TO [file]`.
The command above will execute the passed query and store the results into the defined file. In this PR, it only supports `.csv` extension, so `.parquet` files will be supported in a separated one.

Example:

```
COPY (MATCH (p:person) RETURN id(p), p.ID, p.fName, p.gender, p.isStudent, p.height) TO "person.csv";

COPY (MATCH (a)<-[:knows|:studyAt]-(b:person) RETURN DISTINCT b.fName) to 'names.csv';
```

Related to #1692


src/include/common/copier_config/copier_config.h
```diff
@@ -31,10 +31,12 @@ struct CSVReaderConfig {
 };
 
 struct CopyDescription {
+    enum class Direction : uint8_t { FROM = 0, TO = 1 };
```
andyfengHKU, line 34: Let's call this CopyDirection to be more clear. Don't forget to change

src/parser/transformer.cpp
```diff
@@ -1050,20 +1050,24 @@ std::string Transformer::transformPrimaryKey(CypherParser::KU_CreateNodeConstrai
...
     auto parsingOptions = ctx.kU_ParsingOptions() ?
                               transformParsingOptions(*ctx.kU_ParsingOptions()) :
                               std::unordered_map<std::string, std::unique_ptr<ParsedExpression>>();
-    return std::make_unique<Copy>(std::move(filePaths), std::move(tableName),
+    return std::make_unique<Copy>(direction, std::move(filePaths), std::move(tableName),
         std::move(parsingOptions), common::CopyDescription::FileType::UNKNOWN);
 }
 
 std::unique_ptr<Statement> Transformer::transformCopyNPY(CypherParser::KU_CopyNPYContext& ctx) {
+    auto direction = ctx.FROM() ? common::CopyDescription::Direction::FROM :
```
andyfengHKU, line 1065: You probably don't need to check direction because the grammar only allows COPY FROM. So give Direction::FROM directly to the constructor

andyfengHKU (approved):

src/binder/bind/bind_copy.cpp
```diff
@@ -37,7 +44,8 @@ std::unique_ptr<BoundStatement> Binder::bindCopyClause(const Statement& statemen
         }
     }
     return std::make_unique<BoundCopy>(
-        CopyDescription(boundFilePaths, csvReaderConfig, actualFileType), tableID, tableName);
+        CopyDescription(direction, boundFilePaths, csvReaderConfig, actualFileType), tableID,
```
andyfengHKU, line 47: I think we can have following design for `BoundCopy`. Let's start with the simplified version of only CopyTo node table

```
struct BoundCopyFromInfo {
     string tableName;
     share_ptr<NodeExpression> node
}

struct BoundCopyToInfo {
     share_ptr<NodeOrRelExpression> node
     expression_vector expressionsToCopy
}

class BoundCopy {

private:
    CopyDescription
    unique_ptr<BoundCopyFromInfo>
    unique_ptr<BoundCopyToInfo>
}
```

So you need to create a node Expression through `binder.createQueryNode()` and loop through all node properties through `node.getPropertyExpressions` to construct a `expressionsToCopy`

src/planner/planner.cpp
```diff
@@ -154,25 +155,47 @@ std::unique_ptr<LogicalPlan> Planner::planRenameProperty(const BoundStatement& s
...
+        boundCopyStatement.getStatementResult()->getSingleExpressionToCollect());
     plan->setLastOperator(std::move(copyCSV));
     return plan;
 }
 
+std::unique_ptr<LogicalPlan> Planner::planCopyTo(
+    const catalog::Catalog& catalog, const BoundStatement& statement) {
+    auto plan = std::make_unique<LogicalPlan>();
+    auto& boundCopyStatement = (BoundCopy&)statement;
+    auto copyTo = make_shared<LogicalCopyTo>(boundCopyStatement.getCopyDescription(),
```
andyfengHKU, line 193: If your `boundCopyStatement` has a `nodeExpression`, you can call `joinOrderEnumerator.appendScanNodeID` as the first operator and `queryPlanner.appendScanNodePropIfNecessary` as the second operator. And finally a `CopyTo` on top of it.

src/include/processor/operator/copy_to/copy_to.h
```diff
@@ -0,0 +1,35 @@
...
+    std::string execute(common::TaskScheduler* taskScheduler, ExecutionContext* executionContext);
+
+//    void initLocalStateInternal(ResultSet* resultSet, ExecutionContext* context) override;
+
+    std::unique_ptr<PhysicalOperator> clone() override {
+        return make_unique<CopyTo>(copyDescription, id, paramsString);
+    }
+
+private:
+    common::CopyDescription copyDescription;
```
andyfengHKU, line 31: Apart from copyDescription, you also need what vectors are you going to copy to. Which means you need 2 things

- vector<DataPos> vectorsToCopyPos
   -  You should get this from the constructor 
- vector<ValueVector> vectorsToCopy
   -  Initialize this through initLocalStateInternal

You can take a look at `payloads` in `ResultCollector` as a reference

src/include/processor/operator/copy_to/copy_to.h
```diff
@@ -0,0 +1,35 @@
...
+namespace processor {
+
+class CopyTo : public PhysicalOperator {
+public:
+    CopyTo(const common::CopyDescription& copyDescription, uint32_t id,
+        const std::string& paramsString)
+        : PhysicalOperator{PhysicalOperatorType::COPY_TO, id, paramsString}, copyDescription{copyDescription} {}
+
+    bool getNextTuplesInternal(ExecutionContext* context) override {
+        throw common::InternalException(
```
andyfengHKU, line 18: Here you want do the following

```
if (!children[0].getNext(context)) {
    return false; // No more tuples to pull from child
}
// Here you need to implement the read from vector and copy to csv logic.
```

src/processor/mapper/map_ddl.cpp
```diff
@@ -53,6 +55,15 @@ std::unique_ptr<PhysicalOperator> PlanMapper::mapLogicalCreateRelTableToPhysical
         &storageManager.getRelsStore().getRelsStatistics());
 }
 
+std::unique_ptr<PhysicalOperator> PlanMapper::mapLogicalCopyToToPhysical(
+    LogicalOperator* logicalOperator) {
+    auto copy = (LogicalCopyTo*)logicalOperator;
+    auto tableName = catalog->getReadOnlyVersion()->getTableName(copy->getTableID());
+    // stuff...
```
andyfengHKU, line 62: The goal of mapper is to construct `vector<DataPos> vectorsToCopyPos`. You can do so by 
```
auto childSchema = logicalOperator.getChild(0).getSchema();
vector<DataPos> vectorsToCopyPos;
for (auto& expression : boundCopy.getCopyToInfo.expressionsToCopy ) {
     vectorsToCopyPos.emplace_back(childSchema.getExpressionPos(expression))
}
```

src/binder/bind/bind_copy.cpp
rfdavid: What do you think about these changes?
1. Move `CopyDirection` to `BoundCopy`
2. Add `CopyDescriptionFrom` and `CopyDescriptionTo` to separate the logic of each one. For example, the current `CopyDescription` contains `CSVReaderConfig`, which is not related to `COPY TO`, and also `filePaths` is a vector, which makes sense for `COPY FROM` but not for `COPY TO`. Inside `CopyDescriptionXXX` (or maybe call `BoundCopyXXXInfo`?) we add the specific properties for each one.

andyfengHKU: This design looks good to me.

src/processor/processor.cpp
```diff
@@ -75,6 +75,7 @@ void QueryProcessor::decomposePlanIntoTasks(
     case PhysicalOperatorType::SET_NODE_PROPERTY:
     case PhysicalOperatorType::SET_REL_PROPERTY:
     case PhysicalOperatorType::CREATE_NODE:
+    case PhysicalOperatorType::COPY_TO:
```
andyfengHKU, line 78: move this line under all `updates` and before `STANDALONE_CALL`. So that we have operators of the same type written together

src/antlr4/Cypher.g4
```diff
@@ -17,7 +17,7 @@ oC_Cypher
     : SP ? oC_AnyCypherOption? SP? ( oC_Statement ) ( SP? ';' )? SP? EOF ;
 
 kU_CopyCSV
-    : COPY SP oC_SchemaName SP FROM SP kU_FilePaths ( SP? '(' SP? kU_ParsingOptions SP? ')' )? ;
+    : COPY SP oC_SchemaName SP ( FROM | TO ) SP kU_FilePaths ( SP? '(' SP? kU_ParsingOptions SP? ')' )? ;
```
andyfengHKU, line 20: We probably need to separate `COPY FROM\TO` grammar here. Something like

```
kU_CopyFROM : COPY SP oC_SchemaName SP FROM SP kU_FilePaths ( SP? '(' SP? kU_ParsingOptions SP? ')' )? ;

kU_CopyTO: Copy SP oC_Query SP TO kU_FilePaths (eventually there should also be an option here, though might be different from parsing option)
```

src/include/binder/copy/bound_copy.h
```diff
@@ -11,25 +11,55 @@
...
+};
+
 class BoundCopy : public BoundStatement {
 public:
-    BoundCopy(
-        common::CopyDescription copyDescription, common::table_id_t tableID, std::string tableName)
-        : BoundStatement{common::StatementType::COPY,
-              BoundStatementResult::createSingleStringColumnResult()},
+    // COPY FROM
+    BoundCopy(common::CopyDescription copyDescription, common::table_id_t tableID,
```
andyfengHKU, line 22: Split BoundCopy to BoundCopyFrom and BoundCopyTo. BoundCopyFrom should stay the same as it used to be.

BoundCopyTo should have
```
common::CopyDescription copyDescription;
std::unique_ptr<BoundStatement> regularQuery; // or just call it statement
```

src/common/copier_config/copier_config.cpp
```diff
@@ -7,16 +7,17 @@ using namespace kuzu::utf8proc;
 
 namespace kuzu {
 namespace common {
-CopyDescription::CopyDescription(
-    const std::vector<std::string>& filePaths, CSVReaderConfig csvReaderConfig, FileType fileType)
-    : filePaths{filePaths}, csvReaderConfig{nullptr}, fileType{fileType} {
+CopyDescription::CopyDescription(CopyDirection copyDirection,
+    const std::vector<std::string>& filePaths, FileType fileType, CSVReaderConfig csvReaderConfig)
+    : copyDirection{copyDirection}, filePaths{filePaths}, csvReaderConfig{nullptr}, fileType{
```
andyfengHKU, line 12: Once you split BoundCopyTo&From you probably don't need direction any more.

src/binder/bind/bind_copy.cpp
```diff
@@ -10,16 +10,43 @@ using namespace kuzu::parser;
 namespace kuzu {
 namespace binder {
 
-std::unique_ptr<BoundStatement> Binder::bindCopyClause(const Statement& statement) {
-    auto& copyCSV = (Copy&)statement;
+std::unique_ptr<BoundStatement> Binder::bindCopyToClause(const Copy& copyStatement) {
+    auto boundFilePaths = copyStatement.getFilePaths();
+    auto actualFileType = bindFileType(boundFilePaths);
+    auto tableID = catalog.getReadOnlyVersion()->getTableID(copyStatement.getTableName());
```
andyfengHKU, line 16: once you changed the grammar simply call
```
auto boundFilePaths = copyStatement.getFilePaths();
auto actualFileType = bindFileType(boundFilePaths);
auto boundQuery = bindQuery(copyStatement.getRegularQuery());
return std::make_unique<BoundCopyTo>(std::move(boundQuery),
         CopyDescription(boundFilePaths, actualFileType, CSVReaderConfig{}));
```

test/test_files/exceptions/binder/binder_error.test
```diff
@@ -448,3 +448,8 @@ Binder exception: Invalid option name: thread.
 -STATEMENT CALL threads='abc'
 ---- error
 Binder exception: abc has data type STRING. (INT64) was expected.
+
```
andyfengHKU, line 451: Don't forget to add a test to check copy to result eventually.

src/include/parser/copy.h
```diff
@@ -9,13 +9,15 @@
 namespace kuzu {
 namespace parser {
 
+// This class contains all the information parsed from COPY command.
```
andyfengHKU, line 12: I'm not sure if we should split copy to and from here. This is your call.

src/include/planner/logical_plan/logical_operator/logical_copy_to.h
```diff
@@ -0,0 +1,38 @@
...
+
+    inline void computeFlatSchema() override { copyChildSchema(0); }
+
+    inline std::unique_ptr<LogicalOperator> copy() override {
+        return make_unique<LogicalCopyTo>(children[0]->copy(), expressionsToCopy, copyDescription);
+    }
+
+private:
+    common::CopyDescription copyDescription;
+    binder::expression_vector expressionsToCopy;
```
andyfengHKU, line 34: remove. You should get this from child schema

src/include/planner/logical_plan/logical_operator/logical_copy_to.h
```diff
@@ -0,0 +1,38 @@
...
+
+#include "base_logical_operator.h"
+#include "common/copier_config/copier_config.h"
+#include "planner/logical_plan/logical_plan.h"
+
+namespace kuzu {
+namespace planner {
+
+class LogicalCopyTo : public LogicalOperator {
+
```
andyfengHKU, line 11: remove empty line

src/planner/planner.cpp
```diff
@@ -163,28 +164,56 @@ std::unique_ptr<LogicalPlan> Planner::planRenameProperty(const BoundStatement& s
...
     plan->setLastOperator(std::move(copyCSV));
     return plan;
 }
 
+std::unique_ptr<LogicalPlan> Planner::planCopyTo(const Catalog& catalog,
+    const NodesStatisticsAndDeletedIDs& nodesStatistics, const RelsStatistics& relsStatistics,
+    const BoundStatement& statement) {
+    auto& boundCopyStatement = (BoundCopy&)statement;
+    auto queryGraphCollection = boundCopyStatement.getQueryGraphCollection();
+    auto planner = QueryPlanner(catalog, nodesStatistics, relsStatistics);
```
andyfengHKU, line 207: once you have boundStatement in BoundCopyTo. Simply call
```
auto plan = getBestPlan(..., statement)
// append logical copyTo on top
```

src/processor/mapper/plan_mapper.cpp
```diff
@@ -201,9 +204,18 @@ std::unique_ptr<PhysicalOperator> PlanMapper::appendResultCollectorIfNotCopy(
     Schema* schema) {
     // We have a special code path for executing copy rel and copy npy, so we don't need to append
     // the resultCollector.
-    if (lastOperator->getOperatorType() != PhysicalOperatorType::COPY_REL &&
-        lastOperator->getOperatorType() != PhysicalOperatorType::COPY_NPY) {
+    switch (lastOperator->getOperatorType()) {
+    case PhysicalOperatorType::COPY_REL:
+    case PhysicalOperatorType::COPY_NPY:
+        return lastOperator;
+    case PhysicalOperatorType::COPY_TO:
```
andyfengHKU, line 211: You can leave this for now. Let's revisit this part before merging.

src/include/parser/copy.h
```diff
@@ -33,5 +34,21 @@ class Copy : public Statement {
...
+    explicit CopyTo(std::string filePath, std::unique_ptr<RegularQuery> regularQuery)
+        : Statement{common::StatementType::COPY_TO},
+          regularQuery{std::move(regularQuery)}, filePath{std::move(filePath)} {}
+
+    inline std::string getFilePath() const { return filePath; }
+    inline std::unique_ptr<RegularQuery> getRegularQuery() { return std::move(regularQuery); }
+
+private:
+    std::string filePath;
+    common::CopyDescription::FileType fileType;
```
andyfengHKU, line 48: I guess you only know file type after binding. And it seems this field is not being used. So remove it.

src/include/parser/copy.h
```diff
@@ -33,5 +34,21 @@ class Copy : public Statement {
...
+    std::string tableName;
```
andyfengHKU, line 49: Ditto. Seems that tableName is no longer needed.

src/include/binder/copy/bound_copy.h
```diff
@@ -13,15 +14,24 @@ namespace binder {
 
 class BoundCopy : public BoundStatement {
 public:
+    // COPY FROM
```
andyfengHKU, line 17: Since you have CopyTo and CopyFrom in the parser part. Let's use BoundCopyFrom and BoundCopyTo for consistency.

src/include/common/copier_config/copier_config.h
```diff
@@ -31,22 +31,34 @@ struct CSVReaderConfig {
...
 
     static std::string getFileTypeName(FileType fileType);
 
-    const std::vector<std::string> filePaths;
+    const std::vector<std::string> copyFromFilePaths;
+    const std::string copyToFilePath;
+    const std::vector<std::string> columnNames;
     std::unique_ptr<CSVReaderConfig> csvReaderConfig;
     FileType fileType;
+    CopyDirection copyDirection;
```
andyfengHKU, line 61: I might be wrong, but seems you don't need to store copyDirection.

Also I would prefer to merge `copyFromFilePaths` and `copyToFilePath`. In the copyTo case, the vector should be of size 1.

src/include/processor/mapper/plan_mapper.h
```diff
@@ -2,7 +2,7 @@
 
 #include "binder/expression/node_expression.h"
 #include "common/statement_type.h"
-#include "planner/logical_plan/logical_operator/logical_copy.h"
+#include "planner/logical_plan/logical_operator/logical_copy_from.h"
```
andyfengHKU, line 5: use planner::LogicalOperator rather than planner::LogicalCopy and move this include to cpp file.

src/planner/planner.cpp
```diff
@@ -187,6 +191,21 @@ std::unique_ptr<LogicalPlan> Planner::planCopy(
     return plan;
 }
 
+std::unique_ptr<LogicalPlan> Planner::planCopyTo(const Catalog& catalog,
+    const NodesStatisticsAndDeletedIDs& nodesStatistics, const RelsStatistics& relsStatistics,
+    const BoundStatement& statement) {
+    auto& copyClause = reinterpret_cast<const BoundCopy&>(statement);
+    auto regularQuery = copyClause.getRegularQuery();
+    if (regularQuery->getStatementType() != StatementType::QUERY) {
```
andyfengHKU, line 199: I think the parser has the guarantee for this already. So you can assert here rather than check and throw

src/processor/mapper/plan_mapper.cpp
```diff
@@ -208,6 +211,29 @@ std::unique_ptr<PhysicalOperator> PlanMapper::appendResultCollectorIfNotCopy(
     return lastOperator;
 }
 
+// TODO: re-visit before merge
+// std::unique_ptr<PhysicalOperator> PlanMapper::appendResultCollectorIfNotCopy(
```
andyfengHKU, line 215: Seems you have figured out the way to return a column for CopyTo. So you can just remove this commented code.

andyfengHKU (commented):
Up to planning looks good to me. Good job.

src/processor/operator/copy_to/csv_file_writer.cpp
```diff
@@ -0,0 +1,127 @@
...
+void CSVFileWriter::run(std::vector<common::ValueVector*>& outputVectors) {
+    bool hasData = true;
+    int64_t pos = 0;
+    auto outputVectorsSize = outputVectors.size();
+    if (outputVectorsSize == 0) {
+        return;
+    }
+    // 'pos' represents the position of the row
+    // 'vectorPos' represents the position of the column
+    while (hasData) {
```
ray6080, line 103: As discussed, we should assume all flat input for now. So I don't the logic should be complicated as the one you have.
Generally, it should follow something like this:
```
func writeTuple() {
    for (auto vector : vectors) {
        writeValue(vector)
    }
}

func writeValue(vector) {
    assert(vector is flat)
    write value at flatten pos. selectedPos[0]
}
```

ray6080 (changes requested):

src/include/binder/copy/bound_copy_to.h
```diff
@@ -0,0 +1,38 @@
...
+
+    inline BoundRegularQuery* getRegularQuery() const { return regularQuery.get(); }
+
+    inline common::table_id_t getTableID() const { return tableID; }
+
+    inline std::string getTableName() const { return tableName; }
+
+private:
+    common::CopyDescription copyDescription;
+    common::table_id_t tableID;
```
andyfengHKU, line 32: tableID and tableName no longer seem to be needed

src/include/binder/copy/bound_copy_to.h
```diff
@@ -0,0 +1,38 @@
+#pragma once
+
+#include <string>
+#include <unordered_map>
+#include <vector>
+
+#include "binder/bound_statement.h"
+#include "binder/query/bound_regular_query.h"
+#include "catalog/table_schema.h"
```
andyfengHKU, line 9: doesn't seem to be needed

rfdavid: That's right, only `bound_regular_query.h` is needed. I'm removing the others

src/common/type_utils.cpp
```diff
@@ -75,9 +75,10 @@ std::string TypeUtils::toString(const list_entry_t& val, void* valueVector) {
     auto dataVector = ListVector::getDataVector(listVector);
     for (auto i = 0u; i < val.size; ++i) {
```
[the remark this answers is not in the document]
rfdavid: I think we can also speed up with `ostringstream`

ray6080 (approved):
LGTM!

PR #1719, opened 2023-06-25
Testing framework adjustments

rfdavid:
This PR addresses the feedback from the testing framework presentation.

- Replace `.csv` extension with `.txt` inside `/answers` directory
- Change `-NAME` to `-DEBUG_LOG`. Spaces are allowed in front of the tag. For example, instead of writing `-DEBUG_LOG SingletonNodeBackwardCheck`, it's possible to write `-DEBUG_LOG Check singleton node backwards`
- Remove `-QUERY` statement. Only `-STATEMENT` will be allowed
- Change `-STATEMENT_BLOCK` to `-INSERT_STATEMENT_BLOCK` to make it more explicit
- Replace a long text with `ARANGE` on Unwind test


ray6080 (approved):

ray6080 (commented):
I would consider replace `-DEBUG_LOG` with `-LOG`, which is shorter but convey the same meaning, but up to you.

PR #1725, opened 2023-06-26
Fix use of PreparedStatement in releases

benjaminwinger:
~~Instead of using forward declarations.
The forward declarations prevent using `unique_ptr<PreparedStatement>` without extra includes since the type is incomplete, and the necessary header files are unavailable in the single-file header included in the releases.~~

This is necessary both to compile the rust API against the releases, as well as to use the `Connection::prepare` function in the C++ API from the releases.

~~Were forward declarations being used to try and keep BoundStatementResult and LogicalPlan out of the public API, or was it just to try and reduce the number of includes/deal with cyclic includes?~~

Edit: Explicitly declaring the default destructor fixes the issue without changing the includes.

### DIFF
--- a/src/include/main/prepared_statement.h
+++ b/src/include/main/prepared_statement.h
@@ -43,6 +43,7 @@ public:
     inline std::unordered_map<std::string, std::shared_ptr<common::Value>> getParameterMap() {
         return parameterMap;
     }
+    ~PreparedStatement();
 
 private:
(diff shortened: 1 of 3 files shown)


benjaminwinger:
I think this could also be fixed by explicitly using the default destructor for `PreparedStatement`, which might be the better way.

andyfengHKU (approved):

PR #1726, opened 2023-06-27
Recursive join rel filter

andyfengHKU:


ray6080 (approved):
Just one general comment, I think it's better to closely follow existing MemGraph syntax, though we didn;t support filter predicates on node variables now, still the syntax should be `[e:knows*1..2 (r,n | WHERE list_contains(r.comments, 'rnme'))]` or `[e:knows*1..2 (r,_ | WHERE list_contains(r.comments, 'rnme'))]`

PR #1733, opened 2023-06-28
Testing framework: add ${count} to REPEAT

### LINKED ISSUE #1729: Testing framework: support ${count} on REPEAT
Testing framework `REPEAT` only repeats fixed strings. In some situations, it's useful to generate different names, for example, creating 2000 different column names. A solution is to support representing `${count}` as a count number:

```
-DEFINE COLS REPEAT 2500 "col${count}, "
COLS = "col0, col1, ... col2500, "
```

rfdavid:
This PR adds `${count}` to support using the count number to distinguish names. This might be useful to some use cases, for example, creating a node table with 2000 different columns:

```
-GROUP Test
-DATASET CSV empty

--

-CASE Dummy

-DEFINE COLS REPEAT 2000 "Col${count} STRING, "
-STATEMENT CREATE NODE TABLE Person(${COLS} PRIMARY KEY(Col0))
---- ok
```

closes #1729


acquamarin (approved):
Can you rebase the master and refactor the `NodeTableWith300ColumnsInfo` test using the new testing statement?

PR #1741, opened 2023-06-29
Rust packaging

benjaminwinger:
I fixed a few things to get rust crates working properly:

- The kuzu-src symlink seems to mess with how `cargo package` ignores untracked files in git repositories, which would pull in things like build files, so I added a file `include` list to make sure only the necessary source files are included in the crate.
- Technically rust crates should only modify files within the directory indicated by the `OUT_DIR` environment variable, so it should really be an out of source build. As such, I set it up to run the arrow and kuzu cmake builds manually within build.rs instead of running them through the Makefile.
- I added a way to link against the release binaries using environment variables to point to their path (documented in the root of the rustdocs).

Additionally:
- I noticed that I used the tempdir crate (abandoned and merged into tempfile) instead of the tempfile crate in the tests, so I replaced that.
- I made building the shell and python API optional (enabled by default to match the current behaviour) so that the rust crate doesn't need to build them when building from source (which also removes the python dependency when using the rust API).


benjaminwinger:
I don't think enabling the long paths worked. Apparently the software may need to be "long path aware", which might be the issue.
I've tried reducing the length of the build path in CI instead.

andyfengHKU (approved):

Issue #1736, opened 2023-06-28
You need to improve your google-rating


LowLevelMahn:
you're feature-set seems to fullfill all my dreams - but google buries you deep behind all these Neo4J,Memgraph,etc. links :(

andyfengHKU:
Hi, thanks for your interest. We try to increase our popularities mostly by posting on social medias, e.g. Twitter, hacker news, etc. But looks like attractions come at a relatively low speed. Would be happy to know if you have any suggestions. Also feel free to tell us if you have any feature requests.

LowLevelMahn:
> We try to increase our popularities mostly by posting on social medias, e.g. Twitter, hacker news, etc.

good strategy - maybe it takes some time

> Would be happy to know if you have any suggestions.

builds under Linux out of the box with your build instructions

build-instructions for CMake/VStudio or MSYS2/mingw64 would be nice

but using MSYS2/mingw64 tells me after invoking cmake and ninja

`ninja: error: 'PARQUET_PATH-NOTFOUND', needed by 'src/libkuzu_shared.dll', missing and no known rule to make it`

using cmake+VS2022 generator gives me many include error for apache arrow while building

mewim:
Hi @LowLevelMahn. The error seems to come from Node.js API. For building on Fedora 38, if you would like to build Node.is API, you need to run `npm install --include dev` under `tools/nodejs_api` to install the dependencies for Node.js first. You can also skip building Node.js by using `make release` instead of `make all`.

benjaminwinger:
> for Windows-builds - its a little bit unclear how to build for windows, what compiler is supported:

Build instructions specific to windows do seem to be missing.

So far we've only managed to get it building with MSVC:
- Run inside of a Visual Studio Developer shell: i.e. either cmd after executing `vcvars64.bat` (e.g. see [CI](https://github.com/kuzudb/kuzu/blob/184d8dfaaef25f5063428645681ea84962b34264/.github/workflows/ci-workflow.yml#L123)) or Powershell with Launch-VsDevShell.ps1 (e.g. `'C:\Program Files\Microsoft Visual Studio\2022\Community\Common7\Tools\Launch-VsDevShell.ps1' -Arch amd64 -HostArch amd64`) (with visual studio installed there should be a "developer" terminal that does it automatically).
[...]

PR #1742, opened 2023-06-29
Add windows-specific build instructions to readme

benjaminwinger:
Added a short description of additional requirements for building on windows to the README (see https://github.com/kuzudb/kuzu/issues/1736#issuecomment-1612130209).

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,6 +138,14 @@ To build in parallel, pass `NUM_THREADS` as parameter, e.g., `make NUM_THREADS=8
 
 After build, our CLI binary `kuzu_shell` is available under the directory `build/release/tools/shell/`.
 
+### Building on Windows
+Currently MSVC is the only supported compiler:
+
+- In addition to the dependencies listed above, you will also need GNU Make and Ninja (E.g. with [Chocolatey](https://community.chocolatey.org/): `choco install make ninja`).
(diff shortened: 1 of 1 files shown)


mewim (approved):

LowLevelMahn:
what "Windows" Make should be used for Windows - the 2006 GnuWin32 Version?: https://gnuwin32.sourceforge.net/packages/make.htm

benjaminwinger:
That version might work, depending on the differences between the versions. I don't think it's been tested with that old of a version of make.

What we've used in CI is the version packaged by chocolatey: https://community.chocolatey.org/packages/make.  It looks like the person who packaged that is distributing the binaries themselves along with the package (the exe can be found [here](https://bitbucket.org/xoviat/chocolatey-packages/src/master/make/4.4.1/tools/install/bin/)).

LowLevelMahn:
> What we've used in CI is the version packaged by chocolatey

could you add that info+url to the short build description?

benjaminwinger:
> could you add that info+url to the short build description?

Sure. Or rather, I updated it to link chocolatey and listed the package names in an example install command, since that also includes ninja and might be a little more straightforward than linking the chocolatey packages individually. Is that sufficient?

LowLevelMahn:
> > could you add that info+url to the short build description?
> 
> Sure. Or rather, I updated it to link chocolatey and listed the package names in an example install command, since that also includes ninja and might be a little more straightforward than linking the chocolatey packages individually. Is that sufficient?

yes - seems to be more "complete" then

PR #1750, opened 2023-07-03
Adjust test framework to handle single line result

rfdavid:
Adjust the test framework to deal with single-line results that include new lines. This is necessary to test the results from `EXPLAIN` statements.


rfdavid:
Discussion is still open here: https://github.com/kuzudb/kuzu/issues/1755

Issue #1745, opened 2023-06-30
Suggestions for easier deployment and quality


LowLevelMahn:
only ideas i found in other projects i use and like - i think some of these points are already in work

-CMake only builds (try to get rid of Make in the start) - CMake build should be equally easy
-github auto build for Windows/Linux/Macos, clang, gcc, MSYS2/gcc/clang (see dosbox staging or apache arrow) 
-clear build instruction for the major operating systems/distributions (package builds of debian,ubuntu love prepared stuff)
-vcpkg/conan package - to ease the deployment even further
-coverity/PVS auto checks
-auto runs with ASAN/TSAN
-some project even test the build and stuff for the 3-4 major platforms (to be safe that everything is working on debia, ubuntu, macos etc.)

https://github.com/dosbox-staging/dosbox-staging is doing parts of that very good

mewim:
Hi @LowLevelMahn,

Thanks for the suggestions.

> CMake only builds (try to get rid of Make in the start) - CMake build should be equally easy

When we added Apache Arrow as a dependency, we actually tried to add it as a CMake external project, but unfortuantely we were not able to get it work, so we switched to use Makefile to build Arrow first and then our project. I agree that this approach is not as elegant as only using CMake. Currently, we are only using Arrow to ingest data and in longer term, we plan to replace Arrow with our own CSV and Parquet implementation, which would hopefully get rid of Makefile.

>  github auto build for Windows/Linux/Macos, clang, gcc, MSYS2/gcc/clang (see dosbox staging or apache arrow)

We do have automatic build pipelines for [Windows](https://github.com/kuzudb/kuzu/blob/master/.github/workflows/windows-precompiled-bin-workflow.yml), [Linux](https://github.com/kuzudb/kuzu/blob/master/.github/workflows/linux-precompiled-bin-workflow.yml) and [macOS](https://github.com/kuzudb/kuzu/blob/master/.github/workflows/mac-precompiled-bin-workflow.yml). However, currently we only trigger them manually for releases. We have a plan to also publish nightly builds in our longer term roadmap once the codebase becomes more stable.

> clear build instruction for the major operating systems/distributions 

Good suggestion. We will add more detailed documentations for dependency installation and build steps.

> prepared binary builds for all major platforms

We provide pre-compiled binaries for each of our release for all major platforms, e.g. https://github.com/kuzudb/kuzu/releases/tag/v0.0.4

> vcpkg/conan package - to ease the deployment even further

[...]

PR #1752, opened 2023-07-03
Unified CMake Build

benjaminwinger:
Addresses an issue mentioned in #1736 and also #1745 by running the apache arrow cmake build during the configure step of the kuzu cmake build.

This might be a little unexpected since you don't usually expect a large and slow compilation step when running the configure step, but `USE_SYSTEM_ARROW` and `ARROW_INSTALL` can still be used to avoid building arrow by pointing it at a pre-built version.

I simplified the makefile accordingly, but didn't entirely remove it. There are some situations where it is still helpful to provide default settings, particularly in the case of Windows MSVC builds where (last I checked) the default Visual Studio generator produces build errors. The makefile targets are also being used for CI, though they could be inlined into the CI files.

Note that while this removes the distinction between `make clean` and `make clean-all`, you can still clean without removing arrow by running `make clean` inside the build directory, i.e. on the cmake generated makefile, when using the makefile generator (not quite the same as removing all the files, but it's close enough for most purposes I think).

I also took the liberty of setting `GIT_SHALLOW` on the arrow ExternalProject (cuts the download size from >100MB to ~20MB), and since it requires a named tag, also updated that to 12.0.1, which seems to work fine.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -158,7 +158,21 @@ if (${USE_SYSTEM_ARROW})
 else()
     if (NOT DEFINED ARROW_INSTALL)
         message(STATUS "Configuring arrow for bundled install")
-        set(ARROW_INSTALL ${CMAKE_CURRENT_SOURCE_DIR}/external/build/arrow/install)
+        set(ARROW_INSTALL ${CMAKE_CURRENT_BINARY_DIR}/arrow-build/arrow/install)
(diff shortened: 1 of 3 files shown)


ray6080 (approved):
LGTM

LowLevelMahn:
> There are some situations where it is still helpful to provide default settings

isn't that also possible with CMake only - defaults but overwriteable?

benjaminwinger:
> isn't that also possible with CMake only - defaults but overwriteable?

Not for the generator as far as I can tell, but presumably everything else should be.

PR #1753, opened 2023-07-03
Example Fix and CI

benjaminwinger:
I'd noticed that the C++ example doesn't compile any more (`json_fwd.hpp` has been added into one of the includes, presumably since that example was written). I've fixed it, and added CI jobs for the examples, since while they aren't essential parts of the software, they are less helpful if they get out of date and fail to build when someone tries them out.

### DIFF
--- a/.github/workflows/ci-workflow.yml
+++ b/.github/workflows/ci-workflow.yml
@@ -13,6 +13,9 @@ jobs:
     name: gcc build & test
     needs: [clang-formatting-check]
     runs-on: kuzu-self-hosted-testing
+    env:
+      # Share build cache when building rust API and the example project
+      CARGO_TARGET_DIR: ${{ github.workspace }}/target
     steps:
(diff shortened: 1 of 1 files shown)


examples/cpp/CMakeLists.txt
```diff
@@ -4,6 +4,7 @@ project(example-cpp)
 set(CMAKE_CXX_STANDARD 20)
 
 include_directories(../../src/include)
+include_directories(../../third_party/nlohmann_json)
```
andyfengHKU, line 7: Can you double check if we need this third party any more? I think we removed this dependency since #1751

andyfengHKU (approved):

examples/cpp/CMakeLists.txt
benjaminwinger: No, it's no longer needed. The example now works without any modifications.

PR #1800, opened 2023-07-10
Added deployment workflow for rust API

benjaminwinger:
Note that this will publish the version as recorded in `tools/rust_api/Cargo.toml`, so that will need to be updated before the deployment pipeline is triggered.

Alternatively we could have a script which detects the version and rewrites the file to use the most up to date version, but it would be helpful to keep the version up to date so that anyone adding kuzu as a git dependency sees a meaningful version.

### DIFF
--- a/.github/workflows/build-and-deploy.yml
+++ b/.github/workflows/build-and-deploy.yml
@@ -205,6 +205,33 @@ jobs:
         with:
           password: ${{ secrets.PYPI_TOKEN }}
 
+  deploy-rust:
+    runs-on: kuzu-self-hosted-testing
+    env:
+      CARGO_REGISTRY_TOKEN: ${{ secrets.CARGO_REGISTRY_TOKEN }}
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Update Cargo.toml version
+        run: python3 update_version.py
+        working-directory: tools/rust_api
+
+      - name: Deploy crate to Crates.io
+        run: cargo publish --allow-dirty
+        if: ${{ github.event.inputs.isDeploy == 'true' }}
+        working-directory: tools/rust_api
+
+      - name: Test publishing crate
+        run: cargo publish --dry-run --allow-dirty
+        if: ${{ github.event.inputs.isDeploy != 'true' }}
+        working-directory: tools/rust_api
+
+      - name: Upload crate
+        uses: actions/upload-artifact@v3
+        with:
+          name: kuzu-deploy-crate
+          path: tools/rust_api/target/package/*.crate
+
   build-precompiled-bin-mac:
     uses: ./.github/workflows/mac-precompiled-bin-workflow.yml
     secrets: inherit
--- a/tools/rust_api/Cargo.toml
+++ b/tools/rust_api/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "kuzu"
-version = "0.0.4"
+version = "0.0.5"
 description = "An in-process property graph database management system built for query speed and scalability"
 # Note: 1.63 required for building tests
 rust-version = "1.51"
--- a/tools/rust_api/update_version.py
+++ b/tools/rust_api/update_version.py
@@ -1,0 +1,42 @@
+"""Updates the version in Cargo.toml to match the version in the main CMakeLists.txt"""
+
+import os
+import re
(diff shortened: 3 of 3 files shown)


.github/workflows/build-and-deploy.yml
```diff
@@ -205,6 +205,16 @@ jobs:
         with:
           password: ${{ secrets.PYPI_TOKEN }}
 
+  deploy-rust:
+    runs-on: kuzu-self-hosted-linux-building
```
mewim, line 209: Since this is a very simple job that does not require compilation, consider move it to a GitHub-hosted runner instead of using our own infrastructure.

.github/workflows/build-and-deploy.yml
```diff
@@ -205,6 +205,16 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@v3
+
+      - run: cargo publish
```
mewim, line 213: For cargo publish, is there a way to dry-run or deploy to a test environment similar to npm or pypi? If so, we should do dry run unless `isDeploy` is set to true.

mewim:
> Alternatively we could have a script which detects the version and rewrites the file to use the most up to date version.
Let's add that to the CI, so that the deployment can still be done correctly even if we forget to update the version for rust. You can get the version number from `CMakeLists.txt` similar to Node.js and Python APIs.

.github/workflows/build-and-deploy.yml
benjaminwinger: The choice was deliberate. It's running a full build of the crate before uploading it, which is particularly important since the [rust API only includes certain files when packaging](https://github.com/kuzudb/kuzu/blob/master/tools/rust_api/Cargo.toml#L13-L24) both because the `tools/rust_api/kuzu-src` symlink breaks the way it usually omits untracked git files, and because there's a maximum upload size and including stuff like the dataset directory would make the result too large. There's no verification that it can be built with the included files in CI, so having this last-minute check is important (and while it's . Having the check in CI instead would be slow, but would probably be a lot less so with sccache set up correctly.

.github/workflows/build-and-deploy.yml
benjaminwinger: Yes, there's a `--dry-run` option which skips uploading. I don't think there's a test registry like test.pypi.org, but we could just upload the packaged crate file to github.

mewim: OK. Please add the dry-run based on the flag. We should always upload the packaged files to GitHub as well for debugging purpose in case something goes wrong and the deployment fails.

.github/workflows/build-and-deploy.yml
benjaminwinger: Oh, that I missed. I confused it with `linux-self-hosted-testing`. Since the results of the build aren't what's being published, it probably should just use the testing environment shouldn't it? I don't know what exactly the difference between the two environments are but I'm assuming it's something to do with the libraries it links against.

mewim: Yes the building environment uses a distro with old `glibc` so the prebuilt binaries can have better compatibility. You can use the testing environment since the binaries are not going to be published.

benjaminwinger: I have updated it to use `kuzu-self-hosted-testing`

.github/workflows/build-and-deploy.yml
```diff
@@ -205,6 +205,33 @@ jobs:
...
+      - name: Test publishing crate
+        run: cargo publish --dry-run --allow-dirty
+        if: ${{ github.event.inputs.isDeploy != 'true' }}
+        working-directory: tools/rust_api
+
+      - name: Upload crate
+        uses: actions/upload-artifact@v3
+        with:
+          name: kuzu-deploy-crate
+          path: target/package/*.crate
```
mewim, line 233: It seems that the path here is not correct. I tested run this job once and it’s not able to find any artifact. https://github.com/kuzudb/kuzu/actions/runs/5522371969/jobs/10071856122

benjaminwinger: Whoops. I wrote it relative to `tools/rust_api`, but that's not the working directory for that step.

mewim (approved):

PR #1803, opened 2023-07-11
Table star

andyfengHKU:
Support `... RETURN a.*` as returning all properties of node/rel/struct


src/parser/transformer.cpp
```diff
@@ -746,12 +749,11 @@ std::unique_ptr<ParsedExpression> Transformer::transformPropertyOrLabelsExpressi
     auto raw = ctx.oC_Atom()->getText();
     if (!ctx.oC_PropertyLookup().empty()) {
         auto lookUpCtx = ctx.oC_PropertyLookup(0);
-        auto result = std::make_unique<ParsedPropertyExpression>(
-            transformPropertyLookup(*lookUpCtx), std::move(atom), raw + lookUpCtx->getText());
+        auto child = std::move(atom);
```
andyfengHKU, line 752: inline this

ray6080 (approved):

PR #1819, opened 2023-07-14
Remove openssl linking for tests

benjaminwinger:
No longer appears to be necessary. It will have to be partially added back in by #1582 since it needs to link against the crypto library, which is why I left the `find_openssl_windows` function.


benjaminwinger:
Might still be necessary on windows

acquamarin (approved):

PR #1711, opened 2023-06-22
nTkS Recursive Join scheduler changes

anuchak:
This PR integrates changes for the nTkS scheduler for shortest path queries. 
Currently, only Single Label, No path tracking, Shortest Path recursive queries are supported with this policy. 
The other types of recursive queries are still using the 1T1S policy for work sharing between threads.

PR #1821, opened 2023-07-16
RecursiveJoin changes for nTkS scheduler

anuchak:
Changes specific to the nTkS scheduler used in Recursive Join (shortest path query).
Changes made as per review comments @andyfengHKU from previous PR #1711 -

1) updated name of single source computation struct to `SSSPSharedState` 

2) removed `threadCheckSSSPState` extra variable, now depending on boolean TRUE / FALSE we check combination of global + local state

3) `threadIdx` variable removed, no specific index variable being used anymore

4) check for `xxx == nullptr` changed to specific function in `BaseBFSMorsel` -> `hasSSSPSharedState`


src/include/processor/mapper/plan_mapper.h
```diff
@@ -151,6 +151,7 @@ class PlanMapper {
     storage::MemoryManager* memoryManager;
     ExpressionMapper expressionMapper;
     catalog::Catalog* catalog;
+    uint64_t numThreadsForExecution;
```
anuchak, line 154: Added this variable here because the `MorselDispatcher` needs it to initialize the active Shared State vector at compile time (fixed size)

The numThreads is not stored explicitly anywhere in the operator, since the vector needs to be defined at compile time it needs to be passed from the `PlanMapper`

andyfengHKU: This should be removed though. MorselDispatcher is a global state and it should get numThreads through `initGloablState(context)`. The numThreads is stored in context

benchmark/queries/ldbc-sf100/shortest_path_ldbc100/q37.benchmark
```diff
@@ -0,0 +1,5 @@
+-NAME q37
```
andyfengHKU, line 1: Open a saperate PR to merge these benchmarks into master

src/include/processor/operator/recursive_extend/recursive_join.h
```diff
@@ -122,9 +135,15 @@ class RecursiveJoin : public PhysicalOperator {
...
-    std::unique_ptr<BaseBFSState> bfsState;
     std::unique_ptr<FrontiersScanner> frontiersScanner;
     std::unique_ptr<TargetDstNodes> targetDstNodes;
+
+    /// NEW MEMBERS BEING ADDED
+    std::vector<DataPos> vectorsToScanPos;
+    std::vector<common::ValueVector*> vectorsToScan;
+    std::vector<ft_col_idx_t> colIndicesToScan;
+    std::unique_ptr<BaseBFSMorsel> bfsMorsel;
+    std::shared_ptr<MorselDispatcher> morselDispatcher;
```
andyfengHKU, line 146: move dispatcher to shared state

src/processor/operator/recursive_extend/bfs_scheduler.cpp
```diff
@@ -0,0 +1,192 @@
...
+        bfsMorsel->resetState();
+        auto nodeID = srcNodeIDVector->getValue<common::nodeID_t>(
+            srcNodeIDVector->state->selVector->selectedPositions[0]);
+        bfsMorsel->markSrc(nodeID);
+        return false;
+    }
+    case SchedulerType::nThreadkMorsel: {
+        switch (globalState) {
+        case COMPLETE: {
+            state = {globalState, NO_WORK_TO_SHARE};
```
andyfengHKU, line 39: use COMPLETE directly instead of globalState. Apply the same style across this PR

src/include/processor/operator/recursive_extend/bfs_state.h
```diff
@@ -30,18 +83,85 @@ class TargetDstNodes {
...
+    // Maximum offset of dst nodes.
+    common::offset_t maxOffset;
+    uint64_t upperBound;
+    uint64_t lowerBound;
+    uint32_t numThreadsActiveOnMorsel;
+    uint64_t nextDstScanStartIdx;
+    uint64_t inputFTableTupleIdx;
+};
+
+struct BaseBFSMorsel {
```
andyfengHKU, line 154: Once you finish migrating all type of recursive join. One of your `BaseBFSMorsel` and `SSSPSharedState` should be simplified. Most likely your `BaseBFSMorsel` is the one that only stores a pointer to `SSSPSharedState` plus a range that is locally visible.

src/processor/operator/recursive_extend/bfs_state.cpp
```diff
@@ -0,0 +1,203 @@
...
+        return true;
+    }
+    default:
+        throw common::RuntimeException(
+            &"Unknown local state encountered inside SSSPSharedState: "[ssspLocalState]);
+    }
+}
+
+bool SSSPSharedState::hasWork() const {
+    if (ssspLocalState == EXTEND_IN_PROGRESS && nextScanStartIdx < bfsLevelNodeOffsets.size()) {
```
andyfengHKU, line 74: It's recommended to do 
```
if 
    return
if 
    return
```

rather than
```
if 
   return
else if
   return
```
Apply across PR if you want

src/processor/operator/recursive_extend/bfs_state.cpp
```diff
@@ -0,0 +1,203 @@
...
+    } else if (ssspLocalState == PATH_LENGTH_WRITE_IN_PROGRESS &&
+               nextDstScanStartIdx < visitedNodes.size()) {
+        return true;
+    } else {
+        return false;
+    }
+}
+
+bool SSSPSharedState::finishBFSMorsel(std::unique_ptr<BaseBFSMorsel>& bfsMorsel) {
+    mutex.lock();
```
andyfengHKU, line 85: use `std::unique_lock lck{mutex};`

src/processor/operator/recursive_extend/recursive_join.cpp
```diff
@@ -137,37 +139,137 @@ bool RecursiveJoin::getNextTuplesInternal(ExecutionContext* context) {
...
+        if (checkState) {
+            return false;
+        }
+        computeBFSOneThreadOneMorsel(context);
+        return true;
+    } else {
+        while (true) {
+            if (bfsMorsel->hasSSSPSharedState()) {
+                SSSPLocalState state{NO_WORK_TO_SHARE};
+                bool checkState = bfsMorsel->ssspSharedState->getBFSMorsel(bfsMorsel, state);
```
andyfengHKU, line 215: inline with l216

src/processor/operator/recursive_extend/recursive_join.cpp
```diff
@@ -137,37 +139,137 @@ bool RecursiveJoin::getNextTuplesInternal(ExecutionContext* context) {
...
+                if (!checkState) {
+                    computeBFSnThreadkMorsel(context);
+                    if (bfsMorsel->ssspSharedState->finishBFSMorsel(bfsMorsel)) {
```
andyfengHKU, line 218: this is a bad call chain. Either do `ssspSharedState.finishBFSMorsel(morsel)` or `morsel.finish()`

src/include/processor/operator/recursive_extend/bfs_state.h
```diff
@@ -30,18 +83,85 @@ class TargetDstNodes {
...
+          nextScanStartIdx{0u}, numVisitedNodes{0u}, visitedNodes{std::vector<uint8_t>(
+                                                         maxNodeOffset_ + 1, NOT_VISITED)},
+          pathLength{std::vector<uint8_t>(maxNodeOffset_ + 1, 0u)},
+          bfsLevelNodeOffsets{std::vector<common::offset_t>()}, srcOffset{0u},
+          maxOffset{maxNodeOffset_}, upperBound{upperBound_}, lowerBound{lowerBound_},
+          numThreadsActiveOnMorsel{0u}, nextDstScanStartIdx{0u}, inputFTableTupleIdx{0u} {}
+
+    void reset(TargetDstNodes* targetDstNodes);
+
+    bool getBFSMorsel(std::unique_ptr<BaseBFSMorsel>& bfsMorsel, SSSPLocalState& state);
```
andyfengHKU, line 115: Instead of passing `SSSPLocalState&` why not just return state. Also just give `BaseBFSMorsel*` as parameter change across PR.

src/processor/operator/recursive_extend/bfs_scheduler.cpp
```diff
@@ -0,0 +1,192 @@
...
+        case IN_PROGRESS: {
+            if (numActiveSSSP < activeSSSPSharedState.size()) {
+                auto inputFTableMorsel = inputFTableSharedState->getMorsel();
+                if (inputFTableMorsel->numTuples == 0) {
+                    globalState = (numActiveSSSP == 0) ? COMPLETE : IN_PROGRESS_ALL_SRC_SCANNED;
+                    state = {globalState, NO_WORK_TO_SHARE};
+                    return true;
+                }
+                numActiveSSSP++;
+                auto newSSSPSharedState =
```
andyfengHKU, line 54: I would avoid creating new object and instead always reuse the same SSSPSharedState object.

Think in this way. The maximum SSSPSharedState you can have at the same time equals to the maximum number of threads. You should create the object at compile time. At runtime only write to it. This is database processing principal, avoid creating new object as much as you can.

src/processor/operator/recursive_extend/bfs_scheduler.cpp
```diff
@@ -0,0 +1,192 @@
...
+            size++;
+        }
+        startScanIdxAndSize.first++;
+    }
+    if (size > 0) {
+        dstNodeIDVector->state->initOriginalAndSelectedSize(size);
+        // We need to rescan the FTable to get the source for which the pathLengths were computed.
+        // This is because the thread that scanned FTable initially might not be the thread writing
+        // the pathLengths to its vector.
+        inputFTableSharedState->getTable()->scan(vectorsToScan,
```
andyfengHKU, line 184: Instead of rescanning table. Maybe you should give the src as a parameter. Though I didn't read the parallel writing output part very carefully. I would rather take another look once you finish the write for path as well.

src/processor/operator/recursive_extend/bfs_scheduler.cpp
anuchak: >always reuse the same SSSPSharedState object

Yeah this has 1 major problem, if you have 4 threads working on the same SSSP and then 3 of them finished their work and try to reset the existing SSSP object, the other straggler thread will not be aware of this.
It might still be doing some extension work and end up updating the `visited` vector of this new SSSP.

Something similar could happen let's say for path Length writing step where 3 of them finished writing the lengths to their ValueVectors and then reset the destination vector while the 4th thread was still reading it leading to some values getting lost.

One way I can think of to handle this is to not only check if a morsel is `MORSEL_COMPLETE` but also if `numThreadsActiveOnThread` is 0 or not. This should handle the 1st case. For the 2nd case, there is no thread counter currently for tracking who is helping.

src/include/processor/mapper/plan_mapper.h
anuchak: Removed

src/include/processor/operator/recursive_extend/recursive_join.h
anuchak: done

src/processor/operator/recursive_extend/bfs_scheduler.cpp
anuchak: done

src/include/processor/operator/recursive_extend/bfs_state.h
anuchak: yeah once all shortest, shortest and var len are added

src/processor/operator/recursive_extend/bfs_state.cpp
anuchak: done

src/processor/operator/recursive_extend/recursive_join.cpp
anuchak: changed

benchmark/queries/ldbc-sf100/shortest_path_ldbc100/q37.benchmark
anuchak: Raised a separate PR

src/processor/operator/recursive_extend/bfs_state.cpp
anuchak: changed

src/include/processor/operator/recursive_extend/bfs_state.h
anuchak: changed

src/processor/operator/recursive_extend/recursive_join.cpp
anuchak: changed

src/processor/operator/recursive_extend/bfs_scheduler.cpp
anuchak: As per our discussion, I've changed this part's logic to reuse existing SharedState objects instead of creating a new one everytime. And for thread safety 3 conditions are being checked, before deciding which position to reuse.

src/include/processor/operator/recursive_extend/bfs_state.h
```diff
@@ -30,18 +83,86 @@ class TargetDstNodes {
 
     inline uint64_t getNumNodes() const { return numNodes; }
 
+    inline frontier::node_id_set_t getNodeIDs() const { return nodeIDs; }
```
andyfengHKU, line 86: be careful of this. The caller might be copying.

src/processor/operator/recursive_extend/bfs_scheduler.cpp
```diff
@@ -0,0 +1,204 @@
...
+                        newSharedStateIdx = i;
+                        break;
+                    }
+                    activeSSSPSharedState[i]->mutex.lock();
+                    // 3 conditions to check to decide if an existing SSSPSharedState object is
+                    // reusable or not (to ensure thread safety):
+                    // 1) the local state has been marked as MORSEL_COMPLETE
+                    // 2) no. of threads active (tracks threads doing bfs extension) is 0
+                    // 3) set containing thread_id of threads doing path length writing is empty
+                    if (activeSSSPSharedState[i]->ssspLocalState == MORSEL_COMPLETE &&
```
andyfengHKU, line 70: warp this check in SSSPSharedState

anuchak: done

PR #1827, opened 2023-07-17
Add functions exposing conversion to Arrow ArrayData

### LINKED ISSUE #1804: Support serialization of query results to in-memory arrow data types
Initially brought up on slack.

This is already exposed in the Python API via [PyQueryResult.getAsArrow](https://github.com/kuzudb/kuzu/blob/d6efe97a61eba936b568006509026186d820acf8/tools/python_api/src_cpp/py_query_result.cpp#L211-L223), and could be extended to at least the C++ and Rust APIs.

The kuzu Python API uses the [table](https://arrow.apache.org/docs/cpp/tables.html#tables) data format, which is limited to the C++ arrow API and the APIs built on top of it like the python API.

However it looks like [RecordBatch](https://arrow.apache.org/docs/cpp/tables.html#record-batches) can be sent/shared between implementations, so we could probably use the existing arrow C++ code that the python API uses and then send the data to other APIs.

benjaminwinger:
Fixes #1804

I've added functions to the C and C++ API which allow conversion of query result data to the structures defined in the [Arrow C Data Interface](https://arrow.apache.org/docs/format/CDataInterface.html#structure-definitions).

The rust API takes this a step further and, similarly to the python API, converts it to the `arrow::array::ArrayData` type (which seems a little difficult to use by itself, but converting it to more ergonomic types requires a lot of copying and probably has poor performance, so I thought it best to leave it as ArrayData and let the user decide if they want to use it directly or convert it).
The arrow conversion is hidden behind an `arrow` feature to avoid pulling in the large arrow crate by default.

Rust test will probably fail with the arrow feature. Testing locally the latest version of one of the arrow-rs dependencies requires rust 1.70, which is very recent; I had to manually force it to use an earlier version to get it to build on rust 1.69.

### DIFF
--- a/src/include/c_api/kuzu.h
+++ b/src/include/c_api/kuzu.h
@@ -4,6 +4,61 @@
 #include <stdlib.h>
 #include <string.h>
 
+// The Arrow C data interface.
+// https://arrow.apache.org/docs/format/CDataInterface.html
+
+#include <stdint.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#ifndef ARROW_C_DATA_INTERFACE
+#define ARROW_C_DATA_INTERFACE
+
+#define ARROW_FLAG_DICTIONARY_ORDERED 1
+#define ARROW_FLAG_NULLABLE 2
+#define ARROW_FLAG_MAP_KEYS_SORTED 4
+
+struct ArrowSchema {
+    // Array type description
+    const char* format;
+    const char* name;
+    const char* metadata;
+    int64_t flags;
+    int64_t n_children;
+    struct ArrowSchema** children;
+    struct ArrowSchema* dictionary;
+
+    // Release callback
+    void (*release)(struct ArrowSchema*);
+    // Opaque producer-specific data
+    void* private_data;
+};
+
+struct ArrowArray {
+    // Array data description
+    int64_t length;
+    int64_t null_count;
+    int64_t offset;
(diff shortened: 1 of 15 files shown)


src/include/c_api/kuzu.h
```diff
@@ -4,6 +4,8 @@
 #include <stdlib.h>
 #include <string.h>
 
+#include <common/arrow/arrow.h>
```
benjaminwinger, line 7: This is going to require some packaging changes so that it gets included in the released C API header. Either combining them at release time like with the C++ headers, or maybe just inlined here.

src/include/c_api/kuzu.h
```diff
@@ -439,6 +494,20 @@ KUZU_C_API void kuzu_query_result_write_to_csv(kuzu_query_result* query_result,
...
+ * It is the caller's responsibility to call the release function to release the underlying data
+ */
+struct ArrowSchema kuzu_query_result_get_arrow_schema(kuzu_query_result* query_result);
+
+/**
+ * @return An arrow array representation of the query result
+ *
+ * It is the caller's responsibility to call the release function to release the underlying data
+ */
+struct ArrowArray kuzu_query_result_to_arrow_array(kuzu_query_result* query_result);
```
ray6080, line 509: hmm should this be `kuzu_query_result_get_next_arrow_chunk `?

ray6080 (approved):

PR #1863, opened 2023-07-28
Serial fix node group

aziz-mu:
Fixing serial in node-groups branch. Also added a test/dataset for serial copy which required more than one node group

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,6 @@
 cmake_minimum_required(VERSION 3.11)
 
-project(Kuzu VERSION 0.0.6.2 LANGUAGES CXX)
+project(Kuzu VERSION 0.0.6.3 LANGUAGES CXX)
 
 find_package(Threads REQUIRED)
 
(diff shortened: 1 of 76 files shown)


ray6080 (changes requested):
Just took a quick look. I think there are still something missing in this PR before I dive into more detailed review:
1. You only handled the case for CSV, can you also add the logic to Parquet and NPY?
2. It seems `CombineChunksToBatch` is not performing zero-copy internally (found something [here](https://stackoverflow.com/questions/70692502/converting-apache-arrow-table-to-recordbatch-in-c), double check their source code? The concatenation can be costful. We should avoid it. Are there any other solutions in your mind? if not, can you try to explore my previous suggestion on using `ChunkedArray`? For which you need to change ValueVector to be able to keep `ChunkedArray` instead of `Array`.
3. There is no distinction between forcing order preserving or not (has SERIAL or not) in reading from csv files. This might be fine for CSV, as it has to be in a serial mode anyways, but when you replicate changes to Parquet and NPY, we do need to consider the case that we can read concurrently (which is critical) if there is no SERIAL. So better to differentiate the reading logic of the two cases regarding to preserving order or not.

PR #1870, opened 2023-07-30
Support multi-label delete

andyfengHKU:


test/test_files/tinysnb/update_node/delete.test
```diff
@@ -3,31 +3,21 @@
 
 --
 
--CASE DeleteNodeAfterInsertTest
```
andyfengHKU, line 6: put these test back

src/include/processor/operator/update/delete_manager.h
```diff
@@ -0,0 +1,137 @@
+#pragma once
+
+#include "processor/execution_context.h"
+#include "processor/result/result_set.h"
+#include "storage/store/node_table.h"
+
+namespace kuzu {
+namespace processor {
+
+class NodeDeleteManager {
```
ray6080, line 10: I wouldn't call this Manager, maybe executor is a more appropriate suffix. Another term in my mind is `NodeDeleter`. Not sure if this is a valid word. You might need to take a further look.

ray6080 (approved):

PR #1880, opened 2023-08-01
Enable multi label set

andyfengHKU:


src/optimizer/factorization_rewriter.cpp
```diff
@@ -133,9 +133,10 @@ void FactorizationRewriter::visitFilter(planner::LogicalOperator* op) {
 
 void FactorizationRewriter::visitSetNodeProperty(planner::LogicalOperator* op) {
     auto setNodeProperty = (LogicalSetNodeProperty*)op;
-    for (auto i = 0u; i < setNodeProperty->getNumNodes(); ++i) {
-        auto lhsNodeID = setNodeProperty->getNode(i)->getInternalIDProperty();
-        auto rhs = setNodeProperty->getSetItem(i).second;
+    for (auto& info : setNodeProperty->getInfosRef()) {
+        auto node = (binder::NodeExpression*)info->nodeOrRel.get();
```
ray6080, line 137: - Remove `binder::` prefix.
- Let's use `static_cast`.

src/planner/operator/logical_set.cpp
```diff
@@ -5,10 +5,42 @@
...
+        infosCopy.push_back(info->copy());
+    }
+    return std::make_unique<LogicalSetRelProperty>(std::move(infosCopy), children[0]->copy());
+}
+
+f_group_pos_set LogicalSetRelProperty::getGroupsPosToFlatten(uint32_t idx) {
     f_group_pos_set result;
-    auto rel = rels[setItemIdx];
-    auto rhs = setItems[setItemIdx].second;
+    auto rel = (binder::RelExpression*)infos[idx]->nodeOrRel.get();
```
ray6080, line 42: Ditto. 
- Remove `binder::` prefix.
- Use `static_cast`.

src/processor/operator/update/set_executor.cpp
```diff
@@ -0,0 +1,68 @@
...
+
+void NodeSetExecutor::init(ResultSet* resultSet, ExecutionContext* context) {
+    nodeIDVector = resultSet->getValueVector(nodeIDPos).get();
+    evaluator->init(*resultSet, context->memoryManager);
+    rhsVector = evaluator->resultVector.get();
+}
+
+void SingleLabelNodeSetExecutor::set() {
+    evaluator->evaluate();
+    for (auto i = 0u; i < nodeIDVector->state->selVector->selectedSize; ++i) {
```
ray6080, line 14: This is a bit confusing. Can we not directly use existing write interface of Column? so it can be 
```
column->write(nodeIDVector, rhsVector);
```

Also, what do you mean lhs and rhs here? I don't quite grasp where the convention comes from. I feel nodeIDVector and propertyVector is more clear.

src/processor/operator/update/set_executor.cpp
```diff
@@ -0,0 +1,68 @@
...
+        auto lhsPos = nodeIDVector->state->selVector->selectedPositions[i];
+        auto& nodeID = nodeIDVector->getValue<common::internalID_t>(lhsPos);
+        if (!tableIDToColumn.contains(nodeID.tableID)) {
+            continue;
+        }
+        auto column = tableIDToColumn.at(nodeID.tableID);
+        auto rhsPos = lhsPos;
+        if (rhsVector->state->selVector->selectedSize == 1) {
```
ray6080, line 35: This logic is quite implicit and if we handle the flat/unflat here, it seems to be duplicated with what we do in `Column::write(ValueVector* nodeIDVector, ValueVector* vectorToWriteFrom)`.

Here are two alternative solutions in my mind:
1. Manually flatten nodeIDVector here, so we can reuse `Column::write(ValueVector* nodeIDVector, ValueVector* vectorToWriteFrom)`.
2. Add a new interface `Column::write(offset_t nodeOffset, ValueVector* vectorToWriteFrom)`, which can handle both cases of `vectorToWriteFrom` being flat or unflat.

ray6080 (approved):

PR #1884, opened 2023-08-02
Boolean bitpacking

benjaminwinger:
Implements boolean bitpacking for on-disk booleans and unpacks them when reading into ValueVectors.
Requires #1862.

I also optimized `ColumnChunk::templateCopyArrowArray<bool>` to use copyNullMask instead of setting the bits individually. Arrow's null bits are flipped in comparison to kuzu's, so I modified the copy function to allow the result to be inverted.

### DIFF
--- a/src/include/storage/storage_utils.h
+++ b/src/include/storage/storage_utils.h
@@ -52,7 +52,20 @@ struct PageElementCursor {
 };
 
 struct PageUtils {
-    static uint32_t getNumElementsInAPage(uint32_t elementSize, bool hasNull);
+    static constexpr uint32_t getNumElementsInAPage(uint32_t elementSize, bool hasNull) {
+        assert(elementSize > 0);
+        auto numBytesPerNullEntry = common::NullMask::NUM_BITS_PER_NULL_ENTRY >> 3;
+        auto numNullEntries =
+            hasNull ?
+                (uint32_t)ceil((double)common::BufferPoolConstants::PAGE_4KB_SIZE /
+                               (double)(((uint64_t)elementSize
+                                            << common::NullMask::NUM_BITS_PER_NULL_ENTRY_LOG2) +
+                                        numBytesPerNullEntry)) :
+                0;
+        return (common::BufferPoolConstants::PAGE_4KB_SIZE -
+                   (numNullEntries * numBytesPerNullEntry)) /
+               elementSize;
+    }
 
     // This function returns the page pageIdx of the page where element will be found and the pos of
     // the element in the page as the offset.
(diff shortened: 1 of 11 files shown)


src/storage/store/node_column.cpp
```diff
@@ -437,6 +487,8 @@ std::unique_ptr<NodeColumn> NodeColumnFactory::createNodeColumn(const LogicalTyp
     BMFileHandle* metadataFH, BufferManager* bufferManager, WAL* wal) {
     switch (dataType.getLogicalTypeID()) {
     case LogicalTypeID::BOOL:
+        return std::make_unique<BoolNodeColumn>(
```
ray6080, line 490: Wrap this line around `{}`. We are following the code style of always wrapping case statements inside `{}`.

src/storage/store/node_column.cpp
```diff
@@ -352,17 +389,30 @@ void NodeColumn::scan(transaction::Transaction* transaction, node_group_idx_t no
     scanUnfiltered(transaction, pageCursor, numValuesToScan, resultVector, offsetInVector);
 }
 
+// Page size must be aligned to 8 byte chunks for the 64-bit NullMask algorithms to work
+// without the possibility of memory errors from reading/writing off the end of a page.
+static_assert(PageUtils::getNumElementsInAPage(1, false) % 8 == 0);
```
ray6080, line 394: Add inline comment `/*requireNullColumn*/` after `false`. Same for L404.
We do this to simplify code reading without jumping to the function signature. (especially when reviewing code on github website)

src/include/storage/storage_utils.h
```diff
@@ -52,7 +52,20 @@ struct PageElementCursor {
 };
 
 struct PageUtils {
-    static uint32_t getNumElementsInAPage(uint32_t elementSize, bool hasNull);
+    static constexpr uint32_t getNumElementsInAPage(uint32_t elementSize, bool hasNull) {
```
ray6080, line 55: Why move from cpp to header file? If you want to keep it inside the header file, mark this function as inline.

ray6080 (approved):
LGTM!

src/include/storage/storage_utils.h
benjaminwinger: I'd moved it because so it can be constexpr, and constexpr (according to [cppreference](https://en.cppreference.com/w/cpp/language/constexpr)) implies inline (for cases where it's evaluated at runtime presumably).

Frankly, the static_assert I'd used it for is probably not necessary, but I thought it at least made more sense to make it a single static assert than multiple runtime asserts.

ray6080: Got it. Yeah, this makes sense.

PR #1886, opened 2023-08-03
Fix Serial for node-groups

aziz-mu:
This PR fixes COPY from CSV, Parquet, NPY for tables which include the SERIAL data type. Also added tests. 

This is a continuation of this PR: https://github.com/kuzudb/kuzu/pull/1863, rebase/merge was too complicated so I just added the changes here


ray6080 (changes requested):
Can you check my comments? I think there are still several critical things need to be addressed to make CopyNode correct and performant.

src/storage/copier/read_file_state.cpp
```diff
@@ -75,6 +104,67 @@ void ReadParquetSharedState::countNumRows() {
     }
 }
 
+std::unique_ptr<ReadFileMorsel> ReadParquetSharedState::getMorselSerial() {
```
ray6080, line 107: You seem to forget to rework this function, as the control logic is different from the csv one. Can you redo it as the csv one?

src/common/vector/value_vector.cpp
```diff
@@ -182,6 +182,13 @@ void ArrowColumnVector::setArrowColumn(ValueVector* vector, std::shared_ptr<arro
     arrowColumnBuffer->column = std::move(column);
 }
 
+void ArrowColumnVector::setArrowColumn(
```
ray6080, line 185: Remove the above function `ArrowColumnVector::setArrowColumn(ValueVector* vector, std::shared_ptr<arrow::Array> column)`.

src/include/common/vector/auxiliary_buffer.h
```diff
@@ -56,6 +57,13 @@ class ArrowColumnAuxiliaryBuffer : public AuxiliaryBuffer {
     std::shared_ptr<arrow::Array> column;
 };
 
+class ArrowColumnChunkedAuxiliaryBuffer : public AuxiliaryBuffer {
```
ray6080, line 60: I don't think you should add a new class. Just reuse `ArrowColumnAuxiliaryBuffer`, and replace `arrow::Array` with `arrow::ChunkedArray`.

src/include/common/vector/value_vector.h
```diff
@@ -218,7 +219,13 @@ class ArrowColumnVector {
         return reinterpret_cast<ArrowColumnAuxiliaryBuffer*>(vector->auxiliaryBuffer.get())->column;
     }
 
+    static inline std::shared_ptr<arrow::ChunkedArray> getChunkedArrowColumn(ValueVector* vector) {
```
ray6080, line 222: Ditto. Don't add new function. Instead modify `getArrowColumn `.

src/include/common/vector/value_vector.h
```diff
@@ -218,7 +219,13 @@ class ArrowColumnVector {
...
+        return reinterpret_cast<ArrowColumnChunkedAuxiliaryBuffer*>(vector->auxiliaryBuffer.get())
+            ->chunkedColumn;
+    }
+
     static void setArrowColumn(ValueVector* vector, std::shared_ptr<arrow::Array> column);
+    static void setArrowColumn(ValueVector* vector, std::shared_ptr<arrow::ChunkedArray> column);
```
ray6080, line 228: Ditto. Don't add new function. Instead modify `setArrowColumn` .

src/include/storage/copier/read_file_state.h
```diff
@@ -24,9 +25,9 @@ class ReadFileMorsel {
 };
 
 // For CSV file, we need to read in streaming mode, so we need to read one batch at a time.
-class ReadCSVMorsel : public ReadFileMorsel {
+class ReadSerialMorsel : public ReadFileMorsel {
```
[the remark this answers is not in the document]
ray6080: Make sense.

src/processor/operator/copy/copy_node.cpp
```diff
@@ -82,16 +82,25 @@ void CopyNode::executeInternal(ExecutionContext* context) {
...
         uint64_t numAppendedTuples = 0;
         while (numAppendedTuples < numTuplesToAppend) {
             auto numAppendedTuplesInNodeGroup = localNodeGroup->append(
                 resultSet, copyNodeInfo.dataColumnPoses, numTuplesToAppend - numAppendedTuples);
             numAppendedTuples += numAppendedTuplesInNodeGroup;
             if (localNodeGroup->isFull()) {
-                auto nodeGroupIdx = sharedState->getNextNodeGroupIdx();
+                node_group_idx_t nodeGroupIdx;
+                if (copyNodeInfo.preservingOrder) {
+                    nodeGroupIdx = nodeOffset / StorageConstants::NODE_GROUP_SIZE - 1;
```
ray6080, line 99: As we change the `nodeOffset` to track the start nodeOffset, instead of the end one, we should also change L99 here. You should be able to directly reuse `getNodeGroupIdxFromNodeOffset` function.

ray6080 (approved):

ray6080:
Can you also check tests under `test_files/tck/match` and uncomment them? Those tests should also be able to run now after your fix.

PR #1893, opened 2023-08-05
Support Parquet filetype on COPY TO

rfdavid:
### Implementation

This PR is an extension of #1716 to support copying query results to `.parquet` files. Using the arrow library, there are mainly three ways to write to parquet files (in the decreasing order of level of abstraction): using `StreamWriter`, `WriteTable` and `WriteBatch` [(reference here)](https://arrow.apache.org/docs/cpp/parquet.html). `StreamWriter` doesn't support nested table, and therefore doesn't suit our needs. `WriteTable` provides builders to write logical datatypes (eg, int64 builder), whereas in the low-level API we make use of `WriteBatch` to write the batch of values. The repetition and definition levels [[1]](https://blog.twitter.com/engineering/en_us/a/2013/dremel-made-simple-with-parquet) [[2]](https://static.googleusercontent.com/media/research.google.com/en//pubs/archive/36632.pdf) must be calculated in the code. This PR is concentrated on the low-level API, which gives us more flexibility in the implementation, and it's one step before we create our own parquet writer.

Support `COPY (...) TO 'out.parquet'`

To be addressed:

- Implement ~~`interval`~~, `union` , ~~`internal_id`~~ ,`fixed_list` and `map` data types
- ~~Only int64 is working on `var_list` and struct, the others must be implemented~~
- ~~Fix a bug inside List>Struct~~
- ~~Perform tests on large datasets and check how the flush is actually working~~
- Schema can be improved by making use of more required fields, hence saving more bits
- Implement `BufferedRowGroup` 
- ~~Handle null values inside a nested data type~~


src/binder/bind/bind_copy.cpp
```diff
@@ -19,17 +19,20 @@ std::unique_ptr<BoundStatement> Binder::bindCopyToClause(const Statement& statem
...
     for (auto& column : columns) {
         auto columnName = column->hasAlias() ? column->getAlias() : column->toString();
         columnNames.push_back(columnName);
+        columnTypes.push_back(column->getDataType());
     }
-    if (fileType != CopyDescription::FileType::CSV) {
-        throw BinderException("COPY TO currently only supports csv files.");
+    if (fileType != CopyDescription::FileType::CSV &&
+        fileType != CopyDescription::FileType::PARQUET) {
+        throw BinderException("COPY TO currently only supports csv and parquet files.");
```
ray6080, line 32: Can you refactor the error message as a const of `ExceptionMessage` in `exception.h`?

src/include/processor/operator/persistent/parquet_column_writer.h
```diff
@@ -0,0 +1,80 @@
...
+
+// Kuzu column represents a single column in Kuzu, which may contains multiple Parquet columns
+// writeColumn will be called for each Kuzu column.
+class ParquetColumnWriter {
+public:
+    ParquetColumnWriter(int totalColumns, std::vector<int> maxDefinitionLevels)
+        : maxDefinitionLevels(maxDefinitionLevels), totalColumns(totalColumns){};
+
+    void writeColumn(
+        int kuzuColumn, common::ValueVector* vector, parquet::RowGroupWriter* rowWriter);
```
ray6080, line 17: I think we can just simplify `kuzuColumn` as `columns`. Ditto on the above class comment. 
```
A single column in Kuzu may contain multiple Parquet columns, for example, ...
```

Also, `rowWriter` to `rowGroupWriter`.

src/include/processor/operator/persistent/parquet_column_writer.h
```diff
@@ -0,0 +1,80 @@
...
+
+    // Extract dremel encoding levels
+    int getRepetitionLevel(int currentElementIdx, int parentElementIdx, int depth);
+
+    std::vector<int> maxDefinitionLevels;
+
+    // Properties for nested lists and structs
+    bool isListStarting;
+
+    int currentKuzuColumn;
```
ray6080, line 64: `currentKuzuColumn` -> `currentColumn`

src/processor/operator/persistent/parquet_column_writer.cpp
```diff
@@ -0,0 +1,252 @@
...
+
+using namespace kuzu::common;
+
+namespace kuzu {
+namespace processor {
+
+void ParquetColumnWriter::nextParquetColumn(LogicalTypeID logicalTypeID) {
+    switch (logicalTypeID) {
+    case LogicalTypeID::BOOL:
+        boolWriter = static_cast<parquet::BoolWriter*>(rowWriter->column(currentParquetColumn));
```
ray6080, line 13: Ditto. I think we should only keep one columnWriter pointer, instead of keeping one for each data type. Dynamic casting is fine here.
```
auto columnWriter = rowGroupWriter->column(currentParquetColumn)
    switch (logicalTypeID) {
    case LogicalTypeID::BOOL:
        auto boolWriter = static_cast<parquet::BoolWriter*>(columnWriter);
... ...
```

src/processor/operator/persistent/parquet_file_writer.cpp
```diff
@@ -0,0 +1,194 @@
...
+    //        }
+    //      }
+    //    }
+    case LogicalTypeID::FIXED_LIST:
+    case LogicalTypeID::VAR_LIST: {
+        auto childLogicalType = VarListType::getChildType(&logicalType);
+        auto childNode = kuzuTypeToParquetType(parquetColumnsCount, columnName, *childLogicalType,
+            parquet::Repetition::OPTIONAL, length);
+        auto repeatedGroup =
+            parquet::schema::GroupNode::Make("", parquet::Repetition::REPEATED, {childNode});
```
ray6080, line 156: What's the empty string here? Should we give a const name?

src/processor/processor.cpp
```diff
@@ -36,7 +36,7 @@ std::shared_ptr<FactorizedTable> QueryProcessor::execute(
 }
 
 void QueryProcessor::decomposePlanIntoTasks(
-    PhysicalOperator* op, PhysicalOperator* parent, Task* parentTask, ExecutionContext* context) {
+        PhysicalOperator* op, PhysicalOperator* parent, Task* parentTask, ExecutionContext* context) {
```
ray6080, line 39: hmm revert this change?

src/include/processor/operator/persistent/parquet_file_writer.h
```diff
@@ -0,0 +1,48 @@
...
+// openFile: opens the file and create a arrow::io::FileOutputStream object
+// init:
+//   - generate the schema
+//   - calculate the max definition levels and number of primitive nodes
+//   - initialize parquetColumnWriter
+// writeValues : take a vector of ValueVector and pass to parquetColumnWriter
+class ParquetFileWriter : public CSVParquetWriter {
+public:
+    ParquetFileWriter(){};
+    void openFile(const std::string& filePath) override;
```
ray6080, line 21: Mark this function as `final`. Ditto on other overriden functions in this file.

src/include/processor/operator/persistent/parquet_file_writer.h
```diff
@@ -0,0 +1,48 @@
...
+namespace processor {
+
+// ParquetWriter performs the following:
+// openFile: opens the file and create a arrow::io::FileOutputStream object
+// init:
+//   - generate the schema
+//   - calculate the max definition levels and number of primitive nodes
+//   - initialize parquetColumnWriter
+// writeValues : take a vector of ValueVector and pass to parquetColumnWriter
+class ParquetFileWriter : public CSVParquetWriter {
```
ray6080, line 18: I would also add some comments on the definition level and repetition level, by one or two examples perhaps. Or put a link for external reference material here.

ray6080 (approved):

src/processor/operator/persistent/parquet_file_writer.cpp
```diff
@@ -0,0 +1,182 @@
...
+}
+
+void ParquetFileWriter::init() {
+    std::shared_ptr<parquet::schema::GroupNode> schema;
+    std::unordered_map<int, ParquetColumnDescriptor> columnDescriptors;
+    generateSchema(schema, columnDescriptors);
+    // this is an useful helper to print the generated schema
+    // parquet::schema::PrintSchema(schema.get(), std::cout);
+    parquet::WriterProperties::Builder builder;
+    builder.compression(parquet::Compression::SNAPPY);
```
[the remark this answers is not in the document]
rfdavid: No reason. From all I've seen, it looks like SNAPPY is the most common, but I'm not sure. I was thinking of closing this PR and then implementing options to `COPY TO` statement, the same as we have on `COPY FROM`. What do you recommend?

`COPY (match (p:person) return p.name) TO 'out.parquet' (COMPRESSION=GZIP)`

src/processor/operator/persistent/parquet_file_writer.cpp
```diff
@@ -0,0 +1,182 @@
...
+// DOUBLE: IEEE 64-bit floating point values
+// BYTE_ARRAY: arbitrarily long byte arrays.
+// https://github.com/apache/parquet-cpp/blob/master/src/parquet/column_writer.h
+std::shared_ptr<parquet::schema::Node> ParquetFileWriter::kuzuTypeToParquetType(
+    ParquetColumnDescriptor& columnDescriptor, std::string& columnName,
+    const LogicalType& logicalType, parquet::Repetition::type repetition, int length) {
+    parquet::Type::type parquetType;
+    parquet::ConvertedType::type convertedType;
+    convertedType = parquet::ConvertedType::NONE;
+    switch (logicalType.getLogicalTypeID()) {
```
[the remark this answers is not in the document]
rfdavid: You mean doing something like this?
``` 
switch (logicalType.getLogicalTypeID()) {
    case LogicalTypeID::BOOL: {
        return parquet::schema::PrimitiveNode::Make(
            columnName, repetition, parquet::Type::BOOLEAN, convertedType, length);
    }
    case LogicalTypeID::INTERNAL_ID:
    case LogicalTypeID::STRING: {
        return parquet::schema::PrimitiveNode::Make(
            columnName, repetition, parquet::Type::BYTE_ARRAY, parquet::ConvertedType::UTF8, length);
    }
    case LogicalTypeID::INT64: {
        return parquet::schema::PrimitiveNode::Make(
            columnName, repetition, parquet::Type::INT64, parquet::ConvertedType::INT_64;, length);
    }
```
Or do you have a better idea?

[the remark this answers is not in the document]
rfdavid: Perhaps split primitive and nested into two different functions? 🤔

src/processor/operator/persistent/parquet_column_writer.cpp
rfdavid: I can only see the need of having to static_cast every time. Can this cause overhead?

src/processor/operator/persistent/parquet_file_writer.cpp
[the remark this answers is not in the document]
ray6080: Yeah. splitting primitive and nested sounds good to me.

src/common/types/types.cpp
```diff
@@ -605,6 +605,19 @@ bool LogicalTypeUtils::isNumerical(const LogicalType& dataType) {
     }
 }
 
+bool LogicalTypeUtils::isPrimitive(const LogicalType& dataType) {
```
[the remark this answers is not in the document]
ray6080: Agree on the part of "be more prone to be duplicated without knowing the existence of this". But I'm not sure if another function needs this later, it will end up with the same definition of `isPrimitive`.
`isNested` looks clearer to me. I'm fine to keep it inside `LogicalTypeUtils`.

PR #1895, opened 2023-08-06
Move the initialization of metadada disk arrays to wal-based transaction mechanism

ray6080:
This is a follow up fix to #1802 .
Header pages of metadata DAs are initialized through wal.

Additionally, this PR rewrites some ddl and copy tests that are testing output message and running auto commit inside write transaction.

### DIFF
--- a/src/catalog/catalog.cpp
+++ b/src/catalog/catalog.cpp
@@ -70,10 +70,19 @@ void Catalog::renameTable(table_id_t tableID, const std::string& newName) {
     catalogContentForWriteTrx->renameTable(tableID, newName);
 }
 
-void Catalog::addProperty(
+void Catalog::addNodeProperty(table_id_t tableID, const std::string& propertyName,
+    std::unique_ptr<LogicalType> dataType, std::unique_ptr<MetadataDAHInfo> metadataDAHInfo) {
+    initCatalogContentForWriteTrxIfNecessary();
+    catalogContentForWriteTrx->getTableSchema(tableID)->addNodeProperty(
+        propertyName, std::move(dataType), std::move(metadataDAHInfo));
+    wal->logAddPropertyRecord(
+        tableID, catalogContentForWriteTrx->getTableSchema(tableID)->getPropertyID(propertyName));
+}
+
+void Catalog::addRelProperty(
     table_id_t tableID, const std::string& propertyName, std::unique_ptr<LogicalType> dataType) {
     initCatalogContentForWriteTrxIfNecessary();
-    catalogContentForWriteTrx->getTableSchema(tableID)->addProperty(
+    catalogContentForWriteTrx->getTableSchema(tableID)->addRelProperty(
         propertyName, std::move(dataType));
     wal->logAddPropertyRecord(
         tableID, catalogContentForWriteTrx->getTableSchema(tableID)->getPropertyID(propertyName));
--- a/src/include/storage/storage_structure/disk_array.h
+++ b/src/include/storage/storage_structure/disk_array.h
@@ -231,6 +231,14 @@ public:
     InMemDiskArray(FileHandle& fileHandle, StorageStructureID storageStructureID,
         common::page_idx_t headerPageIdx, BufferManager* bufferManager, WAL* wal);
 
+    static inline common::page_idx_t addDAHPageToFile(BMFileHandle& fileHandle,
+        StorageStructureID storageStructureID, BufferManager* bufferManager, WAL* wal) {
+        DiskArrayHeader daHeader(sizeof(T));
+        return StorageStructureUtils::insertNewPage(fileHandle,
+            StorageStructureID{StorageStructureType::METADATA}, *bufferManager, *wal,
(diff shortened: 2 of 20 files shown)


src/include/storage/storage_structure/disk_array.h
```diff
@@ -231,6 +231,14 @@ class InMemDiskArray : public BaseInMemDiskArray<T> {
     InMemDiskArray(FileHandle& fileHandle, StorageStructureID storageStructureID,
         common::page_idx_t headerPageIdx, BufferManager* bufferManager, WAL* wal);
 
+    static inline common::page_idx_t initializeHeaderPage(BMFileHandle& fileHandle,
```
semihsalihoglu-uw, line 234: This is fine but it seems like this function belongs to a utils class about disk_array, instead of disk_array. That's because it's not really used to modify a disk array, it's used to create a disk array in a file handle. If it's the only function it wouldn't matter but consider it.

semihsalihoglu-uw, line 234: I would rename this to `addDAHPageToFile` or even `addDiskArrayToFile` and explain what it does in comments.

src/processor/operator/ddl/add_rel_property.cpp
```diff
@@ -6,11 +6,13 @@ namespace kuzu {
 namespace processor {
 
 void AddRelProperty::executeDDLInternal() {
-    AddProperty::executeDDLInternal();
+    defaultValueEvaluator->evaluate();
```
semihsalihoglu-uw, line 9: This code looks confusing. Why not just call 1 function, e.g., `getDefaultVal` that does evaluate as its first step (or call `evaluateAndGetDefaultValue` or something. Also it looks like you don't need to get the defaultVal as the first thing. You can get it either inlined (if you make it a single function) or right before calling ` StorageUtils::createFileForRelPropertyWithDefaultVal` on line 14 below.

src/storage/storage_structure/storage_structure_utils.cpp
```diff
@@ -17,6 +17,21 @@ std::pair<BMFileHandle*, page_idx_t> StorageStructureUtils::getFileHandleAndPhys
     }
 }
 
+common::page_idx_t StorageStructureUtils::insertNewPage(BMFileHandle& fileHandle,
```
semihsalihoglu-uw, line 20: So you are creating this function with a function as an argument but you use it only once. Why bother with the generality and refactoring if there are no other callers? Why not get in a uint8_t* and do memcpy as the default operation and you can extract an insertOp function as an argument if there are other callers.  That said,  are there really no other callers that adds a new page and could use this logic?

Update: I went over the code except for the insertOp, it looks like this function is almost identical to the `StorageStructure::addNewPageToFileHandle()` function, which does not return the newOriginalPage and has a noOp. So you should refactor those (maybe make the insertOp a default operator with a noOp and always return the newPageIdx.

semihsalihoglu-uw (approved):
I'm approving but I want to make sure the refactoring of `insertNewPage` and `addNewPageToFileHandle` is done. So ping me if for some reason you won't do that, I'd want to discuss.

src/storage/storage_manager.cpp
```diff
@@ -27,8 +27,12 @@ StorageManager::StorageManager(Catalog& catalog, MemoryManager& memoryManager, W
 
 std::unique_ptr<MetadataDAHInfo> StorageManager::initMetadataDAHInfo(const LogicalType& dataType) {
```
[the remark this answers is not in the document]
ray6080: Renamed to `createMetadataDAHInfo`.

src/catalog/catalog.cpp
```diff
@@ -71,10 +71,19 @@ void Catalog::renameTable(table_id_t tableID, const std::string& newName) {
     catalogContentForWriteTrx->renameTable(tableID, newName);
 }
 
-void Catalog::addProperty(
+void Catalog::addNodeProperty(table_id_t tableID, const std::string& propertyName,
```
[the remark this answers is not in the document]
ray6080: The caller has to `move`, I think it's clear that the ownership has been transferred.

src/include/storage/storage_structure/disk_array.h
ray6080: Renamed to `addDAHPageToFile`.

src/processor/operator/ddl/add_rel_property.cpp
ray6080: Moved the evaluate inside `getDefaultVal`, and inlined the call of `getDefaultVal`.

src/storage/storage_structure/storage_structure_utils.cpp
ray6080: Refactored this altogether with `StorageStructure::addNewPageToFileHandle()` and `NodeColumn:: addNewPageToDataFH()`.

src/include/storage/storage_structure/disk_array.h
ray6080: I think it's fine for now as this is the first static function here. We will extract them out later if we have more needs on more disk array static util functions.

PR #1902, opened 2023-08-07
Rework node table delete interface

ray6080:


src/include/processor/operator/persistent/delete_executor.h
```diff
@@ -9,61 +9,59 @@ namespace processor {
...
+    void init(ResultSet* resultSet, ExecutionContext* context) final;
+    void delete_(ExecutionContext* context) final;
 
     inline std::unique_ptr<NodeDeleteExecutor> copy() const final {
         return std::make_unique<MultiLabelNodeDeleteExecutor>(*this);
     }
 
 private:
     std::unordered_map<common::table_id_t, storage::NodeTable*> tableIDToTableMap;
+    std::unordered_map<common::table_id_t, std::unique_ptr<storage::NodeTable::DeleteState>>
```
andyfengHKU, line 63: We should probably wrap NodeTable and DeleteState into a pair/struct so that we don't need to do 2 lookup at runtime

andyfengHKU (approved):

src/include/processor/operator/persistent/delete_executor.h
ray6080: Yeah, struct or pair has better semantic here, though I wouldn't worry about the extra map lookup. I did try to wrap them into a struct, but awkwardly ended up with that now two fields (table, deleteState) of the struct are separately initialized in the constructor and `init()` call, which is also not a good design. I don't have a better idea in my mind for now, so will keep this as it is and we will revisit as we figure out how to simplify the multi-label and single-label design here.

PR #1906, opened 2023-08-07
Support return after create

andyfengHKU:


src/include/processor/operator/persistent/insert_executor.h
```diff
@@ -42,9 +42,11 @@ class RelInsertExecutor {
 public:
     RelInsertExecutor(storage::RelsStatistics& relsStatistics, storage::RelTable* table,
         const DataPos& srcNodePos, const DataPos& dstNodePos,
+        std::vector<DataPos> lhsVectorPositions,
```
ray6080, line 45: I would recommend adding a comment in insert_executor.h to give a very simple example on what lhs and rhs are. It will be much more easier to understand.

ray6080 (approved):
It seems we can also push the writing of lhs vectors into storage side. Maybe we should explore that option too.

PR #1911, opened 2023-08-08
Fix merge operator mark

andyfengHKU:


ray6080 (approved):

test/test_files/exceptions/binder/binder_error.test
```diff
@@ -5,6 +5,16 @@
 
 -CASE BinderError
 
+-LOG EmptyCreateClause
+-STATEMENT MATCH (a:person) CREATE (a);
+---- error
+Binder exception: Cannot resolve any node or relationship to create.
```
ray6080, line 11: I wonder if we can throw error message same as `Binder exception: Node table PERSON does not exist`. which seems more informative.

PR #1930, opened 2023-08-14
Unnest arbitrary subquery

andyfengHKU:
This PR implements unnest arbitrary subquery with hash join based execution.

### DIFF
--- a/src/binder/expression/expression.cpp
+++ b/src/binder/expression/expression.cpp
@@ -19,14 +19,25 @@ expression_vector Expression::splitOnAND() {
     return result;
 }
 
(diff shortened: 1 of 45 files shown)


src/processor/map/correlated_subquery_unnest_solver.cpp
```diff
@@ -0,0 +1,43 @@
+#include "optimizer/correlated_subquery_unnest_solver.h"
```
andyfengHKU, line 1: move to optimizer

ray6080 (approved):

PR #1932, opened 2023-08-14
Fix Rust deployment

### LINKED ISSUE #1927: Rust deployment failed
I tested build and deploy pipeline in dry run mode and the Rust build was failing with the following error:

```
error: failed to verify package tarball

Caused by:
  package `time-macros v0.2.11` cannot be built because it requires rustc 1.67.0 or newer, while the currently active rustc version is 1.65.0
```

Please refer to https://github.com/kuzudb/kuzu/actions/runs/5838847530/job/15836348078

benjaminwinger:
Fixes #1927.

The rustup install should probably be added to the dockerfile instead of included in the workflow, but I thought it would be easier to test it this way.

I've tested it and it's working: https://github.com/kuzudb/kuzu/actions/runs/5858238195/job/15881791303

I also had to fix the version parsing since` 0.0.6.5` is not a valid [semantic version](https://semver.org/) (so I made the fourth component be turned into a pre-release version, e.g. `0.0.6-5`).

### DIFF
--- a/.github/workflows/build-and-deploy.yml
+++ b/.github/workflows/build-and-deploy.yml
@@ -219,7 +219,6 @@ jobs:
     runs-on: kuzu-self-hosted-testing
     env:
       CARGO_REGISTRY_TOKEN: ${{ secrets.CARGO_REGISTRY_TOKEN }}
-      MSRV: 1.67
     steps:
       - uses: actions/checkout@v3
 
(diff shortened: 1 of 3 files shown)


mewim:
It seems this breaks rust format: https://github.com/kuzudb/kuzu/actions/runs/5858442707/job/15894933233?pr=1931

Let's breifly discuss it tomorrow. Redeploying the Docker containers maybe required to make this work.

benjaminwinger:
Oh, yes with the `minimal` profile it won't install rustfmt. Either dropping the `--profile minimal`, or adding `rustup component add rustfmt` should fix that.

mewim:
I think we should consider seperate the building environment from the regular testing environment. Since the builder now mutates the environment by running `rustup`, we should not run it on the testing container, so we can keep a stable environment for testing.

mewim:
Or, we can just install rust via `rustup` in `Dockerfile` and avoid running it everytime during the build process.

benjaminwinger:
Yes, I think installing it in the Dockerfile is a better idea.

I take it it's not starting up a fresh container per job/workflow then? I hadn't realised that this would propagate to the other jobs.

mewim:
We have a fixed set of containers for each environment instead of launching a new one for each job. I will redeploy the containers to see if it fixes the issue.

mewim:
Verified by https://github.com/kuzudb/kuzu/actions/runs/5868767569/job/15912311132

mewim (approved):

PR #1945, opened 2023-08-16
Skip reading null data if the property is known to have no nulls

benjaminwinger:


src/include/storage/store/table_statistics.h
```diff
@@ -14,17 +15,31 @@ namespace storage {
 using lock_t = std::unique_lock<std::mutex>;
 using atomic_uint64_vec_t = std::vector<std::atomic<uint64_t>>;
 
-class TableStatistics {
+using TablePropertyStats =
```
ray6080, line 18: Usually we give complicated types an alias to simplify its usage, such as nested maps, vectors, etc. But I think this map is pretty straightforward, and I would just use it as it is.

src/include/storage/store/table_statistics.h
```diff
@@ -34,8 +49,20 @@ class TableStatistics {
         numTuples = numTuples_;
     }
 
+    inline PropertyStatistics& getPropertyStatistics(common::property_id_t propertyID) {
```
ray6080, line 52: Add assertion here that the map contains propertyID.

src/include/storage/store/nodes_statistics_and_deleted_ids.h
```diff
@@ -179,10 +177,14 @@ class NodesStatisticsAndDeletedIDs : public TablesStatistics {
...
         catalog::TableSchema* tableSchema) override {
+        TablePropertyStats propertyStats;
+        for (auto property : tableSchema->getProperties()) {
+            propertyStats[property->getPropertyID()] = std::make_unique<PropertyStatistics>();
+        }
         // We use UINT64_MAX to represent an empty nodeTable which doesn't contain
         // any nodes.
         return std::make_unique<NodeStatisticsAndDeletedIDs>(
-            tableSchema->tableID, UINT64_MAX /* maxNodeOffset */);
+            tableSchema->tableID, UINT64_MAX /* maxNodeOffset */, std::move(propertyStats));
```
ray6080, line 187: Instead of moving `propertyStats` into `NodeStatisticsAndDeletedIDs`, you can consider passing the `TableSchema*`, and let the  `NodeStatisticsAndDeletedIDs` to handle the construction of property statistics inside its constructor.

src/storage/store/property_statistics.cpp
```diff
@@ -0,0 +1,45 @@
...
+    common::FileInfo* fileInfo, uint64_t& offset) {
+    bool hasNull;
+    common::SerDeser::deserializeValue<bool>(hasNull, fileInfo, offset);
+    return std::make_unique<PropertyStatistics>(hasNull);
+}
+
+// In a read-only context, readStatistics should always be non-null
+// Read/write statistics cannot be cached since functions like checkpointInMemoryIfNecessary may
+// overwrite them and invalidate the reference
+bool RWPropertyStats::mayHaveNull() {
```
ray6080, line 27: I don't think we should make assumptions on whether to read from the read only version or the write version in this function. Instead, we should take `Transaction*` as a param, and switch based on the transaction type.

Consider this: Merge `TablesStatistics::getWritePropertyStatisticsForTable` and `TablesStatistics::getReadPropertyStatisticsForTable` into a single function `TablesStatistics::getPropertyStatisticsForTable` and that function will take in a transaction pointer as a param. Internally, we will return property statistics based on the transaction type. Other interfaces in `TablesStatistics` should also be refactored to follow this fashion later.

src/include/storage/store/table_statistics.h
```diff
@@ -34,8 +49,20 @@ class TableStatistics {
...
+    inline const TablePropertyStats& getPropertyStatistics() { return propertyStatistics; }
+
+    inline void setPropertyStatistics(
+        common::property_id_t propertyID, PropertyStatistics newStats) {
+        propertyStatistics[propertyID] = std::make_unique<PropertyStatistics>(newStats);
+    }
+
 private:
     uint64_t numTuples;
+    TablePropertyStats propertyStatistics;
```
ray6080, line 65: Though `TablePropertyStats` simply contains a primitive boolean field, it is still good to wrap it into a `unique_ptr` as I can see this data structure will become more complicated soon, and with the `unique_ptr`, we can easily avoid copying it around without notice.

src/storage/store/node_column.cpp
```diff
@@ -143,7 +148,9 @@ void BoolNodeColumn::batchLookup(
 
 void NodeColumn::scan(
     Transaction* transaction, ValueVector* nodeIDVector, ValueVector* resultVector) {
-    nullColumn->scan(transaction, nodeIDVector, resultVector);
+    if (propertyStatistics.mayHaveNull()) {
+        nullColumn->scan(transaction, nodeIDVector, resultVector);
+    }
```
ray6080, line 153: Are we assuming the `resultVector` will always be default to no nulls?
I don't think we should assume that. Instead, we should implicitly set `resultVector`'s null mask to all non null if the property statistics indicates not have nulls.

src/storage/store/node_column.cpp
```diff
@@ -235,7 +242,9 @@ void NodeColumn::scanFiltered(Transaction* transaction, PageElementCursor& pageC
 
 void NodeColumn::lookup(
     Transaction* transaction, ValueVector* nodeIDVector, ValueVector* resultVector) {
-    nullColumn->lookup(transaction, nodeIDVector, resultVector);
+    if (propertyStatistics.mayHaveNull()) {
+        nullColumn->lookup(transaction, nodeIDVector, resultVector);
+    }
```
ray6080, line 247: Ditto here. implicitly set resultVector's null mask to all non null

src/storage/store/node_column.cpp
```diff
@@ -143,7 +148,9 @@ void BoolNodeColumn::batchLookup(
 
 void NodeColumn::scan(
     Transaction* transaction, ValueVector* nodeIDVector, ValueVector* resultVector) {
-    nullColumn->scan(transaction, nodeIDVector, resultVector);
+    if (propertyStatistics.mayHaveNull()) {
+        nullColumn->scan(transaction, nodeIDVector, resultVector);
+    }
     scanInternal(transaction, nodeIDVector, resultVector);
 }
```
ray6080, line 156: There are two scan functions serving for different purposes, one used for string and var list (L157), and the other one for scanning a whole column chunk at a time (L171). Can we also apply the optimization to both interfaces?

ray6080 (changes requested):
This PR looks good to me. But before we merge, can we get some performance number regarding to `COPY` and read-only queries? and we should document them within the PR description so it can be easily traced back. (Thus, I'm holding this PR for now, and also see my comments for necessary changes)

We can run the `COPY` and a set of queries on our local machine (document your configuration when reporting). The dataset is `ldbc SF10`, and if you prefer a larger one, `ldbc SF30` should be good enough. The comparison is between this branch (after you rebase latest master), and the latest master branch. 
We should report the `COPY` time of several relatively large node tables, including Comment, Post, and Person. For read-only queries, we should report end-to-end query evaluation time (use our benchmark tool is the best option) for LSQB queries (see `test_files/lsqb`) and some LDBC queries (see `test_files/ldbc`).

Let's discuss further once you get started on this. Thanks!

src/include/storage/store/table_statistics.h
benjaminwinger: Yes, it's just long and I had to add it to a bunch of different parameter lists where it's mostly just being passed around.

src/include/storage/store/table_statistics.h
benjaminwinger: This is the TablePropertyStats alias, not the PropertyStatistics struct.

src/storage/store/property_statistics.cpp
benjaminwinger: I think that's a good idea. The `Transaction` wasn't actually being passed to the `NodeColumn` when I started working on this, and I wasn't really clear to me how to determine when to use one or the other.

src/storage/store/node_column.cpp
benjaminwinger: Right, I missed that. I think I'd just looked at the one which NullNodeColumn overrides and forgot to add it to the others.

I think it would probably work better if the check is instead done inside the NullNodeColumn::scan functions. That way the nullColumn can be used as usual and it will apply everywhere.

src/storage/store/node_column.cpp
benjaminwinger: Done. 

For both lookup and the scan variant with start and end positions I had it only set the specified parts of the result vector to non-null. Since in this case we know that no part of it is null, it would be safe to set the whole thing, but it might be faster to just set the specific locations, though I do wonder how much, particularly for random-access lookups where setNull could write multiple times to the same word.

src/include/storage/store/table_statistics.h
ray6080: Yeah. But the thing is that right now when we're trying to debug, it's not easy to trace where the exception by `at` was thrown out. `assert` can help in this case.

benjaminwinger: Fair enough

src/storage/store/property_statistics.cpp
benjaminwinger: Using a dummy write transaction gets it compiling, but it doesn't seem to work properly and the tests are failing.

benjaminwinger:
i tried benchmarking on progressively larger versions of the LDBC dataset, but there weren't any consistent speedups, which I eventually realised after examining the `nodes.statistics_and_deleted.ids` file was because every single property was being marked as having nulls, which I'm fairly sure shouldn't be correct, so something is wrong, though I'm not sure what yet.

src/include/storage/store/table_statistics.h
```diff
@@ -15,16 +16,29 @@ using lock_t = std::unique_lock<std::mutex>;
...
+        std::unordered_map<common::property_id_t, std::unique_ptr<PropertyStatistics>>&&
+            propertyStatistics)
+        : numTuples{numTuples}, propertyStatistics{std::move(propertyStatistics)} {
         assert(numTuples != UINT64_MAX);
     }
 
+    explicit TableStatistics(const TableStatistics& other) : numTuples{other.numTuples} {
+        for (auto& propertyStats : other.propertyStatistics) {
+            propertyStatistics[propertyStats.first] =
+                std::make_unique<PropertyStatistics>(propertyStats.second.get());
```
benjaminwinger, line 38: Found the issue: no copy constructor, and this is not being de-referenced anyway, so the pointer is being coerced into a bool (i.e. always true since it's not null).

src/storage/store/property_statistics.cpp
benjaminwinger: The concurrency fix commit has fixed this now

benjaminwinger:
I played a bit more with benchmarks, but I can't see any clear performance benefits which are noticeable through the variation between runs with the LDBC datasets.

However I'm not sure we should really expect a significant performance improvement. The benchmarks are run repeatedly, so the actual cost of reading the data from disk is minimized by OS caching, while null data is packed into a single bit per value, so the total data to read is, well, 64 times smaller than for a 64-bit integer property.

The best I've managed was to create a dataset with an int64 primary key and a bool and a huge number of entries (I was hoping that it might be able to skip reading the primary key, as the query in question was just counting the number of nodes where the boolean is true, but since the performance difference is so small, I guess it must be reading the primary keys too). I ran it twice (50 iterations each, with standard deviation indicated by the black bars), since there is still some significant variation between runs.
[null-benchmark.pdf](https://github.com/kuzudb/kuzu/files/12480257/null-benchmark.pdf).
The performance benefit is small, but noticeable.

Increasing to 4 boolean columns to increase the ratio of null data to regular data makes the difference a little more pronounced:
[null-benchmark-4-bools.pdf](https://github.com/kuzudb/kuzu/files/12480318/null-benchmark-4-bools.pdf).

ray6080 (approved):

PR #1950, opened 2023-08-22
New CSV Reader

aziz-mu:
This (Draft) PR implements a non-arrow CSV reader, using the CSV reader that duckdb developed. The code taken from duckdb is in csv_reader.h and csv_reader.cpp classes, with some changes done to make it work with our data types.

Overall, the design won't change so Semih can take a look. However, there are still issues:

Things that aren't working right now: (Guodong take a look please, maybe there's some things that are easy to help me out with)
 - VAR_LIST and FIXED_LIST aren't working, I think I am reading them into the ValueVector incorrectly.
 - Something with ValueVectors and memory. Could you take a look at how I'm defining the ValueVector and DataChunks, and how I'm setting the values in csv_reader 's `AddValue` ? I fear that I am doing something very wrong here that I only found out once I went to run the benchmark. I am getting a segfault that is either coming from
   (i) the way I am initializing or using DataChunks/ValueVectors, or 
  (ii) My system running out of memory. 
Both of these aren't good - since the benchmark runs on a previous commit on master without any issue, even if the problem is (ii). this means that the csv reader is using too much memory compared to Arrow's csv reader. This likely isn't a problem with duckdb's code but rather inside the data copying I wrote in `Flush`, but I'm not sure how to fix it as of yet. Any tips appreciated.


src/storage/copier/column_chunk.cpp
```diff
@@ -54,17 +54,21 @@ void ColumnChunk::append(common::ValueVector* vector, common::offset_t startPosI
...
 void ColumnChunk::append(
     ValueVector* vector, offset_t startPosInChunk, uint32_t numValuesToAppend) {
-    assert(vector->dataType.getLogicalTypeID() == LogicalTypeID::ARROW_COLUMN);
-    auto chunkedArray = ArrowColumnVector::getArrowColumn(vector).get();
-    for (auto array : chunkedArray->chunks()) {
-        auto numValuesInArrayToAppend =
-            std::min((uint64_t)array->length(), (uint64_t)numValuesToAppend);
-        if (numValuesInArrayToAppend <= 0) {
-            break;
+    if (vector->dataType.getLogicalTypeID() == LogicalTypeID::ARROW_COLUMN) {
```
semihsalihoglu-uw, line 57: I know this is not your code but I'm very confused about this code. What I see is that this function (I mean `ColumnChunk::append(ValueVector* vector, offset_t startPosInChunk, uint32_t numValuesToAppend)`, calls itself recursively but does not do any actual copying. Instead for example the below or above `ColumnChunk::append` do some copying (below one uses memcpy somewhere and the above one calls `copyVectorToBuffer`).

Do you know what's going on or can you ask Guodong?

src/include/common/string_utils.h
```diff
@@ -42,6 +42,8 @@ class StringUtils {
         return c == ' ' || c == '\t' || c == '\n' || c == '\v' || c == '\f' || c == '\r';
     }
 
+    static bool CharacterIsNewline(char c) { return c == '\n' || c == '\r'; }
```
semihsalihoglu-uw, line 45: I'd make sure a similar function does not exist anywhere else in our code base. This looks like a quite basic function.

semihsalihoglu-uw (changes requested):

semihsalihoglu-uw:
@aziz-mu : For your PR comment: `(i) the way I am initializing or using DataChunks/ValueVectors, or
(ii) My system running out of memory.` => First I think it's likely this is because of you not initializing the ValueVectors inside your DataChunk as I mentioned. But I was just curious: Why do you think your system is running out of memory? Segfault does not necessarily mean you're running out of memory? Are you observing a high memory usage?

src/storage/copier/csv_reader.cpp
```diff
@@ -0,0 +1,515 @@
...
+            prev_pos = next_pos + 1;
+        }
+        new_val += old_val.substr(prev_pos, old_val.size() - prev_pos);
+        escape_positions.clear();
+        str_val = new_val;
+    }
+    auto type = parse_chunk.getValueVector(column)->dataType;
+    switch (type.getLogicalTypeID()) {
+    case LogicalTypeID::INT64: {
+        parse_chunk.getValueVector(column)->setValue(columnSizes[column]++,
```
ray6080, line 60: Ditto here, `columnSizes[column]++` is repeating for each column, while it should be the same for all columns. Unless I'm missing something here.

src/storage/copier/csv_reader.cpp
```diff
@@ -0,0 +1,515 @@
...
+    case LogicalTypeID::DOUBLE: {
+        parse_chunk.getValueVector(column)->setValue(columnSizes[column]++,
+            StringCastUtils::castToNum<double_t>(str_val.c_str(), str_val.length()));
+    } break;
+    case LogicalTypeID::BOOL: {
+        parse_chunk.getValueVector(column)->setValue(
+            columnSizes[column]++, StringCastUtils::castToBool(str_val.c_str(), str_val.length()));
+    } break;
+    case LogicalTypeID::STRING: {
+        parse_chunk.getValueVector(column)->setValue(columnSizes[column]++, str_val);
```
ray6080, line 84: This is incorrect. I thought we discussed this before. You should refer to `StringVector::addString`.

src/storage/copier/csv_reader.cpp
```diff
@@ -0,0 +1,515 @@
...
+            return false;
+        }
+    }
+    if (column < return_types.size()) {
+        throw CopyException(StringUtils::string_format(
+            "Error in file {} on line {}: expected {} values per row, but got {}", filePath, linenr,
+            return_types.size(), column));
+    }
+    if (mode == ParserMode::PARSING &&
+        columnSizes[0] >= common::StorageConstants::NODE_GROUP_SIZE) {
```
ray6080, line 149: Are you initializing value vectors to the size of `common::StorageConstants::NODE_GROUP_SIZE` anywhere? Otherwise, by default, if you initialize the value vector, it should be `DEFAULT_VECTOR_CAPACITY`.

src/storage/copier/csv_reader.cpp
```diff
@@ -0,0 +1,515 @@
...
+        insert_chunk.getValueVector(c)->state->selVector->selectedSize =
+            parse_chunk.getValueVector(c)->state->selVector->selectedSize;
+        insert_chunk.insert(c, parse_chunk.getValueVector(c));
+    }
+    InitParseChunk(parse_chunk.getNumValueVectors());
+    return true;
+}
+
+void BaseCSVReader::InitParseChunk(column_id_t numCols) {
+    parse_chunk = DataChunk(numCols);
```
ray6080, line 174: This is not correct. You have to initialize internal value vectors and insert them to parse_chunk.

Also, if this is the correct place for init parse chunk, then you shouldn't init it at the first place in the constructor.

src/storage/copier/csv_reader.cpp
```diff
@@ -0,0 +1,515 @@
...
+    std::vector<kuzu::catalog::Property*> properties, storage::MemoryManager* memoryManager) {
+    // PrepareComplexParser();
+    for (auto property : properties) {
+        return_types.push_back(*property->getDataType());
+    }
+    fd = open(filePath.c_str(), O_RDONLY);
+    ResetBuffer();
+    ReadHeader(csvReaderConfig->hasHeader);
+    // Initializing parse_chunk DataChunk
+    parse_chunk = DataChunk(properties.size());
```
ray6080, line 194: Again, why are we initializing the parse_chunk in multiple places? It looks like you are initializing parse_chunk correctly here. So make sure this is the right place, and clean up other places.

src/storage/copier/csv_reader.cpp
```diff
@@ -0,0 +1,515 @@
...
+    start = 0;
+    cachedBuffers.clear();
+}
+
+void BufferedCSVReader::ReadHeader(bool hasHeader) {
+    if (hasHeader) {
+        // ignore the first line as a header line
+        mode = ParserMode::PARSING_HEADER;
+        DataChunk dummy_chunk(0);
+        ParseCSV(dummy_chunk);
```
ray6080, line 217: I don't understand these two lines here (L216 and L217), what is the dummy chunk for ?

src/storage/copier/csv_reader.cpp
```diff
@@ -0,0 +1,515 @@
...
+
+void BufferedCSVReader::ResetBuffer() {
+    buffer.reset();
+    bufferSize = 0;
+    position = 0;
+    start = 0;
+    cachedBuffers.clear();
+}
+
+void BufferedCSVReader::ReadHeader(bool hasHeader) {
```
ray6080, line 212: This function is a bit confusing, as it is named as `ReadHeader`, and from the function definition, I can tell parsing header is the main stuff, but still this function takes in a parameter to allow skiping reading header. And regardless of has header or not, the function will set `mode` to `PARSING`.
It seems the caller should first switch on `hasHeader` to decide to call this function or not, which is more clear.

src/storage/copier/csv_reader.cpp
```diff
@@ -0,0 +1,515 @@
...
+}
+
+uint64_t BufferedCSVReader::TryParseCSV(
+    ParserMode parserMode, DataChunk& insertChunk, std::string& errorMessage) {
+    mode = parserMode;
+    return TryParseSimpleCSV(insertChunk, errorMessage);
+}
+
+} // namespace storage
+} // namespace kuzu
```
ray6080, line 515: New line here.

src/storage/copier/csv_reader.cpp
```diff
@@ -0,0 +1,515 @@
...
+    if (numRowsRead == -1) {
+        throw CopyException(errorMessage);
+    }
+    return numRowsRead;
+}
+
+uint64_t BufferedCSVReader::TryParseCSV(
+    ParserMode parserMode, DataChunk& insertChunk, std::string& errorMessage) {
+    mode = parserMode;
+    return TryParseSimpleCSV(insertChunk, errorMessage);
```
ray6080, line 511: The way `mode` is setting is confusing to me. Why pass `mode`, the class field, from `ParseCSV` into `TryParseCSV`, and then set it again inside `TryParseCSV`. Is `mode` supposed to change somewhere?

src/include/processor/operator/copy_from/read_csv.h
```diff
@@ -15,14 +15,21 @@ class ReadCSV : public ReadFile {
 
     inline std::shared_ptr<arrow::Table> readTuples(
         std::unique_ptr<storage::ReadFileMorsel> morsel) override {
-        auto csvMorsel = reinterpret_cast<storage::ReadSerialMorsel*>(morsel.get());
-        return csvMorsel->recordTable;
+        return nullptr;
```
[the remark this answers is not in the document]
ray6080: Remove this function. You don't use it anymore. And it's just confusing if you leave it here and `return nullptr`.

src/include/processor/operator/copy_from/read_csv.h
```diff
@@ -15,14 +15,21 @@ class ReadCSV : public ReadFile {
...
+    }
+
+    inline std::shared_ptr<common::DataChunk> readDataChunk(
```
ray6080, line 21: Should this be private? Move it down and make it private.

src/include/processor/operator/copy_from/read_csv.h
```diff
@@ -15,14 +15,21 @@ class ReadCSV : public ReadFile {
...
+        auto csvMorsel = reinterpret_cast<storage::ReadCSVMorsel*>(morsel.get());
+        return csvMorsel->recordOutput;
     }
 
     inline std::unique_ptr<PhysicalOperator> clone() override {
         return std::make_unique<ReadCSV>(
             nodeOffsetPos, dataColumnPoses, sharedState, id, paramsString);
     }
+
+    bool getNextTuplesInternal(ExecutionContext* context) override;
```
ray6080, line 32: Move this function up, before `clone()`.

src/storage/copier/read_file_state.cpp
```diff
@@ -87,6 +87,56 @@ std::unique_ptr<ReadFileMorsel> ReadCSVSharedState::getMorselSerial() {
...
+        if (currFileIdx >= filePaths.size()) {
+            // No more files to read.
+            return nullptr;
+        }
+        auto filePath = filePaths[currFileIdx];
+        if (!reader) {
+            reader = make_shared<BufferedCSVReader>(
+                filePath, &csvReaderConfig, tableSchema, memoryManager);
+        }
+        auto output = std::make_shared<DataChunk>(tableSchema->getNumProperties());
```
ray6080, line 104: Don't create data chunk for each call here. It's unnecessarily adding extra costs. Instead, the `resultSet` already have a data chunk initialized and prepared for you to use, pass it as function param.

src/processor/operator/copy_from/read_csv.cpp
```diff
@@ -0,0 +1,22 @@
+#include "processor/operator/copy_from/read_csv.h"
+
+using namespace kuzu::storage;
+
+namespace kuzu {
+namespace processor {
+bool ReadCSV::getNextTuplesInternal(ExecutionContext* context) {
+    auto morsel = sharedState->getMorselSerial();
```
ray6080, line 8: You can pass in data chunk in resultSet through `getMorselSerial ` here, so you don't need to initialize it for each call inside the function. And should also getting rid of `readDataChunk` function.

src/processor/operator/copy_from/read_csv.cpp
```diff
@@ -0,0 +1,22 @@
...
+    if (morsel == nullptr) {
+        return false;
+    }
+    auto nodeOffsetVector = resultSet->getValueVector(nodeOffsetPos).get();
+    nodeOffsetVector->setValue(
+        nodeOffsetVector->state->selVector->selectedPositions[0], morsel->rowsRead);
+    auto output = readDataChunk(std::move(morsel));
+    resultSet->insert(dataColumnPoses[0].dataChunkPos, output);
+    resultSet->dataChunks[0]->state->setToUnflat();
+    return true;
```
ray6080, line 18: Ditto, if you make the above change, no need to re-insert data chunk at L16. Generally we should't replace data chunk dynamically in resultSet during the operator's runtime, unless very rare cases as i'm not aware of for now.

src/storage/copier/csv_reader.cpp
```diff
@@ -0,0 +1,515 @@
...
+        auto copyDescription =
+            CopyDescription({filePath}, common::CopyDescription::FileType::CSV, *csvReaderConfig);
+        auto value = TableCopyUtils::getArrowFixedListVal(
+            str_val, 1, str_val.length() - 2, type, copyDescription);
+        ValueVector::copyValueToVector(parse_chunk.getValueVector(column)->getData(),
+            parse_chunk.getValueVector(column).get() + columnSizes[column], value.get());
+        columnSizes[column]++;
+    } break;
+    default:
+        throw CopyException("Unsupported data type " +
```
ray6080, line 117: Change `CopyException` to `NotImplementedException`.

src/storage/copier/csv_reader.cpp
```diff
@@ -0,0 +1,515 @@
...
+                            LogicalTypeUtils::dataTypeToString(type.getLogicalTypeID()) +
+                            " inside LIST");
```
ray6080, line 119: What do you mean " inside LIST"?

src/storage/copier/read_file_state.cpp
```diff
@@ -87,6 +87,56 @@ std::unique_ptr<ReadFileMorsel> ReadCSVSharedState::getMorselSerial() {
...
+        }
+        break;
+    }
+    auto result = std::make_unique<ReadCSVMorsel>(
+        currRowIdx, filePaths[0], currRowIdxInCurrFile, rowsRead, std::move(resultOutput));
+    currRowIdx += rows;
+    return result;
+}
+
 std::unique_ptr<ReadFileMorsel> ReadCSVSharedState::getMorsel() {
```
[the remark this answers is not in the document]
ray6080: I don't think this function is needed. @aziz-mu Can you take a closer look to clean this function up?

ray6080 (commented):
I have two high level comments:
1) Can you describe currently in this PR what is supported and what is not supported? Not just data types, but also on csv file format itself, like is there any quotations, delimiters, escapes, or combinations of them we cannot parse right now? I see you are following `TryParseSimpleCSV` code path in DuckDB's buffered csv reader, but I didn't see code path for `TryParseComplexCSV`, maybe you changed it or not yet added. so i didn't go into duckdb's code in details, but instead rely on your description to understand this.
2. This one is for discussion, as we're taking some bulk of code from DuckDB, should we keep different code styles? or merge theirs into ours? (I would prefer latter one actually). Regardless of which way, I do think it's better to have clean comments in the beginning of file to state which version (commit) the code is taking from, and what has changed or added, that would be much more maintainable later.

src/include/processor/operator/copy_from/read_csv.h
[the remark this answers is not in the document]
ray6080: Yeah, you can make the changes.

Issue #744, opened 2022-08-10
Get rid of flat vs unflat distinction in the query processor


semihsalihoglu-uw:
Currently we have many pieces of code that has if (flat) { ... } else { // unflat ... } branches. This does not seem necessary if the flatten operator, which is the only operator that flattens correctly sets the selectedSize of the DataChunk it is flattening to 1 and set the selectedPos correctly. This would get rid of all these branches from many places. However, I don't expect all such branches to go away because some code may still need to identify the flat data chunk. For example in binary comparison expressions one of the two sides are assumed to be flat, so that code still needs to check which side is flat. So we might need to keep the isFlat() function, which would return true if the selectedSize = 1.

The flatten operator would need to change and make sure that if it receives an already selected data chunk to flatten, it copies those original selected vectors. At a high level, its pseudocode should look like this:
```
Flatten : PhysicalOp {
   void getNextTuples() {
      bool[] prevSelectedPos = dataChunkToFlatten->copySelectedState();
      uint64_t prevSelectedSize = dataChunkToFlatten->selectedSize; 
      dataChunkToFlatten->selectedSize = 1;
      for (int i = 0; i < prevSelectedSize; ++i) { // this would not be a for loop but be incremented with each call to getNextTuples()
         dataChunkToFlatten->selectedPos[0] = prevSelectedPos[i];
      }
   } 
}
``` 

And then many but not all if (isFlat()) code paths would be removed.

I'm labeling this as phase 1 but this can be addressed later.

PR #1966, opened 2023-08-27
Remove currIdx from DataChunkState

andyfengHKU:
We take another step in solving #744. This time we remove currIdx and replace with a FactorizationState.


src/include/common/data_chunk/data_chunk_state.h
```diff
@@ -27,21 +28,26 @@ class DataChunkState {
...
+
+    inline uint64_t getNumSelectedValues() const { return selVector->selectedSize; }
 
 public:
-    // The currIdx is >= 0 when vectors are flattened and -1 if the vectors are unflat.
-    int64_t currIdx;
+    std::shared_ptr<SelectionVector> selVector;
+
+private:
+    FactorizationStateType factorizationState = FactorizationStateType::UNFLAT;
```
ray6080, line 43: Initialize this in constructor.

src/include/common/data_chunk/data_chunk_state.h
```diff
@@ -27,21 +28,26 @@ class DataChunkState {
...
+    inline bool isFlat() const { return factorizationState == FactorizationStateType::FLAT; }
+    inline void setToFlat() { factorizationState = FactorizationStateType::FLAT; }
+    inline void setToUnflat() { factorizationState = FactorizationStateType::UNFLAT; }
+
+    inline uint64_t getNumSelectedValues() const { return selVector->selectedSize; }
 
 public:
-    // The currIdx is >= 0 when vectors are flattened and -1 if the vectors are unflat.
-    int64_t currIdx;
+    std::shared_ptr<SelectionVector> selVector;
```
ray6080, line 40: Also move this field `selVector` to private?

test/transaction/transaction_test.cpp
ray6080, line 1: Why remove this test?

src/include/common/data_chunk/data_chunk_state.h
ray6080, line 43: Not related to this PR. but a suggestion is we can simplify `FactorizationStateType` to `FStateType`, which reads easier.

src/processor/operator/hash_join/hash_join_probe.cpp
```diff
@@ -110,6 +110,13 @@ uint64_t HashJoinProbe::getLeftJoinResult() {
         for (auto& vector : vectorsToReadInto) {
             vector->setAsSingleNullEntry();
         }
+        // TODO(Xiyang): We have a bug in LEFT JOIN which should not discard NULL keys. To be more
+        // clear, NULL keys should only be discarded for probe but should not reflect on the vector.
+        // The following for loop is a temporary hack.
```
ray6080, line 115: Maybe open an issue one this if we are not going to fix this immediately?

ray6080 (approved):

src/include/common/data_chunk/data_chunk_state.h
andyfengHKU: This involves a lot of change so should be done separately.

test/transaction/transaction_test.cpp
andyfengHKU: Because it is using currIdx directly. I found it hard to maintain.

PR #1978, opened 2023-08-29
Rdf processing cleanup and removed schema mapping

gaurav8297:


src/include/planner/logical_plan/scan/logical_scan_node.h
```diff
@@ -3,6 +3,8 @@
 #include "binder/expression/node_expression.h"
 #include "planner/logical_plan/logical_operator.h"
 
+using namespace kuzu::binder;
```
andyfengHKU, line 6: don't use namespace in header, this will propagate to other headers

src/planner/plan/append_extend.cpp
```diff
@@ -115,41 +115,28 @@ void QueryPlanner::appendNonRecursiveExtend(std::shared_ptr<NodeExpression> boun
...
-    auto nodeExpression = rdfExtend->getNbrNode();
-
-    // We need to add the following properties to the hash join
-    auto rdfNodeIRIProperty =
-        nodeExpression->getPropertyExpression(common::InternalKeyword::RDF_IRI_PROPERTY_NAME);
-    auto rdfRelIRIProperty =
-        relExpression->getPropertyExpression(common::InternalKeyword::RDF_IRI_PROPERTY_NAME);
-    auto rdfNodeInternalIDProperty = nodeExpression->getInternalIDProperty();
-    auto rdfPredicateIRIOffsetProperty = relExpression->getPropertyExpression(
+    auto nodeExpression = binder.createQueryNode(relExpression->getUniqueName(),
```
andyfengHKU, line 118: So this is where major change happens. I understand what you are doing here. This logic should be done during binding (let me do it, no need to change this PR).

The general design is each module doesn't not expose to each other. Instead the communicate through some protocol. e.g.
```
Parser -> Statement -> Binder
Binder -> BoundStatement -> Planner
Planner -> LogicalPlan -> Processor
Processor -> PhysicalPlan -> Task Schedular 
```

We should always enforce these constraints as more features being added.

I'll modify this and u can review my PR later on.

andyfengHKU (approved):

PR #1980, opened 2023-08-30
Overflow file copy for string PK copy node

KasunaStony:
Implement direct copy of the overflow file from string column chunk to hash index builder


ray6080:
This is closed for now as this is stale now and we plan to rework hash index completely later. Changes here will be considered in later PR.

PR #1981, opened 2023-08-31
Separate String and INT64 from hash_index_builder and remove function pointers, adding partial hash

KasunaStony:


ray6080:
This is closed for now as this is stale now and we plan to rework hash index completely later. Changes here will be considered in later PR.

PR #1995, opened 2023-09-06
share build cache when building rust API and the example project

russell-liu:


benjaminwinger (approved):
From the logs, it looks like it's still using separate build directories. I think that's because the example doesn't use the arrow feature, so it then separately builds kuzu without the arrow feature to use with the example. 

What we could do is have the example build require the arrow feature (or give it a feature that depends on the kuzu crate's arrow feature). So why don't we merge this for now, and I'll add an arrow feature to the example in a separate PR.

PR #2002, opened 2023-09-07
Add arrow feature to rust example so that it can re-use the test build in CI

benjaminwinger:
Following up #1995, this should let the rust example use the previously built version of the kuzu crate in CI, as they'll now use the same features.

I'm tempted to also actually include an example of using the arrow data in the example, but it makes the example a lot more complex and I'm not sure if it really helps provide a good example of kuzu's interfaces as most of it would be dealing with arrow types.

### DIFF
--- a/.github/workflows/ci-workflow.yml
+++ b/.github/workflows/ci-workflow.yml
@@ -88,7 +88,7 @@ jobs:
 
       - name: Rust example
         working-directory: examples/rust
-        run: cargo build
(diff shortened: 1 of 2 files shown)


benjaminwinger:
It does appear to be working correctly. The paths in the logs match, and the rust example compilation time went down from ~4 minutes to 52 seconds.

andyfengHKU (approved):

PR #2004, opened 2023-09-07
Abstract Compression interface and initial integer packing

benjaminwinger:
So far this just sets up an interface for compressing data on disk. *Edit: Final version adds integer packing for int32, int64, uint32 and uint64*

Compressed data must have a fixed number of values per page, but the number of values can vary per ColumnChunk (now stored in the ColumnChunkMetadata, where before it was a fixed value for a given type). 
Additional compression-specific metadata is stored in a small fixed-sized buffer (currently just one byte) inside the ColumnChunkMetadata's new CompressionMetadata substructure.
Compressed data needs to be aligned to the start of each page, since data is read relative to the beginning of a page, so the functions called to compress data work one page at a time.

I've implemented an IntegerBitpacking class for the interface with an implementation using FastPFOR's fastpack/fastunpack functions, combined with DuckDB's SignExtend function used for negative values. Currently this only works on 32 and 64-bit integers as the interface for 16-bit fastpack/fastunpack is separate, but that should still be relatively simple to set up.

The BooleanCompression class is used just for decompression via the same interface, since boolean values are stored packed in memory in BoolColumnChunk, unlike CompressedColumnChunk which stores values uncompressed in memory and compresses them when writing the data (to avoid having to re-compress when collecting data).

~~The Integer Bitpacking isn't currently fully working, so I have integers just using the "CopyCompression" instead. Partially this is because re-compression and relocation isn't implemented when we update with data that can't fit inside the current bit width, but there are also some tests which aren't triggering the exceptions checking that case.~~

TODO (remaining to be moved to issue when this is merged):
- [ ] INT16 and INT8 bitpacking
- [x] Re-compress and relocate values which don't fit in the existing chunk during modifications (either due to difference in bit width, or because the chunk was packed without support for signed values).
- [ ] InternalID could be bitpacked, since it's stored as offset_t (uint32_t), but would need an extra layer to handle widening the results into the internal_id_t struct.
- [ ] Extra values could be packed into the end of each page when the page does not divide evenly into 32-value chunks.
- [ ] Small reads/writes to bitpacked data could be optimized to work on individual data instead of relying on packing/unpacking a 32-value chunk to a temporary array.
- [ ] BooleanBitpacking could make use of integer bitpacking functions with a bitwidth of 1 which might be faster.
- [ ] Set up some small benchmarks (maybe using [google benchmark](https://github.com/google/benchmark), essentially what I did when comparing implementations at the start) so that we can do some fast comparisons of the compression/decompression speed without having to deal with the overhead of the rest of the system (not that such integrated benchmarks aren't also useful).
- [x] Remaining bitpacking bugs? There are some tests still failing for unexplained reasons when bitpacking is enabled for INT64/INT32.


benjaminwinger:
I just noticed that fastpfor's bitpackingaligned is 17000 lines of code. We're not even using it right now, so I've removed the 24 and 32-bit versions. Alternatively, [duckdb implemented a 16-bit version](https://github.com/duckdb/duckdb/commit/68abedc20f290b53a2d1073b82e5cf73aef1e9f7) based on the fastunpack/fastpack in bitpacking.hpp. It might be worth benchmarking and comparing them, but I think duckdb's version would at least be a lot smaller since bitpackingaligned.cpp with just 8 and 16-bit versions is still 5000 lines of code.

benjaminwinger:
I think the remaining issue is with newly created pages, as they don't have the header with the type of compression.
Admittedly that's not really necessary at this point, and could also be moved into the ColumnChunkMetadata, but the same issue will occur with the integer bitpacking, where the bit width and whether or not there are negatives need to be stored somewhere accessible when writing to new pages (but that doesn't really extend to other ColumnChunks and compression types).

I think for now the most straightforward solution would be to have a small buffer (even just one byte) of reserved data in the ColumnChunkMetadata which can be used by the compression algorithm to store and access that sort of thing, and to drop the idea of storing extra metadata along with the data.

ray6080 (commented):
In the middle of review. As discussed, here are todos for this PR:
1. Move from PhysicalMapping to functor-based ideas.
2. Fix compilation on Mx chip mac.
3. Fix remaining bugs.
4. Get benchmark results on COPY performance and compressed file sizes.
5. Increment the project version, so we can trigger a re-run of benchmark loading.
6. Can you open an issue on that we will add a config option `ENABLE_COMPRESSION` in the database level?

Let me know when you finish those todos, and we can do another round. Thanks!

benjaminwinger:
## Copy benchmarks:
Unfortunately I think reading and parsing the CSV files is probably the bulk of the time spent copying at the moment, given the large difference in speed between this and the scan benchmarks and the difference in speed between 0 and 64-bit values, but there is enough of a difference between the results to draw some conclusions.

The dataset I used was generated with 1,000,000 nodes with an id and a random signed integer. The results are an average of 50 runs, benchmarked using hyperfine and kuzu_shell.

### Master Branch
64-bit values: 16MB data, 588.2 ms ±   4.7 ms
0-bit values: 16MB data, 417.8 ms ±   2.6 ms

### Uncompressed
64-bit values: 15MB data, 592.6 ms ±   6.0 ms
0-bit values: 15MB data, 422.9 ms ±   4.0 ms

### Bitpacking
Note that the ID column always has a bit width of 20 for 1000000 nodes), so really we're dealing with 84/52/20 bits of data, compared to 128 bits in the above cases:

64-bit values: 10MB data, 596.1 ms ±   5.3 ms
32-bit values: 6.5MB data, 501.7 ms ±   3.7 ms
0-bit values: 2.7MB data, 434.1 ms ±   2.6 ms

## Scan Benchmarks
It was harder to see the differences here on a small dataset, so I ended up benchmarking a dataset with 100,000,000 nodes. 
[bitpacking-benchmarks.pdf](https://github.com/kuzudb/kuzu/files/12590447/bitpacking-benchmarks.pdf)

There is an odd outlier in the scan benchmarks, which is that scanning positive 63-bit numbers is significantly slower than scanning a mix of positive and negative 63-bit numbers, where I would expect it to be the other way around, given how unpacking signed numbers is implemented. It's possible that is related to an outstanding bug in the implementation.

src/storage/copier/compression.cpp
```diff
@@ -0,0 +1,218 @@
...
+        throw NotImplementedException(
+            "Setting negative values to a chunk stored without negatives is not implemented yet");
+    }
+    if (header.hasNegative && valueSize > header.bitWidth - 1 ||
+        !header.hasNegative && valueSize > header.bitWidth) {
+        throw NotImplementedException(
+            "Setting values larger than the bit width is not implemented yet");
+    }
+
+    U chunk[32];
```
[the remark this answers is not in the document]
benjaminwinger: I made it a class constant named `IntegerBitpacking::CHUNK_SIZE`

src/storage/store/node_column.cpp
```diff
@@ -584,5 +520,7 @@ std::unique_ptr<NodeColumn> NodeColumnFactory::createNodeColumn(const LogicalTyp
     }
 }
 
+NodeColumn::~NodeColumn() = default;
```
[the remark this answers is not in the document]
benjaminwinger: It had been needed because of the unique_ptr containing a forward declaration of the PhysicalMapping, but is no longer necessary since I removed the PhysicalMapping class.

src/include/storage/copier/column_chunk.h
```diff
@@ -173,6 +177,19 @@ inline bool ColumnChunk::getValue(common::offset_t pos) const {
     return common::NullMask::isNull((uint64_t*)buffer.get(), pos);
 }
 
+// Column chunk which is compressed during flushBuffer, but otherwise maintained uncompressed
+class CompressedColumnChunk : public ColumnChunk {
```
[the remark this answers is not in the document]
benjaminwinger: Done

src/include/storage/copier/compression.h
```diff
@@ -0,0 +1,219 @@
...
+// decompress uncompressed data by mistake, and to allow for runtime-configurable compression in
+// future.
+enum class CompressionType : uint8_t {
+    COPY = 0,
+    INTEGER_BITPACKING = 1,
+    BOOLEAN = 2,
+};
+
+struct CompressionMetadata {
+    uint64_t numValuesPerPage;
```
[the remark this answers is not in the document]
benjaminwinger: > I guess it could instead calculate the number of values per page based on the type and the type of compression used via a CompressionMetadata function.

Finished this part

src/include/storage/copier/compression.h
```diff
@@ -0,0 +1,313 @@
...
+        case common::LogicalTypeID::INT64: {
+            integerBitpacking = std::make_unique<IntegerBitpacking<int64_t>>();
+            break;
+        }
+        case common::LogicalTypeID::DATE:
+        case common::LogicalTypeID::INT32: {
+            integerBitpacking = std::make_unique<IntegerBitpacking<int32_t>>();
+            break;
+        }
+        default: {
```
ray6080, line 269: Let's guard the default case by throwing out `NotImplementedException` in advance.

src/include/storage/copier/compression.h
```diff
@@ -0,0 +1,313 @@
...
+        uint64_t dstOffset, uint64_t numValues, const CompressionMetadata& metadata) const final;
+};
+
+class CompressedFunctor {
+public:
+    CompressedFunctor(const CompressedFunctor&) = default;
+
+protected:
+    explicit CompressedFunctor(const common::LogicalType& logicalType) : copy{logicalType} {
+        switch (logicalType.getLogicalTypeID()) {
```
ray6080, line 258: Move the implementation to cpp file.

src/storage/copier/column_chunk.cpp
```diff
@@ -14,12 +15,74 @@ using namespace kuzu::transaction;
...
+    uint64_t numValues, BMFileHandle* dataFH, page_idx_t startPageIdx) {
+    // Since we compress into memory, storage is the same as fixed-sized values,
+    // but we need to mark it as being boolean compressed.
+    FileUtils::writeToFile(dataFH->getFileInfo(), buffer, bufferSize,
+        startPageIdx * BufferPoolConstants::PAGE_4KB_SIZE);
+    return ColumnChunkMetadata(startPageIdx, ColumnChunk::getNumPagesForBytes(bufferSize),
+        numValues, CompressionMetadata(CompressionType::BOOLEAN_BITPACKING));
+}
+
+class CompressedFlushBuffer {
```
ray6080, line 36: hmm where is this class used?

src/storage/copier/column_chunk.cpp
```diff
@@ -14,12 +15,74 @@ using namespace kuzu::transaction;
...
         nullChunk = std::make_unique<NullColumnChunk>();
     }
+    switch (this->dataType.getPhysicalType()) {
+    case PhysicalTypeID::BOOL: {
+        flushBufferFunction = booleanFlushBuffer;
+        break;
+    }
+    default: {
+        flushBufferFunction = fixedSizedFlushBuffer;
+    }
```
ray6080, line 84: Don't we need to switch on compressed data types?

src/storage/store/node_column.cpp
```diff
@@ -15,103 +19,169 @@ using namespace kuzu::transaction;
...
 namespace storage {
 
-void FixedSizedNodeColumnFunc::readValuesFromPage(uint8_t* frame, PageElementCursor& pageCursor,
-    ValueVector* resultVector, uint32_t posInVector, uint32_t numValuesToRead) {
-    auto numBytesPerValue = resultVector->getNumBytesPerValue();
-    memcpy(resultVector->getData() + posInVector * numBytesPerValue,
-        frame + pageCursor.elemPosInPage * numBytesPerValue, numValuesToRead * numBytesPerValue);
-}
+struct FixedSizedNodeColumnFunc {
+    static void readValuesFromPage(uint8_t* frame, PageElementCursor& pageCursor,
```
ray6080, line 23: Can we get rid of `FixedSizedNodeColumnFunc `? Logically, it should be same as `ReadCompressedValuesFromPage` under the UNCOMPRESSED case.

src/storage/copier/compression.cpp
```diff
@@ -0,0 +1,320 @@
...
+uint32_t getDataTypeSizeInChunk(const common::LogicalType& dataType) {
+    using namespace common;
+    switch (dataType.getLogicalTypeID()) {
+    case LogicalTypeID::STRUCT: {
+        return 0;
+    }
+    case LogicalTypeID::STRING: {
+        return sizeof(ku_string_t);
+    }
+    case LogicalTypeID::VAR_LIST: {
```
ray6080, line 41: Merge case for VAR_LIST and INTERNAL_ID.

src/storage/copier/compression.cpp
```diff
@@ -0,0 +1,320 @@
...
+
+using namespace kuzu::common;
+namespace arrow {
+class Array;
+}
+
+namespace kuzu {
+namespace storage {
+
+const LogicalType BooleanBitpacking::LOGICAL_TYPE = LogicalType(LogicalTypeID::BOOL);
```
ray6080, line 26: Are these ever used somewhere?

ray6080 (approved):

ray6080:
Can you document TODOs following this PR somewhere? In my mind, I think we should work on int16 and perhaps int8. And also fix updating bit width to larger size and updating unsigned values to signed.

src/include/storage/copier/compression.h
benjaminwinger: That's not the purpose of the switch though. This is just creating the 64/32-bit IntegerBitpacking class ahead of time in case it is used in `operator()` later, at which point we switch based on the compression type instead.
If anything, we need a way of asserting that any column using integer bitpacking has a type that is covered in this switch.

The IntegerBitpacking class no longer contains any data though, so I guess we could store the PhysicalTypeID instead and switch in `operator()` as the constructor should be more or less a noop. That would also remove the need for virtual dispatch.

src/storage/store/node_column.cpp
benjaminwinger: Yes. Not for InternalIDs yet, but it should be possible to get that working eventually.

benjaminwinger:
> Can you document TODOs following this PR somewhere? 

I've added them to the PR description.

src/storage/copier/column_chunk.cpp
benjaminwinger: I keep testing both without it and with it, since the tests don't all pass with compression enabled I don't want to miss regressions when it's disabled.

src/include/common/constants.h
```diff
@@ -7,7 +7,7 @@
 namespace kuzu {
 namespace common {
 
-constexpr char KUZU_VERSION[] = "v0.0.8.7";
+constexpr char KUZU_VERSION[] = "v0.0.8.8";
```
ray6080, line 10: Can you update this version to latest?

src/include/storage/store/node_column.h
```diff
@@ -8,38 +8,20 @@
 #include "storage/store/column_chunk.h"
 
 namespace kuzu {
+
```
ray6080, line 11: Remove this newline?

PR #2016, opened 2023-09-11
Add test cases match3/4 and new feature "--"

AEsir777:
- added tests match3/match4
- added new feature `MATCH (n) -- (b) RETURN ...`
    - modify '.g4' files
    - add check for 'relDetails' to be nullptr


test/test_files/tck/match/match3.test
```diff
@@ -0,0 +1,415 @@
...
+-STATEMENT CREATE REL TABLE T(FROM A TO B);
+---- ok
+-STATEMENT CREATE (:A)-[:T]->(:B);
+---- ok
+-STATEMENT MATCH (a1)-[r:T]->() 
+           WITH r, a1 
+           MATCH (a1)-[r:T]->(b2) 
+           RETURN a1, r, b2;
+---- error
+Binder exception: Bind relationship r to relationship with same name is not supported.
```
andyfengHKU, line 339: This is probably where we are different from openCypher standard. Can u additionally comment in the test file as well. Something like `# Different result` so that we can quickly locate what are the statements to be fixed. Apply this change to match3&4

andyfengHKU (changes requested):
Remove `/antlr4.` and rebase&squash into 1 commits. Then we are ready to merge!

andyfengHKU (approved):

PR #2038, opened 2023-09-15
Don't self-initialize schema comments

benjaminwinger:
I was getting tons of segfaults on my computer when running tests. It turned out that the new comment field in the TableSchema is being initialized to itself in one of the constructors.

Is the comment supposed to be passed as an argument to this constructor? I just default-initialized it instead, which fixed the crashes.

### DIFF
--- a/src/include/catalog/table_schema.h
+++ b/src/include/catalog/table_schema.h
@@ -17,8 +17,7 @@ public:
         std::vector<std::unique_ptr<Property>> properties)
         : tableName{std::move(tableName)}, tableID{tableID}, tableType{tableType},
           properties{std::move(properties)},
-          nextPropertyID{(common::property_id_t)this->properties.size()}, comment{
-                                                                              std::move(comment)} {}
+          nextPropertyID{(common::property_id_t)this->properties.size()}, comment{} {}
     TableSchema(common::TableType tableType, std::string tableName, common::table_id_t tableID,
         std::vector<std::unique_ptr<Property>> properties, std::string comment,
         common::property_id_t nextPropertyID)


andyfengHKU (approved):

Riolku:
Yikes. Sorry about that, any reason why our tests don't catch this?

> Is the comment supposed to be passed as an argument to this constructor?

Not this one. By default the comment is supposed to be empty, but the second constructor is used as a "more complete" one, for copying, AFAICT.

Riolku (approved):

benjaminwinger:
It was only happening in release mode, I had to compile with the RelWithDebInfo configuration to debug it (which turned a dozen or so crashes into every single database test crashing immediately). So it might vary with the compiler: I think CI is using gcc 11, while I'm building with gcc 13.

[A quick test in godbolt](https://godbolt.org/#z:OYLghAFBqd5QCxAYwPYBMCmBRdBLAF1QCcAaPECAMzwBtMA7AQwFtMQByARg9KtQYEAysib0QXACx8BBAKoBnTAAUAHpwAMvAFYTStJg1DIApACYAQuYukl9ZATwDKjdAGFUtAK4sGe1wAyeAyYAHI%2BAEaYxBIapAAOqAqETgwe3r56icmOAkEh4SxRMVxxdpgOqUIETMQE6T5%2BXLaY9rkM1bUE%2BWGR0bG2NXUNmc0KQ93BvUX9pQCUtqhexMjsHOYAzMHI3lgA1CYbbuPEwcCH2CYaAIKb27uYB0dsLCQAnhdXt2ZbDDte%2B0ObicJ0wrE%2BNy%2BOyYCgUewAamIvI8TAB2Kw3eJeCK0PDIEBfPZEhFIzAQOZ7EB7NAsNiCNEWcboEAgV4ANzJNLpBDmaIAIgz%2BV8mSyTmdqahaYwCIcMbdUQKNnKvl9ggQ9iwmMFyQd0YTiUwvEQ9mzSU8%2BSTvGTeUr9USRSgluqgUCTaSAHRc6VPNyuh2uWiyu17YiYAjLBh7DRByEKjgLWicACsvD8HC0pFQnF9lmsewUSxWKJ%2BPFIBE08YWAGsQGYk/pOJJUxXM5xeAoQHFy%2Bn46Q4LAkJhVBUjSRyJRasAFMpDK0hAhUAB3NOlmnxOhMdozkK0edLtMZtd0frALgbMykI/0YihVhrXhX6IAeSNe%2BXLaHFWuxCnbdIn%2BQap8DTXh%2BEEEQxHYKQZEERQVHUHtSF0ZoDCMFBrGsfQ8AiDtIAWVB4naDsOHbQtVj0Jlgm3OcF3fbheEXYgmHiTgeATZNm0QrMOGwYdkFHYg9lUAAOAA2ABaUTJD2YBkGQPYz3dMw9ggHMrEsUg9lwQgSAOEs5l4bstDmata3rRMOCbUgD14bj207MsKwWftEBAACBPHCBJ2nWdd1olcH0ldcDC3Xy3xsy8guPGJT2gx8bzvP94pfYR/I/Pjv1/Ej/z4oDgj/MDhFEcRoMKuC1BbXQL1Q4wMI03EcPgfDCNSYjbI5YhTiwXDyUWZZyLGAhgOovz91YhimJY%2Bj2I4FNrJbbjeJHIhBJEiSpJkuSFK4d1pNUuqbC0/AVr0jYuAMxyexM0gazMSR3QATie56Xpe0SG0sziMzs2wHKM3sXIgQc%2BI8igvJ/Hyd3C8bIpYYLN1SEbocQ%2BKUBqrhRI0GElDqWHguiW82CSqLrxS5GMwAzK2pyio8pAmCisgiRpDKpQKsQ3QNn0QxatzersJ65qiL/VAOq6zBBb6osKKGqiwrS%2BjSEY5jxpmuaIsWkGTrWyTpOhIwFNE90NGNlTseiIiKX2vnDu0k7NnOwynNMusPqsjW/w7LtnY%2Bswvtsv9/uujrkmcSQgA%3D%3D) shows that it produces a segfault on gcc 12+, and succeeds on gcc <11. It also works fine in clang, which is the only one to produce a warning.

Riolku:
Yet another reason to use clang :wink:

benjaminwinger:
Except there are so many warnings in the codebase that you wouldn't notice with clang normally. If we could clean up all the warnings and turn them into errors, then it could have been caught by the clang CI.

Riolku:
ANTLR4 is particularly loud. I might look into reducing our warnings, ive always liked Werror

PR #2043, opened 2023-09-15
Refactor copy compilation

andyfengHKU:
This PR breaks 

```
LogicalCopyNode
```
as
```
LogicalCopyNode
LogicalScanFile
```

and 
```
LogicalCopyRel
```
as
```
LogicalCopyRel
LogicalIndexScan
LogicalScanFile
```

We are going to reuse logical scan file for our direct scan feature


src/binder/bind/bind_file_scan.cpp
```diff
@@ -0,0 +1,138 @@
...
+#include "common/string_utils.h"
+
+using namespace kuzu::parser;
+using namespace kuzu::binder;
+using namespace kuzu::common;
+
+namespace kuzu {
+namespace binder {
+
+// Start file binding
```
ray6080, line 15: Consider the following comment style?
```cpp
/*
 * Bind file type and paths.
*/
```

src/include/planner/logical_plan/scan/logical_scan_file.h
```diff
@@ -0,0 +1,30 @@
+#pragma once
+
+#include "binder/copy/bound_file_scan_info.h"
+#include "planner/logical_plan/logical_operator.h"
+
+namespace kuzu {
+namespace planner {
+
+class LogicalScanFile : public LogicalOperator {
```
ray6080, line 9: I would name this as `LogicalReader` for now to be consistent with the physical operator. If you make this change, be sure to rename `boundFileScan`, `boundFileScanInfo`, and logical operator type accordingly, too.

ray6080 (approved):

src/include/planner/logical_plan/scan/logical_scan_file.h
andyfengHKU: This PR is not the best fit for renaming (after remaining the diff will be hard to read). I'll do in a separate PR.

PR #2054, opened 2023-09-19
Refactor node table to take in columnID instead of propertyID

ray6080:
Now node table takes in columnID instead propertyID to access a column. Internally node table stores all columns in a vector  (instead of a map), which is aligned with its tableSchema (we skip INTERNAL_ID column).
The front-end (or mapper currently) is responsible for aligning scanned columns in node table with properties in schema.

### DIFF
--- a/src/catalog/table_schema.cpp
+++ b/src/catalog/table_schema.cpp
@@ -5,7 +5,6 @@
 #include "catalog/rel_table_group_schema.h"
 #include "catalog/rel_table_schema.h"
 #include "common/constants.h"
-#include "common/exception/catalog.h"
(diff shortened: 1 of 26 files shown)


src/catalog/table_schema.cpp
```diff
@@ -50,6 +49,16 @@ property_id_t TableSchema::getPropertyID(const std::string& propertyName) const
         "Table: {} doesn't have a property with propertyName={}.", tableName, propertyName));
 }
 
+// TODO(Guodong): Instead of looping over properties, cache a map between propertyID and columnID.
```
andyfengHKU, line 52: Yes let's open an issue about this

PR #2069, opened 2023-09-22
finish list_product

AEsir777:
- list_product([x, x, x])
- Not done yet: 
  - overflow check
  - null

### DIFF
--- a/src/function/vector_list_functions.cpp
+++ b/src/function/vector_list_functions.cpp
@@ -15,6 +15,7 @@
 #include "function/list/functions/list_len_function.h"
 #include "function/list/functions/list_position_function.h"
 #include "function/list/functions/list_prepend_function.h"
+#include "function/list/functions/list_product_function.h"
 #include "function/list/functions/list_range_function.h"
 #include "function/list/functions/list_reverse_sort_function.h"
 #include "function/list/functions/list_slice_function.h"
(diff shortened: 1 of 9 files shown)


src/function/vector_list_functions.cpp
```diff
@@ -634,6 +635,67 @@ std::unique_ptr<FunctionBindData> ListSumVectorFunction::bindFunc(
...
 
+vector_function_definitions ListProductVectorFunction::getDefinitions() {
+    vector_function_definitions result;
+    result.push_back(std::make_unique<VectorFunctionDefinition>(LIST_PRODUCT_FUNC_NAME,
+        std::vector<LogicalTypeID>{LogicalTypeID::VAR_LIST}, LogicalTypeID::INT64, nullptr, nullptr,
+        bindFunc, false /* isVarlength*/));
+    return result;
+}
+
+std::unique_ptr<FunctionBindData> ListProductVectorFunction::bindFunc(
```
andyfengHKU, line 646: I'm sure we should template this together with ListSum because their switch looks identical except the function itself is different.

As you do this refactor, check if other bindFunc can be templated.

src/include/function/list/functions/list_product_function.h
```diff
@@ -0,0 +1,26 @@
...
+        common::ValueVector& inputVector, common::ValueVector& resultVector) {
+        auto inputValues =
+            reinterpret_cast<T*>(common::ListVector::getListValues(&inputVector, input));
+        auto inputDataVector = common::ListVector::getDataVector(&inputVector);
+        result = 1;
+        for (auto i = 0; i < input.size; i++) {
+            if (inputDataVector->isNull(input.offset + i)) {
+                continue;
+            }
+            result *= inputValues[i];
```
andyfengHKU, line 20: This is probably fine. But we are adding new interfaces to avoid direct access of data inside vector. So you should probably do

```
for (auto i = 0; i < input.size; i++) {
       if (inputDataVector->isNull(input.offset + i)) {
                 continue;
         }
          result *= inputDataVector.getValue<T>(input.offset + i);
```

Apply the same change to list_sum

andyfengHKU (approved):

src/function/vector_list_functions.cpp
AEsir777: The error message thrown becomes `LIST_PRODUCT::bindFunc` instead of `ListProductVectorFunctions`

PR #2070, opened 2023-09-22
reader: implement parallel CSV reading

Riolku:
This also refactors the CSVReader class to enable this change.


Riolku:
In testing, LDBC 100 loads in 3 seconds on ac4 (with 128 cores). However, this does not account for the overhead of counting the number of rows. If we are not going to make hash indexes resizable soon, we should add a dedicated CSV row counting function.

Furthermore, the 3 seconds does not account for the overhead of writing to disk, since I ran:

```
LOAD FROM <csv_path> RETURN COUNT(*)
```

On LDBC-10, which is must nicer to benchmark because the serial CSV reader loads it quickly enough, I got these numbers:

Serial: ~12s.
Parallel: ~350ms.

Again, in practice, we see only a 2x speedup since we pay the price of counting the rows.

src/include/common/copier_config/copier_config.h
```diff
@@ -19,18 +19,21 @@ struct CSVReaderConfig {
     char listBeginChar;
     char listEndChar;
     bool hasHeader;
+    bool parallel;
```
andyfengHKU, line 22: Not really related to this PR. I wonder if we should merge CSVReaderConfig with ReaderConfig at some point. Each reader can access a subset of fields from ReaderConfig class.

Riolku: Some attributes can be shared. `Parallel` definitely should be. Others though... don't make sense, right?

src/include/processor/operator/persistent/reader/csv/parallel_csv_reader.h
```diff
@@ -0,0 +1,25 @@
...
+
+#include "base_csv_reader.h"
+
+namespace kuzu {
+namespace processor {
+
+//! ParallelCSVReader is a class that reads values from a stream in parallel.
+class ParallelCSVReader final : public BaseCSVReader {
+public:
+    static const uint64_t PARALLEL_BLOCK_SIZE;
```
andyfengHKU, line 11: Is this expected to not have a default value?

src/processor/operator/persistent/reader/csv/base_csv_reader.cpp
```diff
@@ -17,13 +17,129 @@ using namespace kuzu::common;
...
+            goto escape;
+        } else {
+            position++;
+        }
+    } while (MaybeReadBuffer(nullptr));
+    return rows;
+
+escape:
+    if (!MaybeReadBuffer(nullptr)) {
+        return rows;
```
andyfengHKU, line 135: ditto

src/include/processor/operator/persistent/reader/csv/parallel_csv_reader.h
Riolku: It is defined in the `cpp` file. I mostly did this because it makes testing nicer (changing the block size doesn't cause a giant recompile). I kind of like it in the source file though. What do you think?

src/include/common/copier_config/copier_config.h
```diff
@@ -73,8 +72,13 @@ struct ReaderConfig {
         }
     }
 
-    inline bool parallelRead() const {
-        return fileType != FileType::CSV && fileType != FileType::TURTLE;
+    inline bool csvParallelRead(TableType tableType) const {
```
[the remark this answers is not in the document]
Riolku: Used in `reader.cpp:31`.

src/include/processor/operator/persistent/reader/csv/base_csv_reader.h
```diff
@@ -0,0 +1,99 @@
...
+    void copyStringToVector(common::ValueVector*, std::string);
+    //! Called after a row is finished to determine if we should keep processing.
+    inline bool finishedBlock() {
+        return mode != ParserMode::PARSING || rowToAdd >= common::DEFAULT_VECTOR_CAPACITY ||
+               finishedBlockDetail();
+    }
+
+protected:
+    std::string filePath;
+    common::CSVReaderConfig& csvReaderConfig;
```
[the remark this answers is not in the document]
Riolku:
```
struct ReaderConfig {
  FileType fileType = FileType::UNKNOWN;
  std::vector<std::string> filePaths;
  std::vector<std::string> columnNames;
  std::vector<std::unique_ptr<common::LogicalType>> columnTypes;  
  std::unique_ptr<CSVReaderConfig> csvReaderConfig = nullptr;
```

There are a bunch of fields in here, do you want to make a copy?

dataset/csv-edge-case-tests/bom-and-data.csv
```diff
@@ -0,0 +1,2 @@
+﻿data,moredata
+abc,﻿﻿bomdata﻿
```
acquamarin, line 2: We always add a newline at the end of the csv file.

acquamarin (approved):

dataset/csv-edge-case-tests/bom-and-data.csv
Riolku: It's important to test that files without trailing newlines are also correctly handled.

src/processor/operator/persistent/reader.cpp
```diff
@@ -84,7 +80,7 @@ void Reader::readNextDataChunk() {
         readFunc(*readFuncData, morsel->blockIdx, dataChunk.get());
         if (dataChunk->state->selVector->selectedSize > 0) {
             leftArrowArrays.appendFromDataChunk(dataChunk.get());
-        } else {
+        } else if (readFuncData->emptyBlockImpliesDone()) {
             sharedState->moveToNextFile();
```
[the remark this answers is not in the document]
Riolku: We definitely need this. Otherwise, lines larger than the PARALLEL_BLOCK_SIZE will die. Or, if the header is long enough, we will parse incorrectly.

The specific reason is that if a line is very long and the block end is within this line, the parallel CSV reader will have no data for this block. It will thus return zero. However, that does **not** mean we should move to the next file, in fact, it would be an error to do so.

I can add a test for this if you like.

src/include/processor/operator/persistent/reader/csv/base_csv_reader.h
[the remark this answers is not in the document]
Riolku: Nevermind.

src/processor/operator/persistent/reader.cpp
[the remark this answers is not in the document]
acquamarin: Can you add a comment about the reason?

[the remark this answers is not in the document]
Riolku: Yes, and Ill add a test as well later today.

src/processor/operator/persistent/copy_node.cpp
```diff
@@ -179,9 +179,10 @@ void CopyNode::checkNonNullConstraint(NullColumnChunk* nullChunk, offset_t numNo
 }
 
 void CopyNode::finalize(ExecutionContext* context) {
-    auto numNodes = StorageUtils::getStartOffsetOfNodeGroup(sharedState->getCurNodeGroupIdx()) +
```
ray6080, line 182: Why do we make this change? The change looks like a bug to me. This could just lead to 0 numNodes in statistics.

PR #2071, opened 2023-09-23
Rdf predicate iri processing

andyfengHKU:
This PR address the following issues

- Allow querying node and rel pattern through RDF label instead of RDF resource or triples label. E.g. `MATCH (a:Taxonomy) ...`
- Allow querying predicate IRI property as if it is a regular rel property. Note that we only materialize predicate internal ID in  rel table rather than predicate IRI. Instead, predicate IRI is stored as a node property in resource table. This PR allows querying predicate IRI e.g. `MATCH (s:Taxonomy)-[p:Taxonomy]->(o:Taxonomy) RETURN p.iri` and compiles the property reading as a sequential scan over resource table followed by a hash join over predicate internal ID. The plan can be abstracted as the following

```
             HJ(o._id)
            /         \
        HJ(p._id)     S(o.*)
        /         \   
E(p._id, o._id)     S(p.*)
     |
   S(s.*)
```

We temporarily introduce LogicalFillTableID operator because the table ID of internal ID column is not materialized currently so we don't have a way to map the predicate ID from rel table to the internal ID from node table because their table IDs are different.


src/include/binder/binder.h
```diff
@@ -226,8 +226,13 @@ class Binder {
         const std::string& parsedName, const std::vector<common::table_id_t>& tableIDs);
     void bindQueryNodeProperties(NodeExpression& node);
 
-    std::vector<common::table_id_t> bindNodeTableIDs(const std::vector<std::string>& tableNames);
-    std::vector<common::table_id_t> bindRelTableIDs(const std::vector<std::string>& tableNames);
+    /*** bind table ID ***/
+    // TODO: comment
```
andyfengHKU, line 230: Fix this TODO

src/catalog/catalog_content.cpp
```diff
@@ -270,6 +260,15 @@ bool CatalogContent::containTable(const std::string& tableName, TableType tableT
     return tableSchemas.at(tableID)->tableType == tableType;
 }
 
+bool CatalogContent::containTable(common::TableType tableType) const {
```
andyfengHKU, line 263: contains

src/include/planner/operator/scan/logical_scan_internal_id.h
```diff
@@ -29,5 +29,29 @@ class LogicalScanInternalID : public LogicalOperator {
     std::vector<common::table_id_t> tableIDs;
 };
 
+class LogicalFillTableID : public LogicalOperator {
```
andyfengHKU, line 32: Comment

src/include/processor/operator/scan_node_id.h
```diff
@@ -92,5 +92,29 @@ class ScanNodeID : public PhysicalOperator {
     std::shared_ptr<common::ValueVector> outValueVector;
 };
 
+class FillTableID : public PhysicalOperator {
```
andyfengHKU, line 95: Ditto comment

src/include/planner/operator/scan/logical_scan_internal_id.h
```diff
@@ -29,5 +29,29 @@ class LogicalScanInternalID : public LogicalOperator {
...
+public:
+    LogicalFillTableID(std::shared_ptr<binder::Expression> internalID, common::table_id_t tableID,
```
ray6080, line 34: Let's open an issue on this about removing this by implementing materialized `INTERNAL_ID`.

src/include/planner/operator/scan/logical_scan_internal_id.h
ray6080, line 32: Let's separate this to another file. It's unexpected to see it here.

ray6080 (approved):

PR #2079, opened 2023-09-25
Support map datatype on copy to command

rfdavid:
Simple implementation to add MAP datatype on COPY TO


andyfengHKU:
Closed because we are moving to a parquet writer with compression.

PR #2085, opened 2023-09-25
Int8 and Int16 bitpacking

benjaminwinger:
This uses the 8 and 16-bit extensions to fastunpack from duckdb (external changes to fastpfor are documented in `third_party/fastpfor/README`). fastpfor does actually include 8 and 16-bit functions, but I ran into some issues (which I didn't spend much time investigating) when I briefly tried using them before, and they use a somewhat different interface and are much less compact (~5000 lines of code).

### DIFF
--- a/src/storage/store/compression.cpp
+++ b/src/storage/store/compression.cpp
@@ -84,6 +84,16 @@ bool CompressionMetadata::canUpdateInPlace(
             return IntegerBitpacking<int32_t>::canUpdateInPlace(
                 value, BitpackHeader::readHeader(data));
         }
+        case PhysicalTypeID::INT16: {
+            auto value = vector.getValue<int16_t>(pos);
+            return IntegerBitpacking<int16_t>::canUpdateInPlace(
+                value, BitpackHeader::readHeader(data));
+        }
+        case PhysicalTypeID::INT8: {
+            auto value = vector.getValue<int8_t>(pos);
+            return IntegerBitpacking<int8_t>::canUpdateInPlace(
+                value, BitpackHeader::readHeader(data));
+        }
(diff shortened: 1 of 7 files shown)


src/storage/store/compression.cpp
```diff
@@ -189,6 +220,36 @@ bool IntegerBitpacking<T>::canUpdateInPlace(T value, const BitpackHeader& header
...
+    if constexpr (std::is_same_v<std::make_signed_t<T>, int32_t> ||
+                  std::is_same_v<std::make_signed_t<T>, int64_t>) {
+        FastPForLib::fastpack(in, (uint32_t*)out, bitWidth);
+    } else if constexpr (std::is_same_v<std::make_signed_t<T>, int16_t>) {
+        FastPForLib::fastpack(in, (uint16_t*)out, bitWidth);
+    } else if constexpr (std::is_same_v<std::make_signed_t<T>, int8_t>) {
+        FastPForLib::fastpack(in, (uint8_t*)out, bitWidth);
+    } else {
+        // Unreachable
+        assert(false);
```
ray6080, line 249: Our convention is to not use `assert(false)`, instead `throw NotImplementedException("fastpack")`. mainly because of that in release mode, the `assert(false)` is just simply ignored and triggers compiler complaints.

ray6080 (approved):

src/storage/store/compression.cpp
benjaminwinger: I've moved the last reachable branch into the else and added static_assert instead of its former condition so that we get a compile time failure in future if an unexpected type is used.

What I'd really wanted to use was `static_assert(false)`, but that doesn't work properly because it's not dependent on the template type.

Issue #921, opened 2022-11-05
Disallowing multiple Database instances (e.g., from multiple processes) running on the same database directory 


semihsalihoglu-uw:
Currently we do not have a mechanism to block multiple Database instances to be , i.e., multiple embedded Kuzu instances running on the same database directory. This should not be allowed as long as  we do not have a mechanism to make Database instances read-only. This is because this leads to the following problem if these processes are allowed to write.

Suppose there are 2 separate OS processes: P1 and P2 running Kuzu:
- P1 writes something to the database. For example adds a new User node. Suppose there was 5 User nodes before. Now there is 6.
- P2 will not be able to see this write because parts of the database data, e.g., catalog or node_statistics_and_deleted_ids, as always kept in memory and there is no mechanism for P2 to know that these parts have been modified, and refresh them. In this example, there is no way for P2 to know that node_statistics_and_deleted_ids have been modified. So it will think  that there are still 5 User nodes. 

Probably the most common case this would appear is if there are 2 separate OS processes running Kuzu but this is not really about number of processes. This is about number of `Database` instances. Database here refers to the object in program lines like this: `auto database = make_unique<Database>(*databaseConfig, *systemConfig);`.

We should add a note about this to the documentation's Client APIs and solve this immediately after phase 1 release.

andyfengHKU:
Close due to completion.

PR #2089, opened 2023-09-26
Database access mode

hououou:
This PR is aimed at solving [issue #921](https://github.com/kuzudb/kuzu/issues/921) on non-Windows platforms.

Steps:
- Add a file `.lock` in each database directory.  Use this file to check the status of the database.
- When a database is started with an access mode referring to as` READ_ONLY` or `READ_WRITE` (note I use `READ_WRITE`  when the access mode is not specified), lock the `.lock` using the function` fcntl`.

Limitation
- ` fcntl` can only solve the read-write lock during different processes.  So we cannot solve the read-write lock in the same process.

### DIFF
--- a/src/common/file_utils.cpp
+++ b/src/common/file_utils.cpp
@@ -43,7 +43,8 @@ int64_t FileInfo::getFileSize() {
 #endif
 }
 
-std::unique_ptr<FileInfo> FileUtils::openFile(const std::string& path, int flags) {
+std::unique_ptr<FileInfo> FileUtils::openFile(
+    const std::string& path, int flags, FileLockType lock_type) {
 #if defined(_WIN32)
     auto dwDesiredAccess = 0ul;
     auto dwCreationDisposition = (flags & O_CREAT) ? OPEN_ALWAYS : OPEN_EXISTING;
@@ -62,11 +63,34 @@ std::unique_ptr<FileInfo> FileUtils::openFile(const std::string& path, int flags
         throw Exception(StringUtils::string_format("Cannot open file. path: {} - Error {}: {}",
             path, GetLastError(), std::system_category().message(GetLastError())));
     }
+    if (lock_type != FileLockType::NO_LOCK) {
+        DWORD dwFlags = lock_type == FileLockType::READ_LOCK ?
+                            LOCKFILE_FAIL_IMMEDIATELY :
+                            LOCKFILE_FAIL_IMMEDIATELY | LOCKFILE_EXCLUSIVE_LOCK;
+        OVERLAPPED overlapped = {0};
+        overlapped.Offset = 0;
+        BOOL rc = LockFileEx(handle, dwFlags, 0, 0, 0, &overlapped);
+        if (!rc) {
+            throw Exception("Could not set lock on file : " + path);
+        }
+    }
     return std::make_unique<FileInfo>(path, handle);
 #else
     int fd = open(path.c_str(), flags, 0644);
     if (fd == -1) {
         throw Exception("Cannot open file: " + path);
+    }
+    if (lock_type != FileLockType::NO_LOCK) {
+        struct flock fl;
+        memset(&fl, 0, sizeof fl);
+        fl.l_type = lock_type == FileLockType::READ_LOCK ? F_RDLCK : F_WRLCK;
+        fl.l_whence = SEEK_SET;
+        fl.l_start = 0;
+        fl.l_len = 0;
+        int rc = fcntl(fd, F_SETLK, &fl);
+        if (rc == -1) {
+            throw Exception("Could not set lock on file : " + path);
+        }
     }
     return std::make_unique<FileInfo>(path, fd);
 #endif
--- a/src/include/common/file_utils.h
+++ b/src/include/common/file_utils.h
@@ -15,6 +15,8 @@
 namespace kuzu {
 namespace common {
 
+enum class FileLockType : uint8_t { NO_LOCK = 0, READ_LOCK = 1, WRITE_LOCK = 2 };
+
 struct FileInfo {
 #ifdef _WIN32
     FileInfo(std::string path, const void* handle) : path{std::move(path)}, handle{handle} {}
@@ -36,7 +38,8 @@ struct FileInfo {
 
 class FileUtils {
 public:
-    static std::unique_ptr<FileInfo> openFile(const std::string& path, int flags);
+    static std::unique_ptr<FileInfo> openFile(
+        const std::string& path, int flags, FileLockType lock_type = FileLockType::NO_LOCK);
 
     static void readFromFile(
         FileInfo* fileInfo, void* buffer, uint64_t numBytes, uint64_t position);
--- a/src/include/main/database.h
+++ b/src/include/main/database.h
@@ -10,6 +10,8 @@
 namespace kuzu {
 namespace main {
 
+enum class AccessMode : uint8_t { READ_ONLY = 0, READ_WRITE = 1 };
+
 /**
  * @brief Stores runtime configuration for creating or opening a Database
  */
@@ -21,13 +23,15 @@ struct KUZU_API SystemConfig {
      *        reducing the amount of File I/O
      *  @param maxNumThreads The maximum number of threads to use during query execution
      *  @param enableCompression Whether or not to compress data on-disk for supported types
+     *  @param accessMode Access mode to the database (READ_ONLY or READ_WRITE)
      */
-    explicit SystemConfig(
-        uint64_t bufferPoolSize = -1u, uint64_t maxNumThreads = 0, bool enableCompression = true);
+    explicit SystemConfig(uint64_t bufferPoolSize = -1u, uint64_t maxNumThreads = 0,
+        bool enableCompression = true, AccessMode accessMode = AccessMode::READ_WRITE);
 
     uint64_t bufferPoolSize;
     uint64_t maxNumThreads;
     bool enableCompression;
+    AccessMode accessMode;
 };
 
 /**
@@ -66,7 +70,8 @@ public:
     KUZU_API static void setLoggingLevel(std::string loggingLevel);
 
 private:
-    void initDBDirAndCoreFilesIfNecessary() const;
+    void openLockFile();
+    void initDBDirAndCoreFilesIfNecessary();
     static void initLoggers();
     static void dropLoggers();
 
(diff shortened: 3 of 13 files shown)


examples/cpp/CMakeLists.txt
```diff
@@ -4,7 +4,7 @@ project(example-cpp)
 set(CMAKE_CXX_STANDARD 20)
 
 include_directories(../../src/include)
-link_directories(../../build/release/src)
+link_directories(../../build/debug/src)
```
ray6080, line 7: Revert this after you're done with debugging.

examples/cpp/main.cpp
ray6080, line 1: Ditto. Revert changes in this file after debugging.

src/common/file_utils.cpp
```diff
@@ -72,6 +77,64 @@ std::unique_ptr<FileInfo> FileUtils::openFile(const std::string& path, int flags
...
+		open_flags |= O_DIRECT | O_SYNC;
+#endif
+	}
+    int fd = open (path.c_str(), open_flags, 0666);//(path.c_str(), O_RDWR);
+    // (path.c_str(), flags, 0666);
+    if (fd == -1) {
+        throw Exception("Cannot open file: " + path);
+    }
+
+    if (lock_type != FileLockType::NO_LOCK) {
```
ray6080, line 124: Let's not add a new `openFile` function to `FileUtils`. I believe we can merge the lock logic into existing `openFile` function. By default, `openFile` will opens a file in `NO_LOCK` mode.

Also, remember we should handle windows too (if it is doable).

src/include/common/file_utils.h
```diff
@@ -15,6 +15,24 @@
 namespace kuzu {
 namespace common {
 
+enum class FileLockType : uint8_t { NO_LOCK = 0, READ_LOCK = 1, WRITE_LOCK = 2 };
+
+class FileFlags {
```
ray6080, line 20: This class should be cleaned after you merge locking logic into existing `openFile` function.

src/main/database.cpp
```diff
@@ -68,6 +84,22 @@ void Database::setLoggingLevel(std::string loggingLevel) {
     spdlog::set_level(LoggingLevelUtils::convertStrToLevelEnum(std::move(loggingLevel)));
 }
 
+void Database::CheckAccessMode() const {
```
ray6080, line 87: CheckAccessMode -> checkAccessMode

src/include/main/database.h
```diff
@@ -10,6 +10,8 @@
 namespace kuzu {
 namespace main {
 
+enum class AccessMode : uint8_t { UNDEFINED = 0, AUTOMATIC = 1, READ_ONLY = 2, READ_WRITE = 3};
```
ray6080, line 13: What is `AUTOMATIC` mode? It doesn't seem to be handled. Can we get rid of it for now?

test/runner/test_locking.cpp
ray6080, line 1: Clean up couts.

ray6080 (commented):
Did a pre-review while this is still a draft. please check my comments.
There are still some code not cleaned up. Let me know when it is ready for another review.

src/include/main/database.h
```diff
@@ -88,6 +94,7 @@ class Database {
     std::unique_ptr<transaction::TransactionManager> transactionManager;
     std::unique_ptr<storage::WAL> wal;
     std::shared_ptr<spdlog::logger> logger;
+    std::unique_ptr<common::FileInfo> lockStatus;
```
hououou, line 97: is it ok? I need to hold the `.lock` file information after I open it. otherwise the lock will lose its efficacy

src/main/connection.cpp
```diff
@@ -67,13 +67,25 @@ std::unique_ptr<PreparedStatement> Connection::prepare(const std::string& query)
...
+    }
     return executeAndAutoCommitIfNecessaryNoLock(preparedStatement.get());
 }
 
 std::unique_ptr<QueryResult> Connection::query(
     const std::string& query, const std::string& encodedJoin) {
     lock_t lck{mtx};
     auto preparedStatement = prepareNoLock(query, true /* enumerate all plans */, encodedJoin);
+    bool statementReadOnly = preparedStatement->isReadOnly();
+    bool accessMode = this->database->systemConfig.access_mode == AccessMode::READ_ONLY;
```
ray6080, line 84: accessMode -> isReadOnlyMode.

src/main/connection.cpp
```diff
@@ -67,13 +67,25 @@ std::unique_ptr<PreparedStatement> Connection::prepare(const std::string& query)
...
+    if (accessMode && !statementReadOnly) {
```
ray6080, line 85: L83 - L85:
```cpp
    bool isReadOnlyMode = this->database->systemConfig.access_mode == AccessMode::READ_ONLY;
    if (isReadOnlyMode && !preparedStatement->isReadOnly()) {
```

src/main/database.cpp
```diff
@@ -50,6 +65,7 @@ Database::Database(std::string databasePath, SystemConfig systemConfig)
     bufferManager = std::make_unique<BufferManager>(this->systemConfig.bufferPoolSize);
     memoryManager = std::make_unique<MemoryManager>(bufferManager.get());
     queryProcessor = std::make_unique<processor::QueryProcessor>(this->systemConfig.maxNumThreads);
+    checkAccessMode();
```
ray6080, line 68: Can we move this function inside `initDBDirAndCoreFilesIfNecessary`? You can check access mode after 
```cpp
    if (!FileUtils::fileOrPathExists(databasePath)) {
        FileUtils::createDir(databasePath);
    }
```

ray6080, line 68: Can we move this function inside `initDBDirAndCoreFilesIfNecessary`? You can check access mode after 
```cpp
    if (!FileUtils::fileOrPathExists(databasePath)) {
        FileUtils::createDir(databasePath);
    }
```

src/main/database.cpp
```diff
@@ -68,6 +84,21 @@ void Database::setLoggingLevel(std::string loggingLevel) {
...
 
+void Database::checkAccessMode() {
+    if (!FileUtils::fileOrPathExists(databasePath)) {
+        FileUtils::createDir(databasePath);
+    }
+    int flags;
+    FileLockType lock;
+    auto lockFilePath = StorageUtils::getLockFilePath(databasePath);
+    if (!FileUtils::fileOrPathExists(lockFilePath)) {
+        getFileFlags(systemConfig, flags, lock, true);
```
ray6080, line 95: 1. You're setting both `flags` and `lock`, why name this function as `getFileFlags`?

src/main/database.cpp
```diff
@@ -68,6 +84,21 @@ void Database::setLoggingLevel(std::string loggingLevel) {
     spdlog::set_level(LoggingLevelUtils::convertStrToLevelEnum(std::move(loggingLevel)));
 }
 
+void Database::checkAccessMode() {
+    if (!FileUtils::fileOrPathExists(databasePath)) {
+        FileUtils::createDir(databasePath);
+    }
```
ray6080, line 90: Remove L88 to L90 after you move this function into `initDBDirAndCoreFilesIfNecessary`.

src/main/database.cpp
```diff
@@ -68,6 +84,21 @@ void Database::setLoggingLevel(std::string loggingLevel) {
     spdlog::set_level(LoggingLevelUtils::convertStrToLevelEnum(std::move(loggingLevel)));
 }
 
+void Database::checkAccessMode() {
+    if (!FileUtils::fileOrPathExists(databasePath)) {
```
ray6080, line 88: I think we should check if the access mode is set to READ_ONLY, if so, we should throw an exception "Cannot create an empty database under READ ONLY mode."

src/main/database.cpp
```diff
@@ -21,6 +22,20 @@ using namespace kuzu::transaction;
 namespace kuzu {
 namespace main {
 
+void getFileFlags(SystemConfig systemConfig, int& flags, FileLockType& lock, bool create_new) {
```
ray6080, line 25: 1. Simply pass in `accessMode` is enough, no need to pass in the whole SystemConfig.
2. Make this function static.
3. Rename to `getLockFileFlagsAndType`.
4. Reorder your params, so your input and output are separate and together. better for reading. for example `getFileFlags(SystemConfig systemConfig, bool create_new, int& flags, FileLockType& lock)`

ray6080, line 25: `create_new` -> `createNew`. Be careful with your code style. We follow camelCase, and functions start with lower case char.

src/main/database.cpp
```diff
@@ -21,6 +22,20 @@ using namespace kuzu::transaction;
...
+        assert(!create_new);
+        flags = O_RDONLY;
+        lock = FileLockType::READ_LOCK;
+    } else {
+        flags = O_RDWR;
+        lock = FileLockType::WRITE_LOCK;
+        if (create_new) {
+            flags = O_WRONLY | O_CREAT;
+        }
+    }
```
ray6080, line 36: If it's createNew, do we need to set flags to `O_WRONLY`?, will `O_RDWR` be fine?
If so, I think this piece of code can be re-organized as following:
```cpp
    flags = systemConfig.access_mode == AccessMode::READ_ONLY ? O_RDONLY : O_RDWR;
    if (create_new) {
        assert(flags == O_RDWR);
        flags |= O_CREAT;
    }
    lock = systemConfig.access_mode == AccessMode::READ_ONLY ? FileLockType::READ_LOCK : FileLockType::WRITE_LOCK;
```

src/include/main/database.h
```diff
@@ -5,11 +5,13 @@
 
 #include "common/api.h"
 #include "common/constants.h"
+#include "common/file_utils.h"
```
ray6080, line 8: I don't think you need to include `file_utils.h` in header file. Be careful when adding includes to header files, especially under `common` and `main`, the former one is included in lots of places, and the latter one exposed to clients.
Can you double check if you can remove the include? instead, add forward declaration here https://github.com/kuzudb/kuzu/blob/f05d34893d54b4d75708087a844e6c356907ac84/src/include/main/kuzu_fwd.h#L32

ray6080 (changes requested):
Didn't look into tests yet. Will continue later.

ray6080 (changes requested):
Didn't look into tests yet. Will continue later when you finalize changes to tests.

ray6080 (approved):

PR #2095, opened 2023-09-26
Bitpack internal IDs

benjaminwinger:
I was going to benchmark this, as it adds an extra step to the decompression, but I'm not sure it's actually being used. Are internalIDs actually stored anywhere other than in rels?

### DIFF
--- a/src/storage/store/column_chunk.cpp
+++ b/src/storage/store/column_chunk.cpp
@@ -110,6 +110,7 @@ static std::shared_ptr<CompressionAlg> getCompression(
         return std::make_shared<Uncompressed>(dataType);
     }
     switch (dataType.getPhysicalType()) {
+    case PhysicalTypeID::INTERNAL_ID:
     case PhysicalTypeID::INT64: {
         return std::make_shared<IntegerBitpacking<int64_t>>();
     }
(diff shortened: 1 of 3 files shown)


src/storage/store/column_chunk.cpp
```diff
@@ -122,6 +122,7 @@ ColumnChunk::ColumnChunk(
         break;
     }
     case PhysicalTypeID::VAR_LIST:
+    case PhysicalTypeID::INTERNAL_ID:
```
ray6080, line 125: Can we move the switch case into a separate function? This is becoming quite large now in the constructor.

ray6080 (approved):

ray6080:
> I was going to benchmark this, as it adds an extra step to the decompression, but I'm not sure it's actually being used. Are internalIDs actually stored anywhere other than in rels?

You're right. Currently it's not materialized in node tables.

src/storage/store/column_chunk.cpp
benjaminwinger: Sure. I'd already simplified it somewhat in #2102 ([here](https://github.com/kuzudb/kuzu/pull/2102/files#diff-b7ddec24858a04e0feec0e25f3a5dca30c0672beb96db4147b64e044a84dcad6R95-R144)), so I could move that change into this PR. Is that simplification enough, or would you prefer moving everything into a separate function?

ray6080:
@benjaminwinger , can we hold this PR a bit longer after the rel node group one #2246 ?

benjaminwinger:
Yes, I was thinking it would make sense to wait until that's merged so it can be tested.

ray6080:
I think this can be closed now. @benjaminwinger

benjaminwinger:
Yes, let's close in favour of #2579

PR #2099, opened 2023-09-27
upload benchmark copy query times to benchmark server

russell-liu:


benchmark/serializer.py
```diff
@@ -48,14 +49,22 @@ def serialize(dataset_name, dataset_path, serialized_graph_path):
...
+        # created for each query to avoid memory leaks.
+        process = subprocess.Popen([kuzu_exec_path, serialized_graph_path],
+            stdin=subprocess.PIPE, stdout=subprocess.PIPE)
+        process.stdin.write((s + ";" + "\n").encode("ascii"))
+        process.stdin.close()
+        match = re.match('copy (.*) from', s, re.IGNORECASE)
+        for line in iter(process.stdout.readline, b''):
+            print(line.decode("utf-8"), end='')
+            if match:
+                with open(os.path.join(benchmark_copy_log_dir, match.group(1) + '_log.txt'), 'ab') as f:
```
mewim, line 62: You should consider move this out of the `for` loop. You should not need to reopen the file each time.

benchmark/serializer.py
```diff
@@ -48,14 +49,22 @@ def serialize(dataset_name, dataset_path, serialized_graph_path):
...
-        except subprocess.CalledProcessError as e:
+        # Run kuzu shell one query at a time. This ensures a new process is
+        # created for each query to avoid memory leaks.
+        process = subprocess.Popen([kuzu_exec_path, serialized_graph_path],
+            stdin=subprocess.PIPE, stdout=subprocess.PIPE)
+        process.stdin.write((s + ";" + "\n").encode("ascii"))
+        process.stdin.close()
+        match = re.match('copy (.*) from', s, re.IGNORECASE)
+        for line in iter(process.stdout.readline, b''):
+            print(line.decode("utf-8"), end='')
```
mewim, line 60: Try to switch to `sys.stdout.write` and ` sys.stdout.flush` to see if it solves the out of order issue

russell-liu:
Working now at: https://benchmark.kuzudb.com/run.html#2411

benchmark/serializer.py
```diff
@@ -46,16 +47,36 @@ def serialize(dataset_name, dataset_path, serialized_graph_path):
...
+        # Run kuzu shell one query at a time. This ensures a new process is
+        # created for each query to avoid memory leaks.
+        process = subprocess.Popen([kuzu_exec_path, serialized_graph_path],
+            stdin=subprocess.PIPE, stdout=subprocess.PIPE)
+        process.stdin.write((s + ";" + "\n").encode("ascii"))
+        process.stdin.close()
+        match = re.match('create\s+(.+?)\s+table\s+(.+?)\s*\(', s, re.IGNORECASE)
+        if match:
+            table_types[match.group(2)] = match.group(1).lower()
+            match = False
```
mewim, line 63: This logic is not very readable. 
1. You should run the check for `create` statement before `subprocess.Popen`. If it is a `create` you should simply pass the input via `stdout` and `stderr` redirection instead of manually get the output and print.
2.  Instead of setting `match` to `False` here you should indent line 66-75 so it is under your `else`. If you do this, you can also avoid the `if` check for your `write` and `flush`.
3. Do not reuse your flags. Call it `create_match` and `copy_match` maybe. Reusing a flag-type variable is 
error-prone and make the code hard to understand.

mewim (approved):
Code LGTM now. It should be merged after another round of end-to-end test.

PR #2102, opened 2023-09-27
Optional Compression

benjaminwinger:
~~Not quite done since it's currently only exposed in the C++ and Rust APIs, and the option needs to be passed to NodeColumns~~.

Python and C don't currently have any sort of SystemConfig structure and only take the buffer pool size as an argument. (*Edit: Added for python and C APIs*)
It looks the same for nodejs and java (which I'm much less familiar with).

An optional argument would probably work fine for python, but since C lacks default values it should probably also have a SystemConfig struct that can be passed to `kuzu_database_init`. (*done*)

### DIFF
--- a/tools/java_api/src/main/java/com/kuzudb/KuzuDatabase.java
+++ b/tools/java_api/src/main/java/com/kuzudb/KuzuDatabase.java
@@ -8,6 +8,7 @@ public class KuzuDatabase {
     long db_ref;
     String db_path;
     long buffer_size;
+    boolean enableCompression = true;
     boolean destroyed = false;
 
     /**
@@ -17,7 +18,7 @@ public class KuzuDatabase {
     public KuzuDatabase(String databasePath) {
         this.db_path = databasePath;
         this.buffer_size = 0;
-        db_ref = KuzuNative.kuzu_database_init(databasePath, 0);
+        db_ref = KuzuNative.kuzu_database_init(databasePath, 0, true);
     }
 
     /**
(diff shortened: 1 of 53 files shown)


ray6080 (commented):
Is there a plan to test the configuration inside cpp?

benjaminwinger:
> Is there a plan to test the configuration inside cpp?

Not really. We could maybe expose information about the properties which could include the compression information for each column chunk. That could be useful later for examining the efficiency of the compression (e.g. if it also reports the compressed size and the number of values stored).

ray6080:
We should also add an arg `nocompression` to CLI.

tools/java_api/src/main/java/com/kuzudb/KuzuDatabase.java
```diff
@@ -28,7 +29,19 @@ public KuzuDatabase(String databasePath) {
...
         this.db_path = databasePath;
         this.buffer_size = bufferPoolSize;
-        db_ref = KuzuNative.kuzu_database_init(databasePath, bufferPoolSize);
+        db_ref = KuzuNative.kuzu_database_init(databasePath, bufferPoolSize, true);
+    }
+
+    /**
+    * Creates a database object.
+    * @param databasePath: Database path. If the database does not already exist, it will be created.
+    * @param bufferPoolSize: Max size of the buffer pool in bytes.
```
ray6080, line 38: add param for `enableCompression`

ray6080 (approved):

PR #2115, opened 2023-09-29
fix uint32 on java

Ashleyhx:


Ashleyhx:
added in another pr by @acquamarin

PR #2124, opened 2023-09-30
Inline generated export headers

benjaminwinger:
#2118 broke the single file header generator script because it adds a cmake-generated header which resides in the build directory (probably should add tests to CI for the header generator at some point). This adds the build directory to the list of directories to search for header files when creating the single file header.


andyfengHKU (approved):

benjaminwinger:
Re-written to just inline the headers since this also breaks the single file C header which gets included verbatim without substituting includes.

andyfengHKU (approved):

PR #2130, opened 2023-10-01
Ignore imports and exports in the static rust build

benjaminwinger:
Should fix the rust build failure. The rust build can ignore the imports/exports since it's not using the shared library, but without the define the rust shim is treated as third-party software importing the headers for use with the shared library.


acquamarin (approved):
Can you take a look at the examples compilation error?

PR #2140, opened 2023-10-04
Frame of reference encoding

benjaminwinger:
This modifies the integer bitpacking compression to make use of a fixed offset when values are all large, and if it will reduce the number of bits required to store each value.
The offset is stored in the ColumnChunk metadata (increasing the size of that by 8 bytes), and removed from values when compressing and added back during decompression. 

Decompression is done in-place, but does require an extra pass over the decompressed data when the offset is non-zero. Compression when the offset is non-zero is done using a small temporary buffer.

This should be effective at reducing the size needed for data such as timestamps, which usually fall into a relatively small range of values, but with a relatively high minimum value, and will still have very good random read/write performance. Large numeric IDs with a fixed number of digits should also benefit from this.
If also functionally implements constant compression for integers, as if all values are the same the data can be stored in 0 bits per value with the actual value stored as the offset.

I'll try and put together a quick benchmark with a generated dataset tailored to show the performance difference when this is used.

### DIFF
--- a/src/include/storage/store/compression.h
+++ b/src/include/storage/store/compression.h
@@ -1,5 +1,6 @@
 #pragma once
 
+#include <array>
 #include <cstdint>
 #include <limits>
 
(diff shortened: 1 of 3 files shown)


ray6080 (approved):
LGTM!

benjaminwinger:
I'm informed that this is Frame of Reference encoding (without delta coding): https://lemire.me/blog/2012/02/08/effective-compression-using-frame-of-reference-and-delta-coding/
I've updated some of the terminology in this PR to reflect that.

PR #2142, opened 2023-10-04
refactor: splitup table_functions.h

Riolku:


andyfengHKU (approved):

src/include/function/table_functions/show_tables.h
```diff
@@ -0,0 +1,19 @@
+#pragma once
+
+#include "function/table_functions.h"
+#include "function/table_functions/bind_data.h"
+#include "function/table_functions/bind_input.h"
+
+namespace kuzu::function {
```
andyfengHKU, line 7: Though our convention is to use

namespace kuzu
namespace function

Let's stick to this. Or we have to change every place

Riolku: I strongly prefer this newer syntax. Shall we poll? I can change it everywhere in a subsequent PR.

andyfengHKU: sure let's bring to the standup

Riolku: Reverted namespace changes and will merge when tests pass.

Issue #2136, opened 2023-10-02
Explicit Symbol Exports


benjaminwinger:
In #2118 we switched from exporting all symbols on Windows to explicit export of symbols to avoid hitting the limit in a single windows dll. This change would have a number of benefits on other platforms too (smaller dynamic libraries, faster library loading, no accidental use of private APIs), and needs to be more rigorously tested.

TODO:
- [x] Make default visibility of functions hidden ([by making these statements unconditional](https://github.com/kuzudb/kuzu/blob/ede12a79acc262b8082dad8e927a9802d48a8c5a/CMakeLists.txt#L11-L14))
- [ ] Remove re-exported static dependency symbols from the dynamic libraries on Linux (probably macOS too since I think it's related to ELF). Maybe there are better solutions, but what I could find before was using [linker version scripts](https://www.gnu.org/software/gnulib/manual/html_node/LD-Version-Scripts.html) to filter the public symbols. They support a wildcard syntax which can probably be used to avoid explicitly listing all the symbols. I don't know how it interacts with C++ name mangling, but maybe we could just allow a variation of `kuzu*` to filter everything not in the kuzu namespace (or C functions prefixed with `kuzu_`).
- [x] Get the API tests to link against the dynamic libraries with the restricted symbols so that we can properly test the exported APIs (the main issue is that the testing helper classes use private APIs, so we either need to remove those, or maybe set up an internal static library for those helpers that they can link with alongside the dynamic library).
- [x] Produce the single file header as part of the cmake build (maybe we could set up a cmake install command at the same time).
  - [x] Get the API tests to use the single file header.
  - [ ] Test linking the rust API against the dynamic library and single file header.

Riolku:
From [LD-Version-Scripts](https://www.gnu.org/software/gnulib/manual/html_node/LD-Version-Scripts.html)
```
If you target platforms that do not support linker scripts (i.e., all platforms that doesn’t use GNU LD) you may want to consider a more portable but less powerful alternative: libtool -export-symbols. It will hide internal symbols from your library, but will not add ELF versioning symbols. 
```

We should investigate if the more portable solution is possible. Otherwise, we can't build on Alpine linux, and potentially not on MacOS(?).

Riolku:
It seems the portable solution is a file that lists all symbols to export. I am in favor of this solution, because it's very explicit and we have to explicitly specify when we want to export new symbols (and we don't do this very often, afaik).

benjaminwinger:
[...]

PR #2153, opened 2023-10-05
Link C and C++ API tests against the API-restricted shared library

benjaminwinger:
First pass at this, mostly to see if it passes CI.

Implements part of #2136.

Various functions needed the `KUZU_API` annotation. Some of these are fine, such as `~PreparedStatement` and `StorageDriver`, which should be public. 

However a number of things which should really be private (e.g. `ValueVector::resetAuxiliaryBuffer`) had to be added to the exports since they are used in template functions related to the UDF helpers. While the templates themselves can't be moved into source files, we could probably hide some of their implementations, but I'm not sure if that really does anything other than expose something different which really should be private.

I'm also not convinced that `ValueVector` was really written with external use in mind, as it has various public functions which users probably shouldn't be calling. However it is required for `Connection::createVectorizedFunction` to work. I've only added `KUZU_API` to the bare minimum functions, but note that all inline functions (including `getData()` for direct access to the buffer) would be usable.

TODO: 
- ~~`InMemOverflowBuffer::allocateSpace` could be removed from the exports if the various `operation` functions are moved into cpp files.~~ Done; I threw them into `vector_string_functions.cpp`, though I'm not sure if that's ideal. They could have their own files, but the structure of `src/function` doesn't match `src/include/function` and I wasn't sure where they should go.


src/include/main/connection.h
```diff
@@ -80,7 +80,7 @@ class Connection {
      * @return the result of the query.
      */
     template<typename... Args>
-    KUZU_API inline std::unique_ptr<QueryResult> execute(
+    inline std::unique_ptr<QueryResult> execute(
```
ray6080, line 83: Should we remove `KUZU_API` from this function? As I remember, this is part of our cpp API.

ray6080 (approved):

src/include/main/connection.h
benjaminwinger: MSVC doesn't let you combine inline functions with `dllimport`/`dllexport` presumably because it doesn't actually do anything (they are defined inline in the header files, not in the library, so there's nothing to export/import). clang and gcc don't seem to care.

PR #2155, opened 2023-10-05
change print to sys.stdout.write and flush in benchmark_runner.py

russell-liu:


benchmark/benchmark_runner.py
```diff
@@ -242,7 +242,8 @@ def run_kuzu(serialized_graph_path):
         process = subprocess.Popen(
             tuple(benchmark_cmd), stdout=subprocess.PIPE)
         for line in iter(process.stdout.readline, b''):
-            print(line.decode("utf-8"), end='')
+            sys.stdout.write(line.decode("utf-8"))
+            sys.stdout.flush()
```
benjaminwinger, line 246: You can also use the [flush argument of print](https://docs.python.org/3/library/functions.html#print) to explicitly flush, which is a little more concise.

Riolku:
Why is it sufficient to just flush? I would think that we need a lock.

benchmark/benchmark_runner.py
russell-liu: good idea

russell-liu:
Any idea why MSVC CI is failing? I didn't expect my changes to have any effect on that.

mewim:
> Why is it sufficient to just flush? I would think that we need a lock.

The benchmark queries are executed sequentially, not in parallel, so it seems to be just a printing issue, not a race condition.

mewim:
> Any idea why MSVC CI is failing? I didn't expect my changes to have any effect on that.

The MSVC test might fail randomly. I have relaunched it.

mewim (approved):

PR #2161, opened 2023-10-06
Fix numNodes in CopyNode

ray6080:
There was a change (https://github.com/kuzudb/kuzu/pull/2070/files#r1348216726) to fix edge case for copying empty csv file, which doesn't fully fix the problem with `numNodes`.

### DIFF
--- a/src/processor/operator/persistent/copy_node.cpp
+++ b/src/processor/operator/persistent/copy_node.cpp
@@ -179,10 +179,9 @@ void CopyNode::checkNonNullConstraint(NullColumnChunk* nullChunk, offset_t numNo
 }
 
 void CopyNode::finalize(ExecutionContext* context) {
-    uint64_t numNodes = 0;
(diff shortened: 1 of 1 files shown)


Riolku:
Curious, what problems does the current code give rise to?

ray6080:
> Curious, what problems does the current code give rise to?

The case is that when `sharedState->sharedNodeGroup` is null, we would set `numNodes` to 0, which can be incorrect. `sharedState->sharedNodeGroup` is used only when threads have tuples left to write out. So it could be that each thread writes out all their tuples, while we still set `numNodes` to 0. Does this make sense?

Riolku:
Uh, is NodeGroup a fixed number of nodes stored together on disk? And so if all threads happen to write out a number of nodes that is a multiple of that number we get a totally incorrect statistic?

ray6080:
> Uh, is NodeGroup a fixed number of nodes stored together on disk? And so if all threads happen to write out a number of nodes that is a multiple of that number we get a totally incorrect statistic?

Yep.

andyfengHKU (approved):

PR #2172, opened 2023-10-09
Export map to py dict

ray6080:


mewim:
This does not seem to be a very user-friendly design. The exported dict is in the format of:
```
{
  "key": ["key1", "key2", ...],
  "value": ["val1", "val2", ...]
} 
```

I think the user is expecting a format of 
```
{
  "key1": "val1",
  "key2": "val2",
  ...
}
```

mewim (approved):

PR #2183, opened 2023-10-11
Fix #2174 for Node.js

### LINKED ISSUE #2174: Missing data type export from query result in APIs

mewim:
#2174

### DIFF
--- a/tools/nodejs_api/src_cpp/node_util.cpp
+++ b/tools/nodejs_api/src_cpp/node_util.cpp
@@ -137,6 +137,16 @@ Napi::Value Util::ConvertToNapiObject(const Value& value, Napi::Env env) {
     case LogicalTypeID::INTERNAL_ID: {
         return ConvertNodeIdToNapiObject(value.getValue<nodeID_t>(), env);
     }
+    case LogicalTypeID::MAP: {
+        Napi::Object napiObj = Napi::Object::New(env);
+        for (auto i = 0u; i < NestedVal::getChildrenSize(&value); ++i) {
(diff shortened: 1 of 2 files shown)


tools/nodejs_api/src_cpp/node_util.cpp
```diff
@@ -137,6 +137,16 @@ Napi::Value Util::ConvertToNapiObject(const Value& value, Napi::Env env) {
     case LogicalTypeID::INTERNAL_ID: {
         return ConvertNodeIdToNapiObject(value.getValue<nodeID_t>(), env);
     }
+    case LogicalTypeID::MAP: {
+        Napi::Object napiObj = Napi::Object::New(env);
+        for (auto i = 0u; i < NestedVal::getChildrenSize(&value); ++i) {
+            auto childVal = NestedVal::getChildVal(&value, i);
+            auto key = NestedVal::getChildVal(childVal, 0)->toString();
```
ray6080, line 144: Should we always convert key to string?

mewim: Since NAPI does not support Javascript `Map` type for now, we are just outputting plain Javascript `Object` for now (which is [commonly used by Javascript developers as a dictionary](https://stackoverflow.com/questions/18541940/map-vs-object-in-javascript)). The `Object` type only supports strings as keys.

ray6080 (approved):

PR #2201, opened 2023-10-12
finish cast string to map

AEsir777:
- fix nested type
- typo in union data chunk
- update: fix std::string_view{}
To fix:
- do I need to allow null for struct/map?
- {c:1, c: 2} overwrites the original value   (Acceptable)

|format|Old |New|
|-----|-----|--------|
|`MAP(MAP(INT64, INT64), MAP(STRING, STRING))` x 100000|no support|1768  |
| `MAP(INT64, STRING)`x 50000|   4215.386  | 432.61|


src/common/string_utils.cpp
```diff
@@ -5,6 +5,26 @@
...
 namespace common {
 
+std::vector<std::string> StringUtils::splitComma(
+    const std::string& input, bool ignoreEmptyStringParts) {
+    auto result = std::vector<std::string>();
+    auto currentPos = 0u;
+    auto lvl = 0u;
+    while (currentPos < input.length()) {
+        if (input[currentPos] == '(') {
+            lvl++;
```
andyfengHKU, line 15: Sometimes code coverage is not accurate. So just double check if this part is covered or not.

andyfengHKU (approved):

PR #2224, opened 2023-10-17
Disable buffering on windows when opening files where possible

benjaminwinger:


andyfengHKU (approved):
I don't know this well enough. Can u write some PR description for future reference?

benjaminwinger:
I'm going to close this for now. I'm not convinced it makes a significant difference and might have implications when running on hard-drives since it effectively disables readahead. Duckdb provides this as an option I think, which might be the better way, but it's not clear that it's a useful option at this point anyway so I'm not sure we should bother.

PR #2228, opened 2023-10-17
Reduce number of test query threads

benjaminwinger:
Seems to significantly slow down our tests as they work with many small operations on systems with many processors. Particularly when running many tests in parallel, it seems unnecessary to use the default number of threads.

~~Testing to see what the speedup is for CI~~ Seems like an improvement compared to the last time CI was this quiet; oddly windows doesn't seem any different from the previous run, but that machine doesn't have as many threads, and is running much faster since I stopped the antivirus from scanning the test database files anyway.


mewim (approved):

mewim:
> is running much faster since I stopped the antivirus from scanning the test database files anyway

LOL. I did not expect this to be the root cause of the slowdown.

PR #2229, opened 2023-10-18
Add C, Python, Node.js, and Java API bindings for read-only mode

mewim:


tools/python_api/src_cpp/py_query_result.cpp
```diff
@@ -275,6 +276,10 @@ bool PyQueryResult::isSuccess() const {
     return queryResult->isSuccess();
 }
 
+std::string PyQueryResult::getErrorMessage() const {
+    return queryResult->getErrorMessage();
```
ray6080, line 280: Can we add a test for this function?

ray6080 (approved):

tools/python_api/src_cpp/py_query_result.cpp
mewim: I have tested this in Python: https://github.com/kuzudb/kuzu/blob/98b5c84e06b33f0aea13be5d13dc0479c0854cfe/tools/python_api/test/test_exception.py#L37-L40

I think it is not caught by Codecov

Issue #2137, opened 2023-10-03
Buffer Manager Eviction Rework


ray6080:
## Overview
Currently we adopt a queue-based eviction policy. A concurrent queue is used to keep track of all marked pages as eviction candidates. This poses overheads to `unpin`, as each call of this function needs to switch unlocked pages to marked, and add it to the eviction queue. 

The alternative solution described in [Virtual-Memory Assisted Buffer Management](https://dl.acm.org/doi/abs/10.1145/3588687)[^1] is to store all cached page identifiers in a fixed-size open addressing hash table. The size of the hash table is equal to the number of pages in DRAM (bounded by buffer manager memory limit). Instead of accessing eviction queue in each `unpin`, hash table is not accessed during cache hits and unpin, but only during page faults and eviction.
Moreover, the hash table makes it easier to loop over, and clean up page states from BM as we sometimes need to ` removeFilePagesFromFrames`.

This issue proposes to replace the concurrent queue with a fixed-size hash table.

## Page State and Eviction
The design of our buffer manager follows *vmCache*. See Section 3 in paper [Virtual-Memory Assisted Buffer Management](https://dl.acm.org/doi/abs/10.1145/3588687) for detailed description.

### Page State
Each page is associated with a 64-bit state entry. The entry encodes both state and version counter. 56 bits are used for the version counter. and the left 8 bits are used for state, including LOCKED, UNLOCKED, MARKED, EVICTED. Any page must be in one of these four states. The version counter is used for optimistic read. See our [code comment](https://github.com/kuzudb/kuzu/blob/4aa1d07812665b5b1067e065d40359ad7f037118/src/include/storage/buffer_manager/buffer_manager.h#L71-L162) here for more details .

### Eviction Algorithm
This issue proposes to implement hash table-based CLOCK algorithm.
The key idea is to find an old page to evict, not necessarily the oldest. We loop through cached pages in the hash table until we find one that hasn't been referenced since the last time we check it.

Eviction happens at batches of 64 pages at a time, and is performed in five steps:
1. loop over the hash table, and get a batch of marked candidates from hash table.
2. write dirty pages (optionally, this can be done with async io).
3. try to lock clean page candidates.
4. remove locked pages from page table using `madvise`.
5. remove locked pages from eviction hash table, unlock them.

Note: Currently, we already ensure dirty pages are locked. No need to lock and upgrade dirty pages as described in `vmcache`.

[...]

PR #2231, opened 2023-10-18
ci: use full ASAN

Riolku:
Previously we used LD_PRELOAD to add ASAN to our build. This is at best
imperfect, since the full benefits of ASAN can only be realized by
adding compiler instrumentation.

This does not run the python, nodejs, or java tests with ASAN. Though
Python can be run with ASAN using LD_PRELOAD, it is super slow, and
would need parallelized testing to be viable (which is not worth it
IMO).

Nodejs and Java just die spectacularly. Overall I think the APIs need
ASAN much less.


Riolku:
The lock-free queue is the source of an ASAN failure. Specifically, the
queue does not guarantee strict FIFO circumstances in the name of speed.
This means that our code that loops through the queue is incorrect.

The goal is to replace this queue-based design soon regardless, and
hence we don't want to spend more time on this.

Also, this commit removes a useless logger in the buffer manager.

Makefile
```diff
@@ -113,24 +113,20 @@ clang-tidy-ci:
 	$(call mkdirp,build/release) && cd build/release && \
 	cmake $(GENERATOR) $(FORCE_COLOR) $(SANITIZER_FLAG) -DCMAKE_EXPORT_COMPILE_COMMANDS=1 ../..
 
-pytest:
-	$(MAKE) release
+pytest: release
```
mewim, line 116: If we make this change, will it still rerun `make release` every time  when`make pytest` is run? Since `release` does not have file dependencies configured and is simply a shortcut to invoke CMake, I am not sure if it will rebuild automatically if the code is changed?

mewim (approved):
The CI changes LGTM

Makefile
Riolku: release is a dummy target, so Make will run it everytime pytest is run

ray6080 (approved):
The change to queue is indeed decreasing the performance of BM, as each `unpin` now gets to lock the queue. This should be fixed with issue #2137 , so we move away from the queue-based solution. As for now, we decide to merge this first to get CI asan correctly, and fix it later before the next release.

PR #2241, opened 2023-10-19
ci: use env vars for parallelism

Riolku:
This commit also configures TEST_JOBS to be half of NUM_THREADS, since our tests use 2 threads right now.


Riolku:
Further considerations: should we overcommit TEST_JOBS ?

mewim (approved):

PR #2254, opened 2023-10-24
int128 for apis

Ashleyhx:


Riolku:
Rust is failing because i128 is not a type that can be converted natively to a C++ type AFAICT. We need some sort of bridge to the equivalent C++ type. @Ashleyhx do you want me to look into this bridging?

tools/rust_api/src/value.rs
```diff
@@ -393,6 +417,13 @@ impl TryFrom<&ffi::Value> for Value {
             LogicalTypeID::UINT16 => Ok(Value::UInt16(value.get_value_u16())),
             LogicalTypeID::UINT32 => Ok(Value::UInt32(value.get_value_u32())),
             LogicalTypeID::UINT64 => Ok(Value::UInt64(value.get_value_u64())),
+            LogicalTypeID::INT128 => {
+                let int128_val = ffi::value_get_int128_t(value);
+                Ok(Value::Int128(Int128 {
+                    low: int128_val[0],
+                    high: (int128_val[1] >> 64) as i64,
```
benjaminwinger, line 424: Why is this being shifted? I would think it would make the high part always 0.
I think you could use an `i128` on the rust side, you just need to construct it from the two `u64`s.

I guess there isn't really any builtin functions that will work, particularly with the difference in signedness, but [this code](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=dabb3700b4c0b5561957cc2c5033989e) seems to work

benjaminwinger (approved):

Issue #2198, opened 2023-10-12
Copy from partial columns feature


russell-liu:
Feature:
- Semantics: Allow `COPY FROM` commands to specify specific columns to be copied. Must have: columns in file = columns in copy query ⊆ columns in schema. Columns in schema that are not in copy query are set to NULL.
- Syntax: `COPY schema_name [ ( column_name [, ...] ) ] FROM ...` (the rest is unchanged).

Changes required:
- Parser:
- [x] Change Cypher grammar and regenerate grammar files
- [x] Change transformer to store specified columns in Statement object
- Binder:
- [x] Verify that specified columns are a subset of all columns of specified schema
- [x] Create BoundFileScanInfo with only specified columns instead of all columns of specified schema
- [x] Create BoundCopyFromInfo with columns that need to be set to NULL
- [ ] Implement changes for both node and rel tables
- Planner:
- [x] Insert columns to be set to NULL separately into group and scope
- Processor:
- [x] Set data column positions for columns to be set to NULL
- [x] Change CopyNode physical operator to set required columns to NULL
- Testing:
- [x] Add lots of test cases to cover edge case behavior, including errors

PR #2260, opened 2023-10-24
Copy from partial columns

russell-liu:
#2198


andyfengHKU (changes requested):
The design of `partial column copy` needs to be revised.

If we take a look at the current design. FileScan/Reader will scans a set of columns and put into vectors. CopyNode/Rel will read from these vectors assuming they have one-to-one mapping with the columns being defined in the schema/

This assumption needs to be removed because `partial column copy` no longer guarantees one-to-one mapping. Columns from file can be mapped to any column in schema. And column in schema may not be mapped to a column from file (in which case it should be filled with NULL).

To achieve this, you need to differentiate columns in `FileScanInfo` and `CopyFromInfo`. Columns in `FileScanInfo` (denoted as f_columns below) should be determined only based on input file. Columns in `CopyFromInfo` (denoted as c_columns below) should be determined based on user input. c_columns should always have the same size as described in corresponding schema. Each entry of c_columns either maps to an entry of f_columns or to a NULL column. Copy operator can thus be generated purely based on c_columns.

PR #2263, opened 2023-10-25
Add graph pattern label pruning

andyfengHKU:
This PR replace our previous naive rel pattern label pruning with a new pruning class that prune both node and relationship labels based on topology. 

### Primary use case
In RDF processing, since a join can only happen on resource node, the following query
```
MATCH (a)-[e1]->(b)-[e2]->(c)
```
can be pruned as
```
MATCH (a:resource)-[e1:resource_triple]->(b:resource)-[e2:resource_triple|literal_triple]->(c:resource|literal)
```
The join over `e1,b` will benefit from pruning.

And of course general multi label query will also benefit from this PR.

### Minutia
Technically speaking, label pruning should be performed iteratively until the algorithm converge. Though I don't think it make much difference in practice, plus we want the compilation to be fast. So the current implementation only perform pruning once over each node and rel.

### DIFF
--- a/src/binder/query/query_graph_label_analyzer.cpp
+++ b/src/binder/query/query_graph_label_analyzer.cpp
@@ -1,0 +1,124 @@
+#include "binder/query/query_graph_label_analyzer.h"
+
+#include "catalog/rel_table_schema.h"
+#include "common/exception/binder.h"
+#include "common/string_format.h"
+
+using namespace kuzu::common;
+using namespace kuzu::catalog;
+
+namespace kuzu {
+namespace binder {
+
+void QueryGraphLabelAnalyzer::pruneLabel(const QueryGraph& graph) {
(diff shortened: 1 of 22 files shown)


src/binder/query/query_graph_label_analyzer.cpp
```diff
@@ -0,0 +1,129 @@
...
+}
+
+void QueryGraphLabelAnalyzer::pruneNode(const QueryGraph& graph, NodeExpression& node) {
+    for (auto i = 0u; i < graph.getNumQueryRels(); ++i) {
+        auto queryRel = graph.getQueryRel(i);
+        if (queryRel->isRecursive()) {
+            continue;
+        }
+        common::table_id_set_t candidates;
+        if (*queryRel->getSrcNode() == node) {
```
ray6080, line 29: L38-L43 and L45-L58 are common logic, can we wrap both into a single function?

ray6080 (approved):

PR #2264, opened 2023-10-25
Switch Linux and macOS CI runners to use Ninja build

mewim:


Riolku:
We should probably use gmake in at least one of the jobs. I don't think it particularly matters which, but we should test that gmake still works.

Riolku:
Benefit of doing it in GCC is build with ASAN and benchmark runs sooner. Benefit of doing it in clang means any clang tidy errors are shown earlier.

mewim:
I did not switch the benchmark to use ninja so gmake is being tested in benchmark. Is it sufficient or do you prefer to use gmake in a test?

Riolku:
I think I'd rather have the benchmark use ninja and clang use gmake, because that should give us the fastest overall CI speeds.

mewim:
OK I will redeploy the benchmark runners some time and update this...

andyfengHKU (approved):

mewim:
@Riolku I have updated and redeployed the benchmark runners to use Ninja build and switched clang build job back to using gmake. I also removed several unneeded dependencies from the benchmark runner.

.github/workflows/ci-workflow.yml
```diff
@@ -255,19 +257,13 @@ jobs:
   benchmark:
     name: benchmark
     needs: [gcc-build-test, clang-build-test]
+    env:
+      NUM_THREADS: 30
```
Riolku, line 261: If you use an env var, remove the parameter to make on the later line.

Riolku (approved):

benchmark/Dockerfile
```diff
@@ -3,9 +3,20 @@ FROM ubuntu:22.04
...
+    curl \
+    git \
+    jq \
+    libssl-dev \
+    libcurl4-openssl-dev \
+    ninja-build \
+    python3-dev \
+    python3-pip \
+    python-is-python3
+
```
Riolku, line 19: It sort of looks like we forgot to install npm... https://github.com/kuzudb/kuzu/actions/runs/6655831578/job/18087424945?pr=2273 I have a failure here.

Riolku (changes requested):

benchmark/Dockerfile
Riolku: Nevermind, that's because benchmark no longer needs npm. We should merge this soon then so CI isn't broken for too long.

Riolku (approved):

PR #2273, opened 2023-10-26
tidy: add makefile command

Riolku:
Instead of embedding the clang-tidy command in the workflow file, we should put it in a make subcommand.

Also, this change removes the `build/clangd` generator, since when `make clangd` is run but the `clangd` target is not built, clang-tidy gives false-positives about the java api. Instead, `make clangd` now simply enables exporting compile commands for the `build/release` directory.


Riolku:
The command is `make tidy`.

acquamarin (approved):

PR #2280, opened 2023-10-27
Update LLVM toolchain to version 17 on hosted-runners

mewim:
and also move small jobs to GitHub-hosted runners

### DIFF
--- a/.github/workflows/ci-workflow.yml
+++ b/.github/workflows/ci-workflow.yml
@@ -244,7 +244,7 @@ jobs:
     name: rustfmt check
     runs-on: ubuntu-22.04
     steps:
-      - name: Update rust
+      - name: Update Rust
         run: rustup update
 
       - uses: actions/checkout@v3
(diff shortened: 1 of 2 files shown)


.github/workflows/ci-workflow.yml
```diff
@@ -225,8 +225,13 @@ jobs:
 
   clang-formatting-check:
     name: clang-format check
-    runs-on: kuzu-self-hosted-testing
+    runs-on: ubuntu-22.04
     steps:
+      - name: Install clang-format
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y clang-format-11
```
Riolku, line 233: Are we delaying clang format 17 upgrade?

Riolku (approved):
Maybe mentiom in the commit that youre also moving to some GH runners.

.github/workflows/ci-workflow.yml
mewim: clang-format is not backward compatible. If we use clang-format-17 we need to also edit `.clang-format` to achieve the same format. Since I am not familiar with clang-format, I did not handle it in this PR.

PR #2288, opened 2023-10-27
replace std::assert with KU_ASSERT

### LINKED ISSUE #2283: Replace std::assert with KU_ASSERT
We currently have `KU_ASSERT` in our codebase, as well as `std::assert`. We should switch to only using the former, and provide a make flag to force asserts on for debugging.

Riolku:
This also enables a parameter RUNTIME_CHECKS that can be passed to the Makefile to force these checks to be on. This also enables a CI check to make sure we consistently use KU_ASSERT and not std::assert.

Closes #2283.


ray6080 (approved):
Thanks! LGTM.

Issue #2294, opened 2023-10-28
OPTIONAL MATCH should return `NULL` instead of  dummy/pseudo nodes and relations


sapalli2989:
In Neo4j
```
OPTIONAL MATCH (a)-[b:KNOWS]->(c) return a,b,c;
```
returns `NULL` values for non-existent nodes and relationships (amongst other metadata)
```
    "_fields": [
      null,
      null,
      null
    ],
```

With Kuzu statements
```
CREATE NODE TABLE T(id INT64, PRIMARY KEY(id));
CREATE REL TABLE KNOWS(FROM T TO T);
OPTIONAL MATCH (a)-[b:KNOWS]->(c) return a,b,c;
```
, the API returns
```
[
    [
        {
            "_label": "T",
            "_id": {
                "offset": 18446744073709551615,
                "table": 18446744073709551615
            },
            "id": null
        },
        {
            "_src": {
                "offset": 18446744073709551615,
                "table": 18446744073709551615
            },
            "_dst": {
                "offset": 18446744073709551615,
                "table": 18446744073709551615
            },
            "_label": "KNOWS"
        },
        {
            "_label": "T",
            "_id": {
                "offset": 18446744073709551615,
                "table": 18446744073709551615
            },
            "id": null
        }
    ]
]
```
, which seems too bloated. What about using `NULL` to represent non-existence of entities, instead of returning pseudo-entities with `id: null` and computed internal id?

Also in CLI the output is misleading:
```
kuzu> OPTIONAL MATCH (a)-[b:KNOWS]->(c) return a,b,c;
----------------------------------------------------------
| a             | b                      | c             |
----------------------------------------------------------
| {, _LABEL: T} | ()-{_LABEL: KNOWS}->() | {, _LABEL: T} |
----------------------------------------------------------
(1 tuple)
```
Better would be:
```
----------------------------------------------------------
| a             | b                      | c             |
----------------------------------------------------------
| NULL | NULL | NULL |
----------------------------------------------------------
```

andyfengHKU:
Fixed in #2299

PR #2299, opened 2023-10-30
Fix issue-2294

andyfengHKU:
This PR fixes issue #2294.

#### TODO

- [x] @mewim should remove marked API from C, Python, Node.js, and Java bindings.
- [x] @benjaminwinger should remove marked API from Rust binding.


mewim:
I have removed the deprecated functions from C, Java, Node.js, and Python bindings. I do not know Rust, so @benjaminwinger should take over and remove the APIs from Rust as well.

mewim (approved):

acquamarin (approved):

PR #2317, opened 2023-10-31
Remove CompressionAlg::getValue

benjaminwinger:
It wasn't being used since it's functionally equivalent to decompressFromPage for a single value.

### DIFF
--- a/test/storage/compression_test.cpp
+++ b/test/storage/compression_test.cpp
@@ -33,7 +33,8 @@ void test_compression(CompressionAlg& alg, std::vector<T> src) {
     EXPECT_EQ(decompressed[1], value);
 
     for (int i = 0; i < src.size(); i++) {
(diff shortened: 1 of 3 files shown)


test/storage/compression_test.cpp
```diff
@@ -31,7 +31,7 @@ void test_compression(CompressionAlg& alg, std::vector<T> src) {
     EXPECT_EQ(decompressed[1], value);
 
     for (int i = 0; i < src.size(); i++) {
-        alg.getValue(dest.data(), i, (uint8_t*)decompressed.data(), i, metadata);
+        alg.decompressFromPage(dest.data(), i, (uint8_t*)decompressed.data(), i, 1, metadata);
```
ray6080, line 34: Can u add an inline comment for `1`? Also in L154.

ray6080 (approved):

PR #2335, opened 2023-11-02
build: enable Werror on non-Windows

Riolku:
Because of our changes recently, we can now build the project with -Werror. However, this doesn't enable all warnings. We should add warnings to the list of flagged warnings so as to forbid more and more bad code.

Windows is simply too noisy to enable -Werror for right now.


CMakeLists.txt
```diff
@@ -6,13 +6,18 @@ find_package(Threads REQUIRED)
...
-set(CMAKE_CXX_STANDARD_REQUIRED True)
+set(CMAKE_CXX_STANDARD_REQUIRED TRUE)
 set(CMAKE_POSITION_INDEPENDENT_CODE ON)
 if(NOT APPLE)
     set(CMAKE_CXX_VISIBILITY_PRESET hidden)
     set(CMAKE_VISIBILITY_INLINES_HIDDEN ON)
 endif()
 
+option(ENABLE_WERROR "Treat all warnings as errors" FALSE)
+if(NOT MSVC AND ENABLE_WERROR)
```
benjaminwinger, line 17: Might it make more sense to leave this option available on windows, and just explicitly disable it in the windows CI Job? we might want to manually test enabling this and going through all the errors at some point.

Riolku: Yeah, that makes way more sense.

ray6080 (approved):

PR #2338, opened 2023-11-02
Fix failing tests in LDBC IC and LSQB due to node group changes

ray6080:


benjaminwinger:
This is missing the lines in column_chunk.cpp [here](https://github.com/kuzudb/kuzu/blob/7e8c9c6103832419a87d7568181d0a5ef440c3cc/src/storage/store/column_chunk.cpp#L98) and [here](https://github.com/kuzudb/kuzu/blob/7e8c9c6103832419a87d7568181d0a5ef440c3cc/src/storage/store/column_chunk.cpp#L160).

ray6080:
> This is missing the lines in column_chunk.cpp [here](https://github.com/kuzudb/kuzu/blob/7e8c9c6103832419a87d7568181d0a5ef440c3cc/src/storage/store/column_chunk.cpp#L98) and [here](https://github.com/kuzudb/kuzu/blob/7e8c9c6103832419a87d7568181d0a5ef440c3cc/src/storage/store/column_chunk.cpp#L160).
> 
> Edit: Actually, it looks like enableCompression is hardcoded to false for internal IDs in `createColumnChunk`, so maybe this isn't the issue. I tried running it a few more times and got it to fail again.

Yep. I realized that too last night. The problem lies in scan multi label, it fails only when the multiplicity is x_to_one. btw I think we should refactor the code a bit to reduce switch cases, which is not easy to follow.

acquamarin (approved):

PR #2340, opened 2023-11-03
cleanup: fix all clangd warnings

Riolku:
The bulk of the warnings issued are because of unused includes, but some are because of dead files leftover from #2246. Also, `subplans_table.h` had a `using` directive, which caused symbol pollution.


Riolku:
Some numbers on compilation time. Comparing with master, this branch is 16% faster to compile. For more compilation speed, we need to look into splitting up our bigger headers. First though, we should ensure that we have checks in place so that these kinds of unused headers can't re-enter the codebase. This work will be the subject of a future PR.

PR #2341, opened 2023-11-03
Add typed-literal storage

andyfengHKU:
This PR change RDF literal type from `STRING` to `RDF_VARIANT`.

An `RDF_VARIANT` is a physical `STRUCT{UINT8, BLOB}` where `UINT8` stores the type and `BLOB` stores the actual value. This should be a more performant storage design than `STRING`.

In this PR we start supporting XSD integer, double, decimal, bool and date.


src/include/common/keyword/rdf_keyword.h
ray6080, line 1: Do we have other keywords? It seems like you are adding a bunch of constants here. Consider creating a new directory `constants` instead of `keyword`.

src/include/common/keyword/rdf_keyword.h
```diff
@@ -0,0 +1,28 @@
+#pragma once
+
+#include <string>
+
+namespace kuzu {
+namespace common {
+namespace rdf::keyword {
+
+// static constexpr std::string_view ANONYMOUS = "anonymous";
```
ray6080, line 9: Remove `ANONYMOUS`.

src/include/common/keyword/rdf_keyword.h
```diff
@@ -0,0 +1,28 @@
+#pragma once
+
+#include <string>
+
+namespace kuzu {
+namespace common {
+namespace rdf::keyword {
```
ray6080, line 7: The nested namespace is not a common practice in the codebase. Why are we making these keyword constants special that they need to be under four-level nested namespaces? I don't have a clue on this.

src/include/function/cast/functions/cast_rdf_variant.h
ray6080, line 1: Remove this file.

src/processor/operator/persistent/reader/rdf/rdf_reader.cpp
```diff
@@ -76,10 +70,75 @@ static bool supportSerdType(SerdType type) {
...
+    auto data = reinterpret_cast<const char*>(literalNode->buf);
+    auto length = literalNode->n_bytes;
+    if (typeNode == nullptr) { // No type information available. Treat as string.
+        addVariant(vector, pos, data, length);
+        return;
+    }
+    auto typeData = reinterpret_cast<const char*>(typeNode->buf);
+    auto typeLength = typeNode->n_bytes;
+    bool resolveAsString = true;
+    if (memcmp(XSD.data(), typeData, XSD.size()) == 0) { // Match XSD prefix.
```
ray6080, line 104: Can you create a string view over typeData? That avoids the use of `memcmp` and should make the code more intuitive.

src/processor/operator/persistent/reader/rdf/rdf_reader.cpp
```diff
@@ -76,10 +70,75 @@ static bool supportSerdType(SerdType type) {
...
+        typeData += XSD.size();
+        typeLength -= XSD.size();
+        if (memcmp(XSD_integer.data(), typeData, XSD_integer.size()) == 0) {
+            // XSD:integer
+            int64_t result;
+            if (function::trySimpleIntegerCast<int64_t>(data, length, result)) {
```
ray6080, line 110: If the casting fails, should we ignore and fallback to string? or should we error? This question arises due to that the type of the data is already given in the file, therefore, if silently falling back to string is correct for the data remains unclear to me.

ray6080 (approved):

src/binder/bind/bind_copy.cpp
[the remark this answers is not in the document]
andyfengHKU: I would do this in a separate PR.

src/include/common/keyword/rdf_keyword.h
andyfengHKU: Again let's do this separately, I'm against having one giant const.h file.

src/processor/operator/persistent/reader/rdf/rdf_reader.cpp
andyfengHKU: This is a good point. I will revisit this issue before we deliver RDF.

PR #2344, opened 2023-11-03
ci: check for clangd diagnostics

Riolku:
This is a followup from #2340, which adds a check to CI for clangd diagnostics. It is added to GCC because clang is the slower of the two linux runners right now.

The script itself spawns clangd and speaks Language Server Protocol to it. We open files, and clangd returns diagnostics (if any). This enables a fairly straightforward way of checking an arbitrary number of source files for warnings/errors.

Simply use `make clangd-diagnostics` to run the tool. The variables `CLANGD_DIAGNOSTIC_INSTANCES` and `CLANGD_DIAGNOSTIC_JOBS` control the parallelism.

### DIFF
--- a/scripts/dockerized-ci-tests-runner/Dockerfile
+++ b/scripts/dockerized-ci-tests-runner/Dockerfile
@@ -36,7 +36,8 @@ RUN wget https://apt.llvm.org/llvm.sh && \
     rm llvm.sh && \
     ln /usr/bin/clang++-17 /usr/bin/clang++ && \
     ln /usr/bin/clang-17 /usr/bin/clang && \
-    ln /usr/bin/run-clang-tidy-17 /usr/bin/run-clang-tidy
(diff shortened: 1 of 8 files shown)


scripts/dockerized-ci-tests-runner/Dockerfile
```diff
@@ -36,7 +36,8 @@ RUN wget https://apt.llvm.org/llvm.sh && \
     rm llvm.sh && \
     ln /usr/bin/clang++-17 /usr/bin/clang++ && \
     ln /usr/bin/clang-17 /usr/bin/clang && \
-    ln /usr/bin/run-clang-tidy-17 /usr/bin/run-clang-tidy
+    ln /usr/bin/run-clang-tidy-17 /usr/bin/run-clang-tidy && \
```
mewim, line 39: This was not working due to missing the `&& \`. I added it.

mewim (approved):
I made a small fix to the Dockerfile and redeployed all the Linux runners

PR #2357, opened 2023-11-06
Replace AccessMode in API with boolean flag read_only

ray6080:
Exposing customized num in API is not user-friendly. Instead, native boolean flag is easier to use from users' perspective. Thus, this PR proposes to change `AccessMode accessMode` to `bool readOnly` in various language bindings in similar fashion.

TODO:
- [ ] Update our API doc on access mode.


mewim:
API changed LGTM

mewim (approved):

PR #2361, opened 2023-11-06
common/assert: make KU_ASSERT an expression

Riolku:
This is compliant with the behaviour of std::assert, and avoids warnings about redundant semicolons and potentionally some about duplicate parantheses.


ray6080:
Close this as it is included in #2348.

PR #2375, opened 2023-11-08
common: add lcov exclusions in KU_UNREACHABLE

Riolku:
This allows us to avoid the annoying comments everywhere we have unreachable code in the codebase.


ray6080 (approved):
Nice one. Thanks!

PR #2401, opened 2023-11-13
cleanup: replace NotImplementedException

Riolku:
KU_UNREACHABLE should be used in almost all cases where we have NotImplementedException.


Riolku:
Also cleaned up an instance where invalid `LIST_PRODUCT` returned `NotImplementedException` instead of `BinderException`.

PR #2433, opened 2023-11-16
prepared_statement: move parameters

Riolku:
In the Java API, we had a bug where we take ownership of and free parameters passed into executeWithParams.

Inspecting the method itself, it was taking a shared_ptr, but then performing a deep copy, which is nonsense. Instead, we should take a unique_ptr, since we need to copy the parameters to guarantee that they are not modified for the duration of the query.

This commit also fixes three other issues. First, the Java tests weren't running any tests from ConnectionTest.java, which is why we didn't observe this bug. Additionally, the constructor of KuzuConnection uses an assertion, but assertions are disabled by default, which causes our tests to fail (and if the assertion is skipped, we segfault).

Also, since https://github.com/rust-lang/cc-rs/issues/900 has been fixed, we can remove the version pinning of `cc` on MacOS.

### DIFF
--- a/.github/workflows/ci-workflow.yml
+++ b/.github/workflows/ci-workflow.yml
@@ -359,7 +359,6 @@ jobs:
         run: |
           ulimit -n 10240
           source /Users/runner/.cargo/env
-          cargo update -p cc --precise '1.0.83'
           make rusttest
 
       - name: Rust example
--- a/Makefile
+++ b/Makefile
@@ -128,11 +128,11 @@ nodejstest: nodejs
 javatest: java
 ifeq ($(OS),Windows_NT)
 	$(call mkdirp,tools/java_api/build/test)  && cd tools/java_api/ && \
-	javac -d build/test -cp ".;build/kuzu_java.jar;third_party/junit-platform-console-standalone-1.9.3.jar"  -sourcepath src/test/java/com/kuzudb/test/*.java && \
+	javac -d build/test -cp ".;build/kuzu_java.jar;third_party/junit-platform-console-standalone-1.9.3.jar" src/test/java/com/kuzudb/test/*.java && \
 	java -jar third_party/junit-platform-console-standalone-1.9.3.jar -cp ".;build/kuzu_java.jar;build/test/" --scan-classpath --include-package=com.kuzudb.java_test --details=verbose
 else
 	$(call mkdirp,tools/java_api/build/test)  && cd tools/java_api/ && \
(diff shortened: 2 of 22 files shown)


tools/java_api/src/main/java/com/kuzudb/KuzuValue.java
```diff
@@ -74,6 +74,7 @@ public void destroy() throws KuzuObjectRefDestroyedException {
             KuzuNative.kuzu_value_destroy(this);
             destroyed = true;
         }
+        System.gc();
```
andyfengHKU, line 77: Leave a comment if we decide to keep this.

andyfengHKU (approved):
LGTM

tools/java_api/src/main/java/com/kuzudb/KuzuValue.java
Riolku: Forgot about it. It's covered well by my other change so I'll delete this in the morning.

Riolku:
Should I add a test for the untested exception in the nodejs api?

andyfengHKU:
> Should I add a test for the untested exception in the nodejs api?

Up to you. I thought it should be KU_UNREACHABLE though

mewim (approved):
API changes LGTM

.github/workflows/ci-workflow.yml
```diff
@@ -359,7 +359,6 @@ jobs:
         run: |
           ulimit -n 10240
           source /Users/runner/.cargo/env
-          cargo update -p cc --precise '1.0.83'
```
mewim, line 362: I thought this was required? Could @benjaminwinger comment?

Riolku: Because of https://github.com/rust-lang/cc-rs/issues/900 being fixed, this is no longer required. A note on this is also in the commit message for logging.

Riolku:
> > Should I add a test for the untested exception in the nodejs api?
> 
> Up to you. I thought it should be KU_UNREACHABLE though

Well since we have datatypes that we don't support in NodeJS it's not unreachable :smile: .

It should probably be marked UNREACHABLE in the future when we have better datatype support. I'm leaving it as is for now.

.github/workflows/ci-workflow.yml
mewim: I see. Sorry I did not read it carefully.

Riolku: No worries :)

PR #2453, opened 2023-11-19
Detach delete

ray6080:


src/include/planner/operator/persistent/logical_delete.h
```diff
@@ -6,30 +6,51 @@
 namespace kuzu {
 namespace planner {
 
+struct LogicalDeleteNodeInfo {
+    std::shared_ptr<binder::NodeExpression> node;
+    bool isDetachDelete;
```
andyfengHKU, line 11: Consider use a DeleteType here

andyfengHKU (approved):

PR #2455, opened 2023-11-19
cmake: strip main binaries

Riolku:
Instead of the very complex #2443, this PR is a comparably simple change that strips `libkuzu.so`, `kuzu_shell` and `libkuzu.a`.


mewim (approved):
Tested the generated binaries on Ubuntu 22.04, Fedora 38 and macOS 14 with ARM64 arch and all works fine.

PR #2482, opened 2023-11-22
Add LTO Support

### LINKED ISSUE #2036: Unity builds
Unity builds (where we amalgamate source files before compiling) can offer compilation and runtime speedups. CMake has [builtin support](https://cmake.org/cmake/help/latest/prop_tgt/UNITY_BUILD.html) for it.

Tasks:
- [ ] Ensure every header has `#pragma once` (this is not currently the case).
- [ ] Enable unity builds and test them in CI.
- [ ] Compare unity builds against normal ones and decide which to use for releases.

As for combining all the source files into one, this is generally not recommended, according to the [CMake Documentation](https://cmake.org/cmake/help/latest/prop_tgt/UNITY_BUILD_BATCH_SIZE.html), though it is possible. It is something we could consider trying out, though the memory usage promises to be extremely high. In general, we could play around with batch sizes and see what runtime gains there are to be had.

Riolku:
Link-Time Optimization is a technique whereby the linker performs
    optimizations at the final step, right before generating the final
    executable or shared library. This provides significant performance and
    binary size optimizations, since the linker has access to all the
    context needed, unlike the individual optimizer passes, but in exchange
    linking takes a lot of time and memory.
    
    Based on some reading and some experimentation, Unity Builds, though
    they can offer similar benefits, are not nearly as good for
    optimization, and instead are primarily used for speeding up compile
    times. Additionally, Unity builds are harder to get right and harder to
    debug. LTO on the other hand is a very simple change, and seems to be
    the main area of focus right now.
    
    Closes #2036.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,12 +4,12 @@ project(Kuzu VERSION 0.1.0 LANGUAGES CXX C)
 
 find_package(Threads REQUIRED)
 
-set(CMAKE_FIND_PACKAGE_RESOLVE_SYMLINKS TRUE)
 set(CMAKE_CXX_STANDARD 20)
 set(CMAKE_CXX_STANDARD_REQUIRED TRUE)
-set(CMAKE_POSITION_INDEPENDENT_CODE ON)
 set(CMAKE_CXX_VISIBILITY_PRESET hidden)
 set(CMAKE_C_VISIBILITY_PRESET hidden)
+set(CMAKE_FIND_PACKAGE_RESOLVE_SYMLINKS TRUE)
+set(CMAKE_POSITION_INDEPENDENT_CODE ON)
 set(CMAKE_VISIBILITY_INLINES_HIDDEN ON)
 
 option(ENABLE_WERROR "Treat all warnings as errors" FALSE)
@@ -76,6 +76,7 @@ option(ENABLE_ADDRESS_SANITIZER "Enable address sanitizer." FALSE)
 option(ENABLE_THREAD_SANITIZER "Enable thread sanitizer." FALSE)
 option(ENABLE_UBSAN "Enable undefined behavior sanitizer." FALSE)
 option(ENABLE_RUNTIME_CHECKS "Enable runtime coherency checks (e.g. asserts)" FALSE)
+option(ENABLE_LTO "Enable Link-Time Optimization" FALSE)
 if(MSVC)
     # Required for M_PI on Windows
     add_compile_definitions(_USE_MATH_DEFINES)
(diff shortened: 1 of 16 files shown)


ray6080:
> This provides significant performance and binary size optimizations, since the linker has access to all the context needed.

I'm curious into the performance optimization part. Do we have noticeable performance boosts for some queries? Worth benchmarking a bit if you're sure about this, so we understand the insight here better.

Riolku:
I ran the LTO and non-LTO versions of this branch and timed running our testsuite (i.e., `time ctest -j 60`):

```
LTO: 21.01, 21.59, 21.82, 21.97, 27.3
Non-LTO: 24.09, 24.56, 25.66, 26.39, 27.84
```

Dropping the smallest and largest runtime of both and taking the average, I get:

```
LTO: 21.79
Non-LTO: 25.54
```

Loading LDBC100GB Comment Table and running the following query:

```
MATCH(comment:Comment) WHERE comment.ID < 33980465466560 RETURN comment.ID as ID, count(comment.length) order by ID LIMIT 5;
```

I get:

```
LTO: 10.469s, 10.567s, 11.054s, 11.562s, 12.656s 

Non-LTO:  10.697s, 11.067s, 11.246s, 11.380s, 11.620s
```

Dropping the minimum and maximum and taking the average:

```
LTO: 11.061s
Non-LTO: 11.231s
```

LTO is very good at optimizing tedious things, I think. It's very good at inlining when it otherwise couldn't and optimizing calls that would be across translation units, which explains the major performance increase over the testsuite.

However, LTO will provide little benefit on the critical sections of our large queries, since that code is already in the same translation unit.

Finally, since LTO can help us optimize across translation units (source files), moving code from our headers into our source files should mean we pay little-to-no runtime penalty, which is good overall for code organization.

CMakeLists.txt
```diff
@@ -4,12 +4,12 @@ project(Kuzu VERSION 0.1.0 LANGUAGES CXX C)
...
 find_package(Threads REQUIRED)
 
-set(CMAKE_FIND_PACKAGE_RESOLVE_SYMLINKS TRUE)
 set(CMAKE_CXX_STANDARD 20)
 set(CMAKE_CXX_STANDARD_REQUIRED TRUE)
-set(CMAKE_POSITION_INDEPENDENT_CODE ON)
 set(CMAKE_CXX_VISIBILITY_PRESET hidden)
 set(CMAKE_C_VISIBILITY_PRESET hidden)
+set(CMAKE_FIND_PACKAGE_RESOLVE_SYMLINKS TRUE)
+set(CMAKE_POSITION_INDEPENDENT_CODE ON)
```
ray6080, line 12: Are changes at L11 and L12 related to LTO?

ray6080 (approved):
This looks cool to me. Thanks!

CMakeLists.txt
Riolku: No, I just sorted these `set` statements lexicographically.

Riolku:
Windows fails to build with LTO (https://github.com/kuzudb/kuzu/actions/runs/6960497088), investigating.

Riolku:
Dry running a deploy (https://github.com/kuzudb/kuzu/actions/runs/6961846575) to confirm all is well.

PR #2486, opened 2023-11-22
make: rewrite makefile

Riolku:
The makefile has adopted various technical debt over time. This commit
    rewrites it, with the following major changes:
    
    - Consistent marking of PHONY targets. This is most important for
      `install`, since it matches the name of a directory.
    - Use of `.ONESHELL`. This means that all commands are invoked in the
      same shell, and as such, `cd` is valid without using `&&`. To make
      sure that any failure is fatal, we pass the `-e` flag to the shell.
    - Makes use of `run-cmake-release` to drastically reduce duplication
      throughout the file.
    - Make sure the all target correctly builds everything.
    - Organize targets into sections.
    - Add python and rust target.
    - Remove $(ROOT_DIR) since it is never important. All recipes implicitly
      require the current directory to be the project root.
    
    Also, replaced all instances of `clean-all` with `clean`. `clean-all`
    doesn't exist anymore.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,9 +1,19 @@
-.PHONY: release debug test benchmark all alldebug clean clean-all
+.PHONY: \
+	release debug all alldebug \
+	test lcov \
+	java nodejs python rust \
+	javatest nodejstest pytest rusttest \
+	benchmark example \
+	clangd tidy clangd-diagnostics \
+	install \
+	clean-python-api clean-java clean \
+
+.ONESHELL:
+.SHELLFLAGS = -ec
 
 CLANGD_DIAGNOSTIC_INSTANCES ?= 4
 NUM_THREADS ?= 1
 PREFIX ?= install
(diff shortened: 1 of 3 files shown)


Riolku:
Running a dry run to ensure make works well across all runners. https://github.com/kuzudb/kuzu/actions/runs/6971393645.

benjaminwinger (approved):
Generally looks good to me.

We could also simplify some things by using cmake's `-B` flag instead of mkdir/chdir. That would remove the need to define a platform-specific mkdir.
Similarly, ctest has a `--test-dir` argument, though that wouldn't exactly simplify much.

Riolku:
Ah. That would actually clean up a lot. I'll make those changes.

Riolku:
Updated dry run is https://github.com/kuzudb/kuzu/actions/runs/6971944432.

ray6080 (approved):
LGTM! Thanks!

PR #2489, opened 2023-11-23
tidy: enforce adding override

Riolku:
Turns out only one place in our codebase was missing this. Good job us!

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -18,5 +18,8 @@ Checks:
 
   misc-unused-parameters,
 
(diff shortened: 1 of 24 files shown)


acquamarin (approved):

ray6080 (approved):

Riolku:
Ah nevermind, clang tidy doesnt check header files that have this issue... I really thought it was supposed to...

Riolku:
Updated commit message:

```
    It turns out that `HeaderFilterRegex` uses a Posix Extended Regular
    Expression, which doesn't support lookbacks. The easiest solution is to
    just fix the _one_ error in linenoise.h.
    
    Note that the `run-clang-tidy` regex _does_ support lookbacks, since
    it's a normal Python regex.
```

PR #2491, opened 2023-11-23
cmake: don't build python by default

Riolku:
We shouldn't build anything beyond the core shared and static kuzu libraries by default, and we definitely shouldn't build Python by default but no other APIs.

This changes the default values for these switches to false.


Riolku:
Running a dry run on this as well https://github.com/kuzudb/kuzu/actions/runs/6973481388

Riolku:
After some discussion with @acquamarin , we should build the shell by default since many users use it, but not the Python API. Commit message and title updated.

acquamarin (approved):

benjaminwinger:
Docs will need to be updated. They currently say that the python bindings are built by default: https://kuzudb.com/docusaurus/development/building-kuzu/#build-language-bindings.

PR #2492, opened 2023-11-23
Support disabling CMake options through the makefile

benjaminwinger:
CMake will remember the options and we currently ignore values of anything but 1. Instead I've set it to just pass the argument of `RUNTIME_CHECKS`, `LTO` and `WERROR` directly to cmake, and added options for disabling ASAN, TSAN and UBSAN by setting them to zero.


Riolku (approved):
Yeah this is way more sane. Thanks!

Makefile
```diff
@@ -31,22 +31,26 @@ endif
...
-ifeq ($(UBSAN), 1)
+else ifeq ($(UBSAN), 1)
 	SANITIZER_FLAG=-DENABLE_ADDRESS_SANITIZER=FALSE -DENABLE_THREAD_SANITIZER=TRUE -DENABLE_UBSAN=TRUE
+else ifeq ($(ASAN), 0)
+	SANITIZER_FLAG=-DENABLE_ADDRESS_SANITIZER=FALSE
+else ifeq ($(TSAN), 0)
+	SANITIZER_FLAG=-DENABLE_THREAD_SANITIZER=FALSE
+else ifeq ($(UBSAN), 0)
+	SANITIZER_FLAG=-DENABLE_UBSAN=FALSE
 endif
```
Riolku, line 44: Actually, while we're at it, how do you feel about making UBSAN and TSAN not enable eachother? I think it's really weird behaviour that was to get around problems that this PR fixes.

benjaminwinger: Sure. I hadn't actually noticed that UBSAN is enabling the thread sanitizer too. I thought they were all being enabled exclusively.

Or do you mean to also remove setting them exclusively so that `ASAN=x` is just an alias for `-DENABLE_ADDRESS_SANITIZER=x`?

Riolku: Hm. ASAN and TSAN are incompatible, as are ASAN and UBSAN.

What would be nice would be to be able to specify `UBSAN=1` and have it use UBSAN but not TSAN unless asked, but that's kind of a pain and probably not worth it.

I feel like it makes sense to just be consistent and make `ASAN=x` an alias for `-DENABLE_ADDRESS_SANITIZER=x`, and have users manage what sanitizer they want and don't want in a transparent way.

benjaminwinger: Updated to set and unset them independently.

Riolku (approved):
I'm assuming you tested this :)

benjaminwinger:
I didn't actually test every option, but I tested it enough to make sure that it can be used to both enable and disable the options, and that the `ifdef`s work correctly.

PR #2493, opened 2023-11-23
Pytest cleanup

benjaminwinger:
This removes test_main, which is unnecessary and when tests fail the output indicates test_main.py as the file instead of the actual file containing the test.

I've also removed the relative paths to allow pytest to be run from any directory, and removed the `sys.path` modifications in-code in favour of using the `PYTHONPATH` environment variable (set in the makefile).


Makefile
```diff
@@ -129,8 +129,12 @@ nodejstest: nodejs
...
 
 pytest: python
-	cd tools/python_api/test
-	python3 -m pytest -v test_main.py
+ifeq ($(OS),Windows_NT)
+	set PYTHONPATH=tools/python_api/build
+else
+	export PYTHONPATH=tools/python_api/build
+endif
+	python3 -m pytest -v tools/python_api/test
```
Riolku, line 137: use `cmake -E env PYTHONPATH=tool/python_api/build python3 ...` and remove the OS-specific conditional :)

Riolku (approved):

PR #2499, opened 2023-11-24
Cleanup reading from bool columns

benjaminwinger:
Removed various dedicated functions in place of using the compression interfaces.

One thing I don't entirely like is that it was necessary to modify the `ReadCompressedValuesFromPage` functor to not de-compress bools when reading so that they can be copied directly into the in-place compressed BoolColumnChunk. We could also simplify ColumnChunk somewhat by not compressing bools in-place, which would remove that interface quirk and further simplify things, but that will probably slow down copies a little and increase memory usage.


benjaminwinger:
Merged via #2516

PR #2509, opened 2023-11-28
tidy: include what you use

Riolku:
This change enables [misc-include-cleaner][1], which can be used to clean up our includes. This would allow us to remove the clangd-diagnostics script, and rely entirely on clang-tidy.

In C++, you don't need to include a header for every symbol you use, since some may be included transitively. For instance, if you include a.h, which includes b.h, you don't need to also include b.h to reference symbol B. However, if in the future, a.h removes its dependency on b.h, then your file needs changing. This is one of the reasons transitive includes can make it harder to maintain a project.

The philosophy that all the clang tools follow is to keep headers in an Include What You Use state. This has the following practical benefits:

- No platform-specific issues about transitive includes. The transitive includes differ per-platform, causing failures in CI when the right headers are not included.
- Automatic integration with clang tooling. Clang tidy and clangd both have excellent integration with this paradigm, which means that our editors can automatically add the headers we need as we type.

Finally, note that the headers included this way have no compile time penalty, since they need to be included one way or another.

  [1]: https://clang.llvm.org/extra/clang-tidy/checks/misc/include-cleaner.html


Riolku:
I forgot how hard it is to actually detect used includes properly. clang-tidy suggests I include `bits/stat.h`, which is a private header on linux. These are fundamental flaws with this system that I don't see getting solved without widespread adoption of this system (maybe in a few years this will be easier to use...)

PR #2534, opened 2023-12-01
cmake: generate grammar files

Riolku:
CMake now regenerates the C++ files corresponding to our Cypher.g4 grammar. We still commit these C++ files, in part because Java shouldn't be a build dependency for Kuzu.


Riolku:
For some reason windows still thinks the files differ.

PR #2538, opened 2023-12-02
Replace worker thread busy loop with condition_variable

mewim:


mewim:
Screenshot 2023-12-04 at 7 35 47 PM
This successfully reduces the CPU usage, but the last few runs of benchmark seems show a performance regression. Need to look into it and decide what to do.

mewim:
The performance regression is due to overcommitting the number of threads without considering Docker's affinity setup. It has been fixed now.

src/common/task_system/task_scheduler.cpp
```diff
@@ -81,20 +85,25 @@ void TaskScheduler::removeErroringTask(uint64_t scheduledTaskID) {
...
     }
 }
 
 void TaskScheduler::runWorkerThread() {
+    std::unique_lock<std::mutex> lck{mtx, std::defer_lock};
     while (true) {
-        if (stopThreads.load()) {
-            break;
+        lck.lock();
+        while (taskQueue.empty() && !stopThreads) {
```
ray6080, line 98: Can the while loop be simplified as the following line of code?
```cpp
        cv.wait(lck, [&] { return !taskQueue.empty() || stopThreads; });
```

ray6080 (approved):

src/include/common/task_system/task_scheduler.h
```diff
@@ -66,8 +65,9 @@ class TaskScheduler {
 private:
     std::mutex mtx;
     std::deque<std::shared_ptr<ScheduledTask>> taskQueue;
-    std::atomic<bool> stopThreads{false};
+    volatile bool stopThreads = false;
```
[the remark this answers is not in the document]
mewim: Done

src/common/task_system/task_scheduler.cpp
mewim: Done

Riolku:
image

We still have a busy wait here.

Furthermore, I'm not convinced this PR is actually solving the problem. In an ideal world, we should be notifying the CV exactly when a task completes. Currently we don't do that at all. Instead, we notify the CV when a task is removed from the queue, for some reason?

We can make use of the `Task` structure and have it notify us when it's complete. We can make use of two CVs so that `scheduleTaskAndWaitOrError` need not have a busy wait.

As is, my understanding of the PR is... we are doing busy waits less, but they are not removed entirely. I might be missing something, but there seems to be a lot of `notify_all()` calls, and I can't really figure out why.

Riolku:
Out of band: this PR doesn't fix all the busy waiting, but does fix the main ones. My comments on `volatile` still stand I believe.

As for the `notify_all()` calls, we currently make use of `notify_all()` whenever the task queue is modified, but we should only need to notify when something is pushed to the queue.

src/common/task_system/task_scheduler.cpp
```diff
@@ -33,6 +36,7 @@ void TaskScheduler::scheduleTaskAndWaitOrError(
         scheduleTaskAndWaitOrError(dependency, context);
     }
     auto scheduledTask = scheduleTask(task);
+    cv.notify_all();
```
Riolku, line 39: We should notify inside `scheduleTask`, not here.

mewim: scheduleTask takes the mutex. I think notifying with mutex causes more contention

Riolku: Then we should `unlock()` before notifying, right?

src/include/common/task_system/task_scheduler.h
```diff
@@ -66,8 +65,9 @@ class TaskScheduler {
 private:
     std::mutex mtx;
     std::deque<std::shared_ptr<ScheduledTask>> taskQueue;
-    std::atomic<bool> stopThreads{false};
+    volatile bool stopThreads;
```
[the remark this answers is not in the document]
mewim: Removed volatile

src/common/task_system/task_scheduler.cpp
Riolku: But https://en.cppreference.com/w/cpp/thread/unique_lock/unlock exists, right?

mewim: Yeah but I don’t think unlocking manually, then notifying, then return makes much of a difference from returning, then unlocking automatically, the notifying since returning a pointer should take almost no time.

Riolku: My point is more that `scheduleTask()` is public, and should absolutely be waking up the worker threads, or the method is wrong

mewim: Actually I think scheduleTask could be made private instead. I did a quick search and It seems it is only being called from scheduleTaskAndWaitOrError and scheduleTaskAndWaitOrError is called by processor.

PR #2557, opened 2023-12-08
update Cypher reserved keywords to fix failing tck tests

russell-liu:


scripts/antlr4/Cypher.g4.copy
```diff
@@ -722,7 +722,12 @@ oC_SymbolicName
 
 kU_NonReservedKeywords
     : COMMENT
-        | COUNT ;
+        | COUNT
+        | NODE
+        | REL
+        | BEGIN
```
andyfengHKU, line 728: Can u comment which test is using BEGIN & END as keyword?

andyfengHKU (approved):

PR #2562, opened 2023-12-10
rework table schema constructors

andyfengHKU:


src/catalog/rel_table_group_schema.cpp
```diff
@@ -8,14 +8,20 @@ using namespace kuzu::common;
 namespace kuzu {
 namespace catalog {
 
+RelTableGroupSchema::RelTableGroupSchema(const RelTableGroupSchema& other) : TableSchema{other} {
+    relTableIDs = other.relTableIDs;
+}
```
ray6080, line 13: Double check if this is used or not.

ray6080 (approved):

PR #2567, opened 2023-12-11
implement comparison between nodes and between rels to fix failing tc…

russell-liu:
…k tests


russell-liu:
Not sure why TCKMatch9.Scenario8 is failing; cannot reproduce locally.

PR #2569, opened 2023-12-11
reader/csv: avoid lseek for getting offset

Riolku:
Currently we use lseek() after every row in the parallel CSV reader to get the file offset. I thought this would be very cheap, and figured it was easier than tracking it ourselves. However, it is not cheap at all, as demonstrated by a flamegraph, and it is not hard to track ourselves either.


Riolku:
Benchmark:

Query: `LOAD FROM "100mil_pk.csv" RETURN COUNT(*);`

The file contains 100 million integers, and nothing else.

master: 9.284s.
this branch: 2.427s.

Riolku:
LDBC100 Comment:

master: 80.67s.
this branch: 57.85s.

acquamarin (approved):

PR #2571, opened 2023-12-11
Cleanup logical type construction

benjaminwinger:
This reduces the amount of boiler-plate code that has to be written when creating logical types by moving that into `types.cpp` and providing more specific static methods to create the complex types, particularly since this is part of our public API and users don't really need to handle `ExtraTypeInfo` directly.

E.g. `LogicalType::VAR_LIST(childType);` instead of `LogicalType(LogicalTypeID::VAR_LIST, std::make_unique<VarListTypeInfo>(childType));`.

I'd also cleaned up a couple places where `LogicalType` was being used where `LogicalTypeID` could have been used instead, and removed some indirection from `StructField` to simplify it.

Ideally I think the `ExtraTypeInfo` implementations could be hidden inside `types.cpp`, however there are some more complex places, particularly for `NODE`, `REL` and `RECURSIVE_REL`, which I'd left alone.

One part I'm not sure about is the return types of the `LogicalType` static methods. The simple types all return `unique_ptr` (and the regular constructor is used in places where a unique_ptr is not wanted), but the complex types also need both, and it's a choice between:
1. Provide a separate method for each. E.g. `LogicalType::VAR_LIST_P` to produce a unique_ptr (or a separate method not to, for consistency with the simple types?).
2. Provide just a method which produces a `LogicalType` (what's implemented here), and if you want a unique_ptr wrap it with `make_unique<LogicalType>(...)` to move the object into the `unique_ptr` using the move constructor.
3. Provide just a method which produces a `unique_ptr<LogicalType>` and use `std::move(*LogicalType::VAR_LIST(...))` to move the data out of the `unique_ptr` if you want (unlike (2) this adds an unnecessary heap allocation).

### DIFF
--- a/src/c_api/data_type.cpp
+++ b/src/c_api/data_type.cpp
@@ -3,6 +3,15 @@
 
 using namespace kuzu::common;
 
+namespace kuzu::common {
+struct CAPIHelper {
+    static inline LogicalType* createLogicalType(
+        LogicalTypeID typeID, std::unique_ptr<ExtraTypeInfo> extraTypeInfo) {
+        return new LogicalType(typeID, std::move(extraTypeInfo));
+    }
+};
+} // namespace kuzu::common
+
 kuzu_logical_type* kuzu_data_type_create(
     kuzu_data_type_id id, kuzu_logical_type* child_type, uint64_t fixed_num_elements_in_list) {
     auto* c_data_type = (kuzu_logical_type*)malloc(sizeof(kuzu_logical_type));
@@ -18,7 +27,7 @@ kuzu_logical_type* kuzu_data_type_create(
                                  std::make_unique<FixedListTypeInfo>(
                                      std::move(child_type_pty), fixed_num_elements_in_list) :
                                  std::make_unique<VarListTypeInfo>(std::move(child_type_pty));
-        data_type = new LogicalType(logicalTypeID, std::move(extraTypeInfo));
+        data_type = CAPIHelper::createLogicalType(logicalTypeID, std::move(extraTypeInfo));
     }
     c_data_type->_data_type = data_type;
     return c_data_type;
--- a/src/include/common/types/types.h
+++ b/src/include/common/types/types.h
@@ -10,6 +10,9 @@
 #include "common/types/internal_id_t.h"
 
 namespace kuzu {
+namespace processor {
+class ParquetReader;
+};
 namespace common {
 
 class Serializer;
@@ -159,8 +162,6 @@ public:
     VarListTypeInfo() = default;
     explicit VarListTypeInfo(std::unique_ptr<LogicalType> childType)
         : childType{std::move(childType)} {}
-    explicit VarListTypeInfo(LogicalTypeID childTypeID)
-        : childType{std::make_unique<LogicalType>(childTypeID)} {}
     inline LogicalType* getChildType() const { return childType.get(); }
     bool operator==(const VarListTypeInfo& other) const;
     std::unique_ptr<ExtraTypeInfo> copy() const override;
@@ -206,9 +207,9 @@ public:
 
     void serialize(Serializer& serializer) const;
 
-    static std::unique_ptr<StructField> deserialize(Deserializer& deserializer);
+    static StructField deserialize(Deserializer& deserializer);
 
-    std::unique_ptr<StructField> copy() const;
+    StructField copy() const;
 
 private:
     std::string name;
@@ -218,18 +219,18 @@ private:
 class StructTypeInfo : public ExtraTypeInfo {
 public:
     StructTypeInfo() = default;
-    explicit StructTypeInfo(std::vector<std::unique_ptr<StructField>> fields);
+    explicit StructTypeInfo(std::vector<StructField>&& fields);
     StructTypeInfo(const std::vector<std::string>& fieldNames,
         const std::vector<std::unique_ptr<LogicalType>>& fieldTypes);
 
     bool hasField(const std::string& fieldName) const;
     struct_field_idx_t getStructFieldIdx(std::string fieldName) const;
-    StructField* getStructField(struct_field_idx_t idx) const;
-    StructField* getStructField(const std::string& fieldName) const;
+    const StructField* getStructField(struct_field_idx_t idx) const;
+    const StructField* getStructField(const std::string& fieldName) const;
     LogicalType* getChildType(struct_field_idx_t idx) const;
     std::vector<LogicalType*> getChildrenTypes() const;
     std::vector<std::string> getChildrenNames() const;
-    std::vector<StructField*> getStructFields() const;
+    std::vector<const StructField*> getStructFields() const;
     bool operator==(const kuzu::common::StructTypeInfo& other) const;
 
     static std::unique_ptr<ExtraTypeInfo> deserialize(Deserializer& deserializer);
@@ -239,7 +240,7 @@ private:
     void serializeInternal(Serializer& serializer) const override;
 
 private:
-    std::vector<std::unique_ptr<StructField>> fields;
+    std::vector<StructField> fields;
     std::unordered_map<std::string, struct_field_idx_t> fieldNameToIdxMap;
 };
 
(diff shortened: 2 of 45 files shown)


src/include/function/aggregate_function.h
```diff
@@ -123,12 +123,12 @@ class AggregateFunctionUtil {
     static std::unique_ptr<AggregateFunction> getAvgFunc(const std::string name,
         common::LogicalTypeID inputType, common::LogicalTypeID resultType, bool isDistinct);
     static std::unique_ptr<AggregateFunction> getMinFunc(
-        const common::LogicalType& inputType, bool isDistinct);
+        common::LogicalTypeID inputType, bool isDistinct);
```
andyfengHKU, line 126: I think this is fine in this PR. But in the long run, all functions (scalar, aggregation, table, ...) should have the same base class `Function` whose parameters are `LogicalType` rather than `LogicalTypeID` which means we will be using `LogicalType` almost everywhere for function related code path.

src/include/common/types/types.h
```diff
@@ -287,80 +288,113 @@ class LogicalType {
         const std::vector<std::unique_ptr<LogicalType>>& types);
 
     static std::unique_ptr<LogicalType> BOOL() {
-        return std::make_unique<LogicalType>(LogicalTypeID::BOOL);
+        return std::unique_ptr<LogicalType>(new LogicalType(LogicalTypeID::BOOL));
```
andyfengHKU, line 291: I thought `make_unique` is preferred compared to `std::unique_ptr<>(new object)`. Is there a reason that you are changing?

andyfengHKU (approved):

andyfengHKU:
> One part I'm not sure about is the return types of the LogicalType static methods. The simple types all return unique_ptr (and the regular constructor is used in places where a unique_ptr is not wanted), but the complex types also need both, and it's a choice between:

I think we should just use unique_ptr everywhere or never use it. The former solution seems to be easier since we are already doing so in most places.

src/include/common/types/types.h
benjaminwinger: I'd tried making the constructors private, but only the two-argument one still is. I can revert the change for the simple types, but make_unique doesn't work with private constructors.

src/include/common/types/types.h
```diff
@@ -239,7 +240,7 @@ class StructTypeInfo : public ExtraTypeInfo {
     void serializeInternal(Serializer& serializer) const override;
 
 private:
-    std::vector<std::unique_ptr<StructField>> fields;
+    std::vector<StructField> fields;
```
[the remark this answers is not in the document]
benjaminwinger: There still is no possibility of an unintentional copy. StructField has no copy constructor (deleted implicitly since it contains a `unique_ptr`: [E.g.](https://godbolt.org/#z:OYLghAFBqd5QCxAYwPYBMCmBRdBLAF1QCcAaPECAMzwBtMA7AQwFtMQByARg9KtQYEAysib0QXACx8BBAKoBnTAAUAHpwAMvAFYTStJg1DIApACYAQuYukl9ZATwDKjdAGFUtAK4sGEgMykrgAyeAyYAHI%2BAEaYxBLSAA6oCoRODB7evgGkyamOAqHhUSyx8VK2mPYFDEIETMQEmT5%2BXIF2mA7pdQ0ERZExcQm29Y3N2W0jvf0lZQkAlLaoXsTI7Bzm/mHI3lgA1Cb%2BbgoExGHAh9gmGgCCm9u7mAdHbCwkAJ6X13dmWww7Xn2hzcADdOkRiF9bt9vjsmAoFHs6sQvA4AGJ4KroA4Adist0SXmitDwyBA3z2lKRp1RBAxWIgJ3QIBAJzORj2zDYpD2TJZXgYeAAjl5MAB9RKnYFhAiXPYEd6JTDzPYgTmsTCMgjMkBvMEQLnKxbyxWavm61D6hVK%2BYqkx4%2B0AEW%2BiTOIKYBHYFKp5rZ53VbEO%2BJuVN52v5gpF4slkKOMrl1swQZhOOd/mD3xlexYTDCEDtDtuofNYIcJGByNp9No6Dl/FQyaLVPrADpMCxEgY1mLokxkABrCDmMz14c8805/vigXC0XSwSXCBcW2NkPN1CoFuEhQIHt9wejgCsFi4JkPjvmq6dHEWtE4h94fg4WlIqE4bms1l5y1WT02PFIAhNBvUgICQNAsFwQgSHISgaHoQ1OAA/hBBEMR2AqFD5CUNRgNIXQzH0QxjE/SxKmqdIXAYdxPBaEBCJCMIBlKIYuAADlyFI0gEcY/EIvJuIYGZBnKDiOi6AQejGWjskI8SaikvomNmVixNGJoZL4qZGmEljRMWBQfzWTh7n%2BR5njcJw2UwVgoR%2BP4ASBI4/SMOzTMcv8XnbD43N%2BB5AU8yyUlOGyWDc24s3rCAs3dbxlVxYNQ2ITACBWBg9li0VniufxsD2fwr1Tb4Tg9Uk9jQBgTkwVRXT2LMADUbmCOQ8sOR09jYq8IsEPZe2IaKesy%2BL7USqlktS4h0qGg4zAANj2CBGuavKAFo9mXQrnWhbqCGzXMGHzBLvUpesFq4X5L3TY7eoaM6Ls2kC7w4B9SCfF83w4D9LC/QyVjWGb/DMXggOfG9Fn7EB/H8FsodhuH4dm/ROEkR88I%2B3gFBADRAOAxY4FgcDUA7Og4lgiA0GJ%2Bh4mANouD4OhPWITGIGiPDojCBp3iQ3h2eYYh3gAeWibRwW50gKbYQQBYYWgudB0gsGiLxgDcMRaEx7heCwHMjHEeX8GSrowQ1l9qs6LxPTFmUqjwklomITmPCwPDTjwFhucWKgDGABR6sxAB3AWlSfZDZDQ8RMNkRQVHUeXdDpgwjBQUibDtzHIEWVBJXSDXeEtOIziwdP83I8FKIgVxeL0RjihEvQBJqKu6Yb9JdLmOn5O6dSm9LiTanUtvWO0jSslaYfB/KAyjIw2971R%2BWPr2VQ2NmlbZskPZgGQZB1uhrgFqgiEAeXYHccWBAbKweIS4hqGYfhh%2BocRp6UdetHOAxrGcdBvGYEQEAzbIAtjBCgEAGg%2B2UIYKoQgECoH9iHXgFNEh0FKgISB4RaAwLgW9RBRNkFUxQInGmOINDwiUI0cWeCSbEAiBqD%2BlDKZxAFhbLB8C8KAJuMQH29DAF1HwE%2BXgWFw4YWkFhaOuE456CIcnb6ZE07wEztnAQudXxgmIIXTAxdFgokEG7TAAAVDc6sFFLD%2BhhEY/D0HQNgWwzWpB/YO0SB7JGz157vU4NgVQ5sj7L1XuvcqRD1o4hbBoFs%2B8wEIjiDUBQKoIBfSsGRPYh8SDH3mKfH%2B4NIbQ0fo/Fxr8cGvnoZjbGIMtC2hcUDN%2BC96GlLBqQNRqRnCSCAA%3D)). Given that its purpose is just for creating struct types, I don't think it really needs one either, so we can just leave it as a move-only type.

src/include/function/aggregate_function.h
benjaminwinger: Mostly it seems weird to me to pass around versions of `LogicalType` which have no `extraTypeInfo`, even for complex types, particularly when `LogicalTypeID` is functionally identical.

src/c_api/data_type.cpp
```diff
@@ -3,6 +3,15 @@
 
 using namespace kuzu::common;
 
+namespace kuzu::common {
```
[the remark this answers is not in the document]
benjaminwinger: It's just to give the C API (and there's a similar one for the java API) access to the private LogicalType constructor. I wouldn't expect C users to use the C++ `LogicalType` functions, though maybe we should set up per-type functions similar to the LogicalType ones for constructing types (or at least, for constructing the complex types).

benjaminwinger:
> I think we should just use unique_ptr everywhere or never use it. The former solution seems to be easier since we are already doing so in most places.

I've updated it to have them return `std::unique_ptr`, and also made some of the places which previously took a `LogicalType` without the `unique_ptr` use a `unique_ptr` to compensate.

PR #2573, opened 2023-12-12
Remove busy wait completely in TaskScheduler

mewim:


src/common/task_system/task.cpp
```diff
@@ -22,6 +22,10 @@ void Task::deRegisterThreadAndFinalizeTaskIfNecessary() {
             finalizeIfNecessary();
         } catch (std::exception& e) { setExceptionNoLock(std::current_exception()); }
     }
+    if (isCompletedNoLock()) {
+        mtx.unlock();
```
Riolku, line 26: wait shouldn't this be `lck.unlock()`.

src/include/main/client_context.h
```diff
@@ -55,6 +55,14 @@ class ClientContext {
 
     inline bool isTimeOutEnabled() const { return timeoutInMS != 0; }
 
+    inline uint64_t getTimeoutRemainingInMS() {
+        if (!isTimeOutEnabled()) {
+            return UINT64_MAX;
+        }
```
Riolku, line 61: We could change this into a `KU_ASSERT()`... not sure if that makes more sense. I.e., make it an error to call this method when a timeout is not set.

src/common/task_system/task_scheduler.cpp
```diff
@@ -30,15 +28,32 @@ void TaskScheduler::scheduleTaskAndWaitOrError(
...
             context->clientContext->interrupt();
         }
-        std::this_thread::sleep_for(
-            std::chrono::microseconds(THREAD_SLEEP_TIME_WHEN_WAITING_IN_MICROS));
+        if (timedWait) {
+            task->cv.wait_for(taskLck, std::chrono::milliseconds(timeout));
+        } else {
+            task->cv.wait(taskLck);
+        }
+        taskLck.unlock();
```
Riolku, line 56: One thing we should note in the code as a comment is that completed tasks may sit around in the queue for a bit.

Riolku (approved):

src/common/task_system/task.cpp
Riolku: Via #2574, seems like this is the issue.

Riolku:
image

I'm pretty sure we actually didn't fix the busy wait. In `runWorkerThread()`, we continue if there is no schedulable task. That means we unlock, re-lock, check if the queue is empty (it's not) and then play the same game.

Basically, the issue is that we don't remove tasks from the queue once they have been scheduled on enough threads, I think.

PR #2575, opened 2023-12-12
Optimize scanning of string offsets to skip pages we don't need to read

benjaminwinger:
I'm concerned that this is a little complicated, but it gives a moderate improvement to string scanning performance (roughly 10% on columns with a moderate amount of duplication).


benjaminwinger:
https://github.com/kuzudb/kuzu/actions/runs/7186804200/job/19573028840?pr=2575

Clang-tidy on macos doesn't seem to recognize that `std::ranges::subrange` exists. The code did compile in the macos build job.

ray6080:
> I'm concerned that this is a little complicated, but it gives a moderate improvement to string scanning performance (roughly 10% on columns with a moderate amount of duplication).

Can you share more details on your benchmark? What dataset is used, how large the dictionary is and what are the numbers like?

benjaminwinger:
> Can you share more details on your benchmark? What dataset is used, how large the dictionary is and what are the numbers like?

I've updated the description to include benchmark details, as well as more details about the change

src/storage/store/string_column.cpp
```diff
@@ -28,6 +28,9 @@ StringColumn::StringColumn(std::string name, std::unique_ptr<LogicalType> dataTy
 
 void StringColumn::scanOffsets(Transaction* transaction, const ReadState& state,
     string_offset_t* offsets, uint64_t index, uint64_t numValues, uint64_t dataSize) {
+    // TODO(bmwinger): this causes an issue with structs where a read is attempted on an empty chunk
```
ray6080, line 31: I don't follow here, what's the connection with structs?

ray6080 (approved):

src/storage/store/string_column.cpp
benjaminwinger: I think it was that for some reason structs are reading the first value, even if the numValues in the metadata is 0, so the assertion fails for strings stored in a struct. As long as everything is zero, that value just ends up being an empty string, but it shouldn't be being read. I'm not sure why it was just for structs.

src/storage/store/string_column.cpp
```diff
@@ -154,54 +157,83 @@ void StringColumn::scanUnfiltered(transaction::Transaction* transaction,
     scanValuesToVector(transaction, nodeGroupIdx, offsetsToScan, resultVector, indexState);
 }
 
+std::ranges::subrange<std::vector<std::pair<StringColumn::string_index_t, uint64_t>>::iterator>
+getNextRange(std::vector<std::pair<StringColumn::string_index_t, uint64_t>>& data,
```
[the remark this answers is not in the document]
benjaminwinger: Yes, me too. Maybe it could just be replaced by a pair, but it's a little more convenient and helps simplify the code.
https://en.cppreference.com/w/cpp/compiler_support/20 seems to indicate that it's supported by the four major stdlibs (under "The One Ranges Proposal"). [std::ranges::minmax_result](https://en.cppreference.com/w/cpp/algorithm/ranges/minmax) turned out to not be well supported by apple clang (which is why `std::minmax_result was used instead in the old code here). I'm not sure how much of ranges is.

src/storage/store/string_column.cpp
ray6080: Do you have a test case or example that can reproduce that? It looks like we should fix it now or later, better document it down in an issue?

benjaminwinger: Un-commenting that line, it causes only ` EmptyCreateNodeTest.CreateStructWithWriteTransaction` to fail. That said, after creating the node, the string chunk should contain a single string, so maybe it's the metadata that's wrong, something to do with the scan occurring before the commit probably.

PR #2576, opened 2023-12-12
If the wal file is truncated, don't attempt to read off the end

### LINKED ISSUE #2391: Killing Database during COPY causes Corruption
Sometimes, if the database is killed during a COPY, we fail to recover.

```
kuzu> COPY Comment FROM "~/ldbc10_comment.csv" (HEADER=TRUE, DELIM='|');
fish: Job 1, './kuzu_shell abc' terminated by signal SIGTERM (Polite quit request)
[21:59:09] keenan ~/Projects/kuzu/build/release/tools/shell (master|✔)[SIGTERM]
> ./kuzu_shell abc
Opened the database at path: abc in READ_WRITE mode.
Enter ":help" for usage hints.
Cannot read from file: abc/.wal fileDescriptor: 4 numBytesRead: 0 numBytesToRead: 4096 position: 212992
```

The easiest way to reproduce this was for me to run a long copy (this is LDBC10) and then kill it after most of the copy is done. It's not very consistent... maybe the copy has to be in the finalization step?

benjaminwinger:
Fixes #2391 (hopefully). As far as I can tell, the database is fine, it's just that the wal file is truncated and we crash when determining whether or not  a checkpoint was completed before the interruption since there's a dangling reference to the next page containing the header records.

I was able to start up a database which it was previously failing on successfully, and I've been unable to reproduce it with the changes.

@Riolku, could you double-check this fixes the issue?

TODO: this could be tested by finding an example which writes multiple pages to the wal file after a commit where we skip the checkpoint, manually truncate the wal file to just one page, and then attempt to reopen the database.

### DIFF
--- a/src/storage/wal/wal.cpp
+++ b/src/storage/wal/wal.cpp
@@ -41,6 +41,9 @@ page_idx_t WAL::logPageInsertRecord(DBFileID dbFileID, page_idx_t pageIdxInOrigi
 
 void WAL::logCommit(uint64_t transactionID) {
     lock_t lck{mtx};
+    // Flush all pages before committing to make sure that commits only show up in the file when
+    // their data is also written.
+    flushAllPages();
     WALRecord walRecord = WALRecord::newCommitRecord(transactionID);
     addNewWALRecordNoLock(walRecord);
 }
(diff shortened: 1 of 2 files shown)


benjaminwinger:
Test added, and I added a fix for another issue, where the commit being written before the shadow pages are flushed can lead to a failed recovery (which I noticed when accidentally simulating it by truncating the wal file when it contains the commit within the first page). I think it can only occur if the interruption happens between when the headers are flushed and when the shadow pages are flushed, since the commit shouldn't be flushed until the next time `WAL::flushAllPages` is called, so maybe a better solution than flushing before the commit would be to invert the ordering of the flushes to make sure the header containing the commit is flushed last, but I'm not sure if there's another situation where the commit may be flushed first. 

I don't think it's feasible to test though. Adding a test which kills the database process after a certain amount of time and attempts to recover might trigger it occasionally, but it won't be reliable.

ray6080:
This is a way to fix this, but I think the best way might be to change the logic of logging wal records, we should always cache those records in memory, not write them out to wal file before we call `flushAllPages`. This makes the buffer inside wal grow larger, but it simplifies the logic. We don't flush wal records or shadow pages and do fsync until TransactionManager is committing the transaction.

benjaminwinger:
> always cache those records in memory, not write them out to wal file before we call flushAllPages. This makes the buffer inside wal grow larger, but it simplifies the logic. We don't flush wal records or shadow pages and do fsync until TransactionManager is committing the transaction.

Couldn't that still end up with corrupt data if the interruption occurs during the flush? It should make it much less likely, compared to writing some data before a long copy transaction and the rest afterwards, but I'm not sure it solves the issue by itself, as I think only individual page writes are atomic.

ray6080:
> > always cache those records in memory, not write them out to wal file before we call flushAllPages. This makes the buffer inside wal grow larger, but it simplifies the logic. We don't flush wal records or shadow pages and do fsync until TransactionManager is committing the transaction.
> 
> Couldn't that still end up with corrupt data if the interruption occurs during the flush? It should make it much less likely, compared to writing some data before a long copy transaction and the rest afterwards, but I'm not sure it solves the issue by itself, as I think only individual page writes are atomic.

That's true, we need to also make sure the final flush is atomic in this way.

ray6080 (approved):

Riolku:
Seems to work, thanks!

Issue #2578, opened 2023-12-13
Query returns inconsistent results with multiple runs


sapalli2989:
### Minimal example

![screenshot](https://github.com/kuzudb/kuzu/assets/147163559/9ac0eb04-a53c-4664-9bf7-4129da4162d7)

The idea is to return all longest paths down to the leaf node, given some parent node (`parentid`) and max-depth as user input. Paths  begin with the immediate children of `parentid`. User can also provide `NULL` as input. This means to return all paths starting with the very top nodes.

E.g. 
- `parentid` = `NULL` and max-depth 0 returns `["t1"]`, `["t2"]` and `["t3"]`
- `parentid` = `t1` and max-depth 0 returns `["t11"]`, `["t12"]` and `["t13"]`
- `parentid` = `t1` and max-depth 1 returns `["t11", "t111"]`,`["t11", "t112"]`, `["t12"]`, `["t13"]`

and so on. Third one will be the reference case in this issue.

### Graph reconstruction
```cypher
CREATE NODE TABLE T(id STRING, PRIMARY KEY(id));
CREATE REL TABLE CHILD_OF(FROM T TO T);

CREATE (t1:T {id: "t1"}),
 (t11:T {id: "t11"}),
 (t12:T {id: "t12"}),
 (t13:T {id: "t13"}),
 (t111:T {id: "t111"}),
 (t112:T {id: "t112"}),
 (t1111:T {id: "t1111"}),
 (t2:T {id: "t2"}),
 (t21:T {id: "t21"}),
 (t22:T {id: "t22"}),
 (t3:T {id: "t3"}),
 (t1111)-[:CHILD_OF]->(t111),
 (t111)-[:CHILD_OF]->(t11),
 (t112)-[:CHILD_OF]->(t11),
 (t11)-[:CHILD_OF]->(t1),
 (t12)-[:CHILD_OF]->(t1),
 (t13)-[:CHILD_OF]->(t1),
 (t21)-[:CHILD_OF]->(t2),
 (t22)-[:CHILD_OF]->(t2);
```

### Query

```cypher
// user input
WITH "t1" as parentid
// max-length here 1; also consider single leaf nodes by 0..
MATCH p = (t1:T)<-[e:CHILD_OF*0..1]-(t0:T)
// handle NULL and non-NULL user input for parent node id
WHERE CASE
  WHEN parentid IS NULL THEN NOT EXISTS {MATCH (t1)-[:CHILD_OF]->(:T)}
[...]

PR #2581, opened 2023-12-14
Fix issue-2578

andyfengHKU:
Fix issue #2578 

- Fix a bug when paring predicate in subquery.
- Fix a bug when using CASE in WHERE clause.
- Fix a bug in unary boolean executor.
- Fix a bug when using NULL key in index scan.


ray6080 (approved):
Better put a PR description so it's easy to understand what is happening.

PR #2584, opened 2023-12-14
Enable map type in rel table

### LINKED ISSUE #2184: CREATE REL TABLE with MAP causes "Error: Copy exception: Invalid type for property list creation." + CLI freeze
Works:
```cypher
CREATE NODE TABLE T(name STRING, PRIMARY KEY(name));
CREATE NODE TABLE T2(name STRING, myprops MAP(STRING,STRING), PRIMARY KEY(name));
CREATE REL TABLE KIND_OF(FROM T TO T);
CREATE REL TABLE KIND_OF2(FROM T TO T, mystring STRING);
```
Causes error:
```cypher
kuzu> CREATE REL TABLE KIND_OF3(FROM T TO T, myprops MAP(STRING,STRING));
Error: Copy exception: Invalid type for property list creation.
```
Now subsequent commands afterwards cause CLI freeze/loop:
```
kuzu> MATCH(n) return n;
// CLI freezes / idles
```

(Version: kuzu compiled from today's sources)

benjaminwinger:
Already supported following the update to the rel table storage, but the exception checking that the previously unsupported types aren't used was still in place. Tests have been updated, though not exhaustively added for using the types in rel tables. 

Fixes #2184


ray6080 (approved):

ray6080:
Can you also add a test for COPY with MAP data type on rel tables? you can change one of MANY_MANY in tinysnb. and also test SERIAL with COPY actually.

ray6080:
Also, as we discussed, we don't support serial as a non-pk property for now. So you might need to revert the check on serial on rel table, or you can solve this by supporting it all at once if you want.

andyfengHKU (approved):

benjaminwinger:
I clearly missed the part about serials not being supported as non-primary keys (reverted). But maps do seem to be working in rel tables when copying and inserting.

Issue #2588, opened 2023-12-17
`MATCH ... WHERE false` returns all nodes and has no effect


sapalli2989:
```cypher
CREATE NODE TABLE T (id STRING, PRIMARY KEY(id));
CREATE (:T {id: "t1"}),(:T {id: "t2"});

kuzu> MATCH (t:T) WHERE false RETURN t;
---------------------------------
| t                             |
---------------------------------
| {_ID: 0:0, _LABEL: T, id: t1} |
---------------------------------
| {_ID: 0:1, _LABEL: T, id: t2} |
---------------------------------
(2 tuples)
```

Expected: `false` condition is never satisfied, hence `WHERE false` should lead to 0 results (compared with Neo4j).

Actual: All 2 results are returned.

Background: I was in search for a negated condition in combination with `CASE` given some specific user input value via prepared statement.

andyfengHKU:
Hi @sapalli2989, the bug should now be fixed in PR #2590

sapalli2989:
Much appreciated @andyfengHKU

PR #2590, opened 2023-12-18
Issue 2588

andyfengHKU:
Fix issue #2588


src/binder/expression/expression_util.cpp
```diff
@@ -1,5 +1,7 @@
 #include "binder/expression/expression_util.h"
 
+#include "common/cast.h"
```
ray6080, line 3: Is this the right place for the include?

src/optimizer/filter_push_down_optimizer.cpp
```diff
@@ -46,7 +49,16 @@ std::shared_ptr<planner::LogicalOperator> FilterPushDownOptimizer::visitOperator
 std::shared_ptr<LogicalOperator> FilterPushDownOptimizer::visitFilterReplace(
     const std::shared_ptr<LogicalOperator>& op) {
     auto filter = (LogicalFilter*)op.get();
-    predicateSet->addPredicate(filter->getPredicate());
+    auto predicate = filter->getPredicate();
+    if (predicate->expressionType == ExpressionType::LITERAL) {
+        // Avoid executing child plan is literal is Null or False.
```
ray6080, line 54: // Avoid executing child plan is literal is Null or False.
->

        // Avoid executing child plan **if** literal is Null or False.

src/include/planner/operator/logical_empty_result.h
```diff
@@ -0,0 +1,23 @@
+#pragma once
+
+#include "logical_operator.h"
+
+namespace kuzu {
+namespace planner {
+
+class LogicalEmptyResult final : public LogicalOperator {
```
ray6080, line 8: Can u add a comment here this logical operator is currently only generated when we are doing the filter pushdown optimization? You can point to `FilterPushDownOptimizer::visitFilterReplace`, which I think can help understand this operator if someone runs into here without understanding the optimization.

ray6080 (approved):

Issue #2587, opened 2023-12-17
[Nightly] Kuzu hangs on CREATE, if primary key STRING length is too long (UUID)


sapalli2989:
*This occurs with nightly Kuzu, not release v0.1.0*.

`CREATE` seems to hang without any reported errors (node record is written though). Programs depending on Kuzu API will hang as well, I tested with Python. Issue affects primary key columns only, see below.

In general longer STRING keys are needed to store UUIDs as an alternative to auto-increment SERIAL ids (or we would need separate UUID data type).

Demonstration:
```cypher
CREATE NODE TABLE T (id STRING, descr STRING, PRIMARY KEY (id));

// works
CREATE (:T {id: "foo"});
CREATE (:T {id: "123456789012"});
CREATE (:T {id: "bar", descr: "fe38f9dc-f761-4184-8b8b-ec9a1b5ffb83"});

// hangs
CREATE (:T {id: "ge38f9dc-f751-4174-8b8b-ec9a1b6gab83"});
CREATE (:T {id: "1234567890123"});
```

ray6080:
Hi @sapalli2989 , thanks for reporting! Will look into this soon.

ray6080:
Hi @sapalli2989 , should be fixed in #2592. Can you give a try and verify the fix later (perhaps need to wait for another nightly build release)?

sapalli2989:
Thanks @ray6080! Sure, I'm testing with the next nightly build

sapalli2989:
@ray6080 LGTM (from user perspective)!

PR #2592, opened 2023-12-18
Fix issue 2587

ray6080:
This fixes issue #2587. Also added the case to our test suite.

Also, we should refactor `WALPageIdxAndFrame` to be able to unpin and unlock itself when it is out of scope later.


benjaminwinger (approved):
> Also, we should refactor WALPageIdxAndFrame to be able to unpin and unlock itself when it is out of scope later.

I'd tried adding that when implementing #2516. Note that with MSVC you get hangs if you unlock multiple times, so we need to make sure the destructor is called exactly once by making it a move-only type.

But I think I ended up reverting the change (except the one line re-added here  apparently) since it occurred to me that it might be better to use `DBFileUtils::updatePage`. 
On the other hand, that might need a variant that includes the elemPosInPage or some other form of additional information to take the place of `WALPageIdxPosInPageAndFrame` (maybe by just wrapping it with an additional function, similarly to how `WALPageIdxPosInPageAndFrame` is currently only constructed within the code that needs it and we otherwise just use `createWALVersionIfNecessaryAndPinPage`).

I think there are also some redundant calls to `insertNewPage`, as `createWALVersionIfNecessaryAndPinPage` seems to handle that itself.

PR #2599, opened 2023-12-19
Use updatePage function for write operations

benjaminwinger:
Use the `updatePage` function for write operations instead of manually handling pinning and unpinning.

Following up #2592, particularly https://github.com/kuzudb/kuzu/pull/2592#pullrequestreview-1787781955. 

I was mistaken about insertNewPage being redundant. It's doing some redundant work, however we need to call at least `fileHandle.addNewPage()` and `wal.logPageInsertRecord(...)` as far as I understand.
One option could be to use insertNewPage's insertOp so that when inserting we use that exclusively instead of `updatePage`, though I'm mildly concerned that we're not actually pinning the desired page, but rather the added one, which isn't necessarily the same. Otherwise, maybe we should just add a stripped-down version of the function, or have updatePage do that automatically when you pass `isInsertingNewPage` as true (or even move the check to determine that into `updatePage`).

There is one test which fails about half the time with these changes. I'm not really sure why, as this should be identical to the original; I must have made a mistake somewhere. But I thought I'd open this so it isn't forgotten and in case I'm missing something obvious.

### DIFF
--- a/src/storage/storage_structure/disk_overflow_file.cpp
+++ b/src/storage/storage_structure/disk_overflow_file.cpp
@@ -6,6 +6,7 @@
 #include "common/exception/message.h"
 #include "common/type_utils.h"
 #include "common/types/types.h"
+#include "storage/storage_structure/db_file_utils.h"
 #include "storage/storage_utils.h"
 
 using lock_t = std::unique_lock<std::mutex>;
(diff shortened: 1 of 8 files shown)


src/storage/storage_structure/disk_overflow_file.cpp
```diff
@@ -49,14 +51,19 @@ void DiskOverflowFile::setStringOverflowWithoutLock(
...
+    auto originalPageCursor = PageUtils::getPageElementCursorForPos(
         nextBytePosToWriteTo, BufferPoolConstants::PAGE_4KB_SIZE);
-    memcpy(updatedPageInfoAndWALPageFrame.frame + updatedPageInfoAndWALPageFrame.posInPage,
-        srcRawString, len);
-    DBFileUtils::unpinWALPageAndReleaseOriginalPageLock(
-        updatedPageInfoAndWALPageFrame, *fileHandle, *bufferManager, *wal);
-    TypeUtils::encodeOverflowPtr(diskDstString.overflowPtr,
-        updatedPageInfoAndWALPageFrame.originalPageIdx, updatedPageInfoAndWALPageFrame.posInPage);
+    bool insertingNewPage = false;
+    if (originalPageCursor.pageIdx >= fileHandle->getNumPages()) {
```
ray6080, line 57: Isn't the logic here serving similar functionality with `addNewPageIfNecessaryWithoutLock(len);` ? Why separate them in different places?
Also, I think you also need to check if the `originalPageCursor.elemPosInPage + len` is exceeding 4KB.

ray6080 (approved):

src/storage/storage_structure/disk_overflow_file.cpp
benjaminwinger: Yes, indeed. I was trying to keep things as close as possible to the old code since it wasn't working, however we may as well clean up that too.
I think the existing checks are sufficient. It's checking that the length is no greater than 4KB, and if `originalPageCursor.elemPosInPage + len` is greater than 4KB `addNewPageIfNecessaryWithoutLock` will add a new page.

src/storage/store/column.cpp
```diff
@@ -520,21 +500,20 @@ PageCursor Column::getPageCursorForOffsetInGroup(offset_t offsetInChunk, const R
...
     } else if (originalPageCursor.pageIdx >= dataFH->getNumPages()) {
         KU_ASSERT(originalPageCursor.pageIdx == dataFH->getNumPages());
         DBFileUtils::insertNewPage(*dataFH, dbFileID, *bufferManager, *wal);
         insertingNewPage = true;
     }
-    auto walPageIdxAndFrame = DBFileUtils::createWALVersionIfNecessaryAndPinPage(
-        originalPageCursor.pageIdx, insertingNewPage, *dataFH, dbFileID, *bufferManager, *wal);
-    return {walPageIdxAndFrame, originalPageCursor.elemPosInPage};
+    DBFileUtils::updatePage(*dataFH, dbFileID, originalPageCursor.pageIdx, insertingNewPage,
+        *bufferManager, *wal, [&](auto frame) { writeOp(frame, offsetInChunk); });
```
benjaminwinger, line 516: Issue was passing the offsetInChunk instead of the `originalPageCursor.elemPosInPage`.

PR #2602, opened 2023-12-20
Copy rdf graph

andyfengHKU:
This PR rework `COPY` statement for RDF graph so that all internal tables are copied in one plan.

The plan looks like the following (detailed operators like indexLookup are omitted)
```
                                CopyRDF
       CopyRel    CopyRel   CopyNode    CopyNode
       Scan       Scan      Scan        Scan
```

This is an out-of-memory solution where we scan file for each Copy pipeline (in total 4 times).

**Note**
CopyNodeSharedState is exposed to IndexLookup and Partition in this PR because
- We need to perform lookup before hash index is persist on disk (which happens in WAL).
- We need to know num of nodes before statistics is available (which happens in WAL).


src/include/processor/operator/persistent/copy_node.h
```diff
@@ -35,22 +35,27 @@ class CopyNodeSharedState {
 private:
     inline common::offset_t getNextNodeGroupIdxWithoutLock() { return currentNodeGroupIdx++; }
 
+    void calculateNumTuples();
```
ray6080, line 38: Ideally, you don't need to calculate num tuples. Statistics can provide the information for you after `finalize`.

src/include/processor/operator/persistent/copy_rdf.h
```diff
@@ -0,0 +1,36 @@
...
+#include "processor/operator/sink.h"
+
+namespace kuzu {
+namespace processor {
+
+struct CopyRdfSharedState {
+    std::shared_ptr<FactorizedTable> fTable;
+};
+
+class CopyRdf : public Sink {
```
ray6080, line 12: Consider making this operator as a general CopySink op. The pipeline of copy rel is facing the same problem that there are three pipelines running, and the last pipeline is serving as the dummy CopySink now to flush information to the final shared fTable, but can be changed to use `CopySink` as the dummy final op to flush information out.

src/include/processor/operator/persistent/copy_rel.h
```diff
@@ -48,14 +48,14 @@ class CopyRelSharedState {
     }
     inline void updateRelsStatistics() { relsStatistics->setNumTuplesForTable(tableID, numRows); }
 
-    inline std::shared_ptr<FactorizedTable> getFTable() { return fTable; }
```
ray6080, line 51: Why move the get function and expose this as public here?

src/planner/operator/persistent/logical_copy_from.cpp
```diff
@@ -6,15 +6,29 @@ namespace kuzu {
 namespace planner {
 
 void LogicalCopyFrom::computeFactorizedSchema() {
-    copyChildSchema(0);
+    if (getNumChildren() == 1) {
```
ray6080, line 9: hmm do we really have to check num children to identify if this is copy rdf? If so, better to keep a field inside LogicalCopyFrom to indicate of copy type (NODE, REL, RDF).

src/processor/operator/partitioner.cpp
```diff
@@ -31,6 +31,9 @@ void PartitionerSharedState::initialize() {
     KU_ASSERT(maxNodeOffsets.size() == 2);
     numPartitions.resize(maxNodeOffsets.size());
     for (auto i = 0u; i < maxNodeOffsets.size(); i++) {
+        if (!copyNodeSharedStates.empty()) {
```
ray6080, line 34: Can you leave a todo for me to also move the setting of maxNodeOffsets inside `PartitionerSharedState `? I think we can unify the case for rdf and property graph tables.

src/processor/operator/persistent/copy_node.cpp
```diff
@@ -162,19 +162,28 @@ void CopyNode::checkNonNullConstraint(NullColumnChunk* nullChunk, offset_t numNo
...
+    if (sharedNodeGroup) {
+        numTuples += sharedNodeGroup->getNumRows();
+    }
+}
+
 void CopyNode::finalize(ExecutionContext* context) {
-    uint64_t numNodes = StorageUtils::getStartOffsetOfNodeGroup(sharedState->getCurNodeGroupIdx());
+    sharedState->calculateNumTuples();
+    //    uint64_t numNodes =
+    //    StorageUtils::getStartOffsetOfNodeGroup(sharedState->getCurNodeGroupIdx());
```
ray6080, line 175: Remove commented code. You can keep using `calculateNumTuples`.

ray6080 (approved):

src/include/processor/operator/persistent/copy_node.h
andyfengHKU: hmmm you do need to calculate in order to update Statistics. This is a refactor of existing code.

src/include/processor/operator/index_lookup.h
```diff
@@ -1,29 +1,35 @@
...
 namespace processor {
 
 struct IndexLookupInfo {
-    common::table_id_t tableID;
     std::unique_ptr<common::LogicalType> pkDataType;
-    storage::PrimaryKeyIndex* pkIndex; // NULL if the PK data type is SERIAL.
+    storage::PrimaryKeyIndex* index; // NULL if the PK data type is SERIAL.
+    // In copy rdf, we need to perform lookup before primary key is persist on disk. So we need to
+    // use index builder.
+    std::shared_ptr<CopyNodeSharedState> copyNodeSharedState;
```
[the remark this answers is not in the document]
andyfengHKU: As we discussed, we shouldn't do this because it's not a good practice to create IndexBuilder at mapping stage.

src/include/processor/operator/persistent/copy_rdf.h
andyfengHKU: As discussed we will do this in a different PR

src/include/processor/operator/persistent/copy_rel.h
andyfengHKU: We will try to solve this with comment above together.

PR #2608, opened 2023-12-22
libuv-async

AEsir777:
- possible improvement: thread pool size
- buffer for request: manage life cycle (think its bugged in this draft PR)
- uv_run use NO_WAIT but need to make sure the number of returned requests is exactly corresponding to one node group

test failed due to:
https://github.com/nodejs/node/issues/50717


ray6080:
Close for now. See #2729

Issue #1663, opened 2023-06-12
OpenCypher regression test


andyfengHKU:
We should complete openCypher regression test (aka, Technology Compatibility Kit (TCK)) in order to quantify how close we are in terms of feature complete towards open cypher.

## Clauses

- [ ] call
- [ ] create
- [ ] delete
- [x] match where
- [x] match
- [ ] merge
- [ ] remove
- [ ] return order by
- [x] return skip limit
- [x] return
- [ ] set
- [ ] union
- [x] with order by
- [x] with skip limit
- [x] with where
- [x] with

## Expressions

- [ ] aggregation
- [ ] boolean
- [ ] comparison
- [ ] conditional
- [ ] existentialSubquery
- [ ] graph
- [ ] list
- [ ] literals
- [ ] map
- [ ] mathematical
- [ ] null
- [ ] path
- [ ] pattern
- [ ] precedence
- [ ] quantifier
- [ ] string
- [ ] temporal
- [ ] typeConversion

AEsir777:
## Issues
1. [Matching3.feature scenario4](https://github.com/opencypher/openCypher/blob/master/tck/features/clauses/match/Match3.feature): `rel table` does not support multiple `FROM table1 TO table2` declarations.

russell-liu:
## Issues, continued
- [Return 2 Scenario 3](https://github.com/opencypher/openCypher/blob/master/tck/features/clauses/return/Return2.feature#L60): Missing node property throws exception.
- [Return 2 Scenario 5](https://github.com/opencypher/openCypher/blob/master/tck/features/clauses/return/Return2.feature#L92): Missing rel property throws exception.
- [Return 2 Scenario 12](https://github.com/opencypher/openCypher/blob/master/tck/features/clauses/return/Return2.feature#L200): Cannot create list with parameter types of both NODE and REL.
- [Return 2 Scenario 13](https://github.com/opencypher/openCypher/blob/master/tck/features/clauses/return/Return2.feature#L217): Parser exception when key in map is set to `rel`, probably because it is a reserved word.
- [Return 2 Scenario 15](https://github.com/opencypher/openCypher/blob/master/tck/features/clauses/return/Return2.feature#L252): Returning property of deleted node does not fail.
- [Return 2 Scenario 16](https://github.com/opencypher/openCypher/blob/master/tck/features/clauses/return/Return2.feature#L266): Returning label of deleted node does not fail.
[...]

PR #2609, opened 2023-12-22
Open cypher expression other

Ashleyhx:
this test is currently not supporting


Ashleyhx:
https://github.com/kuzudb/kuzu/issues/1663

PR #2612, opened 2023-12-23
common: add MPSC Queue

Riolku:
This queue is core to the new parallel hash index. It implements the design in this [blog][1], which is also used by Rust. The design is quite simple, and producers are wait-free. The main disadvantage is it is non-blocking, but adding blocking pop() support is tricky, and currently not needed.

  [1]: https://www.1024cores.net/home/lock-free-algorithms/queues/non-intrusive-mpsc-node-based-queue

### DIFF
--- a/src/include/common/mpsc_queue.h
+++ b/src/include/common/mpsc_queue.h
@@ -1,0 +1,107 @@
+#pragma once
+
+// Based on the following design:
+// https://www.1024cores.net/home/lock-free-algorithms/queues/non-intrusive-mpsc-node-based-queue
+
+#include <atomic>
+
+#include "common/assert.h"
+#include "common/copy_constructors.h"
+
+namespace kuzu {
+namespace common {
+
+// Producers are completely wait-free.
+template<typename T>
+class MPSCQueue {
+    struct Node {
+        T data;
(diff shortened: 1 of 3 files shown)


src/include/common/mpsc_queue.h
```diff
@@ -0,0 +1,106 @@
...
+public:
+    MPSCQueue() : head(nullptr), tail(nullptr), _approxSize(0) {
+        // Allocate a dummy element.
+        Node* stub = new Node(T());
+        head = stub;
+
+        // Ordering doesn't matter.
+        tail.store(stub, std::memory_order_relaxed);
+    }
+    MPSCQueue(const MPSCQueue& other) = delete;
```
ray6080, line 39: hmm Should we use macros in `copy_constructors` instead?

ray6080 (approved):

src/include/common/mpsc_queue.h
Riolku: Probably. Ill update it later

Riolku: I also rewrote some of that file, since it was a bit unpolished.

PR #2613, opened 2023-12-23
common: add StaticVector

Riolku:
StaticVector is a class that behaves like a vector in that it is dynamically-sized, but unlike std::vector in that the capacity is fixed and inline.

### DIFF
--- a/src/include/common/static_vector.h
+++ b/src/include/common/static_vector.h
@@ -1,0 +1,88 @@
+#pragma once
+
+#include <cstring>
+#include <memory>
+#include <utility>
+
+#include "common/assert.h"
+#include "common/copy_constructors.h"
+
+namespace kuzu {
+namespace common {
+
+template<typename T>
+class MaybeUninit {
+public:
+    T& assumeInit() { return *ptr(); }
+    const T& assumeInit() const { return *ptr(); }
+    T* ptr() { return reinterpret_cast<T*>(&data); }
+    const T* ptr() const { return reinterpret_cast<const T*>(&data); }
+
+private:
+    alignas(T) std::byte data[sizeof(T)];
+};
+
+template<typename T, size_t N>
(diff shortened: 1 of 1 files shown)


ray6080 (approved):

src/include/common/static_vector.h
```diff
@@ -0,0 +1,86 @@
...
+#include <utility>
+
+#include "common/assert.h"
+
+namespace kuzu {
+namespace common {
+
+template<typename T>
+class MaybeUninit {
+    std::aligned_storage_t<sizeof(T), alignof(T)> data;
```
ray6080, line 15: Run into this https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/p1413r2.pdf, it seems `aligned_storage` will be deprecated. Any concerns over this?

Riolku: I read https://stackoverflow.com/questions/71828288/why-is-stdaligned-storage-to-be-deprecated-in-c23-and-what-to-use-instead and it makes me sad because they dont provide alternatives.

The only real alternative is super boring, it's to do all the manual work and create a properly aligned byte array.

Anyway. I guess I'll do the future-proof solution.

src/include/common/static_vector.h
```diff
@@ -0,0 +1,86 @@
...
+public:
+    T& assumeInit() { return *ptr(); }
+    const T& assumeInit() const { return *ptr(); }
+    T* ptr() { return reinterpret_cast<T*>(&data); }
+    const T* ptr() const { return reinterpret_cast<const T*>(&data); }
+};
+
+template<typename T, size_t N>
+class StaticVector {
+    MaybeUninit<T> items[N];
```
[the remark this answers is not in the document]
Riolku: Will do.

PR #2614, opened 2023-12-23
common: add RAII mutex type

Riolku:
Modelled after Rust, this change adds a Mutex type that more closely follows RAII. Specifically:

- The Mutex hides the underlying data; it is innacessible.
- lock() returns a guard, which provides access to the underlying data.
- The guard unlocks the Mutex when dropped.

### DIFF
--- a/src/include/common/mutex.h
+++ b/src/include/common/mutex.h
@@ -1,0 +1,61 @@
+#pragma once
+
+#include <mutex>
+#include <optional>
+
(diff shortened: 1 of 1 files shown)


ray6080 (approved):
LGTM! I think we should also gradually replace some other use cases with this.

Riolku:
> LGTM! I think we should also gradually replace some other use cases with this.

I strongly agree - and I'll be doing so in the future.

PR #2615, opened 2023-12-23
Parallel Hash Index

Riolku:
Depends on #2612, #2613, and #2614. Disables 3 RDF tests. Commit log is below.

    processor: use queue-based index building
    
    This also moves index building to its own file. Future work may move it
    to its own standalone operator.
    
    These changes break RDF tests, so they have been disabled. They cause
    higher memory usage, so LDBC and LSQB buffer pool sizes have been
    adjusted. They vastly increase the performance - ingesting 100 million
    integers from a parquet file with 64 threads takes about 90 seconds on
    master, but about 5 seconds with this change.

    storage: use parallel hash index
    
    The design is quite simple: every hash index is now represented
    internally as 256 hash indexes. This way, when copying, we can easily
    operator on multiple indexes at once without locking.

    function: use splitmix64 for hashing
    
    SplitMix64 is an excellent integer hashing function. According to [this
    blog][1], it is the main function to beat in terms of hashing. It is
    simple and provides much better output than our previous ones.
    
    In particular, this function does a good job of shuffling the higher
    bits of the output, a property critical for the new hash index design.


src/storage/index/hash_index_builder.cpp
```diff
@@ -126,12 +121,13 @@ template<typename T>
...
 bool HashIndexBuilder<T>::lookupOrExistsInSlotWithoutLock(
     Slot<T>* slot, const uint8_t* key, offset_t* result) {
-    for (auto entryPos = 0u; entryPos < slotCapacity; entryPos++) {
-        if (!slot->header.isEntryValid(entryPos)) {
-            continue;
-        }
+    auto guard = inMemOverflowFile ?
+                     std::make_optional<MutexGuard<InMemFile>>(inMemOverflowFile->lock()) :
+                     std::nullopt;
+    auto memFile = guard ? guard->get() : nullptr;
```
ray6080, line 127: Can you open an issue to optimize locks on memFile for string keys insertion? We can discuss how to do this. One way is to let each thread grab a page at a time, so reducing the total num of lock acquisitions.

src/include/main/database.h
```diff
@@ -44,6 +44,7 @@ struct KUZU_API SystemConfig {
  */
 class Database {
     friend class EmbeddedShell;
+    friend class ClientContext;
```
ray6080, line 47: What's this change here for?

src/include/processor/operator/persistent/index_builder.h
```diff
@@ -0,0 +1,135 @@
...
+// RAII for producer counting.
+class ProducerToken {
+public:
+    explicit ProducerToken(std::shared_ptr<IndexBuilderSharedState> sharedState)
+        : sharedState(std::move(sharedState)) {
+        this->sharedState->addProducer();
+    }
+    NO_COPY(ProducerToken);
+
+    void quit() {
```
ray6080, line 94: mark as inline. we took the convention to mark function implementations as `inline`,  though we missed some places I guess.

src/include/processor/operator/persistent/index_builder.h
```diff
@@ -0,0 +1,135 @@
...
+    explicit IndexBuilder(std::unique_ptr<storage::PrimaryKeyIndexBuilder> pkIndex);
+
+    IndexBuilder clone() { return IndexBuilder(sharedState); }
+
+    void initGlobalStateInternal(ExecutionContext* /*context*/) {}
+    void initLocalStateInternal(ExecutionContext* context);
+    void insert(
+        storage::ColumnChunk* chunk, common::offset_t nodeOffset, common::offset_t numNodes);
+
+    ProducerToken getProducerToken() const { return ProducerToken(sharedState); }
```
ray6080, line 122: mark as inline too.

src/include/storage/index/hash_index_builder.h
```diff
@@ -12,6 +13,7 @@ namespace storage {
 static constexpr common::page_idx_t INDEX_HEADER_ARRAY_HEADER_PAGE_IDX = 0;
 static constexpr common::page_idx_t P_SLOTS_HEADER_PAGE_IDX = 1;
 static constexpr common::page_idx_t O_SLOTS_HEADER_PAGE_IDX = 2;
+static constexpr common::page_idx_t HEADER_PAGES = 3;
```
ray6080, line 16: HEADER_PAGES -> NUM_HEADER_PAGES ?

src/processor/operator/persistent/index_builder.cpp
```diff
@@ -0,0 +1,197 @@
...
+void IndexBuilderSharedState::quitProducer() {
+    if (producers.fetch_sub(1, std::memory_order_relaxed) == 1) {
+        done.store(true, std::memory_order_relaxed);
+    }
+}
+
+IndexBuilder::IndexBuilder(std::unique_ptr<PrimaryKeyIndexBuilder> pkIndex)
+    : IndexBuilder(std::make_shared<IndexBuilderSharedState>(std::move(pkIndex))) {}
+
+void IndexBuilder::initLocalStateInternal(ExecutionContext* /*context*/) {}
```
ray6080, line 147: Move to header file.

src/processor/operator/persistent/index_builder.cpp
```diff
@@ -0,0 +1,197 @@
...
+    }
+}
+
+IndexBuilderSharedState::IndexBuilderSharedState(std::unique_ptr<PrimaryKeyIndexBuilder> pkIndex)
+    : globalQueues(std::move(pkIndex)) {}
+
+IndexBuilder::IndexBuilder(std::shared_ptr<IndexBuilderSharedState> sharedState)
+    : sharedState(std::move(sharedState)), localBuffers(this->sharedState->globalQueues) {}
+
+void IndexBuilderSharedState::consume() {
```
ray6080, line 130: Move to header and inline explicitly? Same for `addProducer`.

ray6080 (approved):
LGTM! Thanks! It will be good to keep a summary of our benchmark result here as a record, which can be revisited easily.

src/include/main/database.h
Riolku: This is to make ClientContext get the correct number of threads (i.e. the number from the system config).

src/include/processor/operator/persistent/index_builder.h
Riolku: From https://en.cppreference.com/w/cpp/language/inline:
> A function defined entirely inside a [class/struct/union definition](https://en.cppreference.com/w/cpp/language/classes), whether it's a member function or a non-member friend function, is implicitly an inline function unless it is attached to a [named module](https://en.cppreference.com/w/cpp/language/modules#Module_declarations)(since C++20).

Do we still want to inline it?

Riolku:
Forgot to bump the storage version..

src/include/processor/operator/persistent/index_builder.h
ray6080: I see. I'm fine with not explicitly inline it.

Riolku:
Dirty Benchmark Results on loading of 100 million integers from a parquet file:

master, 64 threads: 90s.
branch, 1 thread: 30s (cache locality!)
branch, 2 threads: 18s
branch, 4 threads: 14s
branch: 8 threads: 10s
branch, 64 threads: 8s

LDBC 10GB:
master, 64 threads: 15s
branch, 64 threads: 10s

LDBC 10GB Parquet:
master, 64 threads: 14s
branch, 64 threads: 7.2s

Notably, the remaining bottleneck in our copy pipeline is counting the rows in the CSV file.

PR #2619, opened 2023-12-30
In mem copy rdf graph

andyfengHKU:
This PR adds `in_memory` option when copying rdf dataset. In memory version will first cache entire rdf dataset into memory before copying them into different tables. This approach avoids repeated parsing (which is a significant overhead) of rdf dataset.

#### Additional Changes
- Remove ExtraCopierConfig and store them inside each scan function's bind data respectively. In fact, we only need additional config for CSV reader.
- Regiester 9 rdf reader functions. Out-of-memory version should use the first 4 scan functions. In-memory version uses the fifth function to first scan into memory and then re-scan memory with the first 4 functions (in-memory version). 
  - (in memory) scan resource
  - (in memory) scan literal
  - (in memory) scan rrr
  - (in memory) scan url
  - scan all


src/processor/operator/persistent/reader/rdf/rdf_reader.cpp
```diff
@@ -229,27 +178,59 @@ SerdStatus RdfReader::rrlHandle(void* handle, SerdStatementFlags, const SerdNode
...
     } break;
-    default: {
+    case RdfReaderMode::RESOURCE:
+    case RdfReaderMode::LITERAL: {
         sVector = dataChunk->getValueVector(0).get();
     } break;
+    default:
+        break;
     }
     vectorSize = 0;
```
ray6080, line 235: Double check if we really need `vectorSize`. Updating selectedSize might be doable inside reader.

ray6080, line 235: In L240, let's rewrite `printf("warning\n");` to a proper exception.
Also, can you add a TODO or open an issue to handle this by adding a rdf store cache inside RdfReader?

test/test_files/rdf/rdfox_example_in_memory.test
```diff
@@ -0,0 +1,39 @@
+-GROUP RdfoxExampleInMemory
```
ray6080, line 1: Can we also test "spb1k" dataset?

ray6080 (approved):

PR #2629, opened 2024-01-03
Rdf test fix

andyfengHKU:
- Fix rdf test
- Differentiate local global index builder. 
   - Global index builder: created during `initGlobalState` and stored in shared state. Responsible for 
      -  creating `IndexBuilderSharedState`
      -  inserting and flushing last node group during finalize 
   - Local index builder: created during `initLocalState` and stored in operator. Responsible for
     -   local inserting and flushing
 - Move copy node finalize logic to shared state.

### DIFF
--- a/src/include/processor/operator/persistent/index_builder.h
+++ b/src/include/processor/operator/persistent/index_builder.h
@@ -20,7 +20,7 @@ using StringBuffer = common::StaticVector<std::pair<std::string, common::offset_
 
 class IndexBuilderGlobalQueues {
 public:
-    explicit IndexBuilderGlobalQueues(std::unique_ptr<storage::PrimaryKeyIndexBuilder> pkIndex);
+    explicit IndexBuilderGlobalQueues(storage::PrimaryKeyIndexBuilder* pkIndex);
 
     void flushToDisk() const;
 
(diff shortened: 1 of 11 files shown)


src/include/processor/operator/persistent/index_builder.h
```diff
@@ -20,7 +20,7 @@ using StringBuffer = common::StaticVector<std::pair<std::string, common::offset_
 
 class IndexBuilderGlobalQueues {
 public:
-    explicit IndexBuilderGlobalQueues(std::unique_ptr<storage::PrimaryKeyIndexBuilder> pkIndex);
+    explicit IndexBuilderGlobalQueues(storage::PrimaryKeyIndexBuilder* pkIndex);
```
Riolku, line 23: Curious, why can't the global queues have ownership of the index anymore? We have to be careful that accesses to the index in any other fashion are properly synchronized, right?

src/include/processor/operator/persistent/index_builder.h
```diff
@@ -67,13 +67,14 @@ class IndexBuilderSharedState {
     friend class IndexBuilder;
 
 public:
-    explicit IndexBuilderSharedState(std::unique_ptr<storage::PrimaryKeyIndexBuilder> pkIndex);
-    void consume() { globalQueues.consume(); }
-    void flush() { globalQueues.flushToDisk(); }
+    explicit IndexBuilderSharedState(storage::PrimaryKeyIndexBuilder* pkIndex)
+        : globalQueues{pkIndex} {}
+    inline void consume() { globalQueues.consume(); }
```
Riolku, line 72: Every method defined inside a class definition is implicitly inline. 
Screenshot_20240103-130453_1

Riolku (approved):
Otherwise LGTM, thanks for fixing these up.

src/include/processor/operator/persistent/index_builder.h
andyfengHKU: Because RDF copy requires CopyNode and CopyRel in a single physical plan. And CopyRel needs to access `PrimaryKeyIndexBuilder` before it being flushed to disk. It's easier to access if the ownership is in CopyNode instead of GlobalQueue

PR #2633, opened 2024-01-05
Add UUID support to java & c & nodejs API

hououou:
Add UUID support to java & c & nodejs API


mewim (commented):
C and Node.js changes LGTM. For Java, there is `java.util.UUID` type. We should consider binding to this type if possible.

mewim (approved):

PR #2642, opened 2024-01-09
Add MD5 Hashing to Test Framework

mxwli:
This is to allow for the incorporation of a CypherLogicTest into the framework.


test/include/test_runner/test_group.h
```diff
@@ -29,6 +29,9 @@ struct TestStatement {
     std::string batchStatmentsCSVFile;
     std::optional<std::string> connName;
     bool reloadDBFlag = false;
+    bool expectHash = false;
+    uint64_t expectedNumValuesHashed;
```
ray6080, line 33: Double check if you can reuse `expectedNumTuples`.

test/include/test_runner/test_group.h
```diff
@@ -29,6 +29,9 @@ struct TestStatement {
...
+    std::string hashValueExpected;
```
ray6080, line 34: `hashValueExpected` -> `expectedHashValue`.

src/include/common/md5.h
```diff
@@ -0,0 +1,98 @@
+#pragma once
+/*
+** This code taken from the SQLite test library (can be found at
+** https://www.sqlite.org/sqllogictest/doc/trunk/about.wiki).
+** Originally found on the internet. The original header comment follows this comment.
+** The code has been heavily modified and reformatted
```
ray6080, line 6: Better to say `The code has been refactored, but the algorithm stays the same.`?

test/test_files/md5testing/all_shortest_path_md5.test
ray6080, line 1: Instead of keeping a separate test file using md5, we should directly replace existing tests, which can reduce duplicated tests.

test/test_files/md5testing/order_by_md5.test
ray6080, line 1: Ditto. Let's directly replace existing tests.

test/test_runner/test_runner.cpp
```diff
@@ -129,6 +150,19 @@ std::vector<std::string> TestRunner::convertResultToString(
     return actualOutput;
 }
 
+std::pair<std::string, uint64_t> TestRunner::convertResultToMD5Hash(QueryResult& queryResult) {
+    uint64_t numValuesHashed;
```
ray6080, line 154: Instead of comparing numValues, why not compare numTuples, since you are hashing on tuples? Comparing of num tuples looks simpler (for users) and consistent with what we are currently doing.

test/test_files/md5testing/md5.test
ray6080, line 1: Can you add a test case where we have null values in test?

ray6080 (approved):
Generally looks good! Thanks! 
Let's update the documentation on the testing framework.

Can you also open an issue that we can support `md5` utility function with what you have under `common`?

test/test_files/md5testing/order_by_md5.test
mxwli: Should we delete or comment out original expected outputs? I think that keeping the plain text there would still help with debugging if the hash test case fails.

ray6080: Agree. Can we keep them as comments?

test/test_runner/test_runner.cpp
mxwli: Storing numTuples is probably the better option. It originally compared numValues because that's what SQLLogicTest does.

ray6080: I see. Do you think if we change it to comparing by numTuples, it will the transformation harder?

mxwli: It looks like in their tests, the number of values is equivalent to the number of tuples multiplied by the number of columns in the output, which shouldn't be hard to transform.

test/test_files/md5testing/md5.test
mxwli: Yep. I also want to note that since SQLLogicTest outputs null values as `"NULL"` instead of the empty string, null values will be hashed as `"NULL"` in the next commit.

PR #2645, opened 2024-01-09
Fix issue 2643

ray6080:
There was a bug when collecting struct child fields from local vector collection.

In the longer term, we should revisit and refactor the use of `ValueVector`'s state in the local storage. The current usage is not intuitive.

### DIFF
--- a/src/storage/local_storage/local_table.cpp
+++ b/src/storage/local_storage/local_table.cpp
@@ -55,6 +55,7 @@ std::unique_ptr<LocalVectorCollection> LocalVectorCollection::getStructChildVect
     for (auto i = 0u; i < numRows; i++) {
         auto fieldVector =
             common::StructVector::getFieldVector(getLocalVector(i)->getVector(), idx);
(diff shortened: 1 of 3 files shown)


src/storage/local_storage/local_table.cpp
```diff
@@ -55,6 +55,7 @@ std::unique_ptr<LocalVectorCollection> LocalVectorCollection::getStructChildVect
     for (auto i = 0u; i < numRows; i++) {
         auto fieldVector =
             common::StructVector::getFieldVector(getLocalVector(i)->getVector(), idx);
+        fieldVector->state->selVector->selectedPositions[0] = i & (DEFAULT_VECTOR_CAPACITY - 1);
```
andyfengHKU, line 58: We probably want to change append interface to take `ValueVector*, DataChunkState*`. Let's open an issue or leave a TODO here.

andyfengHKU (approved):

PR #2659, opened 2024-01-12
Added a few translated tests from SQLLogicTest

mxwli:
Test case statuses:
|Name|Status|Pass to Fail Ratio|Comments|
|-|-|-|-|
Select1|Passed|1:0|
Select2|Failed|855:37|failed due to bugs and coalesce feature support
Select3|Failed|3096:224|failed due to bugs
Select5|Failed|unknown|running the test case causes memory consumption to constantly increase, eventually causing the process to be killed, and no log to be produced. test disabled.
IndexBetween1|Segfaulted|1480:30|not sure where segfault occurs. output is definitely incorrect.
IndexBetween10|Segfaulted|4320:20|not sure where segfault occurs. output is definitely incorrect.


mxwli:
closed due to changes in test formatting

Issue #2675, opened 2024-01-17
Relation schema does not respect direction / no error on wrong node labels / `CREATE` rewrites user-provided node label


sapalli2989:
Version: v0.1.0

Schema:
```cypher
CREATE NODE TABLE V1 (id STRING, PRIMARY KEY(id));
CREATE NODE TABLE V2 (id STRING, PRIMARY KEY (id));
CREATE REL TABLE E(FROM V1 TO V2);
```
Case 1:
```cypher
kuzu> CREATE (v21: V2 {id: "v21"})-[:E]->(v11:V1 {id: "v11"});
-
(0 tuples)

kuzu> MATCH (v:V2) RETURN v;
-----------------------------------
| v                               |
-----------------------------------
| {_ID: 1:0, _LABEL: V2, id: v21} |
-----------------------------------
(1 tuple)

kuzu> MATCH (v:V1) RETURN v;
-----------------------------------
| v                               |
-----------------------------------
| {_ID: 0:0, _LABEL: V1, id: v11} |
-----------------------------------
(1 tuple)

kuzu> MATCH (v2:V2)-[e:E]->(v1:V1) RETURN v2,e,v1;
Error: Binder exception: Cannot find a label for node v2 that connects to all of its neighbour relationships.

kuzu> MATCH (v1:V1)-[e:E]->(v2:V2) RETURN v1,e,v2;
----------------------------------------------------------------------------------------------------------
| v1                              | e                                  | v2                              |
----------------------------------------------------------------------------------------------------------
| {_ID: 0:0, _LABEL: V1, id: v11} | (0:0)-{_LABEL: E, _ID: 2:0}->(1:0) | {_ID: 1:0, _LABEL: V2, id: v21} |
----------------------------------------------------------------------------------------------------------
(1 tuple)
```

Expected: `CREATE` errors, as the direction `V2 -> V1` is wrong, schema is `FROM V1 TO V2`.
Actual: Both nodes are created with the inverse direction. Querying the actual user-provided direction results in not-understandable `Error: Binder exception`.

Case 2:
```cypher
kuzu> CREATE (v22:V2 {id: "v22"})-[:E]->(v23: V2 {id: "v23"});
-
(0 tuples)

kuzu> MATCH (v2a:V2)-[e:E]->(v2b:V2) RETURN v2a,e,v2b;
Error: Binder exception: Cannot find a label for node v2a that connects to all of its neighbour relationships.
```
[...]

PR #2680, opened 2024-01-17
Fix issue-2675

andyfengHKU:
Fix issue #2675.

The bug is because insertion didn't check query graph pattern against schema constraint. It makes me think the storage backend should assert inserted data (specifically `table_id`) while inserting.

This PR also improves error message to be more readable.


ray6080:
Just to leave a note here, as discussed, I will add sanity checks on the storage side for these cases, too.

ray6080 (approved):

PR #2681, opened 2024-01-17
Check unused `WriteInternalIDValuesToPage`, and fix copy of rel id

ray6080:
`WriteInternalIDValuesToPage` is not correctly copying offsets to column pages. I added a test case that can trigger `offsetInVector` not equal to 0, which triggers incorrectly writing of rel IDs.

### DIFF
--- a/src/storage/store/column.cpp
+++ b/src/storage/store/column.cpp
@@ -43,21 +43,17 @@ private:
 
 struct WriteInternalIDValuesToPage {
     WriteInternalIDValuesToPage() : compressedWriter{LogicalType(LogicalTypeID::INTERNAL_ID)} {}
-    void operator()(uint8_t* frame, uint16_t posInFrame, const uint8_t* data, uint32_t dataOffset,
-        offset_t numValues, const CompressionMetadata& metadata) {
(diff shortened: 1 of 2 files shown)


src/storage/store/column.cpp
```diff
@@ -43,21 +43,17 @@ struct ReadInternalIDValuesToVector {
...
+        KU_ASSERT(false);
     }
     void operator()(uint8_t* frame, uint16_t posInFrame, ValueVector* vector,
         uint32_t offsetInVector, const CompressionMetadata& metadata) {
         KU_ASSERT(vector->dataType.getPhysicalType() == PhysicalTypeID::INTERNAL_ID);
-        compressedWriter(
-            frame, posInFrame, vector->getData(), offsetInVector, 1 /*numValues*/, metadata);
+        auto internalIDValues = reinterpret_cast<const internalID_t*>(vector->getData());
+        compressedWriter(frame, posInFrame,
+            reinterpret_cast<const uint8_t*>(&internalIDValues[offsetInVector].offset),
```
benjaminwinger, line 55: Maybe `vector->getValue(offsetInVector)` would be better here, if only to delegate the cast to ValueVector.

benjaminwinger (approved):

Issue #2672, opened 2024-01-16
Segmentation fault when CASE evaluates to NULL


mxwli:
The first query fails an assertion, and the second query causes a segfault.

```
kuzu> return case 25 when 25 then NULL else 25 end;
Error: Assertion failed in file "src/common/types/types.cpp" on line 99: KU_UNREACHABLE
kuzu> return case 25 when 25 then null else 25 end;
Segmentation fault (core dumped)
```

mxwli:
note: after further observation it seems like queries with the expression `NULL IS NULL` fail the assertion as well.

PR #2690, opened 2024-01-18
Fix issue-2672

andyfengHKU:
Fix issue #2672


mxwli:
Could you try the following queries?
```
CREATE NODE TABLE tab0(ID SERIAL,pk INT32, col0 INT32, col1 FLOAT, col2 STRING, col3 INT32, col4 FLOAT, col5 STRING, PRIMARY KEY(ID));
MATCH (cor0:tab0) WHERE NOT Null IS Null RETURN cor0.col4;
```
The match query appears to still give the same KU_UNREACHABLE error in this branch

acquamarin (approved):

PR #2707, opened 2024-01-23
Don't use bitpacking for the full bit width

### LINKED ISSUE #2711: SignExtend incorrect behaviour when `width == sizeof(T_U)`
## Overview
In the `SignExtend` function in `src/include/storage/compression/sign_extend.h`, when the provided parameter `width` is equal to the size of `T_U`, the bitshift `(T_U(1) << width)` will overflow and (at least on clang 17.0.1) the value will be zeroed out instead extracting its `width` lowest bits.

## Steps to reproduce
I've created a case that causes the SignExtend function to fail [here](https://godbolt.org/z/TPM8KrYrY) as well as a potential fix.

I haven't tried this directly but the adding the following case to `test/storage/compression_test.cpp` should also reproduce the error (bitpacking where the bitWidth is the same as the original size of the element to pack):

```
TEST(CompressionTests, IntegerPackingTest8Small) {
    auto length = 3;
    std::vector<int8_t> src(length, 0);
    src[0] = 0;
    src[1] = -127;
    src[2] = 128;
    auto alg = IntegerBitpacking<int8_t>();
    test_compression(alg, src);
}
```

## Expected behaviour:
The above test case should pass, and `SignExtend` should return the correct sign-extended value (and not 0) even when `width == sizeof(T_U)`

benjaminwinger:
SignExtend produces undefined behaviour if the bitwidth is equal to the size of the type in bits, and it's also somewhat slower than treating the data as uncompressed.

I've just added it as a special case in the IntegerBitpacking compression code since that was simple. Decompression already uses the metadata to determine how to decompress, but we should probably have something similar for compression, in addition to a more generic way of determining the type of compression to use rather than just having algorithm-specific functions.

Fixes #2711 (or at least, it avoids calling SignExtend under those circumstances; it's better for performance reasons to treat the data as uncompressed anyway).

### DIFF
--- a/src/include/storage/compression/compression.h
+++ b/src/include/storage/compression/compression.h
@@ -139,6 +139,7 @@ class Uncompressed : public CompressionAlg {
 public:
     explicit Uncompressed(const common::LogicalType& logicalType)
         : numBytesPerValue{getDataTypeSizeInChunk(logicalType)} {}
+    explicit Uncompressed(uint8_t numBytesPerValue) : numBytesPerValue{numBytesPerValue} {}
 
     Uncompressed(const Uncompressed&) = default;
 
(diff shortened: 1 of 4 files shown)


ray6080 (approved):
Do we have any test cases that already covered this edge case? It would be good to add one to validate future changes on this if not.

benjaminwinger:
Sure, I've put together a few tests which store and retrieve min/max signed values for all the integer data types. 

Oddly, only the 32 and 64-bit ones reproduce the errors (without these changes), though the test example in #2711 does reproduce the issue with uint8_t (noting that 128 is -128 in an int8_t). That test uses slightly different parameters than what's actually done internally though.

PR #2727, opened 2024-01-26
Added CLI highlighting for missing keywords

MSebanc:

### DIFF
--- a/tools/shell/embedded_shell.cpp
+++ b/tools/shell/embedded_shell.cpp
@@ -37,7 +37,7 @@ struct ShellCommand {
 
 const char* TAB = "    ";
 
-const std::array<const char*, 82> keywordList = {"CALL", "CREATE", "DELETE", "DETACH", "EXISTS",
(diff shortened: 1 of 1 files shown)


tools/shell/embedded_shell.cpp
```diff
@@ -37,15 +37,17 @@ struct ShellCommand {
 
 const char* TAB = "    ";
 
-const std::array<const char*, 82> keywordList = {"CALL", "CREATE", "DELETE", "DETACH", "EXISTS",
+const std::array<const char*, 104> keywordList = {"CALL", "CREATE", "DELETE", "DETACH", "EXISTS",
```
andyfengHKU, line 40: LGTM.

Can u also open an issue about synchronizing our Lexer rule with keyword highlighting in the shell. Ideally, we should maintain our keywords in one place that use it as the source of truth to generate our Lexer and keywords.

andyfengHKU (approved):

PR #2728, opened 2024-01-26
Support most simple types in the hash index

benjaminwinger:
This includes support for all integer types (including int128_t) and floating point types. I've omitted bool as it seems mostly pointless, and interval because the implementation is a little ambiguous at the moment (values may be different, but equivalent; and while there is some normalization, it's only in certain operators).
The HashIndex also now uses PhysicalTypes instead of LogicalTypes, so any LogicalTypes using those physical types are also supported (~~TODO: I guess I've omitted them from the tests though~~) (this is breaking change for the storage format since the type IDs are stored in the hash index header).

Floats are going to be a little fragile, but seem to be functional. Looking them up directly works, but any arithmetic could introduce error which will affect the representation of the value enough to make lookups fail. Floats use std::hash, which doesn't appear to perform too badly (I wrote up some small benchmarks just looking at lookup speed and lookups for doubles were 67ns compared to 52ns for int64). There were already floating point hash functions in hash_functions.h, but they were just converting the values to integers and using the integer version, which is not a very good way of hashing floats.

I've done some work unifying the implementations of the HashIndex and the HashIndexBuilder by moving more things into the BaseHashIndex, however that's not really complete yet and they are still mostly implemented independently. Given that this is already a rather large change, I figured it could wait.

I've also removed templating from the DiskArray implementations (though left it in the interfaces). This PR by itself adds about 500kb of what is presumably mostly templated code to the binary. The templated DiskArray implementations added another ~400KB (with the new types) and the templating wasn't really used internally anyway.

To help deal with large switch cases with identical or very similar code in each case, I added a `TypeUtils::visit` function, which behaves similarly to [std::visit](https://en.cppreference.com/w/cpp/utility/variant/visit), but instead maps physical/logical types to their C++ data type using a dummy argument, and has built-in support for passing multiple overloaded functions. The compiler will optimize the result the same as a manually written switch statement, but this is much more concise (more or less what I was attempting in #2611 before ending up with some ugly macros). There are a number of other places in the code where this could probably also be used.


src/processor/operator/index_lookup.cpp
```diff
@@ -63,60 +66,59 @@ void IndexLookup::fillOffsetArraysFromVector(transaction::Transaction* transacti
...
-        if (info.copyNodeSharedState == nullptr) {
-            for (auto i = 0u; i < numKeys; i++) {
-                auto pos = keyVector->state->selVector->selectedPositions[i];
-                auto key = keyVector->getValue<int64_t>(pos);
-                if (!info.index->lookup(transaction, key, offsets[i])) {
-                    throw RuntimeException(
-                        ExceptionMessage::nonExistentPKException(std::to_string(key)));
+    TypeUtils::visit(
+        info.pkDataType->getPhysicalType(),
+        [&](ku_string_t) {
```
ray6080, line 71: Can we move these two lambdas for `ku_string_t` and primitive types to functions? That is more readable.

src/include/storage/index/base_hash_index.h
```diff
@@ -0,0 +1,112 @@
...
+static constexpr common::page_idx_t INDEX_HEADER_ARRAY_HEADER_PAGE_IDX = 0;
+static constexpr common::page_idx_t P_SLOTS_HEADER_PAGE_IDX = 1;
+static constexpr common::page_idx_t O_SLOTS_HEADER_PAGE_IDX = 2;
+static constexpr common::page_idx_t NUM_HEADER_PAGES = 3;
+static constexpr uint64_t INDEX_HEADER_IDX_IN_ARRAY = 0;
+
+template<typename T, typename S = T>
+class BaseHashIndex {
+public:
+    explicit BaseHashIndex() : slotCapacity(getSlotCapacity<T>()) {}
```
ray6080, line 33: For `getSlotCapacity`, it seems we didn't refresh it for newly supported data types. (I cannot add comment at the place where the code is, so note it here).
```cpp
template<typename T>
static constexpr uint8_t getSlotCapacity() {
    if (std::is_same<T, int64_t>::value) {
        return common::HashIndexConstants::INT64_SLOT_CAPACITY;
    } else {
        return common::HashIndexConstants::STRING_SLOT_CAPACITY;
    }
}
```

src/include/storage/storage_structure/disk_array.h
```diff
@@ -276,5 +366,16 @@ class InMemDiskArrayBuilder : public BaseInMemDiskArray<T> {
...
 
+template<typename U>
+class InMemDiskArrayBuilder : public InMemDiskArrayBuilderInternal {
+public:
+    InMemDiskArrayBuilder(FileHandle& fileHandle, common::page_idx_t headerPageIdx,
+        uint64_t numElements, bool setToZero = false)
+        : InMemDiskArrayBuilderInternal(
+              fileHandle, headerPageIdx, numElements, sizeof(U), setToZero) {}
+
+    inline U& operator[](uint64_t idx) { return *(U*)BaseInMemDiskArray::operator[](idx); }
```
ray6080, line 377: This function should be marked as `override`.

src/storage/index/base_hash_index.cpp
```diff
@@ -0,0 +1,97 @@
+#include "storage/index/base_hash_index.h"
+
+#include "common/types/int128_t.h"
+#include "common/types/ku_string.h"
+#include "transaction/transaction.h"
+
+using namespace kuzu::common;
+using namespace kuzu::transaction;
+
+namespace kuzu::storage {
```
ray6080, line 10: Let's not use nested namespace here for now. We should change to nested namespace all for once if we decide to move to this style later, which I don't find it's important to do.

ray6080 (approved):
LGTM!

src/include/storage/storage_structure/disk_array.h
benjaminwinger: It doesn't override anything. `BaseInMemDiskArray::operator[]` has a different signature; this is just exposing a typed version of that function. I've changed `BaseInMemDiskArray::operator[]` to protected so that it's not accessible from `InMemDiskArrayBuilder`/`InMemDiskArray` except through the typed versions.

src/include/storage/index/base_hash_index.h
ray6080: Right.

src/include/storage/storage_structure/disk_array.h
ray6080: ah. got it.

benjaminwinger: Setting it to protected doesn't actually work because I'd initially written it so that `InMemDiskArrayBuilder` inherits from `BaseInMemDiskArray`, but `InMemDiskArray` uses composition (i.e. needs public functions). I've made it so they both use composition (like `BaseDiskArray` too) as that keeps the interfaces cleaner.

PR #2730, opened 2024-01-29
Packed CSR

ray6080:
This PR introduces a gapped array-based design to optimize insertions/deletions for CSR-based rel tables.
The change can greatly reduce disk space usage for `data.kz` file as it reduces most of unnecessary out-of-place rewrites of column chunks for insertions/deletions.
~~Kept it as draft for now as I will have to profile and benchmark it.~~
Detailed benchmark numbers will be updated later.
Right now, the sliding in the leaf segment is done in a naive way, which I'm going to add some optimizations. (Left TODOs in the PR).

Still several tests are failing, which I skipped in this PR, as I think they are more deep rooted in existing code, but get exposed now. I will fix them after this PR.

TODOs:
- [ ] Rework the var_list storage layout to be more write-friendly. My thought right now is to add a `length` column, and we can allow in place updates when the data column has enough space left for updated list values. Any writes to the var list will try to append the updated list values to the end of data column, and update the offset and length to point to the end of the data column. Whenever we resize the data column chunk, or perform out-of-place commit, we can seize the chance to rewrite the chunk by ascending ordering of offsets.
- [ ] Rework local storage to use ColumnChunk for local data. Currently we use ValueVectors, which cause interface bloats in Column.


src/storage/store/column.cpp
```diff
@@ -582,10 +648,52 @@ void Column::prepareCommitForChunk(Transaction* transaction, node_group_idx_t no
...
+    for (auto& [offset, rowIdx] : insertInfo) {
+        if (offset > maxOffset) {
+            maxOffset = offset;
+        }
+    }
+    if ((metadata.compMeta.numValues(BufferPoolConstants::PAGE_4KB_SIZE, *dataType) *
+            metadata.numPages) <= maxOffset) {
+        // Note that for constant compression, `metadata.numPages` will be equal to 0. Thus, this
+        // function will always return false.
+        return false;
```
benjaminwinger, line 695: Don't we want constant compression to be able to do in-place commits if the values happen to be identical to the existing ones? Or is that handled somewhere else?

src/storage/compression/compression.cpp
```diff
@@ -394,6 +394,8 @@ void IntegerBitpacking<T>::setValuesFromUncompressed(const uint8_t* srcBuffer, o
     auto chunkStart = (uint8_t*)getChunkStart(dstBuffer, posInDst, header.bitWidth);
     auto startPosInChunk = posInDst % CHUNK_SIZE;
     U chunk[CHUNK_SIZE];
+    // TODO(bmwinger): Now I'm a bit confused here. Why do we always need to unpack the chunk first?
+    //      Can't we just unpack the chunk if we need to modify only partial of it?
```
benjaminwinger, line 398: Yes, I guess this should only be the behaviour for partial chunks, and we can avoid unpacking when writing whole chunks. This was originally written to work on individual values and I hadn't spent too much time optimizing it.

benjaminwinger (approved):

src/storage/store/column.cpp
```diff
@@ -461,17 +504,38 @@ void Column::writeValue(const ColumnChunkMetadata& chunkMeta, node_group_idx_t n
...
+        }
         DBFileUtils::unpinWALPageAndReleaseOriginalPageLock(
             walPageInfo, *dataFH, *bufferManager, *wal);
-        throw;
+        numValuesWritten += numValuesToWriteInPage;
     }
-    DBFileUtils::unpinWALPageAndReleaseOriginalPageLock(walPageInfo, *dataFH, *bufferManager, *wal);
 }
 
 offset_t Column::appendValues(
```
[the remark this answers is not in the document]
ray6080: Should be doable. I think you did that also to update `metadata.numPages`.

src/storage/store/column.cpp
ray6080: Nice catch. Changed the logic here to enable potential in-place commits for constant compression.

src/storage/compression/compression.cpp
ray6080: OK. I will keep the TODO comment there.

PR #2743, opened 2024-01-30
Avoid unnecessary rewrites of catalog and stats files

ray6080:
Currently, for each write transaction, we will always rewrite the catalog, node and rel stats file. In most cases, some of these rewrites are unnecessary. For small transactions, the rewrites cause performance degrades.

### DIFF
--- a/src/include/catalog/catalog.h
+++ b/src/include/catalog/catalog.h
@@ -98,9 +98,15 @@ public:
 private:
     CatalogContent* getVersion(transaction::Transaction* tx) const;
 
-    bool hasUpdates() { return readWriteVersion != nullptr; }
+    inline bool hasUpdates() const { return isUpdated; }
+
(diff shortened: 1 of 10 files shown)


src/include/catalog/catalog.h
```diff
@@ -98,9 +98,13 @@ class Catalog {
...
     CatalogContent* getVersion(transaction::Transaction* tx) const;
 
-    bool hasUpdates() { return readWriteVersion != nullptr; }
+    inline bool hasUpdates() const { return isUpdated; }
+
+    inline void setToUpdated() { isUpdated = true; }
+    inline void resetToNotUpdated() { isUpdated = false; }
 
 protected:
+    bool isUpdated;
```
andyfengHKU, line 107: Let's comment the rationale of this field.

andyfengHKU (approved):

PR #2776, opened 2024-02-02
Support rehashing in the HashIndexBuilder

benjaminwinger:
Still needs to be integrated with the CSV reader.

The performance has taken a bit of a hit, `HashIndexBuilder::append` is roughly 60% slower than before, mostly I think due to some unnecessary copying when iterating over slots (it was worse before I fixed `InMemDiskArrayBuilder` to directly access the header in `getNumElements` instead of going through the unnecessary locking inherited from `BaseDiskArray`). I'm going to try and restructure things to remove the copies and speed things up.


benjaminwinger:
Performance is fixed, with a slight improvement from before.

Testing on a csv containing 1 to 60 million in a random order, copying this from csv as a primary key takes on average on my 12-core machine (over six runs):
- 3053ms when read as an int64 column
- 21023ms when read as a string column

Compared to master:
- 3075ms when read as an int64 column
- 26002ms when read as a string column

The change on the int column is mostly within the range of noise. On the other hand, when I'd benchmarked HashIndexBuilder::append by itself, 120 million inserts (with bulk reserve) was taking ~15s, compared to ~20s on master, so other parts of the copy pipeline may be dominating the runtime (strings took ~37s vs 40s; the main difference in parallel is likely due to avoiding re-acquiring the mutex for the shared overflow file).

I think string performance could be further improved by using an rwlock instead of a mutex, since access to the overflow file is usually a read-only check to see if a key already exists in the index and could be done concurrently.

benjaminwinger:
I did a quick test, and with an `std::deque` that only locks when appending, the end-to-end string benchmark I'd done in the previous comment improved to 12s, and an `std::list` with an append lock took only 9.8s. Presumably lists are sufficiently simpler to perform better, and we already have one level of indirection.

That should be better than an rwlock, as concurrent reads can be safely done even when writing and appending new pages (since we never delete anything and the only thing changing is the pointer at the end). I'll have to try some larger tests though to see how well it scales, since finding a particular index would be slower, so the fastest in this test may end up being slower with datasets which are much larger. Deques may scale better sometimes, but the MSVC `std::deque` implementation uses a block size so small that it's effectively an `std::list`, so I'm not keen on having to do something custom to get similar performance on all platforms, but it would at least be helpful to know if there is much of a performance gap with larger datasets.

src/include/storage/index/hash_index_builder.h
```diff
@@ -57,49 +56,67 @@ class HashIndexBuilder final : public InMemHashIndex, BaseHashIndex<T, S> {
...
-        return BaseHashIndex<T, S>::insert(key, entry, offset);
+    inline void insert(
+        T key, Slot<S>* slot, uint8_t entryPos, common::offset_t value, InMemFile* /*memFile*/) {
+        auto entry = slot->entries[entryPos].data;
+        memcpy(entry, &key, sizeof(T));
+        memcpy(entry + sizeof(T), &value, sizeof(common::offset_t));
+        updateForNewEntry(slot, entryPos);
+    }
+    void copy(const uint8_t* oldEntry, slot_id_t newSlotId);
+    inline void insertToNewOvfSlot(
```
ray6080, line 95: This function doesn't seem likely to end as inline? Consider moving to cpp file then.

ray6080 (approved):
Looks good to me! I love the new iterator design 😄

PR #2797, opened 2024-02-04
Add RDF_VARIANT bindings for Node.js and Python APIs

mewim:


tools/python_api/src_cpp/include/py_query_result.h
```diff
@@ -6,10 +6,12 @@
 #include "arrow_array.h"
 #include "common/arrow/arrow.h"
 #include "common/types/internal_id_t.h"
+#include "common/types/value/value.h"
 #include "main/kuzu.h"
 #include "pybind_include.h"
 
 using namespace kuzu::main;
+using namespace kuzu::common;
```
andyfengHKU, line 14: This probably doesn't matter much, but we usually avoid using namespace in header files because it will be propagated through other includes. I would remove `using namespace kuzu::common`

tools/python_api/src_cpp/include/py_query_result.h
```diff
@@ -56,9 +58,12 @@ class PyQueryResult {
 
     bool getNextArrowChunk(const std::vector<std::unique_ptr<DataTypeInfo>>& typesInfo,
         py::list& batches, std::int64_t chunk_size);
+
     py::object getArrowChunks(
         const std::vector<std::unique_ptr<DataTypeInfo>>& typesInfo, std::int64_t chunkSize);
 
+    static py::object convertRdfVariantToPyObject(const Value& value);
```
andyfengHKU, line 65: Seems that you can hide this in cpp file. If so, you don't have to include Value in header anymore.

tools/nodejs_api/src_cpp/include/node_util.h
```diff
@@ -16,4 +16,5 @@ class Util {
     static Napi::Object ConvertNodeIdToNapiObject(const nodeID_t& nodeId, Napi::Env env);
     static Value TransformNapiValue(
         Napi::Value napiValue, LogicalType* expectedDataType, const std::string& key);
+    static Napi::Value ConvertRdfVariantToNapiObject(const Value& value, Napi::Env env);
```
andyfengHKU, line 19: ditto. Seems you only need to declare this in cpp.

andyfengHKU (approved):

PR #2799, opened 2024-02-04
Add dictionary abstraction

ray6080:


src/include/storage/store/string_column_chunk.h
```diff
@@ -59,11 +49,7 @@ class StringColumnChunk : public ColumnChunk {
 private:
     // String data is stored as a UINT8 chunk, using the numValues in the chunk to track the number
     // of characters stored.
```
benjaminwinger, line 51: Since the implementation has moved, this comment should probably also be moved into the dictionary chunk.

benjaminwinger (approved):
Looks good to me!

Issue #2762, opened 2024-02-01
RDF Blank Nodes


semihsalihoglu-uw:
When importing blank nodes from Turtle files, we should have the following rules:

1. When importing from multiple files within the same COPY statement, e.g., `COPY UniKG  FROM "/path/*.ttl"`, any blank node with label `_:foo` will be recognized as the same blank node and get 1 generated iri.
2. When we support multiple COPY statements , if the same label, so `_:foo` is used again across different COPY statements, it will be recognized as a different blank node. 
3. The convention for the generated blank node iris is: `_i:bj`, where i and j are two integers. So "_i:bj" is our prefix for blank node IRIs.
4. If a user in a CREATE statement uses `_i:bj` IRI for some node, we do not do something special, i.e., if `_i:bj` is the IRI of some node, we do our default behavior (either error saying there is a duplicate IRI or merge if a new relationship is being added etc.)
5. When we support exporting into different RDF formats, then we export blank nodes as blank nodes instead of regular nodes with IRIs. That is if we are exporting into Turtle, then we export with `_i:bj`. If we export to RDF/XML, then we omit the IRI or use whatever is the blank node specification convention for the exported file's format. 
6. We should not allow a blank node to appear as a predicate (so ignore a triple in the form of "ex:foo _:xyz ex:bar"). Only allow blank nodes to appear in the subjects and objects. We already avoid adding blank nodes as predicates when loading from Turtle files. Let's disable this in CREATE statements as well.

andyfengHKU:
This issue is mostly solved in PR #2827, except for 

> When we support multiple COPY statements , if the same label, so _:foo is used again across different COPY statements, it will be recognized as a different blank node.

> When we support exporting into different RDF formats, then we export blank nodes as blank nodes instead of regular nodes with IRIs. That is if we are exporting into Turtle, then we export with _i:bj. If we export to RDF/XML, then we omit the IRI or use whatever is the blank node specification convention for the exported file's format.

Because we don't support multi copy not exporting into rdf format. I'm closing this issue because we don't have the plan to support these features yet.

PR #2827, opened 2024-02-06
Fix issue-2762

andyfengHKU:
Fix issue #2762

### DIFF
--- a/dataset/rdf/blank/copy.cypher
+++ b/dataset/rdf/blank/copy.cypher
@@ -1,1 +1,1 @@
-COPY R FROM "dataset/rdf/blank/data.ttl";
+COPY R FROM ["dataset/rdf/blank/data.ttl", "dataset/rdf/blank/data2.ttl"];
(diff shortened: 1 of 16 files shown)


semihsalihoglu-uw:
@andyfengHKU: We should not allow `_:ibj` where i and j are integers also when creating new predicates. That does not seem to be done. That's because one day when we export, we'll actually check that pattern and export it as a blank node, node the `_:b` ones (though if you want to restrict that, that's fine too).

mewim (approved):

andyfengHKU:
> @andyfengHKU: We should not allow `_:ibj` where i and j are integers also when creating new predicates. That does not seem to be done. That's because one day when we export, we'll actually check that pattern and export it as a blank node, node the `_:b` ones (though if you want to restrict that, that's fine too).

I don't think we should enforce `_:ibj`. This will lead to inconsistency against sqarql's definition of blank nodes. We should follow the convention that any string starts with `_:` is a blank node.

semihsalihoglu-uw:
The important thing is to block IRIs of the form `_:ibj` entering the system because that's the type of blank nodes we store. When we export databases we will search for that pattern primarily. But as per our discussion, you observed that `_:` is more restrictive and covers patterns of the form `_:ibj`, so this is fine.

Issue #2625, opened 2024-01-03
Parallel Hash Index Database Bloat


Riolku:
After #2615, because we use three disk arrays per index, we use a ton of space for hash indexes on small tables. There are a few opportunities for optimization here, primarily removing the header for the hash index.

Riolku:
As discussed with Guodong, he will take this on, not me.

PR #2857, opened 2024-02-09
Hash Index parallel strings

### LINKED ISSUE #2626: Per-index Memfile Locking Contention
After #2615, we lock a shared in-memory file for string keys in the hash index. Though this works, it is likely (?) a point of contention when strings are long. As mentioned in [this comment](https://github.com/kuzudb/kuzu/pull/2615#discussion_r1435770346), one solution is to let each thread grab a page at a time.

benjaminwinger:
Allows string primary keys to be written in parallel without locking.
Fixes #2626

Instead of a single shared `InMemFile` used to store the string data, each of the concurrent hash indexes stores its own `InMemFile`, and a shared atomic counter is used to reserve pages. Each page stores the next page from that particular index in the last four bytes at the end so that strings split across pages can be pieced together. The `InMemFile` stores pages in an `std::unordered_map` since they are no longer contiguous by index and we need to be able to lookup the pages by their on-disk index when doing in-memory comparisons (this leads to a slight performance decrease compared to the initial test I'd reported in #2626, but not very significant).

A small amount of space is wasted due to how the `DiskOverflowFile` currently works, as it appends new strings starting at the beginning of a fresh page, so the empty space in the 256 partial pages written during the copy will be lost (similar to before, but it's now 256 pages instead of 1). This could be a bit of a bloating issue on small databases (#2625) (though on very small databases not all 256 `InMemFile`s will be used), but it's <1MB, so on large databases it's not too significant.

Performance is very good compared to before, giving speedups of about 5.5x-15x on the 12-core machine I was testing on (and it should scale well on machines with more parallelism). Using the same tests mentioned in https://github.com/kuzudb/kuzu/issues/2626#issuecomment-1934522147, that is, a CSV with 60M lines and a table with just a primary key column, copy takes:
- ~5.5 seconds for a dataset of only short strings (which don't touch the overflow file, but before still used the lock), compared to ~30 seconds on master
- ~16 seconds for a dataset of only long strings (~29 characters long) compared to ~239 seconds on master.

### DIFF
--- a/src/storage/storage_structure/disk_overflow_file.cpp
+++ b/src/storage/storage_structure/disk_overflow_file.cpp
@@ -5,6 +5,8 @@
 #include "common/constants.h"
 #include "common/exception/message.h"
 #include "common/type_utils.h"
+#include "common/types/types.h"
+#include "storage/storage_utils.h"
 
 using lock_t = std::unique_lock<std::mutex>;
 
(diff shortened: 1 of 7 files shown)


src/storage/storage_structure/disk_overflow_file.cpp
```diff
@@ -27,27 +29,37 @@ std::string DiskOverflowFile::readString(TransactionType trxType, const ku_strin
...
-            auto numBytesToReadInPage = std::min(static_cast<uint64_t>(remainingLength),
-                BufferPoolConstants::PAGE_4KB_SIZE - cursor.elemPosInPage);
+            auto numBytesToReadInPage = std::min(
+                static_cast<uint32_t>(remainingLength), END_OF_PAGE - cursor.elemPosInPage);
+            page_idx_t nextPage;
             bufferManager->optimisticRead(*fileHandleToPin, pageIdxToPin, [&](uint8_t* frame) {
                 retVal +=
                     std::string_view(reinterpret_cast<const char*>(frame) + cursor.elemPosInPage,
                         numBytesToReadInPage);
+                nextPage = *(page_idx_t*)(frame + END_OF_PAGE);
```
ray6080, line 39: Can u add a sanity check that `nextPage` is within the range?

ray6080 (approved):

PR #2858, opened 2024-02-09
Create interactive-v1 benchmark

mxwli:
Added LDBC SNB Interactive v1 benchmark workflow that is manually triggered. Average runtime is about 10 minutes. Results for 0.2.0 can be found in the shared results spreadsheet.

Disabled queries:
- Complex 1
- Complex 5
- Complex 8
- Complex 10
- Complex 13
- Complex 14
- Short 7

### DIFF
--- a/.github/workflows/interactive-v1-benchmark.yml
+++ b/.github/workflows/interactive-v1-benchmark.yml
@@ -1,0 +1,82 @@
+name: LDBC-SNB-Interactive-V1-Benchmark
+
+env:
+  RUNTIME_CHECKS: 1
+  WERROR: 1
+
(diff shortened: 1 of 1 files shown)


mxwli:
Changes to the script:
1. Steps are more broken down. There is now Update Driver, Update Implementation, Load Database, Run Benchmark in place of just Run Benchmark
2. The script now makes use of $SERIALIZED_DIR to keep persistent changes and prevent constantly cloning repositories.
3. Data sets aren't grabbed every time they're required, and are now assumed to already exist on the machine.

mewim (approved):

PR #2859, opened 2024-02-09
NULL is NULL Fix

mxwli:
Exact same PR as [2830](https://github.com/kuzudb/kuzu/pull/2830), this time from a branch instead of a fork so as to pass CI tests.

### DIFF
--- a/src/binder/bind_expression/bind_null_operator_expression.cpp
+++ b/src/binder/bind_expression/bind_null_operator_expression.cpp
@@ -12,7 +12,12 @@ std::shared_ptr<Expression> ExpressionBinder::bindNullOperatorExpression(
     const ParsedExpression& parsedExpression) {
     expression_vector children;
     for (auto i = 0u; i < parsedExpression.getNumChildren(); ++i) {
-        children.push_back(bindExpression(*parsedExpression.getChild(i)));
+        auto boundChild = bindExpression(*parsedExpression.getChild(i));
+        if (boundChild->getDataType().getLogicalTypeID() == LogicalTypeID::ANY) {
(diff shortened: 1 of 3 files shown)


src/binder/bind_expression/bind_null_operator_expression.cpp
```diff
@@ -12,7 +13,16 @@ std::shared_ptr<Expression> ExpressionBinder::bindNullOperatorExpression(
     const ParsedExpression& parsedExpression) {
     expression_vector children;
     for (auto i = 0u; i < parsedExpression.getNumChildren(); ++i) {
-        children.push_back(bindExpression(*parsedExpression.getChild(i)));
+        if (parsedExpression.getChild(i)->getExpressionType() == common::ExpressionType::LITERAL) {
```
andyfengHKU, line 16: We probably can reuse existing code path to resolve ANY data type for NULL literals
```
auto boundChild = bindExpression(*parsedExpression.getChild(i));
// if boundChild.dataType == ANY
//     implicitCastIfNecessary
```

andyfengHKU (changes requested):

src/binder/bind_expression/bind_null_operator_expression.cpp
mxwli: I've applied the requested change in the new force push

andyfengHKU (approved):

src/binder/bind_expression/bind_null_operator_expression.cpp
andyfengHKU: I end up making a minor change to refactor the code.

PR #2865, opened 2024-02-12
Remove unnecessary transaction APIs

ray6080:


acquamarin (approved):

tools/rust_api/src/connection.rs
```diff
@@ -67,12 +67,12 @@ pub struct PreparedStatement {
 /// a transaction and committed (even if the query is READ_ONLY).
 /// If the connection is in MANUAL transaction mode, which happens only if an application
 /// manually begins a transaction (see below), then an application has to manually commit or
-/// rollback the transaction by calling commit() or rollback().
+/// rollback the transaction by issuing queries `COMMIT` or `ROLLBACK`.
```
benjaminwinger, line 70: Maybe this section should be removed from here now since it's no longer a part of the rust API.

PR #2875, opened 2024-02-13
Use lockfile to pin all versions used by rust build in main pipeline

benjaminwinger:
Also removed pinning from the other jobs, and bumped MSRV to 1.60 (required for features with the dep: prefix).

I've removed the workspace insertion to share build directories between the example and the rust API since that necessarily moves the lock file into the root. I think the better solution may be to move the example into `tools/rust_api`, and maybe symlink `examples/rust` to that, so that the workspace can exist permanently in `tools/rust_api`.

As far as I can tell, ignoring dependency MSRVs ours should be 1.60. Without the arrow feature, the latest version of our dependencies will require 1.65, and with the arrow feature that increases to 1.72. I think it might be appropriate to leave 1.60 as the MSRV listed in `Cargo.toml`, but to internally treat it as 1.72, mark this in a comment, and make sure all the CI machines use at least that version. Anyone using kuzu will see error messages such as `ahash v0.8.8 cannot be built because it requires rustc 1.72.0 or newer...`, which is easy to understand.

I'm going to leave the arrow dependency at version 43, since newer than that requires rust 1.70 explicitly. When we've updated CI we can bump that to version 50 (the latest).

### DIFF
--- a/.github/workflows/ci-workflow.yml
+++ b/.github/workflows/ci-workflow.yml
@@ -123,21 +123,14 @@ jobs:
     steps:
       - uses: actions/checkout@v3
 
-      - name: Rust share build
-        # Share build cache when building rust API and the example project
-        run: echo $'[workspace]\nmembers = ["tools/rust_api","examples/rust"]' > Cargo.toml
-
       - name: Rust test
         working-directory: tools/rust_api
         run: |
-          cargo update -p half@2.3.1 --precise '2.2.1'
-          cargo update -p time --precise '0.3.23'
-          cargo update -p ahash --precise '0.8.7'
-          cargo test --features arrow -- --test-threads=1
+          cargo test --locked --features arrow -- --test-threads=1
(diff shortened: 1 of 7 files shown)


benjaminwinger:
I don't think this will break the nightly pipelines yet, as I think they will also avoid updating the lock file if they can. But the build machines should be updated to use rust 1.72, and then we can explicitly run cargo update in the nightly pipelines to make sure that the build works with the latest versions of the dependencies (and also target arrow 50).

benjaminwinger:
I don't think this will break the nightly pipelines yet, as I think they will also avoid updating the lock file if they can. But the build machines should be updated to use rust 1.72, and then we can explicitly run cargo update in the nightly pipelines to make sure that the build works with the latest versions of the dependencies (and also target arrow 50).

ray6080 (approved):

PR #2889, opened 2024-02-15
add support for multiple query statements

hououou:
This PR is to add support for multiple query statements.

Currently, we split multiple query statements into several query statements. Each query statement then triggers an auto transaction. This is because we do not support DDL in manual transactions. We should support multiple query statements in one transaction after we support DDL in manual transactions.


src/include/main/query_result.h
```diff
@@ -79,6 +96,12 @@ class QueryResult {
      * @return whether there are more tuples to read.
      */
     KUZU_API bool hasNext() const;
+    /**
+     * The next result (if any)
+     */
+    KUZU_API std::unique_ptr<QueryResult> nextQueryResult;
```
andyfengHKU, line 102: why are we exposing l102?

andyfengHKU (changes requested):
Overall LGTM.

But I want to take another look after you change the grammar.

Issue #2287, opened 2023-10-27
Hash Index Rework


ray6080:
## Overview
Hash index is the main bottleneck during COPY. There are two main problems:
1. It doesn't scale to multiple threads.
2. It requires pre-counting, doesn't support rehash dynamically.

To fix these, this issue proposes to rework the current hash index in two ways: 
1) Implement the lock-free design to make it scalable to multiple thread insertions (with mixed lookups); 
2) Implement the splitting logic to enable it to handle insertions without `bulkReserve` with an accurate num of nodes, which can leave `Reader` the freedom to change its `count` to an estimated one.

The new design is to follow main ideas presented in Dash[^1]: segment based hash table; fingerprints; optimistic locks; and stashed buckets + balanced inserts + displacement to handle collisions.
We still stick to the idea of linear hashing, which is our current implementation, that only one segment will be split at one time when the load factor hits a pre-defined threshold.

## Layout
(TBD)

## TODO Steps
1) change the physical layout of the hash index to be segment based, and implement optimistic locks and fingerprints. At this stage, we still require `bulkReserve` before insertions, and assume that no segment splitting is triggered during insertions.
2) support segment splitting, and also add optimizations for load balancing (balanced insertion, displacement).

[^1]: [Dash: Scalable Hashing on Persistent Memory](http://www.vldb.org/pvldb/vol13/p1147-lu.pdf)

PR #2908, opened 2024-02-16
Persistent hash index performance improvements

benjaminwinger:
This adds the fingerprinting optimization mentioned in #2287. In some isolated and not very general tests, fingerprinting improved lookup times by about 10% for INT64, 20% for short strings (roughly the same length and with a common prefix) and 40% for long strings (about 24 characters, all roughly the same length and with a common prefix, so close to worst case for string comparison short of making the strings longer).

I also made a couple of other small optimizations:
1. Disk array bounds checking has been changed to only occur with runtime checks enabled (disk array performance is a major bottleneck for the on-disk hash index).
2. The hash index header is cached in memory for write transactions like it is for read transactions, and only written to the disk array/buffer manager at the end of `prepareCommit` (to reduce unnecessary load on the buffer manager for a small piece of fixed-sized data).

### DIFF
--- a/src/include/common/constants.h
+++ b/src/include/common/constants.h
@@ -117,7 +117,7 @@ struct StorageConstants {
 
 // Hash Index Configurations
 struct HashIndexConstants {
-    static constexpr uint8_t SLOT_CAPACITY_BYTES = 240;
+    static constexpr uint16_t SLOT_CAPACITY_BYTES = 256;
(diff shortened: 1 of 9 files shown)


ray6080 (approved):
LGTM. I wonder how the lookup performance will be like if we get rid of `ku_string_t` in storage, which should save a bit on file space usage. Also, are there other known options for calculating the fingerprint? Should we play a bit with that, too?

PR #2941, opened 2024-02-24
Fix generate_random_uuid bug

andyfengHKU:
`generate_random_uuid` is a random function so we need to generate a new random value for each tuple. In such case, we cannot apply factorization technique to reduce duplicate intermediate result.

Our solution is to fall back to tuple at a time evaluation for random function.

### DIFF
--- a/src/binder/expression_visitor.cpp
+++ b/src/binder/expression_visitor.cpp
@@ -1,10 +1,12 @@
 #include "binder/expression_visitor.h"
 
 #include "binder/expression/case_expression.h"
+#include "binder/expression/function_expression.h"
 #include "binder/expression/node_expression.h"
 #include "binder/expression/property_expression.h"
 #include "binder/expression/rel_expression.h"
 #include "binder/expression/subquery_expression.h"
+#include "common/cast.h"
 
 using namespace kuzu::common;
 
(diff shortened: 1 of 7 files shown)


acquamarin (approved):
Can you update the GenRandomUUID operation as well?
```
struct GenRandomUUID {
    static void operation(common::ku_uuid_t& input, void* dataPtr) {
        input = common::UUID::generateRandomUUID(
            reinterpret_cast<main::ClientContext*>(dataPtr)->getRandomEngine());
    }
};
```

And executor:
```
     template<typename RESULT_TYPE, typename OP>
     static void execute(common::ValueVector& result, void* dataPtr) {
+        if (result.state->selVector->isUnfiltered()) {
+            for (auto i = 0u; i < result.state->selVector->selectedSize; i++) {
+                OP::operation(result.getValue<RESULT_TYPE>(i), dataPtr);
+            }
+        } else {
+            for (auto i = 0u; i < result.state->selVector->selectedSize; i++) {
+                auto pos = result.state->selVector->selectedPositions[i];
+                OP::operation(result.getValue<RESULT_TYPE>(pos), dataPtr);
+            }
+        }
     }
```

PR #2952, opened 2024-02-26
Replace std::hash with our custom hash function

### LINKED ISSUE #2943: Index lookups on string primary keys fail on Linux when the database files are created on MacOS
The bug can be reproduced as following:

1. Create kuzu db on MacOS.
```
CREATE NODE TABLE DimensionNode(model_key STRING, technical_name STRING, name STRING, aliases STRING[], PRIMARY KEY (model_key));
COPY DimensionNode FROM "<file>" (HEADER=true, DELIM=',');
```
[dimNode.csv](https://github.com/kuzudb/kuzu/files/14400033/dimNode.csv)

2. Transport the db files on Mac to a Linux machine (I tried it with `Ubuntu 22.04.3 LTS`).

3. Open the transported db files and run the query, which returns empty.
```
match (n:DimensionNode) where n.model_key='businessEntity::tpcds_1.tpcds.store' return n;
```

while, regex matching on scanned tuples can correctly return the matching tuple
```
kuzu> match (n:DimensionNode) where n.model_key=~'businessEntity::tpcds_1.tpcds.store' return n;
-----------------------------------------------------------------------------------------------------------------------------------------------------
| n                                                                                                                                                 |
-----------------------------------------------------------------------------------------------------------------------------------------------------
| {_ID: 0:11, _LABEL: DimensionNode, model_key: businessEntity::tpcds_1.tpcds.store, technical_name: tpcds_1.tpcds.store, name: Store, aliases: []} |
-----------------------------------------------------------------------------------------------------------------------------------------------------
(1 tuple)
(1 column)
```

Possibly, this bug can be reproduced between other platforms, but I haven't checked yet.

benjaminwinger:
std::hash is implementation dependent and not guaranteed to be the same between different stdlibs, which means our current hash index layout is not always portable.

`std::hash` also apparently is not required to produce the same result for the same input in different runs of the same program, and while there hasn't been any evidence of that yet, it could cause issues in future.

From https://en.cppreference.com/w/cpp/utility/hash:
> Hash functions are only required to produce the same result for the same input within a single execution of a program; this allows salted hashes that prevent collision denial-of-service attacks. 

Fixes #2943.

I did a quick benchmark on a long string (first five paragraphs of lorem ipsum), and this took 123ns/iteration on average, compared to 117ns/iteration with std::hash, which is fairly reasonable.

### DIFF
--- a/src/include/function/hash/hash_functions.h
+++ b/src/include/function/hash/hash_functions.h
@@ -9,6 +9,7 @@
 #include "common/types/int128_t.h"
 #include "common/types/interval_t.h"
 #include "common/types/ku_string.h"
+#include "common/types/types.h"
 #include "common/vector/value_vector.h"
 
 namespace kuzu {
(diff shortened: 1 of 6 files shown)


src/include/function/hash/hash_functions.h
```diff
@@ -133,31 +134,52 @@ inline void Hash::operation(
 template<>
 inline void Hash::operation(
     const double& key, common::hash_t& result, common::ValueVector* /*keyVector*/) {
-    result = std::hash<double>()(key);
+    // 0 and -0 are not byte-equivalent, but should have the same hash
+    if (key == 0) {
```
ray6080, line 138: I wonder if adding 0.0 can also solve the "0 and -0 are not byte-equivalent" problem without an `if/else` branch.
```
auto newKey = key + 0.0;
result = murmurhash64(*reinterpret_cast<const uint64_t*>(&newKey));
```

ray6080 (approved):
LGTM

ray6080:
@benjaminwinger I merged it, so it can be in our dev build to be verified by the bug reporter.

src/include/function/hash/hash_functions.h
benjaminwinger: That does appear to also work.

Issue #2956, opened 2024-02-27
Add maxDBSize config to all supported language bindings


ray6080:
See here https://github.com/kuzudb/kuzu/pull/2948 for details.

andyfengHKU:
Close due to finish

PR #2960, opened 2024-02-27
Address #2956 for Java and Node.js APIs

mewim:


ray6080 (approved):
Can you see if we can also replicate some tests I did in https://github.com/kuzudb/kuzu/pull/2948 to verify maxDBSize should be power of 2 and must be larger than 4MB?

PR #2963, opened 2024-02-27
Support specifying max DB size in rust

benjaminwinger:
#2956  for the Rust API

### DIFF
--- a/tools/rust_api/src/database.rs
+++ b/tools/rust_api/src/database.rs
@@ -36,7 +36,11 @@ pub struct SystemConfig {
     /// When true, new columns will be compressed if possible
     /// Defaults to true
     enable_compression: bool,
+    /// Whether or not the database can be written to from this instance
+    /// Defaults to not read-only
     read_only: bool,
+    /// The maximum size of the database. Defaults to 8TB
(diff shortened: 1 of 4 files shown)


tools/rust_api/src/database.rs
```diff
@@ -46,6 +50,8 @@ impl Default for SystemConfig {
             max_num_threads: 0,
             enable_compression: true,
             read_only: false,
+            // This is a little weird, but it's a temporary interface
+            max_db_size: u32::MAX as u64,
```
ray6080, line 54: Why not initialize `max_db_size` in a similar way to `buffer_pool_size`?

ray6080 (approved):

tools/rust_api/src/database.rs
benjaminwinger: I was trying to match the default argument for the constructor, but it guess the rust API isn't using the C++ SystemConfig constructor anyway (it probably should, but I think it was written when the constructor was different).

On the other hand, I think both bufferPoolSize should default to zero instead since `-1u` is a valid value that just gets turned into auto-detection.

PR #2964, opened 2024-02-27
Import database

hououou:
This PR is to add support for importing databases. We joint queries based on the `schema.cypher`, `copy.cypher` and `macro.cypher`(files created by export database), getting the query like `"create node table xxx; create rel table xxx;copy from xxxx;"`. Then we execute the generated multiple query statements. Currently, due to we do not support DDL and copy statements in one transaction, each single query statement in generated multiple query statement has its own auto transaction. This should be refactored after we support DDL and Copy statements in one transaction.


src/processor/operator/persistent/import_db.cpp
```diff
@@ -0,0 +1,22 @@
+#include "processor/operator/persistent/import_db.h"
+
+#include "catalog/catalog.h"
+#include "catalog/catalog_entry/node_table_catalog_entry.h"
```
andyfengHKU, line 4: Are these includes needed?

src/parser/transform/transform_port_db.cpp
```diff
@@ -16,5 +16,12 @@ std::unique_ptr<Statement> Transformer::transformExportDatabase(
     return exportDB;
 }
 
+std::unique_ptr<Statement> Transformer::transformImportDatabase(
+    CypherParser::KU_ImportDatabaseContext& ctx) {
+    std::string filePath = transformStringLiteral(*ctx.StringLiteral());
+    auto importDB = std::make_unique<ImportDB>(std::move(filePath));
+    return importDB;
```
andyfengHKU, line 23: inline l22 and l23

src/binder/bind/bind_import_database.cpp
```diff
@@ -0,0 +1,39 @@
...
+using namespace kuzu::transaction;
+using namespace kuzu::storage;
+
+namespace kuzu {
+namespace binder {
+
+std::unique_ptr<BoundStatement> Binder::bindImportDatabaseClause(const Statement& statement) {
+    auto& importDatabaseStatement = ku_dynamic_cast<const Statement&, const ImportDB&>(statement);
+    auto boundFilePath = importDatabaseStatement.getFilePath();
+    std::vector<std::string> files = {"schema.cypher", "copy.cypher", "macro.cypher"};
```
andyfengHKU, line 23: Let's make these file names constant somewhere?

src/binder/bind/bind_import_database.cpp
```diff
@@ -0,0 +1,39 @@
...
+    std::string finalQueryStatements;
+    for (auto& file : files) {
+        auto filePath = vfs->joinPath(boundFilePath, file);
+        std::ifstream cypherFile(filePath);
```
andyfengHKU, line 27: Let's open with our file system. And I think it should handle the case if a file cannot be opened.

src/include/main/client_context.h
```diff
@@ -17,6 +17,10 @@
 
 namespace kuzu {
 
+namespace parser {
```
andyfengHKU, line 20: Apparently this is not needed.

src/main/client_context.cpp
```diff
@@ -425,5 +427,28 @@ void ClientContext::commitUDFTrx(bool isAutoCommitTrx) {
     }
 }
 
+void ClientContext::runQuery(std::string query) {
+    if (transactionContext->hasActiveTransaction()) {
```
andyfengHKU, line 431: Leave a comment explaining this case.

src/main/client_context.cpp
```diff
@@ -232,11 +232,13 @@ std::unique_ptr<PreparedStatement> ClientContext::prepareNoLock(
     try {
         // parsing
         if (parsedStatement->getStatementType() != StatementType::TRANSACTION) {
-            auto txContext = this->transactionContext.get();
-            if (txContext->isAutoTransaction()) {
-                txContext->beginAutoTransaction(preparedStatement->readOnly);
+            if (transactionContext->isAutoTransaction()) {
+                if (transactionContext->hasActiveTransaction()) {
```
andyfengHKU, line 236: Ditto explain the case causing this if statement. We will most likely change it later on.

andyfengHKU (approved):
Let's also open an issue about removing ImportDB operator and instead reuse StandaloneCall operator and we make importDB as a pragma function

PR #2979, opened 2024-02-29
Re-write partitioner to use ColumnChunks instead of ValueVectors

### LINKED ISSUE #2863: Buffer manager exception: Failed to claim a frame: Python API: Bulk Load kuzu==0.0.12 v  kuzu==0.2.0
Python API:

Executed on a Mac local client: 2.6 GHz 6-Core Intel Core i7. AMD Radeon Pro 5300M 4 GB. Intel UHD Graphics 630 1536 MB. Memory 32 GB 2667 MHz DDR4

Scenario:
A parquet snappy compressed file to bulk load a relationship table, file size: 1.97 GB from disk:
Version of Kuzu:
kuzu==0.0.12
Result:
Loads successfully

Version of Kuzu:
kuzu==0.2.0
Result: ERROR:
Buffer manager exception: Failed to claim a frame.

### LINKED ISSUE #2957: Rework Partitioner to use ColumnChunk instead of ValueVector for buffering partitioned data
The CopyRel pipeline relies on `Partitioner`, which take scanned DataChunk from source (csv/parquet files) as input, and partition tuples according to their src/dst node offset (which are populated through IndexLookup, omitted here). 

Partitioner will copy each tuple into a buffering space, inside which each node group consists of a set of DataChunks currently. Each ValueVector inside a DataChunk corresponds to a column/property.

### Problem
When there are string columns/properties, each string ValueVector will by default allocate a 256KB memory block from BufferManager, while the ValueVector can only store 2048 values by default (to fully fill the memory block, the ValueVector has to have strings of average length `256KB/2048`). This behaviour can potentially lead to unexpected out of memory error in BM when avg length of string values are small, while num of tuples are quite large, leading to a lot ValueVectors (thus lots of nearly empty 256KB blocks).

### Solution
The correct data structure to buffer tuples should be `ColumnChunk`, instead of `ValueVector`.
`ColumnChunk` is more memory efficient in the above case.

The change described here should be able to solve the issue https://github.com/kuzudb/kuzu/issues/2863.

benjaminwinger:
ValueVectors have high memory fragmentation, and allocate strings in 256KB chunks for only 2048 strings.
ColumnChunks can have a much larger capacity, and also support string de-duplication.

Fixes #2863 (peak memory usage when copying [the dataset](https://github.com/saschamcdonald/ch_06_kuzudb_tests) was 13GB (36GB before I enabled string de-duplication in the partitioner's column chunks).
Fixes #2957 

I'm not sure if we should leave string compression enabled in the Partitioner, as it can increase memory usage for datasets with little duplication. At the least it should probably follow the global compression setting (which I don't think is easily accessible to the Partitioner).

### DIFF
--- a/src/include/processor/operator/partitioner.h
+++ b/src/include/processor/operator/partitioner.h
@@ -1,7 +1,7 @@
 #pragma once
 
-#include "common/data_chunk/data_chunk_collection.h"
 #include "processor/operator/sink.h"
+#include "storage/store/column_chunk.h"
 
 namespace kuzu {
 namespace storage {
@@ -20,7 +20,14 @@ struct PartitionerFunctions {
 // partitioning methods. For example, copy of rel tables require partitioning on both FWD and BWD
 // direction. Each partitioning method corresponds to a PartitioningState.
 struct PartitioningBuffer {
-    std::vector<std::unique_ptr<common::DataChunkCollection>> partitions;
+    using ColumnChunkCollection = std::vector<std::unique_ptr<storage::ColumnChunk>>;
+    struct Partition {
+        // One chunk for each column, grouped into a list
+        // so that groups from different threads can be quickly merged without copying
+        // E.g. [(a,b,c), (a,b,c)] where a is a chunk for column a, b for column b, etc.
+        std::vector<ColumnChunkCollection> chunks;
+    };
+    std::vector<Partition> partitions;
 
     void merge(std::unique_ptr<PartitioningBuffer> localPartitioningStates);
 };
@@ -49,11 +56,11 @@ struct PartitionerSharedState {
     void resetState();
     void merge(std::vector<std::unique_ptr<PartitioningBuffer>> localPartitioningStates);
 
-    inline common::DataChunkCollection* getPartitionBuffer(
+    inline PartitioningBuffer::Partition& getPartitionBuffer(
         common::vector_idx_t partitioningIdx, common::partition_idx_t partitionIdx) {
         KU_ASSERT(partitioningIdx < partitioningBuffers.size());
         KU_ASSERT(partitionIdx < partitioningBuffers[partitioningIdx]->partitions.size());
-        return partitioningBuffers[partitioningIdx]->partitions[partitionIdx].get();
+        return partitioningBuffers[partitioningIdx]->partitions[partitionIdx];
     }
 };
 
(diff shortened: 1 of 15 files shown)


benjaminwinger:
I've done some optimizations and memory usage is now ~16GB without compression on that dataset (compression on the chunks used in the partitioner has been disabled for now. It may be worthwhile to look into compressing ColumnChunks in-memory more broadly when we're accumulating intermediate values, but it just makes this change more complicated).

ColumnChunk capacity was left at 2048 (same as the previous implementation using `DataChunkCollection`/`ValueVector`) since each thread needs to allocate a chunk for each node group for each property in both directions, which adds up quickly and if it's too high can lead to much more data being allocated when initializing the per-thread state than is necessary to store the data.

src/include/processor/operator/partitioner.h
```diff
@@ -20,7 +20,13 @@ struct PartitionerFunctions {
 // partitioning methods. For example, copy of rel tables require partitioning on both FWD and BWD
 // direction. Each partitioning method corresponds to a PartitioningState.
 struct PartitioningBuffer {
-    std::vector<std::unique_ptr<common::DataChunkCollection>> partitions;
+    struct NodeGroup {
```
ray6080, line 23: I'd rather not use the term `NodeGroup` here, just not to be confused with the one in `storage/store/node_group.h`. You can consider using `Partition`.

You can also wrap `std::vector<std::unique_ptr<storage::ColumnChunk>>` as `ColumnChunkCollection`.

src/processor/operator/partitioner.cpp
```diff
@@ -146,18 +154,28 @@ void Partitioner::executeInternal(ExecutionContext* context) {
...
+        auto& partition =
+            localState->getPartitioningBuffer(partitioningIdx)->partitions[partitionIdx];
+        if (partition.chunks.empty() || partition.chunks.back()[0]->getNumValues() + 1 >
+                                            partition.chunks.back()[0]->getCapacity()) {
+            partition.chunks.emplace_back();
+            partition.chunks.back().reserve(chunkToCopyFrom->getNumValueVectors());
+            for (auto i = 0u; i < chunkToCopyFrom->getNumValueVectors(); i++) {
+                partition.chunks.back().emplace_back(ColumnChunkFactory::createColumnChunk(
+                    chunkToCopyFrom->getValueVector(i)->dataType, false /*enableCompression*/,
+                    2048));
```
ray6080, line 171: Can you turn `2048` into a constant of Partitioner?

src/include/storage/store/column_chunk.h
```diff
@@ -151,12 +151,12 @@ class BoolColumnChunk : public ColumnChunk {
               enableCompression, hasNullChunk) {}
 
     void append(common::ValueVector* vector) final;
+    void appendOne(common::ValueVector* vector, common::vector_idx_t pos) final;
```
ray6080, line 154: This is fine temporarily, but we should refactor the two interfaces `append(common::ValueVector* vector)` and `appendOne(common::ValueVector* vector, common::vector_idx_t pos)` into a single one by passing SelVector as an argument,

```
void append(common::ValueVector* vector, SelVector& sel);
```

ray6080 (approved):

PR #2985, opened 2024-03-02
Support Polars DataFrame export from QueryResult

### LINKED ISSUE #1662: Support for working with DataFrames in Polars
[Polars](https://pola-rs.github.io/polars-book/user-guide/) is a great alternative to pandas to do data manipulation via DataFrames. It offers the following key benefits over pandas for users that may want it:

* Speed (there are numerous benchmarks online, but in general Polars seems way faster because of Rust)
* Has a lazy API for larger datasets
* Much cleaner and simpler API than pandas (more similar to Spark's DataFrame API)
* [No indexes](https://pola-rs.github.io/polars-book/user-guide/migration/pandas/#polars-does-not-have-a-multi-indexindex), reducing boilerplate and repetitive code like in pandas
* MUCH better support for `null` data type for missing values (pandas just defaults any missing value to `NaN`, leading to weird, inconsistent data types in certain columns)
* Concurrency and parallelism (column transforms are run concurrently, not sequentially like in Pandas)

## Use case

* For users that have a large CSV or parquet file, and need to perform some wrangling to get the data in the format they need, Polars can be massively faster and more performant
* For cases where there are `null` values in a column, Polars has a very good `null` data type that's different from the `NaN` that's assigned by default in pandas. Working with missing data is very common for a lot of users, so avoiding data type inconsistencies ( Even a single `NaN` automatically promotes an `int` column to a `float` in pandas, which is incredibly annoying)

## How DuckDB does it

[DuckDB allows users the option](https://duckdb.org/docs/api/python/result_conversion.html) to convert its internal representations to either Pandas (`.df()`) or Polars (`.pl()`), depending on what they want to work with.

alexander-beedie:
Closes #1662 (allow export to Polars as well as Pandas).

Hi, thought this might be helpful; I'm one of the Polars devs, so took a look at it. As @pankajp mentioned, Polars can zero-copy init from Arrow, so _very_ little new code was required to enable this. Instead I put most of the focus into generalising the existing Arrow unit tests so that they can _also_ validate the Polars results, ensuring good code coverage. Hope this is welcome!

FYI: I'm also enabling `kuzu` support on the Polars side (via our `read_database` init method); there's a nice symmetry (and broader discoverability) in having the functionality available in both APIs 👌 (See: https://github.com/pola-rs/polars/pull/14822, which was just merged).

(Looking at `kuzu` I think there may be some good use-cases for it at work too, so I'm going to turn a few people on to this; thanks for the great work  :)

### DIFF
--- a/tools/python_api/requirements_dev.txt
+++ b/tools/python_api/requirements_dev.txt
@@ -1,6 +1,7 @@
 pybind11>=2.6.0
 pytest
 pandas
+polars
 networkx~=3.0.0
 numpy
 pyarrow==14.0.1
--- a/tools/python_api/src_py/query_result.py
+++ b/tools/python_api/src_py/query_result.py
@@ -98,6 +98,19 @@ class QueryResult:
 
         return self._query_result.getAsDF()
 
+    def get_as_pl(self):
+        """
+        Get the query result as a Polars DataFrame.
+
+        Returns
+        -------
+        polars.DataFrame
+            Query result as a Polars DataFrame.
+        """
+
+        import polars as pl
+        return pl.from_arrow(data=self.get_as_arrow(10_000))
+
     def get_as_arrow(self, chunk_size):
         """
         Get the query result as a PyArrow Table.
--- a/tools/python_api/test/test_arrow.py
+++ b/tools/python_api/test/test_arrow.py
@@ -1,237 +1,401 @@
-import time
-
-import kuzu
+import polars as pl
 import pyarrow as pa
-import datetime
 import ground_truth
+from datetime import date, datetime, timedelta
 from decimal import Decimal
 from pandas import Timestamp
 import pytz
 
 
+# enable polars' decimal support
+pl.Config.activate_decimals(True)
+
+
+_expected_dtypes = {
+    # ------------------------------------------------
+    # person
+    # ------------------------------------------------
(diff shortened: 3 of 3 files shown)


mewim:
LGTM. The CI will not pass because we used a few internal secrets in our tests which external contributors do not have access to. We might have to reorder the CI pipeline to make sure that at least the Python tests pass on the platforms we support to merge this.

semihsalihoglu-uw:
Thanks a lot for the contribution @alexander-beedie. Happy to chat with you and your colleagues if you want to use Kuzu at your work. Please reach out to us over Discord.

prrao87 (approved):
These changes look great! Thanks a lot for the PR!

prrao87:
@alexander-beedie I took a look at the `ConnectionExecutor` and it looks great! Once it's in the next release I'll do a thread describing its use and here's looking forward to bringing more Polars users to Kùzu and vice versa :)

alexander-beedie:
> @alexander-beedie I took a look at the `ConnectionExecutor` and it looks great! Once it's in the next release I'll do a thread describing its use and here's looking forward to bringing more Polars users to Kùzu and vice versa :)

Sounds good to me; was one of our users that pointed me in this direction, and it looks really promising. I'm sure there will be some use-cases where clean interop between the two will lead to some very powerful pipelines/capabilities 😎

PR #2989, opened 2024-03-03
Support multiple COPY statements on rel tables

ray6080:
Experimental feature for now. This can be enabled by `CALL` setting.
```
call enable_multi_copy=true;
```

There is a temp hack for setting rel offset in `PlanMapper::mapCopyRelFrom`. Details can be found in #3033.


benjaminwinger (approved):
Looks fine to me

PR #2994, opened 2024-03-04
More efficient ColumnChunk string dictionary caching

benjaminwinger:
Something from #2979 when I was testing in-memory compression, as the original implementation of the DictionaryChunk's index performed poorly on that dataset (largely due to the size and number of the chunks in the partitioner magnifying the overhead of the dictionary's cache).

This changes the storage of the DictionaryChunk's indexTable to store just the index and use custom comparison and hashing functions instead of storing a second copy of the string. In addition to removing one extra copy of the string data, it reduces the memory overhead of the cache from 32 bytes per entry (`sizeof(std::string)`, though this may also include the string data for short strings, so this change won't necessarily reduce memory by `>=28` bytes per string in the dictionary) to just 4 bytes per entry (plus whatever overhead is required for `std::unordered_set`, but that hasn't really changed).

I did a few benchmarks, and when copying 60000000 strings with random numbers, performance improved from ~25s to ~16s for long strings (~20 characters long), and 12s to 10s for short strings (both should have no duplication, so this tests the worst case where all entries get added to the dictionary; the table has one string column with a serial primary key).

### DIFF
--- a/src/include/storage/store/dictionary_chunk.h
+++ b/src/include/storage/store/dictionary_chunk.h
@@ -1,6 +1,5 @@
 #pragma once
 
-#include "common/string_utils.h"
 #include "storage/store/column_chunk.h"
 
 namespace kuzu {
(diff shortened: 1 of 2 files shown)


ray6080 (approved):
LGTM.

Issue #2938, opened 2024-02-23
Hash Index for multiple copy


benjaminwinger:
The current `HashIndexBuilder` assumes that we are creating the index from scratch, and the persistent `HashIndex` is not well-optimized for inserting a large number of entries at once. To support multiple `COPY` statements one of these things needs to change, and it seems simplest to combine the two into a single HashIndex which can be as fast as the `HashIndexBuilder` when there is no existing data (and ideally not much slower when doing subsequent copies).

After looking at doing some sort of disk array caching to try and reduce the amount of WAL reads+writes that would currently be done through the `BaseDiskArray` if we were to use the current `HashIndex`, I think the solution is to do as much work as possible in the hash index's in-memory `HashIndexLocalStorage`, and then write the data in a single pass, caching the page to write all 16 slots together, or as many as are modified within that page, as there would be no need to read or write the same page again afterwards (for existing overflow slots this may still require reading/writing the page as many as 16 times since overflow slots aren't necessarily grouped the way primary slots are, but that should still be much better than once per insert).

This should be very similar to what the HashIndexBuilder is already doing, except that the layout is different in the in-memory part, and our `Slot`s aren't created until the data is flushed.

### Optimizing Hash index Local Storage for fast updates
The HashIndexLocalStorage could be replaced with a structure that groups elements to be inserted (and deleted, but that's not relevant to copy) by their primary slot ID, with the number of primary slots calculated using the total number of entries in both local storage and on-disk (and re-hashing as necessary to keep the number of primary slots equal to what will be needed when writing to disk).

I suspect that a vector of vectors will perform best, though it does have significantly more overhead than the current `unordered_map` (1 24-byte vector per primary slot, plus small allocations for each vector's buffer when keys are inserted). It might be interesting to compare this to the `HashIndexBuilder`, as that could be adapted for use as a purely in-memory index, and has the benefit already of using the exact same slot mechanism and supporting rehashing, but wouldn't be well-optimized for doing a small number of inserts into a large existing index (since we'd need to more or less allocate space for the entire index in-memory).

### Disk Array
To let the `HashIndex` read and update slots one page at a time (unless we change the slot size to one page), and to bypass the WAL file when writing new pages, there will need to be a new version of the disk array, or maybe just some new functions added to the existing `BaseDiskArray`.

### Copy Pipeline optimizations
[...]

PR #2997, opened 2024-03-05
Insert into the hash index builder one chunk at a time

benjaminwinger:
First part of #2938.

There is a small performance improvement, though it's obscured somewhat by the amount of variance I was seeing between runs. It was roughly 100ms improvement (copying a table with just 60 million integer primary keys, went from ~2.3 to ~2.2 seconds).

This replaces the one at a time append in `HashIndexBuilder` with an append function that takes a StaticVector (since that's what's used in the IndexBuilder's queues). Resizing the index and calculating the hashes is done on the 1024 values in the StaticVector all at once, before each value is inserted one by one.
I also removed the second type from the HashIndexBuilder since the StaticVector stores a `std::string` and it makes more sense to just hardcode places which take a `std::string`/`std::string_view` using `std::conditional` than to add more template parameters (and there is only really one variable parameter anyway).

### DIFF
--- a/src/include/storage/index/hash_index_builder.h
+++ b/src/include/storage/index/hash_index_builder.h
@@ -1,6 +1,9 @@
 #pragma once
 
+#include "common/static_vector.h"
 #include "common/type_utils.h"
+#include "common/types/internal_id_t.h"
+#include "common/types/ku_string.h"
 #include "common/types/types.h"
 #include "storage/index/hash_index_header.h"
 #include "storage/index/hash_index_slot.h"
@@ -17,6 +20,10 @@ public:
     virtual void flush() = 0;
     virtual void bulkReserve(uint32_t numEntries) = 0;
 };
+
+constexpr size_t BUFFER_SIZE = 1024;
+template<typename T>
+using IndexBuffer = common::StaticVector<std::pair<T, common::offset_t>, BUFFER_SIZE>;
 
 /**
  * Basic index file consists of three disk arrays: indexHeader, primary slots (pSlots), and overflow
@@ -44,9 +51,10 @@ public:
  *
  *  */
 
-// T is the key type used to access values
-// S is the stored type, which is usually the same as T, with the exception of strings
-template<common::IndexHashable T, typename S = T>
+// T is the key type stored in the slots.
+// For strings this is different than the type used when inserting/searching
+// (see BufferKeyType and Key)
+template<common::IndexHashable T>
 class HashIndexBuilder final : public InMemHashIndex {
     static_assert(getSlotCapacity<T>() <= SlotHeader::FINGERPRINT_CAPACITY);
     // Size of the validity mask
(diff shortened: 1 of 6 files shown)


src/include/storage/index/hash_index_builder.h
```diff
@@ -173,10 +176,20 @@ class PrimaryKeyIndexBuilder {
...
-    template<typename T, typename S = T>
-    inline HashIndexBuilder<T, S>* getTypedHashIndex(uint64_t indexPos) {
-        return common::ku_dynamic_cast<InMemHashIndex*, HashIndexBuilder<T, S>*>(
-            hashIndexBuilders[indexPos].get());
+    template<common::IndexHashable T>
+    using HashIndexType =
+        typename std::conditional<std::same_as<T, std::string_view> || std::same_as<T, std::string>,
+            common::ku_string_t, T>::type;
+    template<typename T>
+    inline HashIndexBuilder<HashIndexType<T>>* getTypedHashIndex(uint64_t indexPos) {
```
ray6080, line 184: I wonder if you can simplify the `if constexpr` here by making use of `HashIndexType`.
maybe something like following
```cpp
    template<common::IndexHashable T>
    using HashIndexType = typename std::conditional<std::same_as<T, std::string_view> || std::same_as<T, common::ku_string_t> || std::same_as<T, std::string>, common::ku_string_t, T>::type;
    template<typename T>
    inline HashIndexBuilder<HashIndexType<T>>* getTypedHashIndex(uint64_t indexPos) {
        if constexpr (std::same_as<HashIndexType<T>, common::ku_string_t>) {
            return common::ku_dynamic_cast<InMemHashIndex*, HashIndexBuilder<common::ku_string_t>*>(
                hashIndexBuilders[indexPos].get());
        } else {
            return common::ku_dynamic_cast<InMemHashIndex*, HashIndexBuilder<T>*>(
                hashIndexBuilders[indexPos].get());
        }
    }
```

src/include/storage/index/hash_index_builder.h
```diff
@@ -62,13 +69,21 @@ class HashIndexBuilder final : public InMemHashIndex {
     // Reserves space for at least the specified number of elements.
     void bulkReserve(uint32_t numEntries) override;
 
-    bool append(T key, common::offset_t value);
-    bool lookup(T key, common::offset_t& result);
+    using AppendKeyType =
```
ray6080, line 72: Maybe `BufferKeyType` is more accurate here? It looks to me that the type is more about the type info of stored keys inside the IndexBuffer.

src/include/storage/index/hash_index_builder.h
```diff
@@ -62,13 +69,21 @@ class HashIndexBuilder final : public InMemHashIndex {
...
+        typename std::conditional<std::same_as<T, common::ku_string_t>, std::string, T>::type;
+    // Appends the buffer to the index. Returns the number of values successfully inserted.
+    // I.e. if a key fails to insert, its index will be the return value
+    size_t append(const IndexBuffer<AppendKeyType>& buffer);
+    using KeyView =
```
ray6080, line 77: Consider `Key` here. `KeyView` is a bit confusing then the type is not string.

ray6080 (approved):

PR #2999, opened 2024-03-06
Avoid moving DictionaryChunks

benjaminwinger:
There was a regression in #2994 which wasn't caught by our test suite. On some datasets it was causing a segfault when copying rel tables.

`DictionaryChunk` now stores pointers in its internal cache which would be invalidated when moving and can't be easily updated (which is understandable, changing the hash function and equals functions in an unordered_set usually would be expected to break things, even if in this case it would be safe).

### DIFF
--- a/src/include/storage/store/string_column_chunk.h
+++ b/src/include/storage/store/string_column_chunk.h
@@ -31,11 +31,11 @@ public:
 
     uint64_t getStringLength(common::offset_t pos) const {
         auto index = ColumnChunk::getValue<DictionaryChunk::string_index_t>(pos);
-        return dictionaryChunk.getStringLength(index);
+        return dictionaryChunk->getStringLength(index);
     }
 
-    inline DictionaryChunk& getDictionaryChunk() { return dictionaryChunk; }
-    inline const DictionaryChunk& getDictionaryChunk() const { return dictionaryChunk; }
+    inline DictionaryChunk& getDictionaryChunk() { return *dictionaryChunk; }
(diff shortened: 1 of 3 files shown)


src/include/storage/store/string_column_chunk.h
```diff
@@ -31,11 +31,11 @@ class StringColumnChunk : public ColumnChunk {
 
     uint64_t getStringLength(common::offset_t pos) const {
         auto index = ColumnChunk::getValue<DictionaryChunk::string_index_t>(pos);
-        return dictionaryChunk.getStringLength(index);
+        return dictionaryChunk->getStringLength(index);
     }
 
-    inline DictionaryChunk& getDictionaryChunk() { return dictionaryChunk; }
-    inline const DictionaryChunk& getDictionaryChunk() const { return dictionaryChunk; }
+    inline DictionaryChunk& getDictionaryChunk() { return *dictionaryChunk; }
```
ray6080, line 37: I'm thinking to form a convention of naming functions returning non-const reference with suffix `UnSafe` (or maybe other suffix to be more explicit). What do you think? Is this necessary?

ray6080 (approved):

ray6080:
Do we have a minimal example to reproduce this? so we can add to our test suite.

ray6080:
@benjaminwinger I'm merging it so we can ship to for today's nightly build.

benjaminwinger:
I'll see if I can come up with an example. It's caused by accessing strings after `StringColumnChunk::finalize` (where the move happens), which I think only occurs if we're copying multiple node groups, so maybe we don't have any tests where we copy multiple node groups of strings.

PR #3009, opened 2024-03-08
Support use of `QueryResult` as a context manager, and add a `get_schema` method

alexander-beedie:
Currently `close()` is called on `QueryResult` either manually or via `__del__` (which is not deterministic as it relies on GC being triggered and may _never_ actually be called).

A cleaner pattern is to use context manager syntax; this both guarantees immediate connection close on scope exit, and gracefully handles the case where an error is raised before the connection is manually closed.

**Note:** this syntax is optional. If you don't use it, you get the existing behaviour (it's a non-breaking addition).

# Example

### Manual:

Requires a potentially fallible call to `close()`.
```python
res = conn.execute("MATCH (a:person) WHERE a.ID = 0 RETURN a")
df = res.get_as_pl()  # <-- if this raises "close" will NOT be called
res.close()
```
### Context manager:

Scope exit guarantees that `close()` will be called successfully.
```python
with conn.execute("MATCH (a:person) WHERE a.ID = 0 RETURN a") as res:
    df = res.get_as_pl()  # <-- if this raises "close" WILL still be called
````

# Also

* Added a handy `get_schema()` method (combines `get_column_names()` and `get_column_data_types()` as a dict).
* Made the chunk_size used by `get_as_pl()` adaptive, minimising the chance that Polars needs to rechunk the result.
* Added test coverage for all of the above updates.
* Added some "See Also" docstring sections.

# CLA

I have read and agree to the terms under `CLA.md` 👌

### DIFF
--- a/tools/python_api/src_py/query_result.py
+++ b/tools/python_api/src_py/query_result.py
@@ -23,6 +23,12 @@ class QueryResult:
         self._query_result = query_result
         self.is_closed = False
 
+    def __enter__(self):
+        return self
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        self.close()
+
     def __del__(self):
         self.close()
 
@@ -73,17 +79,21 @@ class QueryResult:
         Close the query result.
         """
 
-        if self.is_closed:
-            return
-        self._query_result.close()
-        # Allows the connection to be garbage collected if the query result
-        # is closed manually by the user.
-        self.connection = None
-        self.is_closed = True
+        if not self.is_closed:
+            # Allows the connection to be garbage collected if the query result
+            # is closed manually by the user.
+            self._query_result.close()
+            self.connection = None
+            self.is_closed = True
 
     def get_as_df(self):
         """
         Get the query result as a Pandas DataFrame.
+
+        See Also
+        --------
+        get_as_pl : Get the query result as a Polars DataFrame.
+        get_as_arrow : Get the query result as a PyArrow Table.
 
         Returns
         -------
@@ -102,6 +112,11 @@ class QueryResult:
         """
         Get the query result as a Polars DataFrame.
 
+        See Also
+        --------
+        get_as_df : Get the query result as a Pandas DataFrame.
+        get_as_arrow : Get the query result as a PyArrow Table.
+
         Returns
         -------
         polars.DataFrame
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: tools/python_api/src_py/query_result.py [hunk 1]
```diff
@@ -102,14 +112,26 @@ def get_as_pl(self):
...
             Query result as a Polars DataFrame.
         """
 
         import polars as pl
-        return pl.from_arrow(data=self.get_as_arrow(10_000))
+
+        target_n_elems = (
+            10_000_000  # adaptive chunk_size; target 10m elements per chunk
+        )
+        target_chunk_size = max(target_n_elems // len(self.get_column_names()), 10)
```
prrao87, line 131: This makes sense - for most cases in this setting, we'd have a single chunk that helps Polars retrieve it more efficiently, and it's also in line with the DuckDB heuristically selected value of 1M.

tools/python_api/src_py/query_result.py
```diff
@@ -23,6 +23,12 @@ def __init__(self, connection, query_result):
         self._query_result = query_result
         self.is_closed = False
 
+    def __enter__(self):
```
prrao87, line 26: @alexander-beedie Switching to the context manager protocol would count as a breaking change, correct? As in, the way it's done now `conn.execute(...)` would fail and the user is required to specify `with conn.execute` regardless of whether or not Polars is the desired output format?

alexander-beedie: Fortunately not; it's completely optional 😉 If you don't use it as a context manager then there's no change; you get identical behaviour to what happens now.

tools/python_api/src_py/query_result.py
alexander-beedie: Yup; each chunk will be broadly equivalent to 10 cols of 1m rows each.

mewim (approved):

mewim:
@alexander-beedie

Thanks for your contribution. 

Could you please take a look at our [CLA](https://github.com/kuzudb/kuzu/blob/master/CLA.md) and add “I have read and agree to the terms under CLA.md” to your PR description if you agree with the CLA?

alexander-beedie:
> Could you please take a look at our [CLA](https://github.com/kuzudb/kuzu/blob/master/CLA.md) and add “I have read and agree to the terms under CLA.md” to your PR description if you agree with the CLA?

Done ;)

PR #3012, opened 2024-03-08
Rewrite the Hash Index overflow file to support multiple copies

benjaminwinger:
See #2938

This merges the `InMemFile` (previously used by the `HashIndexBuilder` to write to the string overflow file) and the `DiskOverflowFile` (used by the persistent `HashIndex`) into a single `OverflowFile`, which combines elements of both into a system that can continue writing from where it left off before with all 256 concurrent hash indices.
The overflow file next byte pos record in the wal file has been replaced with a header page in the overflow file which stores the 256 `PageCursor`s that do the same thing, but support 256 concurrent writers (the header also stores a page counter that records the next page to be written to, which is necessary since newly added pages bypass the WAL file and get written directly to the overflow file, but in the case that the transaction is interrupted while flushing those pages will still remain in the file, so the counter in the header gets used instead of the size of the file).

I've benchmarked this and it has more or less identical copy performance (for the first copy). For additional copies we will see once support is finished, but additional copies should be more or less the same as the first, with the exception that writing to the header, and the first incomplete page for each of the 256 writers will go through the WAL file.


src/include/storage/storage_structure/overflow_file.h
```diff
@@ -0,0 +1,159 @@
...
+        const common::ku_string_t& keyInEntry) const;
+
+    common::ku_string_t writeString(std::string_view rawString);
+    inline common::ku_string_t writeString(const char* rawString) {
+        return writeString(std::string_view(rawString));
+    }
+
+    void prepareCommit();
+    inline void checkpointInMemory() { pageCache.clear(); }
+    inline void rollbackInMemory(PageCursor nextPosToWriteTo) {
```
ray6080, line 49: `nextPosToWriteTo` -> `nextPosToWriteTo_`.

src/include/storage/storage_structure/overflow_file.h
```diff
@@ -0,0 +1,159 @@
...
+    }
+
+    // Handles contain a reference to the overflow file
+    OverflowFile(OverflowFile&& other) = delete;
+
+    void rollbackInMemory();
+    void prepareCommit();
+    void checkpointInMemory();
+
+    static inline DBFileIDAndName constructDBFileIDAndName(
```
ray6080, line 107: Can you double check if this function needs to be public?

src/include/storage/storage_structure/overflow_file.h
```diff
@@ -0,0 +1,159 @@
...
+    }
+
+    inline OverflowFileHandle* addHandle() {
+        KU_ASSERT(handles.size() < NUM_HASH_INDEXES);
+        handles.emplace_back(
+            std::make_unique<OverflowFileHandle>(*this, header.cursors[handles.size()]));
+        return handles.back().get();
+    }
+
+    inline common::page_idx_t getNewPageIdx() {
```
ray6080, line 122: Ditto here on public.

src/include/storage/storage_structure/overflow_file.h
```diff
@@ -0,0 +1,159 @@
...
+        // If this isn't the first call reserving the page header, then the header flag must be set
+        // prior to this
+        KU_ASSERT(pageCounter == 0 || headerChanged);
+        return pageCounter.fetch_add(1);
+    }
+
+    inline void setHeaderChanged() { headerChanged = true; }
```
ray6080, line 129: Ditto here on public.

src/include/storage/storage_structure/overflow_file.h
```diff
@@ -0,0 +1,159 @@
...
+
+    inline BufferManager& getBM() { return *bufferManager; }
```
ray6080, line 131: Can you double check if `getBM()` and `getWAL()` is used anywhere? I don't think OverflowFile is the right place to get access to a BM or WAL.

src/include/storage/storage_structure/overflow_file.h
```diff
@@ -0,0 +1,159 @@
...
+
+    inline WAL& getWAL() { return *wal; }
+
+    inline BMFileHandle* getBMFileHandle() const {
+        return common::ku_dynamic_cast<FileHandle*, BMFileHandle*>(fileHandle.get());
+    }
+
+    inline DBFileID& getFileID() { return dbFileID; }
```
ray6080, line 139: Remove this function.

src/storage/storage_structure/overflow_file.cpp
```diff
@@ -0,0 +1,232 @@
...
+    }
+}
+
+ku_string_t OverflowFileHandle::writeString(std::string_view rawString) {
+    ku_string_t result;
+    result.len = rawString.length();
+    auto shortStrLen = ku_string_t::SHORT_STR_LENGTH;
+    auto inlineLen = std::min(shortStrLen, (uint64_t)result.len);
+    memcpy(result.prefix, rawString.data(), inlineLen);
+    overflowFile.setHeaderChanged();
```
ray6080, line 139: I don't follow L139. Why do we set header to changed here? Shouldn't we set it inside `setStringOverflow` when the string is not short?

src/storage/storage_structure/overflow_file.cpp
```diff
@@ -0,0 +1,232 @@
...
+    }
+}
+
+void OverflowFile::checkpointInMemory() {
+    headerChanged = false;
+    numPagesOnDisk = pageCounter;
+}
+
+void OverflowFile::rollbackInMemory() {
+    readFromDisk(transaction::TransactionType::READ_ONLY, 0,
```
ray6080, line 222: Can you turn `0` into a constant, e.g., `HEADER_PAGE` ?

src/include/storage/storage_structure/overflow_file.h
```diff
@@ -0,0 +1,159 @@
...
+        const std::function<void(uint8_t*)>& func) const;
+
+private:
+    static const common::page_idx_t END_OF_PAGE =
+        common::BufferPoolConstants::PAGE_4KB_SIZE - sizeof(common::page_idx_t);
+    // This is the index of the last free byte to which we can write.
+    PageCursor& nextPosToWriteTo;
+    OverflowFile& overflowFile;
+    // Cached pages which have been written in the current transaction
+    std::unordered_map<common::page_idx_t, std::unique_ptr<InMemPage>> pageCache;
```
ray6080, line 69: You can rename `pageCache` to `pageWriteCache` to be more explicit that the cache is only triggered by page writes, but not reads.

ray6080 (approved):

src/include/storage/storage_structure/overflow_file.h
benjaminwinger: For this one, and `setHeaderChanged`, they are used by the handles. I guess the handles could be made friend classes if we want to have a more restricted interface for the hash index to use.

src/storage/storage_structure/overflow_file.cpp
benjaminwinger: Yes

PR #3019, opened 2024-03-11
Rework VAR_LIST

hououou:
addtionally store length to var list

### DIFF
--- a/src/include/storage/store/var_list_column.h
+++ b/src/include/storage/store/var_list_column.h
@@ -25,19 +25,19 @@ namespace kuzu {
 namespace storage {
 
 struct ListOffsetInfoInStorage {
-    common::offset_t prevNodeListOffset;
+    common::offset_t numTotal;
     std::vector<std::unique_ptr<common::ValueVector>> offsetVectors;
+    std::vector<std::unique_ptr<common::ValueVector>> sizeVectors;
 
-    ListOffsetInfoInStorage(common::offset_t prevNodeListOffset,
-        std::vector<std::unique_ptr<common::ValueVector>> offsetVectors)
-        : prevNodeListOffset{prevNodeListOffset}, offsetVectors{std::move(offsetVectors)} {}
+    ListOffsetInfoInStorage(common::offset_t numTotal,
+        std::vector<std::unique_ptr<common::ValueVector>> offsetVectors,
+        std::vector<std::unique_ptr<common::ValueVector>> sizeVectors)
+        : numTotal{numTotal}, offsetVectors{std::move(offsetVectors)}, sizeVectors{std::move(
+                                                                           sizeVectors)} {}
 
-    common::offset_t getListOffset(uint64_t nodePos) const;
-
-    inline uint64_t getListLength(uint64_t nodePos) const {
-        KU_ASSERT(getListOffset(nodePos + 1) >= getListOffset(nodePos));
-        return getListOffset(nodePos + 1) - getListOffset(nodePos);
-    }
+    uint64_t getListLength(uint64_t nodePos) const;
+    common::offset_t getListEndOffset(uint64_t nodePos) const;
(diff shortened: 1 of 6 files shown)


src/include/storage/store/var_list_column.h
```diff
@@ -25,19 +25,19 @@ namespace kuzu {
...
-        : prevNodeListOffset{prevNodeListOffset}, offsetVectors{std::move(offsetVectors)} {}
+    ListOffsetInfoInStorage(common::offset_t numTotal,
+        std::vector<std::unique_ptr<common::ValueVector>> offsetVectors,
+        std::vector<std::unique_ptr<common::ValueVector>> sizeVectors)
+        : numTotal{numTotal}, offsetVectors{std::move(offsetVectors)}, sizeVectors{std::move(
+                                                                           sizeVectors)} {}
 
+    uint64_t getListLength(uint64_t nodePos) const;
+    common::offset_t getEndListOffset(uint64_t nodePos) const;
     common::offset_t getListOffset(uint64_t nodePos) const;
```
ray6080, line 40: `getListOffset` -> `getListStartOffset`

src/include/storage/store/var_list_column.h
```diff
@@ -25,19 +25,19 @@ namespace kuzu {
...
-        std::vector<std::unique_ptr<common::ValueVector>> offsetVectors)
-        : prevNodeListOffset{prevNodeListOffset}, offsetVectors{std::move(offsetVectors)} {}
+    ListOffsetInfoInStorage(common::offset_t numTotal,
+        std::vector<std::unique_ptr<common::ValueVector>> offsetVectors,
+        std::vector<std::unique_ptr<common::ValueVector>> sizeVectors)
+        : numTotal{numTotal}, offsetVectors{std::move(offsetVectors)}, sizeVectors{std::move(
+                                                                           sizeVectors)} {}
 
+    uint64_t getListLength(uint64_t nodePos) const;
+    common::offset_t getEndListOffset(uint64_t nodePos) const;
```
ray6080, line 39: `getEndListOffset` -> `getListEndOffset`

src/storage/store/var_list_column.cpp
```diff
@@ -12,12 +12,24 @@ namespace kuzu {
...
-        return offsetVector->getValue<offset_t>((nodePos - 1) % DEFAULT_VECTOR_CAPACITY);
-    }
+    return getEndListOffset(nodePos) - getListLength(nodePos);
+}
+
+offset_t ListOffsetInfoInStorage::getEndListOffset(uint64_t nodePos) const {
+    if (offsetVectors.size() == 0)
+        return offset_t(0);
+    if (nodePos == numTotal)
+        return getEndListOffset(nodePos - 1);
```
ray6080, line 22: I don't think these two if conditions make sense.
For the first one, I don't think we should allow accessing under this case `offsetVectors.size() == 0`;
For the second one, should also not happen.

src/storage/store/var_list_column.cpp
```diff
@@ -145,19 +182,26 @@ void VarListColumn::scanFiltered(Transaction* transaction, node_group_idx_t node
         auto pos = resultVector->state->selVector->selectedPositions[i];
         auto startOffsetInStorageToScan = listOffsetInfoInStorage.getListOffset(pos);
         auto endOffsetInStorageToScan = listOffsetInfoInStorage.getListOffset(pos + 1);
+        auto dataVector = ListVector::getDataVector(resultVector);
+        if (!dataVector->state) {
+            dataVector->setState(std::make_unique<DataChunkState>());
```
ray6080, line 187: Double check here to see when `dataVector->state` is nullptr.
We might have set state for dataVector somewhere else. better understand this better before we add an ad-hoc `if` check here.

ray6080:
See #3093.

Issue #3022, opened 2024-03-11
Build fails with VS2019 due to error C3855


mewim:
See https://github.com/kuzudb/kuzu/actions/runs/8229430092/job/22500623754

The compilation fails on VS2019 due to:

```
error C3855: 'kuzu::storage::HashIndexBuilder<T,S>': template parameter 'T' is incompatible with the declaration
```

We should either fix this for VS2019 or update the [docs](https://docs.kuzudb.com/development/building-kuzu) to reflect that we require VS2022 at least on Winodws.

PR #3030, opened 2024-03-12
Remove the constraint on HashIndexBuilder's template parameter

benjaminwinger:
The constraint seems to cause issues with MSVC 19.27 (#3022), and I don't think it's necessary any more.

This fixes the compilation failure (see https://github.com/kuzudb/kuzu/actions/runs/8250693752), however there are now some tests failing on the VS2019 job, I'm not really sure why.

~~I've pulled a change to the typing (merging the two template parameters) out of #2997 as I didn't really want to modify the template parameters twice.~~

### DIFF
--- a/src/include/storage/index/hash_index_builder.h
+++ b/src/include/storage/index/hash_index_builder.h
@@ -2,7 +2,6 @@
 
 #include "common/static_vector.h"
 #include "common/type_utils.h"
-#include "common/types/internal_id_t.h"
 #include "common/types/ku_string.h"
 #include "common/types/types.h"
 #include "storage/index/hash_index_header.h"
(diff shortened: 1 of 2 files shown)


src/include/storage/index/hash_index_builder.h
```diff
@@ -62,13 +63,15 @@ class HashIndexBuilder final : public InMemHashIndex {
     // Reserves space for at least the specified number of elements.
     void bulkReserve(uint32_t numEntries) override;
 
-    bool append(T key, common::offset_t value);
-    bool lookup(T key, common::offset_t& result);
+    using KeyView =
```
ray6080, line 66: I think `Key` is fine here. Same from this comment https://github.com/kuzudb/kuzu/pull/2997#discussion_r1521748116

src/include/storage/index/hash_index_builder.h
```diff
@@ -173,10 +174,20 @@ class PrimaryKeyIndexBuilder {
     common::PhysicalTypeID keyTypeID() const { return keyDataTypeID; }
 
 private:
-    template<typename T, typename S = T>
-    inline HashIndexBuilder<T, S>* getTypedHashIndex(uint64_t indexPos) {
-        return common::ku_dynamic_cast<InMemHashIndex*, HashIndexBuilder<T, S>*>(
-            hashIndexBuilders[indexPos].get());
+    template<common::IndexHashable T>
```
ray6080, line 177: Same comment here https://github.com/kuzudb/kuzu/pull/2997#discussion_r1521738453.

ray6080 (approved):

PR #3032, opened 2024-03-12
Optimise Python unit test runtime (~7x speedup)

alexander-beedie:
## Update

* Each of the 88 tests was recreating a fresh Database from scratch, reloading/parsing all data, etc. 
* Optimised this by creating as a read-only DB fixture _once_, and caching it: 
  * `conn_db_readonly`.
* Only the 3 tests that require a writable DB get a different database; there are two writable fixtures, one of which is also cached (and this is the one that gets used, as the current set of tests requiring write-access don't step on each other):
  * `conn_db_writable_cached`
  * `conn_db_writable_new`.

## Speedup
_(These timings were generated locally on Apple Silicon M2)_
```
Before: ~37 secs
 After: ~5½ secs 🚀 
```
**Note:** not worth adding `xdist` parallelism, as the initial setup cost leads to a higher total runtime (8-9 secs).


mewim:
@alexander-beedie Thanks for your contribution. I think it makes a lot of sense to cache database for testing instead of create a new one per test. For write tests though, I think there might be some risks in caching the database. It is not easy to tell whether a write test would conflict with other ones or not and may cause confusing errors when modifications on a test causes other tests to fail. So I changed the code a bit to only keep `conn_db_readonly` and `conn_db_readwrite` (which always create a new database for writing). I also added a minor workaround and notes as read only mode is not currently supported properly on Windows. I think this is now ready to be merged.

mewim (approved):

alexander-beedie:
> @alexander-beedie I changed the code a bit to only keep `conn_db_readonly` and `conn_db_readwrite` (which always create a new database for writing). 

Good call; as there are only three tests that need it the performance difference is marginal, and it's better to always do the right thing rather than chase a few % more speed in the tests ;)

PR #3034, opened 2024-03-12
First-pass lint/format for Python `shell` tests

alexander-beedie:
Along the same lines as https://github.com/kuzudb/kuzu/pull/3023.

* Applies formatting and a subset of the lint rules (a follow-up can handle the remaining lint).
* Does not apply any function signature or return  typing yet.


mewim:
@MSebanc could you please review this as you wrote most of the shell tests?

MSebanc (approved):
LGTM! 🎉

PR #3051, opened 2024-03-13
Implemented Progress Bar

MSebanc:


src/processor/operator/scan_node_id.cpp
```diff
@@ -73,6 +74,8 @@ bool ScanNodeID::getNextTuplesInternal(ExecutionContext* context) {
         setSelVector(context, state, startOffset, endOffset);
     } while (outValueVector->state->selVector->selectedSize == 0);
     metrics->numOutputTuple.increase(outValueVector->state->selVector->selectedSize);
+    context->clientContext->progressBar->finishJobsInPipeline(
```
andyfengHKU, line 77: I suggest let's not worry about other source operators but only implement for `ScanNodeID` in our initial version.

I would add an `double getProgress()` interface in `PhysicalOperator` and each source operator implement this interface. That means each source operator need to maintain its own progress (usually in shared state). For `ScanNodeID`, you just need to additionally track how many nodeIDs have been scanned. The total number of nodeIDs is available in the table.

If we do this then you most likely do not need `addJobsToPipeline` and `finishJobsInPipeline`

src/processor/processor.cpp
```diff
@@ -25,12 +25,16 @@ std::shared_ptr<FactorizedTable> QueryProcessor::execute(
...
     initTask(task.get());
+    context->clientContext->progressBar->startProgress();
     taskScheduler->scheduleTaskAndWaitOrError(task, context);
     return resultCollector->getResultFactorizedTable();
 }
 
 void QueryProcessor::decomposePlanIntoTask(
     PhysicalOperator* op, Task* task, ExecutionContext* context) {
+    if (op->isSource()) {
+        context->clientContext->progressBar->addPipeline();
```
andyfengHKU, line 36: This is irrelevant to this PR but I think it might be better to introduce Pipeline as another abstraction level.

andyfengHKU (commented):
One thing that I didn't pick up from this draft is that at what frequency are we printing the progress bar. Maybe let's discuss that.

src/common/task_system/task.cpp
```diff
@@ -23,6 +23,7 @@ void Task::deRegisterThreadAndFinalizeTask() {
         } catch (std::exception& e) { setExceptionNoLock(std::current_exception()); }
     }
     if (isCompletedNoLock()) {
+        updateProgress();
```
andyfengHKU, line 26: As we discussed, this may not be a good place to update progress. So let's refresh each time we update but only if an amount of work (say 1%) has been finished since last refresh.

src/include/processor/operator/scan_node_id.h
```diff
@@ -25,6 +25,8 @@ class NodeTableScanState {
 
     std::pair<common::offset_t, common::offset_t> getNextRangeToRead();
 
+    common::offset_t getCurrentNodeOffset() const { return currentNodeOffset; }
```
andyfengHKU, line 28: remove

src/include/processor/operator/scan_node_id.h
```diff
@@ -42,6 +44,7 @@ class ScanNodeIDSharedState {
     }
     inline uint32_t getNumTableStates() const { return tableStates.size(); }
     inline NodeTableScanState* getTableState(uint32_t idx) const { return tableStates[idx].get(); }
+    inline NodeTableScanState* getCurrentTableState();
```
andyfengHKU, line 47: ditto

src/include/processor/processor_task.h
```diff
@@ -14,14 +14,18 @@ class ProcessorTask : public common::Task {
...
 private:
     static std::unique_ptr<ResultSet> populateResultSet(
         Sink* op, storage::MemoryManager* memoryManager);
 
+    void updateProgress() const override;
+
 private:
     bool sharedStateInitialized;
     Sink* sink;
+    std::vector<PhysicalOperator*> sourceOps;
```
andyfengHKU, line 28: By definition, a process task can only have 1 source. So just use `PhysicalOperator*`

src/include/processor/processor_task.h
```diff
@@ -14,14 +14,18 @@ class ProcessorTask : public common::Task {
 
     void run() override;
     void finalizeIfNecessary() override;
+    void addSource(PhysicalOperator* source);
```
andyfengHKU, line 17: instead of addSource, you might wanna just change the constructor

MSebanc: I implemented this so I could update the progress bar in the `Task::deRegisterThreadAndFinalizeTask()` so this will be removed as it will be unused after I move the `updateProgress()` function

src/include/processor/processor_task.h
MSebanc: Same as the reply to the comment below

src/processor/operator/scan_node_id.cpp
```diff
@@ -98,5 +106,14 @@ void ScanNodeID::setSelVector(ExecutionContext* context, NodeTableScanState* tab
         context->clientContext->getTx(), outValueVector);
 }
 
+double ScanNodeID::getProgress(ExecutionContext* context) const {
+    if (!sharedState->getCurrentTableState()) {
```
andyfengHKU, line 110: As we discussed, let's keep `numNodes` and `numNodesScanned` in shared state. `numNodes` should be updated once only in `initGlobalStateInternal` and `numNodesScanned` is updated each getMorsel is called.

src/include/processor/operator/physical_operator.h
```diff
@@ -136,6 +136,7 @@ class PhysicalOperator {
         }
         metrics->executionTime.start();
         auto result = getNextTuplesInternal(context);
+        context->clientContext->progressBar->updateProgress(getProgress(context));
```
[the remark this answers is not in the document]
MSebanc: That should be handled as prevCurPipelineProgress would also be 0 if the progress is 0 so it won't print

PR #3060, opened 2024-03-14
Rework VAR_LIST storage layout

hououou:
Previously, we only stored the end offset in the VarList storage layout. 
For an example layout for four lists of INT64:  [4,7,8,12], null, [2, 3], []. We store as follows: 
```
Offset column: [4, 4, 6, 6]
data column: [4, 7, 8, 12, 2, 3]
```
We need to ensure that the offset column must be in ascending order. So when we update [4,7,8,12] to [4,5,6,7,8], we need to rewrite the whole varlist column chunk, both offset and data column chunk. 
```
Offset column: [5, 5, 7, 7]
data column: [4, 5, 6, 7, 8, 2, 3]
```

It will take time to rewrite the column chunk. To solve it, we additionally store the length of VarList. So the data layout is just  as follows. 
```
Offset column: [4, 4, 6, 6]
Size column: [4, 0, 2, 0]
data column: [4, 7, 8, 12, 2, 3]
```
When we update, we just append data in the back of the data column chunk and only update the offset and size column chunk. The varlist column chunk will be
```
Offset column: [10, 4, 6, 6]
Size column: [5, 0, 2, 0]
data column: [4, 7, 8, 12, 2, 3, 4, 5, 6, 7, 8]
```

### DIFF
--- a/src/include/storage/store/var_list_column.h
+++ b/src/include/storage/store/var_list_column.h
@@ -1,6 +1,7 @@
 #pragma once
 
 #include "column.h"
+#include "var_list_column_chunk.h"
 
 // List is a nested data type which is stored as two chunks:
 // 1. Offset column (type: INT64). Using offset to partition the data column into multiple lists.
@@ -24,20 +25,21 @@
 namespace kuzu {
 namespace storage {
 
-struct ListOffsetInfoInStorage {
-    common::offset_t prevNodeListOffset;
-    std::vector<std::unique_ptr<common::ValueVector>> offsetVectors;
+struct ListOffsetSizeInfo {
+    common::offset_t numTotal;
+    std::unique_ptr<ColumnChunk> offsetColumnChunk;
+    std::unique_ptr<ColumnChunk> sizeColumnChunk;
 
-    ListOffsetInfoInStorage(common::offset_t prevNodeListOffset,
-        std::vector<std::unique_ptr<common::ValueVector>> offsetVectors)
-        : prevNodeListOffset{prevNodeListOffset}, offsetVectors{std::move(offsetVectors)} {}
+    ListOffsetSizeInfo(common::offset_t numTotal, std::unique_ptr<ColumnChunk> offsetColumnChunk,
+        std::unique_ptr<ColumnChunk> sizeColumnChunk)
+        : numTotal{numTotal}, offsetColumnChunk{std::move(offsetColumnChunk)},
+          sizeColumnChunk{std::move(sizeColumnChunk)} {}
 
-    common::offset_t getListOffset(uint64_t nodePos) const;
+    uint32_t getListLength(uint64_t pos) const;
+    common::offset_t getListEndOffset(uint64_t pos) const;
+    common::offset_t getListStartOffset(uint64_t pos) const;
 
-    inline uint64_t getListLength(uint64_t nodePos) const {
-        KU_ASSERT(getListOffset(nodePos + 1) >= getListOffset(nodePos));
-        return getListOffset(nodePos + 1) - getListOffset(nodePos);
-    }
+    bool isOffsetSortedAscending(uint64_t startPos, uint64_t endPos) const;
 };
 
 class VarListColumn : public Column {
@@ -69,18 +71,11 @@ protected:
     void append(ColumnChunk* columnChunk, uint64_t nodeGroupIdx) override;
 
 private:
-    inline common::offset_t readListOffsetInStorage(transaction::Transaction* transaction,
-        common::node_group_idx_t nodeGroupIdx, common::offset_t offsetInNodeGroup) {
-        return offsetInNodeGroup == 0 ?
-                   0 :
-                   readOffset(transaction, nodeGroupIdx, offsetInNodeGroup - 1);
-    }
-
     void scanUnfiltered(transaction::Transaction* transaction,
         common::node_group_idx_t nodeGroupIdx, common::ValueVector* resultVector,
-        const ListOffsetInfoInStorage& listOffsetInfoInStorage);
+        const ListOffsetSizeInfo& listOffsetInfoInStorage);
     void scanFiltered(transaction::Transaction* transaction, common::node_group_idx_t nodeGroupIdx,
-        common::ValueVector* offsetVector, const ListOffsetInfoInStorage& listOffsetInfoInStorage);
+        common::ValueVector* offsetVector, const ListOffsetSizeInfo& listOffsetInfoInStorage);
 
     inline bool canCommitInPlace(transaction::Transaction*, common::node_group_idx_t,
         const ChunkCollection&, const offset_to_row_idx_t&, const ChunkCollection&,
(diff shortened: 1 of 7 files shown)


src/include/storage/store/var_list_column.h
```diff
@@ -102,13 +98,19 @@ class VarListColumn : public Column {
...
+
     ListOffsetInfoInStorage getListOffsetInfoInStorage(transaction::Transaction* transaction,
         common::node_group_idx_t nodeGroupIdx, common::offset_t startOffsetInNodeGroup,
         common::offset_t endOffsetInNodeGroup,
         const std::shared_ptr<common::DataChunkState>& state);
 
 private:
+    std::unique_ptr<Column> sizeColumn;
     std::unique_ptr<Column> dataColumn;
+    std::unique_ptr<VarListDataColumnChunk> tmpDataColumnChunk;
```
ray6080, line 113: Why do you need a temp column chunk as a class field? Should remove it.

src/storage/stats/table_statistics_collection.cpp
```diff
@@ -97,6 +97,8 @@ std::unique_ptr<MetadataDAHInfo> TablesStatistics::createMetadataDAHInfo(
         }
     } break;
     case PhysicalTypeID::VAR_LIST: {
+        metadataDAHInfo->childrenInfos.push_back(
+            createMetadataDAHInfo(*LogicalType::INT64(), metadataFH, bm, wal));
```
ray6080, line 101: UINT32

src/storage/store/var_list_column.cpp
```diff
@@ -11,13 +11,37 @@ using namespace kuzu::transaction;
 namespace kuzu {
 namespace storage {
 
-offset_t ListOffsetInfoInStorage::getListOffset(uint64_t nodePos) const {
-    if (nodePos == 0) {
-        return prevNodeListOffset;
-    } else {
-        auto offsetVector = offsetVectors[(nodePos - 1) / DEFAULT_VECTOR_CAPACITY].get();
-        return offsetVector->getValue<offset_t>((nodePos - 1) % DEFAULT_VECTOR_CAPACITY);
+offset_t ListOffsetInfoInStorage::getListStartOffset(uint64_t nodePos) const {
```
ray6080, line 14: `nodePos` -> `pos`. Ditto on all `nodePos`

src/include/storage/store/var_list_column.h
```diff
@@ -25,19 +26,21 @@ namespace kuzu {
...
-    common::offset_t getListOffset(uint64_t nodePos) const;
+    uint64_t getListLength(uint64_t nodePos) const;
+    common::offset_t getListEndOffset(uint64_t nodePos) const;
+    common::offset_t getListStartOffset(uint64_t nodePos) const;
 
-    inline uint64_t getListLength(uint64_t nodePos) const {
-        KU_ASSERT(getListOffset(nodePos + 1) >= getListOffset(nodePos));
-        return getListOffset(nodePos + 1) - getListOffset(nodePos);
-    }
+    bool checkOffsetOder(uint64_t startPos, uint64_t endPos) const;
```
ray6080, line 43: `checkOffsetOder` -> `isOffsetSortedAscending`

src/storage/store/var_list_column.cpp
```diff
@@ -11,13 +11,37 @@ using namespace kuzu::transaction;
...
+    return offsetVector->getValue<offset_t>((nodePos) % DEFAULT_VECTOR_CAPACITY);
+}
+
+uint64_t ListOffsetInfoInStorage::getListLength(uint64_t nodePos) const {
+    KU_ASSERT(nodePos < sizeVectors.size() * common::DEFAULT_VECTOR_CAPACITY);
+    auto sizeVector = sizeVectors[nodePos / common::DEFAULT_VECTOR_CAPACITY].get();
+    return sizeVector->getValue<uint64_t>(nodePos % common::DEFAULT_VECTOR_CAPACITY);
+}
+
+bool ListOffsetInfoInStorage::checkOffsetOder(uint64_t startPos, uint64_t endPos) const {
```
ray6080, line 34: Why not just check if end offsets are sorted ascending? This code should be much simplified if so.

src/include/storage/store/var_list_column.h
```diff
@@ -25,19 +26,21 @@ namespace kuzu {
 namespace storage {
 
 struct ListOffsetInfoInStorage {
```
ray6080, line 28: `ListOffsetInfoInStorage` -> `ListOffsetSizeInfo`

src/include/storage/store/var_list_column.h
```diff
@@ -25,19 +26,21 @@ namespace kuzu {
...
-    common::offset_t prevNodeListOffset;
+    common::offset_t numTotal;
     std::vector<std::unique_ptr<common::ValueVector>> offsetVectors;
+    std::vector<std::unique_ptr<common::ValueVector>> sizeVectors;
```
ray6080, line 31: You can rework `offsetVectors` and `sizeVectors` into ColumnChunk, instead of a vector of ValueVector. Using ColumnChunk should simplify code in this class.

src/storage/store/var_list_column_chunk.cpp
```diff
@@ -195,34 +223,17 @@ void VarListColumnChunk::copyListValues(const list_entry_t& entry, ValueVector*
 }
 
 void VarListColumnChunk::finalize() {
```
ray6080, line 225: You can remove the implementation of `finalize` from `VarListColumnChunk`.

src/storage/store/var_list_column_chunk.cpp
```diff
@@ -94,78 +104,95 @@ void VarListColumnChunk::append(ValueVector* vector, SelectionVector& selVector)
...
 }
 
 void VarListColumnChunk::write(
     ValueVector* vector, offset_t offsetInVector, offset_t offsetInChunk) {
-    needFinalize = true;
-    if (!indicesColumnChunk) {
-        initializeIndices();
+    auto selVector = std::make_shared<SelectionVector>(1);
+    selVector->resetSelectorToValuePosBuffer();
+    selVector->selectedPositions[0] = offsetInVector;
```
ray6080, line 151: Is `selVector` used anywhere?

src/storage/store/var_list_column_chunk.cpp
```diff
@@ -29,6 +29,8 @@ VarListColumnChunk::VarListColumnChunk(
     LogicalType dataType, uint64_t capacity, bool enableCompression, bool inMemory)
     : ColumnChunk{std::move(dataType), capacity, enableCompression, true /* hasNullChunk */},
       needFinalize{false} {
+    sizeColumnChunk = ColumnChunkFactory::createColumnChunk(
```
ray6080, line 32: UINT32()

src/storage/store/var_list_column_chunk.cpp
```diff
@@ -37,25 +39,31 @@ VarListColumnChunk::VarListColumnChunk(
...
+        auto startOffset = otherListChunk->getListStartOffset(startPosInOtherChunk + i);
+        auto appendLen = otherListChunk->getListLen(startPosInOtherChunk + i);
+        varListDataColumnChunk->dataColumnChunk->append(
+            otherListChunk->varListDataColumnChunk->dataColumnChunk.get(), startOffset, appendLen);
+    }
 }
 
 void VarListColumnChunk::resetToEmpty() {
     ColumnChunk::resetToEmpty();
+    sizeColumnChunk = ColumnChunkFactory::createColumnChunk(
```
ray6080, line 64: just `resetToEmpty` should be fine for sizeColumnChunk.

src/storage/store/var_list_column_chunk.cpp
```diff
@@ -94,78 +104,95 @@ void VarListColumnChunk::append(ValueVector* vector, SelectionVector& selVector)
 }
 
 void VarListColumnChunk::appendNullList() {
-    auto nextListOffsetInChunk = getListOffset(numValues);
+    offset_t nextListOffsetInChunk = varListDataColumnChunk->getNumValues();
     auto offsetBufferToWrite = (offset_t*)(buffer.get());
+    sizeColumnChunk->setValue(int64_t(0), numValues);
```
ray6080, line 109:
```
    sizeColumnChunk->setValue<uint32_t>(0, numValues);
```

ray6080:
Can you put some benchmark numbers here?
1. COPY node/rel with a var list property.  You can populate some csv files of 10M tuples.
2. See if you can also benchmark inserting 1000 of nodes/rels after COPY.

ray6080:
See #3093.

PR #3074, opened 2024-03-18
Remove unnecessary components for pip package

mewim:


ray6080 (approved):

mewim:
Fix verified by https://github.com/kuzudb/kuzu/actions/runs/8326125142
It shows Rust deployment as broken, which I opened as another issue: #3075

PR #3077, opened 2024-03-18
Make Property properly use explicit copy

benjaminwinger:
It should have a private copy constructor, and `vector<Property>` can be copied using `copyVector`

### DIFF
--- a/src/include/catalog/property.h
+++ b/src/include/catalog/property.h
@@ -20,9 +20,6 @@ public:
         common::property_id_t propertyID, common::table_id_t tableID)
         : name{std::move(name)}, dataType{std::move(dataType)},
           propertyID{propertyID}, tableID{tableID} {}
-    Property(const Property& other)
-        : name{other.name}, dataType{other.dataType->copy()},
-          propertyID{other.propertyID}, tableID{other.tableID} {}
     EXPLICIT_COPY_DEFAULT_MOVE(Property);
 
     std::string getName() const { return name; }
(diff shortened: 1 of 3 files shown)


andyfengHKU (approved):

src/include/catalog/property.h
```diff
@@ -41,6 +38,11 @@ class Property {
     static void toCypher(
         const std::vector<kuzu::catalog::Property>& properties, std::stringstream& ss);
 
+private:
+    Property(const Property& other)
```
andyfengHKU, line 42: I understand this change.

But why was a public constructor causing a problem on Windows. I thought it will always allow copy assignment everywhere.

andyfengHKU (approved):

src/include/catalog/property.h
benjaminwinger: It wasn't. The issue on windows was that it was generating a default `operator=` for `TableCatalogEntry`, which stores a `vector<Property>` and the default uses the deleted `Property::operator=` (maybe it always generates the default with `dllexport`/`KUZU_API`).

But I also saw this and thought it should be corrected to match the pattern, particularly since it's a little odd to have a deleted `operator=` and a public copy constructor.
Though now in #3071 I guess `TableCatalogEntry` will have the same issue, but that has a separate `copy` function that produces a `unique_ptr` which would conflict with the one produced by the `EXPLICIT_COPY_METHOD` macro. I suppose it might be able to also define `operator=` manually using `copyVector` and the move constructor instead of `operator=` for assigning the properties, but that's more work and it might be worth considering replacing the explicit copy producing a `unique_ptr` with the explicit copy from `copy_constructors.h` to be more consistent.

benjaminwinger:
Changes from this were incorporated into #3071

PR #3078, opened 2024-03-18
Add scan function replacement

andyfengHKU:
Add `ScanReplacementData` or external object scanning. Each external scan should provide a `scan_replace_func_t` that can return its own scan (table) function.


andyfengHKU:
This was merged in #3058

PR #3084, opened 2024-03-19
Arrow chunk_size as keyword argument

### LINKED ISSUE #2998: Make `chunk_size` in `get_as_arrow` an optional keyword argument
Currently, `chunk_size` is a [positional argument](https://github.com/kuzudb/kuzu/blob/9e2399572926d4d65c18f81558c39e71e139025b/tools/python_api/src_py/query_result.py#L114) in Python. This requires a new user to experience a runtime error before realizing this fact. Pandas has no such requirement (`get_as_df` has no arguments). We could instead make the `chunk_size` an optional kwarg, and set it to a default of 10000.

In fact, this is what Polars has done as per [this discussion](https://github.com/pola-rs/polars/pull/14822#discussion_r1510166298). What do you think about this @ray6080? It won't break any existing functionality and 10000 seems like a reasonable default.

prrao87:
Closes #2998.

As per @ray6080's [comment](https://github.com/kuzudb/kuzu/issues/2998#issuecomment-1983000247), setting the arrow record batch size (which we call `chunk_size`) to 1M is a reasonable default because DuckDB does the same. In the majority of cases, a larger record batch size is favourable, and the user can always bring down the `chunk_size` if necessary.

I also fixed the pyarrow tests to not specify the `chunk_size` argument as low integer values.


prrao87:
@mewim what do you think about moving the `get_as_polars` adaptive chunk size estimation logic over to the `get_as_arrow` method? That way Polars could just benefit from the arrow logic directly and not have to specify it independently.

prrao87:
> moving the get_as_pl adaptive chunk size estimation logic for polars over to the get_as_arrow method

I just made those changes and built & tested locally, works well 👌🏽.

alexander-beedie:
> fixed the pyarrow tests to not specify the `chunk_size` argument as low integer values.

FYI: I believe they were specified this way in order to validate result acquisition across chunk boundaries 🤔

prrao87:
Hi @alexander-beedie, great points.

1. It makes sense to revert to the small chunk size numbers in the tests so as to validate result acquisition across chunk boundaries as you mentioned
2. For the opt-in behaviour for `get_as_pl` as you mentioned (where `chunk_size=None`, we would still perform adaptive logic in the `get_as_arrow` method, correct?
3. I'm not sure what you meant by the last comment (the "new default") not being a breaking change. Do you mean breaking change w.r.t the old approach that you committed, or w.r.t. this new way where we do the adaptive logic inside `get_as_arrow`?

In any case, I think we're converging towards an agreeable and better solution, so thanks again :)

mewim:
I would probably keep the chunk size as referring to number of rows and use adaptive chunk size only if chunk size is set to None / 0 / -1.

prrao87:
@mewim I reworked it according to your latest comment. What do you think?

tools/python_api/src_py/query_result.py
```diff
@@ -166,7 +171,13 @@ def get_as_arrow(self, chunk_size: int) -> pa.Table:
         """
         self.check_for_query_result_close()
 
-        return self._query_result.getAsArrow(chunk_size)
+        if chunk_size <= 0 or chunk_size is None:
+            # adaptive chunk_size; target number of elements per chunk_zise
```
mewim, line 175: `chunk_zise` is a typo

tools/python_api/src_py/query_result.py
```diff
@@ -166,7 +171,13 @@ def get_as_arrow(self, chunk_size: int) -> pa.Table:
...
+            target_chunk_size = max(chunk_size // len(self.get_column_names()), 10)
```
mewim, line 176: Since in this case `chunk_size` is set to `0` / `-1` / `None`, this is not useful here, as `chunk_size // len(self.get_column_names())` will always be something invalid, so this is equivalent to `target_chunk_size=10`

prrao87: Yeah the test is failing in any case because None can't be compared against an integer. Will fix.

alexander-beedie:
> I would probably keep the chunk size as referring to number of rows and use adaptive chunk size only if chunk size is set to None / 0 / -1.

I'd reserve `chunk_size = 0/-1` for a new mode that guarantees only a _single_ chunk (aka: no chunks?) is produced. This allows `None` to be adaptive, an integer to continue doing what it currently does, and if/when the functionality is added lower down to produce an unchanged Arrow result we could enable 0/-1 for that 🤔 

> I'm not sure what you meant by the last comment (the "new default") not being a breaking change.

Just that the way the PR was initially written changed the meaning of `chunk_size` so that everyone currently using it would start getting back results that were not chunked the same way. If the new default was `None` then all _current_ usage would continue to behave identically (so the change in default would not be breaking); only new usage that omits an integer `chunk_size` would get the new behaviour.

prrao87:
@alexander-beedie how does -1 produce an "unchunked" result? To my understanding arrow will always return record batches, i.e., chunks of records?

alexander-beedie:
> @alexander-beedie how does -1 produce an "unchunked" result? To my understanding arrow will always return record batches, i.e., chunks of records?

It doesn't/can't at the moment as the current low-level code will always produce chunked results. The idea would be to produce a single chunk representing the _complete_ result set.

We'd use such a mode in Polars, as otherwise we typically have to explicitly rechunk Arrow data to achieve the same result (it's more optimal for us to operate on contiguous data).

prrao87:
Hmm, so it's a polars-specific setting where we'd document that setting `chunk_size` as -1 is an option. But as per the polars [docs](https://docs.pola.rs/py-polars/html/reference/api/polars.from_arrow.html) for the `from_arrow` method, this would require that we specify `rechunk=False` when the user specifies -1 to `get_as_pl`, correct?

prrao87:
I think the last commit addresses your comments @alexander-beedie. We can use the `get_num_tuples()` method to return the entire results as a single chunk when the user specifies 0 or -1.

alexander-beedie:
> as per the polars [docs](https://docs.pola.rs/py-polars/html/reference/api/polars.from_arrow.html) for the `from_arrow` method, this would require that we specify `rechunk=False` when the user specifies -1 to `get_as_pl`, correct?

It's a no-op unless n_chunks > 1, so no need to set explicitly 😎

tools/python_api/test/test_arrow.py
```diff
@@ -470,7 +470,7 @@ def test_to_arrow_complex(conn_db_readonly: ConnDB) -> None:
     def _test_node(_conn: kuzu.Connection) -> None:
         query = "MATCH (p:person) RETURN p ORDER BY p.ID"
         query_result = _conn.execute(query)
-        arrow_tbl = query_result.get_as_arrow(12)
+        arrow_tbl = query_result.get_as_arrow()
```
mewim, line 473: I think we should not remove the arg from all of `get_as_arrow` call. I would probably use number, -1, 0, no arg in different test cases to cover all of the branches.

prrao87: What do you think about Alex's earlier comment:
> FYI: I believe they were specified this way in order to validate result acquisition across chunk boundaries.

Do we need to validate the result acquisition across chunk boundaries by specifying a small value for certain cases?

mewim: I am not sure what this means?

prrao87: Ok, done.

mewim (approved):

mewim:
@prrao87 Note that when merging to the master, the commits needs to be squashed into one.

prrao87:
Shall I go ahead and squash-merge?

mewim:
Yeah I think it is ready to merge

alexander-beedie:
> We can use the `get_num_tuples()` method to return the entire results as a single chunk

@prrao87: Somehow I completely missed this method; great solution ✌️😎

PR #3088, opened 2024-03-19
Hash index cleanup

benjaminwinger:
A couple of minor changes to the hash index. I've updated the HashIndex to use one template parameter like the HashIndexBuilder, and I've made SlotEntry store key/value instead of just a `data` byte array. We can still use memcpy to copy both together, but it provides an easier to read way of accessing and updating the key and value.


ray6080 (approved):
LGTM

PR #3092, opened 2024-03-19
Added customizable delay before displaying progress bar

MSebanc:


src/common/task_system/progress_bar.cpp
```diff
@@ -4,18 +4,18 @@ namespace kuzu {
 namespace common {
 
 void ProgressBar::startProgress() {
+    queryTimer->start();
```
andyfengHKU, line 7: We should start timer only if we decide to track progress right?

andyfengHKU (approved):

PR #3093, opened 2024-03-20
Improve the performance of VAR_LIST storage layout

hououou:
This PR is to improve the performance of var  #3060. 

First, we now support in-place commits for VAR_LIST. We separately commit the offset column, size column, and data column. Previously, we do not support in-place commit for VAR_LIST. Thus, every time we update an item, it will trigger an out-place commit: scanning the whole column, updating the column, and writing the updated column into the persistence storage. 

Second, the design of #3060 is good for writing performance. But it is bad for scan performance since it may cause random access to the data column. If the offset is not in ascending order, we need to scan the list one by one. Worse, the data column will be increased with every update/insert. To balance write and read performance, we rewrite the whole var list column in ascending order at some point. Now we rewrite it when we do an out-place commit and the size of the data column is larger than a threshold(Currently, we use capacity/2).

There are still some spaces we can optimize the storage layout of VAR_LIST.

1. We now can only do out-place commits for the offset column because of some interfaces. A better way is to store the offset in another offset column but not in the buffer of the VAR_LIST column itself, just like the size column.
2. We should support scanning the data in non-empty column chunk. Then we can move `tmpDataColumnChunk` in `VarListColumn::scan(Transaction* transaction, node_group_idx_t nodeGroupIdx,
    kuzu::storage::ColumnChunk* columnChunk, offset_t startOffset, offset_t endOffset)` .
3. If a list is not in ascending order, we now scan the list items one by one. A better way is to scan in groups, trying to scan out those small parts of contiguous data together as much as possible.

I also ran a benchmark to test the performance of the current VAR_LIST storage layout. The dataset only has one person(id int64, age int64[]) table and one knows table(from person to person, length int64) table. The dataset has 10000000 nodes(around 1GB) and 100000000 edges (around 11GB). The results are as follows. The writing performance has improved a lot. The reading performance is similar to the previous implementation except for scanning the whole edge table (maybe caused by the above second bottleneck).
 
var_list_storage

### DIFF
--- a/src/include/storage/store/var_list_column.h
+++ b/src/include/storage/store/var_list_column.h
@@ -1,43 +1,45 @@
 #pragma once
 
 #include "column.h"
+#include "var_list_column_chunk.h"
 
-// List is a nested data type which is stored as two chunks:
+// List is a nested data type which is stored as three chunks:
 // 1. Offset column (type: INT64). Using offset to partition the data column into multiple lists.
-// 2. Data column. Stores the actual data of the list.
+// 2. Size column. Stores the size of each list.
+// 3. Data column. Stores the actual data of the list.
 // Similar to other data types, nulls are stored in the null column.
 // Example layout for list of INT64:
 // Four lists: [4,7,8,12], null, [2, 3], []
 // Offset column: [4, 4, 6, 6]
+// Size column: [4, 0, 2, 0]
 // data column: [4, 7, 8, 12, 2, 3]
-// When reading the data, we firstly read the offset column and utilize the offset column to find
-// the data column partitions.
-// 1st list(offset 4): Since it is the first element, the start offset is a constant 0, and the end
-// offset is 4. Its data is stored at position 0-4 in the data column.
-// 2nd list(offset 4): By reading the null column, we know that the 2nd list is null. So we don't
-// need to read from the data column.
-// 3rd list(offset 6): The start offset is 4(by looking up the offset for the 2nd list), and the end
-// offset is 6. Its data is stored at position 4-6 in the data column.
-// 4th list(offset 6): The start offset is 6(by looking up the offset for the 3rd list), and the end
-// offset is 6. Its data is stored at position 6-6 in the data column (empty list).
+// When updating the data, we first append the data to the data column, and then update the offset
+// and size accordingly. Besides offset column, we introduce an extra size column here to enable
+// in-place updates of a list column. In a list column chunk, offsets of lists are not always sorted
+// after updates. This is good for writes, but it introduces extra overheads for scans, as lists can
+// be scattered, and scans have to be broken into multiple small reads. To achieve a balance between
+// reads and writes, during updates, we rewrite the whole var list column chunk in ascending order
+// when the offsets are not sorted in ascending order and the size of data column chunk is larger
+// than half of its capacity.
 
 namespace kuzu {
 namespace storage {
 
-struct ListOffsetInfoInStorage {
-    common::offset_t prevNodeListOffset;
-    std::vector<std::unique_ptr<common::ValueVector>> offsetVectors;
+struct ListOffsetSizeInfo {
+    common::offset_t numTotal;
+    std::unique_ptr<ColumnChunk> offsetColumnChunk;
+    std::unique_ptr<ColumnChunk> sizeColumnChunk;
 
-    ListOffsetInfoInStorage(common::offset_t prevNodeListOffset,
-        std::vector<std::unique_ptr<common::ValueVector>> offsetVectors)
-        : prevNodeListOffset{prevNodeListOffset}, offsetVectors{std::move(offsetVectors)} {}
+    ListOffsetSizeInfo(common::offset_t numTotal, std::unique_ptr<ColumnChunk> offsetColumnChunk,
+        std::unique_ptr<ColumnChunk> sizeColumnChunk)
+        : numTotal{numTotal}, offsetColumnChunk{std::move(offsetColumnChunk)},
+          sizeColumnChunk{std::move(sizeColumnChunk)} {}
 
-    common::offset_t getListOffset(uint64_t nodePos) const;
+    common::list_size_t getListSize(uint64_t pos) const;
+    common::offset_t getListEndOffset(uint64_t pos) const;
+    common::offset_t getListStartOffset(uint64_t pos) const;
 
-    inline uint64_t getListLength(uint64_t nodePos) const {
-        KU_ASSERT(getListOffset(nodePos + 1) >= getListOffset(nodePos));
-        return getListOffset(nodePos + 1) - getListOffset(nodePos);
-    }
+    bool isOffsetSortedAscending(uint64_t startPos, uint64_t endPos) const;
 };
 
 class VarListColumn : public Column {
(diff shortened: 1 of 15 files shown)


src/include/storage/store/var_list_column.h
```diff
@@ -1,43 +1,45 @@
...
-// 4th list(offset 6): The start offset is 6(by looking up the offset for the 3rd list), and the end
-// offset is 6. Its data is stored at position 6-6 in the data column (empty list).
+// When updating the data, we first append the data to the data column, and then update the offset
+// and size. For example, we want to update [4,7,8,12] to [1,3,4]. We first append [1,3,4] to the
+// data column, and then update the offset and size. The layout after the update: Offset column: [9,
+// 4, 6, 6] Size column: [3, 0, 3, 0] data column: [4, 7, 8, 12, 2, 3, 1, 3, 4] This design is good
+// for writing performance. But it is bad for scan performance since it may cause random access to
+// the data column. Worse, the data column will be increased every update/insert. To balance the
+// write performance and read performance, we rewrite the whole var list column in ascending order
+// when the size of data column is larger than a threshold(Currently we use capacity/2).
```
ray6080, line 23: // When updating the data, we first append the data to the data column, and then update the offset
// and size accordingly. Besides offset column, we introduce an extra size column here to enable in-place updates of a list column.
// In a list column chunk, offsets of lists are not always sorted after updates.
// This is good for writes, but it introduces extra overheads for scans, as lists can be scattered, and scans have to be broken into multiple small reads. To achieve a balance between reads and writes, during updates, we rewrite the whole var list column chunk in ascending order
// when the offsets are not sorted in ascending order and the size of data column chunk is larger than half of its capacity.

src/include/storage/store/var_list_column.h
```diff
@@ -69,45 +71,37 @@ class VarListColumn : public Column {
...
+        const offset_to_row_idx_t& insertInfo, const ChunkCollection& localUpdateChunks,
+        const offset_to_row_idx_t& updateInfo, const offset_set_t& deleteInfo) override;
+    void prepareCommitForChunk(transaction::Transaction* transaction,
+        common::node_group_idx_t nodeGroupIdx, const std::vector<common::offset_t>& dstOffsets,
+        ColumnChunk* chunk, common::offset_t startSrcOffset) override;
 
 private:
+    std::unique_ptr<Column> sizeColumn;
     std::unique_ptr<Column> dataColumn;
+    std::unique_ptr<VarListDataColumnChunk> tmpDataColumnChunk;
```
ray6080, line 104: Add a TODO here for me. This should be moved to table states.

src/storage/store/var_list_column.cpp
```diff
@@ -39,39 +73,85 @@ void VarListColumn::scan(Transaction* transaction, node_group_idx_t nodeGroupIdx
...
     offset_t listOffsetInVector =
         offsetInVector == 0 ? 0 :
                               resultVector->getValue<list_entry_t>(offsetInVector - 1).offset +
                                   resultVector->getValue<list_entry_t>(offsetInVector - 1).size;
     auto offsetToWriteListData = listOffsetInVector;
     auto numValues = endOffsetInGroup - startOffsetInGroup;
+    KU_ASSERT(numValues >= 0);
     for (auto i = 0u; i < numValues; i++) {
-        auto length = listOffsetInfoInStorage.getListLength(i);
+        uint64_t length = listOffsetInfoInStorage.getListLength(i);
```
ray6080, line 86: should use `list_length/size_t` here

src/storage/store/var_list_column.cpp
```diff
@@ -39,39 +73,85 @@ void VarListColumn::scan(Transaction* transaction, node_group_idx_t nodeGroupIdx
...
+        if (isOffsetSortedAscending) {
+            varListColumnChunk->resizeDataColumnChunk(std::bit_ceil(resizeNumValues));
+            offset_t startVarListOffset = varListColumnChunk->getListStartOffset(0);
+            offset_t endVarListOffset =
+                varListColumnChunk->getListStartOffset(columnChunk->getNumValues());
+            dataColumn->scan(transaction, nodeGroupIdx, varListColumnChunk->getDataColumnChunk(),
+                startVarListOffset, endVarListOffset);
+            varListColumnChunk->resetOffset();
+        } else {
+            varListColumnChunk->resizeDataColumnChunk(std::bit_ceil(resizeNumValues));
```
ray6080, line 139: This code path is not well covered by tests. Can you add tests that first perform a bunch of in place updates, then perform reads?

ray6080 (commented):
Let's also bump the last digit of the CMake project version.

Haven't finished reviewing all of these changes, I will continue later.

src/storage/store/var_list_column.cpp
```diff
@@ -175,29 +275,110 @@ offset_t VarListColumn::readOffset(
...
+        auto appendListDataNum = 0u;
+        auto startVarListOffset = varListChunk->getListStartOffset(startSrcOffset);
+        std::vector<common::offset_t> dstOffsetsInDataColumn;
+        for (auto i = 0u; i < appendListNum; i++) {
+            auto appendLen = varListChunk->getListLen(startSrcOffset + i);
+            for (auto j = 0u; j < appendLen; j++) {
+                dstOffsetsInDataColumn.push_back(dataColumnSize + appendListDataNum + j);
+            }
+            appendListDataNum += appendLen;
+        }
```
ray6080, line 362: Why not just
```
        for (auto i = 0u; i < appendListNum; i++) {
            for (auto j = 0u; j < appendLen; j++) {
                dstOffsetsInDataColumn.push_back(dataColumnSize++);
            }
        }
```

btw, you can reuse this function under `rel_table_data.h`
```
    inline void fillSequence(std::span<common::offset_t> offsets, common::offset_t startOffset) {
        for (auto i = 0u; i < offsets.size(); i++) {
            offsets[i] = i + startOffset;
        }
    }
```

src/storage/store/var_list_column.cpp
```diff
@@ -175,29 +275,110 @@ offset_t VarListColumn::readOffset(
...
+        auto varListChunk = ku_dynamic_cast<ColumnChunk*, VarListColumnChunk*>(chunk);
+        sizeColumn->prepareCommitForChunk(transaction, nodeGroupIdx, dstOffsets,
+            varListChunk->getSizeColumnChunk(), startSrcOffset);
+        auto dataColumnSize =
+            dataColumn->getMetadata(nodeGroupIdx, transaction->getType()).numValues;
+        auto dataColumnChunk = varListChunk->getDataColumnChunk();
+        auto appendListNum = std::min(chunk->getNumValues(), (uint64_t)dstOffsets.size());
+        auto appendListDataNum = 0u;
+        auto startVarListOffset = varListChunk->getListStartOffset(startSrcOffset);
+        std::vector<common::offset_t> dstOffsetsInDataColumn;
```
ray6080, line 355: reserve `dstOffsetsInDataColumn` first.

src/storage/store/var_list_column.cpp
```diff
@@ -175,29 +275,110 @@ offset_t VarListColumn::readOffset(
...
+        dataColumn->prepareCommitForChunk(
+            transaction, nodeGroupIdx, dstOffsetsInDataColumn, dataColumnChunk, startVarListOffset);
+        // we need to update the offset since we do not do in-place list data update but append data
+        // in the end of list data column we need to plus to original data column size to get the
+        // new offset
+        // TODO(Jiamin): A better way is to store the offset in a offset column, just like size
+        // column. Then we can reuse prepareCommitForChunk interface for offset column.
+        auto offsetChunkMeta = getMetadata(nodeGroupIdx, transaction->getType());
+        auto offsetColumnChunk = ColumnChunkFactory::createColumnChunk(*dataType.copy(),
+            enableCompression, 1.5 * std::bit_ceil(offsetChunkMeta.numValues + dstOffsets.size()));
```
ray6080, line 372: ditto, why `1.5*` here?

src/storage/store/var_list_column.cpp
```diff
@@ -175,29 +275,110 @@ offset_t VarListColumn::readOffset(
...
+            }
+            appendListDataNum += appendLen;
+        }
+        dataColumn->prepareCommitForChunk(
+            transaction, nodeGroupIdx, dstOffsetsInDataColumn, dataColumnChunk, startVarListOffset);
+        // we need to update the offset since we do not do in-place list data update but append data
+        // in the end of list data column we need to plus to original data column size to get the
+        // new offset
+        // TODO(Jiamin): A better way is to store the offset in a offset column, just like size
+        // column. Then we can reuse prepareCommitForChunk interface for offset column.
```
ray6080, line 369: We should apply this optimization. You can separate it in another PR, but we should do it soon.

src/storage/store/var_list_column_chunk.cpp
```diff
@@ -27,35 +27,57 @@ void VarListDataColumnChunk::resizeBuffer(uint64_t numValues) {
...
+    sizeColumnChunk = ColumnChunkFactory::createColumnChunk(
+        *common::LogicalType::UINT32(), enableCompression, capacity);
     varListDataColumnChunk = std::make_unique<VarListDataColumnChunk>(
         ColumnChunkFactory::createColumnChunk(*VarListType::getChildType(&this->dataType)->copy(),
             enableCompression, 0 /* capacity */, inMemory));
+    checkOrder = true;
     KU_ASSERT(this->dataType.getPhysicalType() == PhysicalTypeID::VAR_LIST);
 }
 
+bool VarListColumnChunk::isOffsetSortedAscending(uint64_t startPos, uint64_t endPos) const {
```
ray6080, line 40: Re-read on this, actually this function is not just check sorting order, but also checking if offsets are consecutive or not.
So the name is confusing. we should rename it. `isOffsetsConsecutiveAndSortedAscending` is a candidate.

src/storage/store/var_list_column_chunk.cpp
```diff
@@ -214,35 +276,69 @@ void VarListColumnChunk::copyListValues(const list_entry_t& entry, ValueVector*
...
+    offset_t offsetInChunk = 0;
+    offset_t currentIndex = 0;
+    for (auto i = 0u; i < numValues; i++) {
+        if (nullChunk->isNull(i)) {
             newVarListChunk->appendNullList();
         } else {
-            auto index = indicesColumnChunk->getValue<offset_t>(i);
-            newColumnChunk->append(this, index, 1);
+            auto startOffset = getListStartOffset(i);
+            auto appendLen = getListLen(i);
```
ray6080, line 319: appendLen -> listLen

src/storage/store/var_list_column_chunk.cpp
```diff
@@ -214,35 +276,69 @@ void VarListColumnChunk::copyListValues(const list_entry_t& entry, ValueVector*
...
+            dataColumnChunk->append(
+                varListDataColumnChunk->dataColumnChunk.get(), startOffset, appendLen);
+            offsetInChunk += appendLen;
+            newVarListChunk->getNullChunk()->setNull(currentIndex, false);
+            newVarListChunk->sizeColumnChunk->getNullChunk()->setNull(currentIndex, false);
+            newVarListChunk->sizeColumnChunk->setValue<uint32_t>(appendLen, currentIndex);
```
ray6080, line 325: use `length_t` here.

src/storage/store/var_list_column_chunk.cpp
```diff
@@ -109,83 +139,113 @@ void VarListColumnChunk::lookup(
...
+    KU_ASSERT(sizeColumnChunk->getNullChunk()->getNumValues() == numValues);
+    KU_ASSERT(sizeColumnChunk->sanityCheck());
 }
 
 void VarListColumnChunk::write(
     ValueVector* vector, offset_t offsetInVector, offset_t offsetInChunk) {
-    needFinalize = true;
-    if (!indicesColumnChunk) {
-        initializeIndices();
+    checkOrder = false;
```
ray6080, line 196: Why not set `checkOrder` to true when we have writes? so we need to "check order" when it is set to true.

ray6080 (commented):
Let me take another look after you've addressed comments.

ray6080:
Can we close #3060 and #3019 ?

ray6080 (approved):

PR #3099, opened 2024-03-20
Update rustdoc to show how to enable parallel compilation

prrao87:
Running `cargo build` after setting `export CMAKE_BUILD_PARALLEL_LEVEL=20` (or some similar high value) significantly speeds up the compilation process. Because we can't know what a reasonable number of threads might be for users (it depends on the arch and the OS), I've added an additional step in the rustdoc to explain that this setting helps.

On my Mac, setting `CMAKE_BUILD_PARALLEL_LEVEL` to 20 results in a 5x speedup in the `cargo build` step, so we should definitely document this on the Kùzu docs as well. Will make a note of this and add it there.

### DIFF
--- a/tools/rust_api/src/lib.rs
+++ b/tools/rust_api/src/lib.rs
@@ -23,13 +23,15 @@
 //!
 //! ## Safety
 //!
-//! Generally, use of of this API is safe, however creating multiple databases in the same
-//! scope is not safe.
+//! Generally, use of this API is safe - however creating multiple databases in the same
+//! scope is not considered safe.
 //! If you need to access multiple databases you will need to do so in separate processes.
 //!
 //! ## Building
 //!
 //! By default, the kuzu C++ library will be compiled from source and statically linked.
+//! Set the `CMAKE_BUILD_PARALLEL_LEVEL` environment variable to a higher value than 1 (e.g., 8-20)
+//! to speed up compilation of the C++ library.
 //!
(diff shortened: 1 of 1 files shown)


tools/rust_api/src/lib.rs
```diff
@@ -23,13 +23,15 @@
...
-//! scope is not safe.
+//! Generally, use of this API is safe - however creating multiple databases in the same
+//! scope is not considered safe.
 //! If you need to access multiple databases you will need to do so in separate processes.
 //!
 //! ## Building
 //!
 //! By default, the kuzu C++ library will be compiled from source and statically linked.
+//! Set the `CMAKE_BUILD_PARALLEL_LEVEL` environment variable to a higher value than 1 (e.g., 8-20)
+//! to speed up compilation of the C++ library.
```
benjaminwinger, line 34: The parallelism should be passed to cmake by default by the `cmake-rs` crate. It sounds like there are sometimes issues with this, but I think the wording should be adjusted to reflect that this is in case of problems with that, rather than something that is normally required.

Maybe, something like:
> If the kuzu C++ library is not being built using multiple threads by default, you can use the `CMAKE_BUILD_PARALLEL_LEVEL` environment variable to manually set the number of threads for the build

prrao87: Sure, I rephrased it to make it seem totally optional rather than something the user needs to do.

benjaminwinger (approved):

PR #3109, opened 2024-03-21
Cache DiskArray write header in-memory

benjaminwinger:
Still a bunch of tests failing.

Additionally, as far as I can tell, the `InMemDiskArray` isn't actually doing anything meaningful. operator[] is unused (except in the `HashIndexBuilder` via `InMemDiskArrayBuilder`) and get/update are just using the `BaseDiskArray` implementation. The description suggests that it's supposed to cache the entirety of the metadataDA in memory, but I don't think it's done that for a while (hence our attempts to cache metadata locally when updating columns).

The cached header was creating some issues with the InMemDiskArray code, and while I think I was able to fix them in the initial version of this (or at least fix the crashes), being unused the code is not well tested. I've replaced InMemDiskArray with a thin shim for now, as it was doing some unnecessary work reading the pages into memory.


src/storage/store/var_list_column.cpp
```diff
@@ -246,6 +246,11 @@ void VarListColumn::scanFiltered(Transaction* transaction, node_group_idx_t node
         listOffset += resultVector->getValue<list_entry_t>(pos).size;
     }
 }
```
ray6080, line 248: Add an empty line here.

src/storage/store/node_table.cpp
```diff
@@ -160,6 +160,15 @@ void NodeTable::prepareCommit(Transaction* transaction, LocalTable* localTable)
...
+    tableData->prepareCommit();
+    wal->addToUpdatedTables(tableID);
+}
+
+void NodeTable::prepareCommit() {
+    if (pkIndex) {
+        pkIndex->prepareCommit();
+    }
+    tableData->prepareCommit();
     wal->addToUpdatedTables(tableID);
```
ray6080, line 172: I don't think you need L172.

src/storage/storage_manager.cpp
```diff
@@ -127,6 +127,11 @@ void StorageManager::dropTable(table_id_t tableID) {
 
 void StorageManager::prepareCommit(transaction::Transaction* transaction) {
     transaction->getLocalStorage()->prepareCommit();
+    for (auto tableID : wal->getUpdatedTables()) {
```
ray6080, line 130: Can you add a brief comment here to explain the changes? Better for readability.

ray6080 (approved):

src/storage/store/var_list_column.cpp
benjaminwinger: Hmm. apparently clang-format >=14 can enforce this: https://clang.llvm.org/docs/ClangFormatStyleOptions.html#separatedefinitionblocks. We should probably look at upgrading at some point.

PR #3110, opened 2024-03-22
QueryResult `get_as_pl` should always return a single chunk

alexander-beedie:
Small follow-up to #3084.

* `get_as_pl()` should not offer a configurable chunk size, as Polars always wants a single chunk ;)
* Slightly streamlined the adaptive "chunk_size" in `get_as_arrow()`.
* Integrated an explicit `num_chunks` assertion into the Arrow tests.


mewim (approved):
LGTM

PR #3116, opened 2024-03-22
Enable compression for INTERNAL_ID

ray6080:
This now brings data.kz file size of ldbc100 down from 126GB to 69GB.

### DIFF
--- a/src/storage/store/column_chunk.cpp
+++ b/src/storage/store/column_chunk.cpp
@@ -125,6 +125,7 @@ static std::shared_ptr<CompressionAlg> getCompression(
     case PhysicalTypeID::INT8: {
         return std::make_shared<IntegerBitpacking<int8_t>>();
     }
+    case PhysicalTypeID::INTERNAL_ID:
     case PhysicalTypeID::VAR_LIST:
     case PhysicalTypeID::UINT64: {
         return std::make_shared<IntegerBitpacking<uint64_t>>();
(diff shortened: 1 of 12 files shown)


src/storage/store/column_chunk.cpp
```diff
@@ -550,6 +560,16 @@ class InternalIDColumnChunk final : public ColumnChunk {
         }
     }
 
+    void copyInt64VectorToBuffer(
+        ValueVector* vector, offset_t startPosInChunk, common::SelectionVector& selVector) {
+        for (auto i = 0u; i < selVector.selectedSize; i++) {
```
benjaminwinger, line 565: An assert that it actually is an int64 vector might be helpful (I suppose it's in the function name, so it should be hard to mistake, but it's probably better to add those assertions whenever something isn't meant to work for all types).

benjaminwinger (approved):

src/storage/store/column_chunk.cpp
ray6080: sg.

src/storage/store/string_column_chunk.cpp
```diff
@@ -82,7 +82,6 @@ void StringColumnChunk::write(
 void StringColumnChunk::write(
     ColumnChunk* chunk, ColumnChunk* dstOffsets, RelMultiplicity /*multiplicity*/) {
     KU_ASSERT(chunk->getDataType().getPhysicalType() == PhysicalTypeID::STRING &&
-              dstOffsets->getDataType().getPhysicalType() == PhysicalTypeID::INT64 &&
```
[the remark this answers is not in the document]
ray6080: true. will add back the assertions.

PR #3120, opened 2024-03-22
Added progress for in_query_call operators

MSebanc:

### DIFF
--- a/src/processor/operator/persistent/reader/csv/parallel_csv_reader.cpp
+++ b/src/processor/operator/persistent/reader/csv/parallel_csv_reader.cpp
@@ -124,6 +124,7 @@ static offset_t tableFunc(TableFuncInput& input, TableFuncOutput& output) {
             auto result = parallelCSVLocalState->reader->continueBlock(outputChunk);
             outputChunk.state->selVector->selectedSize = result;
             if (result > 0) {
+                parallelCSVSharedState->fileOffset = parallelCSVLocalState->reader->getFileOffset();
                 return result;
             }
         }
@@ -141,9 +142,11 @@ static offset_t tableFunc(TableFuncInput& input, TableFuncOutput& output) {
         auto numRowsRead = parallelCSVLocalState->reader->parseBlock(blockIdx, outputChunk);
         outputChunk.state->selVector->selectedSize = numRowsRead;
         if (numRowsRead > 0) {
+            parallelCSVSharedState->fileOffset = parallelCSVLocalState->reader->getFileOffset();
             return numRowsRead;
         }
         if (parallelCSVLocalState->reader->isEOF()) {
+            parallelCSVSharedState->fileOffset = parallelCSVLocalState->reader->getFileOffset();
             parallelCSVSharedState->setFileComplete(parallelCSVLocalState->fileIdx);
             parallelCSVLocalState->reader = nullptr;
         }
@@ -168,8 +171,12 @@ static std::unique_ptr<TableFuncSharedState> initSharedState(TableFunctionInitIn
     auto bindData = reinterpret_cast<ScanBindData*>(input.bindData);
     auto csvConfig = CSVReaderConfig::construct(bindData->config.options);
     row_idx_t numRows = 0;
-    return std::make_unique<ParallelCSVScanSharedState>(bindData->config.copy(), numRows,
+    auto sharedState = std::make_unique<ParallelCSVScanSharedState>(bindData->config.copy(), numRows,
         bindData->columnNames.size(), bindData->context, csvConfig.copy());
+    auto reader = std::make_unique<ParallelCSVReader>(sharedState->readerConfig.filePaths[0],
+        sharedState->csvReaderConfig.option.copy(), sharedState->numColumns, sharedState->context);
+    sharedState->fileSize = reader->getFileSize();
+    return sharedState;
 }
 
 static std::unique_ptr<TableFuncLocalState> initLocalState(TableFunctionInitInput& /*input*/,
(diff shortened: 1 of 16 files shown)


src/processor/operator/persistent/reader/csv/serial_csv_reader.cpp
```diff
@@ -115,20 +116,34 @@ static std::unique_ptr<TableFuncSharedState> initSharedState(TableFunctionInitIn
...
+    return sharedState;
 }
 
 static std::unique_ptr<TableFuncLocalState> initLocalState(TableFunctionInitInput& /*input*/,
     TableFuncSharedState* /*state*/, storage::MemoryManager* /*mm*/) {
     return std::make_unique<TableFuncLocalState>();
 }
 
+static double progressFunc(TableFuncSharedState* sharedState) {
+    auto state = reinterpret_cast<SerialCSVScanSharedState*>(sharedState);
```
andyfengHKU, line 133: Did you consider the case where we scan from multiple csv files?

src/processor/operator/persistent/reader/csv/parallel_csv_reader.cpp
```diff
@@ -124,6 +124,7 @@ static offset_t tableFunc(TableFuncInput& input, TableFuncOutput& output) {
             auto result = parallelCSVLocalState->reader->continueBlock(outputChunk);
             outputChunk.state->selVector->selectedSize = result;
             if (result > 0) {
+                parallelCSVSharedState->fileOffset = parallelCSVLocalState->reader->getFileOffset();
```
andyfengHKU, line 127: I feel that we could alternatively just use `blockIdx` to calculate progress. Each block guarantees to have size `PARALLEL_BLOCK_SIZE`. So you could calculate your fileOffset from `blockIdx`. If so, you don't need to write `parallelCSVSharedState->fileOffset` in multiple places.

Also I would try to enforce progress being calculated in shared state as long as it's doable.

src/processor/operator/persistent/reader/parquet/parquet_reader.cpp
```diff
@@ -601,9 +602,11 @@ static common::offset_t tableFunc(TableFuncInput& input, TableFuncOutput& output
     do {
         parquetScanLocalState->reader->scan(*parquetScanLocalState->state, outputChunk);
         if (outputChunk.state->selVector->selectedSize > 0) {
+            parquetScanSharedState->numRowsRead += outputChunk.state->selVector->selectedSize;
```
andyfengHKU, line 605: Ditto I think we can use `blockIdx` and `getNumRowsGroups()` in `parquetScanSharedState`. Also try to make the change inside `parquetScanSharedState`.

andyfengHKU (changes requested):

src/processor/operator/persistent/reader/csv/parallel_csv_reader.cpp
```diff
@@ -182,11 +189,24 @@ static std::unique_ptr<TableFuncLocalState> initLocalState(TableFunctionInitInpu
     return localState;
 }
 
+static double progressFunc(TableFuncSharedState* sharedState) {
+    auto state = reinterpret_cast<ParallelCSVScanSharedState*>(sharedState);
```
andyfengHKU, line 193: always use ku_dynamic_case instead of `reinterpret_cast` or `static_cast`

src/processor/operator/persistent/reader/csv/serial_csv_reader.cpp
```diff
@@ -115,20 +115,42 @@ static std::unique_ptr<TableFuncSharedState> initSharedState(TableFunctionInitIn
...
+    if (state->totalSize == 0) {
+        return 0.0;
+    } else if (state->fileIdx >= state->readerConfig.getNumFiles()) {
+        return 1.0;
+    }
+    uint64_t totalReadSize = 0;
+    for (auto i = 0u; i < state->fileIdx; i++) {
```
andyfengHKU, line 142: I would suggest track a totalReadSizeByFile in sharedState and update it each time we move to a new file. In such design you don't need to calculate the accumulation of file size here again and again.

src/processor/operator/persistent/reader/csv/parallel_csv_reader.cpp
```diff
@@ -182,11 +189,24 @@ static std::unique_ptr<TableFuncLocalState> initLocalState(TableFunctionInitInpu
...
+    if (state->totalSize == 0) {
+        return 0.0;
+    } else if ((state->blockIdx * CopyConstants::PARALLEL_BLOCK_SIZE) * (state->fileIdx + 1) >=
```
andyfengHKU, line 196: Do we have the guarantee that all files will have the same block? I don't quite understand why we do `* (state->fileIdx + 1)`.

You probably have a wrong understanding of parallel csv reader. We don't parallelize by allocating each thread to a file. Instead, we parallelize by assigning all threads to one file and proceed to the next only when current file has finished.

So you could do this in a similar fashion as serial csv reader, i.e. track numBlocksReadByFile in shared state.

src/processor/operator/persistent/reader/parquet/parquet_reader.cpp
```diff
@@ -674,11 +679,23 @@ static std::unique_ptr<function::TableFuncLocalState> initLocalState(
...
 
+static double progressFunc(TableFuncSharedState* state) {
+    auto parquetScanSharedState = reinterpret_cast<ParquetScanSharedState*>(state);
+    if (parquetScanSharedState->fileIdx >= parquetScanSharedState->readerConfig.getNumFiles()) {
+        return 1.0;
+    }
+    if (parquetScanSharedState->totalRowsGroups == 0) {
+        return 0.0;
+    }
+    return (double)(parquetScanSharedState->blockIdx * (parquetScanSharedState->fileIdx + 1)) /
```
andyfengHKU, line 690: ditto. Parquet parallelism is done the same way as parallel csv. So you need to change this as well.

andyfengHKU (approved):
Although the part for estimating progress of parallel csv and parquet is wrong. I believe you get the idea already so I don't need to take a look at this PR again.

Let me know if you need to discuss something.

PR #3126, opened 2024-03-24
Support read after update

andyfengHKU:
We now support read after update.

Since we already support read write in one transaction. There is no reason we couldn't support read after update in one plan. Also we need to enforce is that update pipeline is executed before read pipeline.

### DIFF
--- a/src/include/planner/operator/logical_operator.h
+++ b/src/include/planner/operator/logical_operator.h
@@ -58,9 +58,9 @@ enum class LogicalOperatorType : uint8_t {
     IMPORT_DATABASE,
 };
 
-class LogicalOperatorUtils {
-public:
+struct LogicalOperatorUtils {
     static std::string logicalOperatorTypeToString(LogicalOperatorType type);
+    static bool isUpdate(LogicalOperatorType type);
 };
 
 class LogicalOperator;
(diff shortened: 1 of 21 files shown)


src/include/planner/operator/logical_operator.h
```diff
@@ -58,45 +58,45 @@ enum class LogicalOperatorType : uint8_t {
...
+    static bool isUpdate(LogicalOperatorType type);
 };
 
 class LogicalOperator;
 using logical_op_vector_t = std::vector<std::shared_ptr<LogicalOperator>>;
 
 class LogicalOperator {
 public:
-    // Leaf operator.
+    // Constructors.
```
ray6080, line 71: Remove the comment.

src/include/planner/operator/logical_operator.h
```diff
@@ -58,45 +58,45 @@ enum class LogicalOperatorType : uint8_t {
...
     explicit LogicalOperator(LogicalOperatorType operatorType, const logical_op_vector_t& children);
 
     virtual ~LogicalOperator() = default;
 
-    inline uint32_t getNumChildren() const { return children.size(); }
-
-    inline std::shared_ptr<LogicalOperator> getChild(uint64_t idx) const { return children[idx]; }
-    inline std::vector<std::shared_ptr<LogicalOperator>> getChildren() const { return children; }
-    inline void setChild(uint64_t idx, std::shared_ptr<LogicalOperator> child) {
+    // Children.
```
ray6080, line 81: Remove the comment.

src/planner/operator/logical_operator.cpp
```diff
@@ -129,6 +143,21 @@ LogicalOperator::LogicalOperator(
     }
 }
 
+bool LogicalOperator::hasUpdateRecursive() {
+    if (LogicalOperatorUtils::isUpdate(operatorType)) {
+        return true;
+    }
+    if (children.size() == 0) {
+        return LogicalOperatorUtils::isUpdate(operatorType);
+    }
```
ray6080, line 152: I don't think L150-L152 is necessary.

ray6080 (approved):

Issue #3065, opened 2024-03-15
[PROPOSAL] Prepare Statement Improvement


andyfengHKU:
At a high level, our query execution is split into 2 stages.
- Stage 1: Convert string to preparedStatement. This stage includes parsing, binding and planning
- Stage 2: Convert preparedStatement into result. This stage includes mapping and executing. Note that we put mapping into Stage 2 instead of 1 because we don't have a mechanism to reset operator state yet. The problem of resetting is mostly about testing.
For auto-committed transactions, state 1 will create an auto-committed transaction and stage 2 will end that transaction.

For manual transactions, we don't need to anything special because transaction behaviour is controlled by the user through transaction statements.
## Transaction
This problem with current transaction over prepared statement happens in the following 2 cases. We assume auto-committed transaction in the following description.
**Case 1**: prepare multiple times
```
PREPARE stmt1;
PREPARE stmt2;
EXECUTE stmt2;
```
When preparing stmt1, we will start a tx1. Then when preparing stmt2, we will try to start a tx2 without committing tx1 which leads to an exception.
**Case 2**: prepare once and execute multiple times
```
PREPARE stmt;
EXECUTE stmt;
EXECUTE stmt;
```
When executing stmt for the first time. We will commit the tx generated in stage 1. And when we execute stmt for the second time. We can no longer find a valid tx.

### Prepare with Transaction
The core idea is to separate tx logic between stage 1 and stage 2. And force the invariant that at stage 2 there is always a correct tx if a tx is needed. (The case where tx may not be needed will be explained later). 

We should provide the following interfaces
```
public
     
QueryResult query(string) {
    Parse string as multiple stmts
    For each stmt
         pResult = prepareInternal(stmt)
         result = executeInternal(pResult)
}

PreparedStmt prepare(string) {
    Parse string as single stmt. 
    Return prepareInternal(stmt)
}

QueryResult execute(preparedStmt) {
    Begin tx if auto
    execute(preparedStmt)
}

private

PreparedStmt prepareInternal(stmt) {
    Begin tx if auto
    Parse
    Bind
    Plan
    Commit tx if auto
}

// Stage 2
QueryResult execute(preparedStmt) {
    Execute
    End tx depend on exec state
}
```

Now that if revisit previous cases
**Case 1**: prepare multiple times
```
PREPARE stmt1;
PREPARE stmt2;
EXECUTE stmt2;
```
[...]

PR #3140, opened 2024-03-26
Prepare Statement Improvement

hououou:
see #3065 

A little different implementation: for a query statement, we always need to prepare, and we will start a transaction.
-  we will commit the transaction opened in prepare if `prepare()` is a standalone call. (requiredNewTx=true)
- we will not commit the transaction opened in `prepare()` called in `query()`. In `query() `, `execute()` is always followed with `prepare()`, we will close the transaction in `execute()`. (requiredNewTx=false)

 Similar logic is used in `execute()`.

### DIFF
--- a/src/include/binder/expression/parameter_expression.h
+++ b/src/include/binder/expression/parameter_expression.h
@@ -10,8 +10,8 @@ class ParameterExpression : public Expression {
 public:
     explicit ParameterExpression(
         const std::string& parameterName, std::shared_ptr<common::Value> value)
-        : Expression{common::ExpressionType::PARAMETER,
-              common::LogicalType(common::LogicalTypeID::ANY), createUniqueName(parameterName)},
+        : Expression{common::ExpressionType::PARAMETER, common::LogicalType(*value->getDataType()),
+              createUniqueName(parameterName)},
           parameterName(parameterName), value{std::move(value)} {}
 
     void cast(const common::LogicalType& type) override;
--- a/src/include/main/client_context.h
+++ b/src/include/main/client_context.h
@@ -106,6 +106,11 @@ public:
     std::unique_ptr<QueryResult> query(std::string_view queryStatement);
     void runQuery(std::string query);
 
+    // TODO(Jiamin): should remove after supporting ddl in manual tx
+    std::unique_ptr<PreparedStatement> prepareTest(std::string_view query);
+    // only use for test framework
+    std::vector<std::shared_ptr<parser::Statement>> parseQuery(std::string_view query);
+
 private:
     std::unique_ptr<QueryResult> query(
         std::string_view query, std::string_view encodedJoin, bool enumerateAllPlans = true);
@@ -114,10 +119,15 @@ private:
 
     std::unique_ptr<PreparedStatement> preparedStatementWithError(std::string_view errMsg);
 
-    std::vector<std::unique_ptr<parser::Statement>> parseQuery(std::string_view query);
-
-    std::unique_ptr<PreparedStatement> prepareNoLock(parser::Statement* parsedStatement,
-        bool enumerateAllPlans = false, std::string_view joinOrder = std::string_view());
+    // when we do prepare, we will start a transaction for the query
+    // when we execute after prepare in a same context, we set requireNewTx to false and will not
+    // commit the transaction in prepare when we only prepare a query statement, we set requireNewTx
+    // to true and will commit the transaction in prepare
+    std::unique_ptr<PreparedStatement> prepareNoLock(
+        std::shared_ptr<parser::Statement> parsedStatement, bool enumerateAllPlans = false,
+        std::string_view joinOrder = std::string_view(), bool requireNewTx = true,
+        std::optional<std::unordered_map<std::string, std::shared_ptr<common::Value>>> inputParams =
+            std::nullopt);
 
     template<typename T, typename... Args>
     std::unique_ptr<QueryResult> executeWithParams(PreparedStatement* preparedStatement,
@@ -133,7 +143,7 @@ private:
         const std::unordered_map<std::string, std::unique_ptr<common::Value>>& inputParams);
 
     std::unique_ptr<QueryResult> executeAndAutoCommitIfNecessaryNoLock(
-        PreparedStatement* preparedStatement, uint32_t planIdx = 0u);
+        PreparedStatement* preparedStatement, uint32_t planIdx = 0u, bool requiredNexTx = true);
 
     void addScalarFunction(std::string name, function::function_set definitions);
 
--- a/src/main/client_context.cpp
+++ b/src/main/client_context.cpp
@@ -191,10 +191,27 @@ std::string ClientContext::getEnvVariable(const std::string& name) {
 
 std::unique_ptr<PreparedStatement> ClientContext::prepare(std::string_view query) {
     auto preparedStatement = std::unique_ptr<PreparedStatement>();
+    if (query.empty()) {
+        return preparedStatementWithError("Connection Exception: Query is empty.");
+    }
     std::unique_lock<std::mutex> lck{mtx};
-    auto parsedStatements = std::vector<std::unique_ptr<Statement>>();
+    auto parsedStatements = std::vector<std::shared_ptr<Statement>>();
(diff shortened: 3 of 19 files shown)


src/main/connection.cpp
```diff
@@ -53,11 +53,12 @@ std::unique_ptr<PreparedStatement> Connection::preparedStatementWithError(std::s
...
 
 std::unique_ptr<PreparedStatement> Connection::prepareNoLock(
-    Statement* parsedStatement, bool enumerateAllPlans, std::string_view encodedJoin) {
+    std::shared_ptr<Statement> parsedStatement, bool enumerateAllPlans,
+    std::string_view encodedJoin) {
     return clientContext->prepareNoLock(parsedStatement, enumerateAllPlans, encodedJoin);
 }
 
-std::vector<std::unique_ptr<Statement>> Connection::parseQuery(std::string_view query) {
+std::vector<std::shared_ptr<Statement>> Connection::parseQuery(std::string_view query) {
```
andyfengHKU, line 61: I have a feeling we don't need this function.

src/main/connection.cpp
```diff
@@ -53,11 +53,12 @@ std::unique_ptr<PreparedStatement> Connection::preparedStatementWithError(std::s
 }
 
 std::unique_ptr<PreparedStatement> Connection::prepareNoLock(
-    Statement* parsedStatement, bool enumerateAllPlans, std::string_view encodedJoin) {
+    std::shared_ptr<Statement> parsedStatement, bool enumerateAllPlans,
```
andyfengHKU, line 56: ditto

src/include/binder/expression/parameter_expression.h
```diff
@@ -11,7 +11,8 @@ class ParameterExpression : public Expression {
     explicit ParameterExpression(
         const std::string& parameterName, std::shared_ptr<common::Value> value)
         : Expression{common::ExpressionType::PARAMETER,
-              common::LogicalType(common::LogicalTypeID::ANY), createUniqueName(parameterName)},
+              common::LogicalType(value->getDataType()->getLogicalTypeID()),
```
andyfengHKU, line 14: this is not safe because logical type can be nested. Just do `common::LogicalType(value->getDataType())`

src/common/types/value/value.cpp
```diff
@@ -15,7 +15,6 @@ namespace kuzu {
 namespace common {
 
 void Value::setDataType(const LogicalType& dataType_) {
-    KU_ASSERT(dataType->getLogicalTypeID() == LogicalTypeID::ANY);
```
andyfengHKU, line 18: Why are we removing this?

src/include/main/client_context.h
```diff
@@ -106,6 +106,11 @@ class ClientContext {
     std::unique_ptr<QueryResult> query(std::string_view queryStatement);
     void runQuery(std::string query);
 
+    void prepareInternal(PreparedStatement& preparedStatement, bool enumerateAllPlans = false,
```
andyfengHKU, line 109: why is a function named as internal but in public scope?

src/main/client_context.cpp
```diff
@@ -254,8 +254,65 @@ std::unique_ptr<PreparedStatement> ClientContext::preparedStatementWithError(
     return preparedStatement;
 }
 
+void ClientContext::prepareInternal(PreparedStatement& preparedStatement, bool enumerateAllPlans,
```
andyfengHKU, line 257: prepareInternal & prepareNoLock is creating ambiguity. Consider rename.

src/main/client_context.cpp
```diff
@@ -369,11 +394,6 @@ void ClientContext::bindParametersNoLock(PreparedStatement* preparedStatement,
         if (value->getDataType()->getLogicalTypeID() == LogicalTypeID::ANY) {
```
andyfengHKU, line 394: This should no longer be needed.

test/main/prepare_test.cpp
```diff
@@ -7,6 +7,48 @@ static void checkTuple(kuzu::processor::FlatTuple* tuple, const std::string& gro
...
+    auto preparedStatement = conn->prepare("MATCH (a:NodeOne { id: $a_id }),"
+                                           "(b:NodeTwo { id: $b_id }),"
+                                           "(c: NodeOne{ id: $c_id } )"
+                                           " MERGE"
+                                           " (a)-[:RelA]->(c),"
+                                           " (b)-[r:RelB { name: $my_param }]->(c)"
+                                           " return r.name;");
+    auto result = conn->execute(preparedStatement.get(), std::make_pair(std::string("a_id"), 1),
+        std::make_pair(std::string("b_id"), 3), std::make_pair(std::string("c_id"), 2),
+        std::make_pair(std::string("my_param"), "null"));
```
andyfengHKU, line 37: hmmm did you check if DuckDB allow passing a `NULL` parameter?

Plus we are passing a string "null" here and everything is a string operation. If you are testing for string parameter use a different value.

src/main/client_context.cpp
```diff
@@ -355,6 +368,18 @@ std::unique_ptr<QueryResult> ClientContext::executeWithParams(PreparedStatement*
...
+    // rebind
+    KU_ASSERT(preparedStatement->parsedStatement != nullptr);
+    auto& parameterMap = preparedStatement->parameterMap;
+    preparedStatement->requiredNewTx = transactionContext->hasActiveTransaction() ? false : true;
+    try {
+        prepareInternal(*preparedStatement, false, "", parameterMap);
+    } catch (std::exception& exception) {
+        preparedStatement->success = false;
+        preparedStatement->errMsg = exception.what();
+        this->transactionContext->rollback();
```
andyfengHKU, line 380: The more I read this code, the more concerned I have with our existing design. PrepareInternal opens a transaction but not commit/rollback it. Meaning whoever calls prepareInternal need to take care of the dangling tx.

src/main/client_context.cpp
```diff
@@ -355,6 +368,18 @@ std::unique_ptr<QueryResult> ClientContext::executeWithParams(PreparedStatement*
         std::string errMsg = exception.what();
         return queryResultWithError(errMsg);
     }
+    // rebind
+    KU_ASSERT(preparedStatement->parsedStatement != nullptr);
+    auto& parameterMap = preparedStatement->parameterMap;
+    preparedStatement->requiredNewTx = transactionContext->hasActiveTransaction() ? false : true;
```
andyfengHKU, line 374: Is the right behaviour to not acquire any tx? I thought the right behaviour is to begin a new tx.

andyfengHKU (changes requested):

src/main/client_context.cpp
```diff
@@ -323,6 +345,9 @@ std::unique_ptr<PreparedStatement> ClientContext::prepareNoLock(
         } else {
             preparedStatement->logicalPlans = std::move(plans);
         }
+        if (transactionContext->isAutoTransaction() && requireNewTx) {
+            this->transactionContext->commit();
```
andyfengHKU, line 349: shouldn't we have a similar check when rollback?

src/main/client_context.cpp
```diff
@@ -383,12 +404,11 @@ void ClientContext::bindParametersNoLock(PreparedStatement* preparedStatement,
...
 
 std::unique_ptr<QueryResult> ClientContext::executeAndAutoCommitIfNecessaryNoLock(
-    PreparedStatement* preparedStatement, uint32_t planIdx) {
+    PreparedStatement* preparedStatement, uint32_t planIdx, bool requiredNexTx) {
     if (!preparedStatement->isSuccess()) {
         return queryResultWithError(preparedStatement->errMsg);
     }
-    if (preparedStatement->preparedSummary.statementType != common::StatementType::TRANSACTION &&
-        this->getTx() == nullptr) {
+    if (preparedStatement->parsedStatement->requireTx() && requiredNexTx && getTx() == nullptr) {
```
andyfengHKU, line 411: Can u branch out from this branch and try commit after prepare and we open a new transaction during execute. In this design,  we should no longer need to pass `requiredNexTx`

andyfengHKU (approved):
Overall looks good to me. You can merge if the CI passes.

Issue #3070, opened 2024-03-17
Nodes/Rels `COLLECT` & `UNWIND`


bigluck:
I would like to be able to COLLECT relationships or nodes and UNWIND them later to create additional relationships, but this seems not to be supported yet. 

I have a query like this:

```cypher
MATCH
    (p:RPD)-[pi:II]->(fn:Fn),
    (fn)-[:PPS]->(dip:PPDIP)
WITH
    DISTINCT
        CONCAT(CONCAT(CONCAT(CONCAT(dip.name, ' @ '), p.name), '=='), pi.version) AS id,
        dip,
        p,
        pi.version AS pv
WITH
    *, gen_random_uuid() AS task_id
CREATE
    (:PPT {
        id: task_id,
        type: 'p_install'
    })-[:PPHT]->(t:PPI {
        id: id,
        task_id: task_id,
        di: dip.name,
        name: p.name,
        version: p_version,
        pv: dip.pv
    }),
    (p)-[:PPS]->(t),
    (dip)-[:PPS]->(t)
```

The goal of this query is to create a distinct node, and it works.

Now I would like to create a relationship between all the Fn nodes that contributed to the creation of this new node and the node itself, so I changed the query to:

```cypher
MATCH
    (p:RPD)-[pi:II]->(fn:Fn),
    (fn)-[:PPS]->(dip:PPDIP)
WITH
    DISTINCT
        CONCAT(CONCAT(CONCAT(CONCAT(dip.name, ' @ '), p.name), '=='), pi.version) AS id,
        dip,
        p,
        pi.version AS pv,
        COLLECT(fn) AS fns // NEW
WITH
    *, gen_random_uuid() AS task_id
CREATE
    (:PPT {
        id: task_id,
        type: 'p_install'
    })-[:PPHT]->(t:PPI {
        id: id,
        task_id: task_id,
        di: dip.name,
        name: p.name,
        version: p_version,
        pv: dip.pv
    }),
    (p)-[:PPS]->(t),
    (dip)-[:PPS]->(t)
WITH t, fns  // NEW
UNWIND fns AS fn  // NEW
CREATE (fn)-[:PPS]->(t) // NEW
```

But this new query fails with:

> Cannot bind fn as node pattern.

But the alternative solution (WITH ... MATCH) fails with:

> Read after update is not supported. Try query with multiple statements.

bigluck:
It works starting from `v0.3.3.dev17`, thanks so much @andyfengHKU

PR #3153, opened 2024-03-27
Support extend from unwind node

andyfengHKU:
Fix issue #3070


ray6080 (approved):
Can you open an issue on the new design that directly scan from accumulated unwind result as we discussed?

PR #3156, opened 2024-03-27
List Auxiliary Buffer NullMask Fix

mxwli:
The List Auxiliary Buffer allocates 64 times the necessary amount of null bitmap memory when resizing the buffer. This one-line change was made in its own PR so as to be easy to revert.


src/common/vector/auxiliary_buffer.cpp
```diff
@@ -52,7 +52,9 @@ void ListAuxiliaryBuffer::resizeDataVector(ValueVector* dataVector) {
     auto buffer = std::make_unique<uint8_t[]>(capacity * dataVector->getNumBytesPerValue());
     memcpy(buffer.get(), dataVector->valueBuffer.get(), size * dataVector->getNumBytesPerValue());
     dataVector->valueBuffer = std::move(buffer);
-    dataVector->nullMask->resize(capacity); // note: allocating 64 times what is needed
+    dataVector->nullMask->resize(
```
andyfengHKU, line 55: can u check if we are making the same mistake in other places when resizing nullMask? If so, this logic should be hidden inside NullMask class it self.

andyfengHKU (approved):

PR #3160, opened 2024-03-27
Resolve ANY Resolution Occurring at End of Python List Parameter

mxwli:
`[1, 2, 3, None]` and `{'a': 1, 'b': None}` should correctly resolve to `INT64[]` and `MAP(STRING, INT64)` now.

Additionally, `[]` resolves to `ANY[]` instead of `STRING[]` so that `[[], [1, 2, 3]]` resolves to `INT64[][]` instead of `STRING[][]`


acquamarin (approved):

mxwli:
Fixes the additional bug brought up at #3054

Issue #1061, opened 2022-11-23
Potential bug in ScanNodeID after deletion.


andyfengHKU:
There is probably a bug when trying to scan node ID with semi mask + node deletion.

andyfengHKU:
Solved in #3176

PR #3176, opened 2024-03-30
Fix scan after delete bug

andyfengHKU:
This should also solve issue #1061

### DIFF
--- a/src/storage/stats/node_table_statistics.cpp
+++ b/src/storage/stats/node_table_statistics.cpp
@@ -104,33 +104,32 @@ void NodeTableStatsAndDeletedIDs::deleteNode(offset_t nodeOffset) {
 
 // Note: this function will always be called right after scanNodeID, so we have the guarantee
 // that the nodeOffsetVector is always unselected.
-void NodeTableStatsAndDeletedIDs::setDeletedNodeOffsetsForMorsel(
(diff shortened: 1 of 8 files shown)


src/storage/stats/node_table_statistics.cpp
```diff
@@ -104,33 +104,32 @@ void NodeTableStatsAndDeletedIDs::deleteNode(offset_t nodeOffset) {
...
+        auto originalSize = nodeIDVector->state->getOriginalSize();
         auto itr = deletedNodeOffsets.begin();
-        sel_t nextDeletedNodeOffset = *itr - morselBeginOffset;
-        uint64_t nextSelectedPosition = 0;
-        for (auto pos = 0u; pos < nodeOffsetVector->state->getOriginalSize(); ++pos) {
-            if (pos == nextDeletedNodeOffset) {
+        common::sel_t numSelectedValue = 0;
+        auto selectedBuffer = nodeIDVector->state->selVector->getSelectedPositionsBuffer();
+        for (auto pos = 0u; pos < nodeIDVector->state->getOriginalSize(); ++pos) {
+            KU_ASSERT(pos == nodeIDVector->state->selVector->selectedPositions[pos]);
```
ray6080, line 118: I think you can directly assert that nodeIDVector is unfiltered here.

ray6080 (approved):

PR #3177, opened 2024-03-31
Refactor sel vector interface

andyfengHKU:
This PR ends up being a pure renaming PR. 

We rename `resetSelectorToUnselected` to `setToUnfiltered` to be consistent with `isUnfiltered`. And `resetSelectorToValuePosBuffer` to `setToMutable`. The reasons are as follow

- `reset` has a very specific meaning of setting to initial state. Out interface in `SelVector` clearly not qualify this notion.
- `resetSelectorToValuePosBuffer` is quite confusing because developer do not know what `ValuePosBuffer` is suppose to be. We should avoid name an interface by naively describing a code block (e.g. `resetSelectorToValuePosBuffer ` is just describing `selectedPositions = selectedPositionsBuffer.get();`). Instead, call it `setToMutable` makes more sense because by default `selectedPositions` points to a static array that is immutable.

I also take the chance to examine if we were using `SelVector` wrongly. I spot a few pieces where @ray6080 should comment as it is storage code.

### DIFF
--- a/src/include/common/data_chunk/sel_vector.h
+++ b/src/include/common/data_chunk/sel_vector.h
@@ -11,27 +11,23 @@ class SelectionVector {
 public:
     explicit SelectionVector(sel_t capacity) : selectedSize{0} {
         selectedPositionsBuffer = std::make_unique<sel_t[]>(capacity);
-        resetSelectorToUnselected();
+        setToUnfiltered();
     }
 
-    inline bool isUnfiltered() const {
-        return selectedPositions == (sel_t*)&INCREMENTAL_SELECTED_POS;
-    }
-    inline void resetSelectorToUnselected() {
-        selectedPositions = (sel_t*)&INCREMENTAL_SELECTED_POS;
-    }
-    inline void resetSelectorToUnselectedWithSize(sel_t size) {
+    bool isUnfiltered() const { return selectedPositions == (sel_t*)&INCREMENTAL_SELECTED_POS; }
+    void setToUnfiltered() { selectedPositions = (sel_t*)&INCREMENTAL_SELECTED_POS; }
+    void setToUnfiltered(sel_t size) {
         selectedPositions = (sel_t*)&INCREMENTAL_SELECTED_POS;
         selectedSize = size;
     }
-    inline void resetSelectorToValuePosBuffer() {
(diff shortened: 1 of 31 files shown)


src/include/common/data_chunk/sel_vector.h
```diff
@@ -11,27 +11,22 @@ class SelectionVector {
...
+    void setToUnfiltered(sel_t size) {
         selectedPositions = (sel_t*)&INCREMENTAL_SELECTED_POS;
         selectedSize = size;
     }
-    inline void resetSelectorToValuePosBuffer() {
-        selectedPositions = selectedPositionsBuffer.get();
-    }
-    inline void resetSelectorToValuePosBufferWithSize(sel_t size) {
+
+    void setBufferToMutable() { selectedPositions = selectedPositionsBuffer.get(); }
```
ray6080, line 24: I would still prefer to call this `setToFiltered()` instead of `setBufferToMutable()`. The former one is more consistent with `setToUnfiltered()`, though we should comment that "filtered" not only refer to filtered out selected values, but any changes to selVector so it is no longer a sequence from 0 (e.g., we might sort values)

src/storage/store/column_chunk.cpp
```diff
@@ -318,13 +318,16 @@ void ColumnChunk::resize(uint64_t newCapacity) {
 }
 
 void ColumnChunk::populateWithDefaultVal(ValueVector* defaultValueVector) {
+    // TODO(Guodong): don't set vector state to a new one. Default vector is shared across all
+    // operators on the pipeline so setting its state will affect others.
+    // You can only set state for vectors that is local to this class.
```
ray6080, line 323: make sense to me. will refactor later.

src/storage/store/var_list_column_chunk.cpp
```diff
@@ -126,8 +126,9 @@ void VarListColumnChunk::append(ValueVector* vector, const SelectionVector& selV
     }
     varListDataColumnChunk->resizeBuffer(nextListOffsetInChunk);
     auto dataVector = ListVector::getDataVector(vector);
+    // TODO(Guodong): we should not set vector to a new state.
```
ray6080, line 129: make sense to me.

ray6080 (approved):

PR #3189, opened 2024-04-01
Hash index multi copy

benjaminwinger:
This doesn't enable multi-copy yet, but restructures the HashIndex to support it.

See #2938 for details, including benchmarks. Currently the performance is close to the original for the initial copy.

The HashIndexBuilder has been left mostly as-is to use as the in-memory local storage for bulk inserts. . Currently the number of primary slots is forced to be the same as the number of primary slots after merging with the on-disk index, which is fine for the first copy, but should be revisited when multi-copy is enabled as it will use more memory than is necessary.

I've made `DBFileUtils::createWALVersionIfNecessaryAndPinPage` public again so it can be used by `BaseDiskArrayInternal::Iterator` to cache pinned pages, with a note that `updatePage`/`insertNewPage` should generally be preferred.

The `PrimaryKeyIndex::prepareCommit` needs to run before the `COPY TABLE` record is logged to the WAL file (since when checkpointing it reloads the index from disk on reading that record), but also gets run after the IndexBuilder has finished during the prepareCommit stage as a result of #3109 (which is also necessary when inserts/deletes are done). I've added a flag to make sure it only executes once before checkpointing but I think there should be a better way of handling that.

The disk array can bypass the WAL, but is configured only to do so in the hash index to avoid wasting pages when rolling back since they won't be removed/truncated from the original file. 

This should generally extend well to additional copies, with the exception that resizing an existing HashIndex is probably going to be slow, since the current `splitSlot` method was not optimized here and doesn't scale well to doing many splits at once. It should be possible to update it to use the new disk array write iterator to split multiple slots at once relatively efficiently.


src/storage/index/hash_index.cpp
```diff
@@ -208,6 +220,7 @@ void HashIndex<T>::insertIntoPersistentIndex(Key key, offset_t value) {
         ;
     copyKVOrEntryToSlot<Key, false /* insert kv */>(
         iter.slotInfo, iter.slot, key, value, fingerprint);
+    updateSlot(iter.slotInfo, iter.slot);
```
ray6080, line 223: Why do `updateSlot` here again? I thought you already did that inside `copyKVOrEntryToSlot`.

src/storage/index/hash_index.cpp
```diff
@@ -363,6 +418,198 @@ void HashIndex<T>::copyEntryToSlot(slot_id_t slotId, const T& entry, uint8_t fin
...
+                            diskSlot = &*diskOverflowSlotIterator;
+                            diskEntryPos = 0;
+                        }
+                    }
+                    KU_ASSERT(diskEntryPos < getSlotCapacity<T>());
+                    copyAndUpdateSlotHeader<const T&, true>(*diskSlot, diskEntryPos,
+                        localSlot.slot->entries[entryPos].key, UINT32_MAX,
+                        localSlot.slot->header.fingerprints[entryPos]);
+                    KU_ASSERT([&]() {
+                        auto hash = hashStored(
```
ray6080, line 535: I feel we should directly store the hash inside localSlot, instead of recomputing it. Can you give this a try as an optimization later?

src/storage/index/hash_index.cpp
```diff
@@ -363,6 +418,198 @@ void HashIndex<T>::copyEntryToSlot(slot_id_t slotId, const T& entry, uint8_t fin
...
+            // builder
+            continue;
+        }
+        diskSlotIterator.seek(slotId);
+        slot_id_t diskEntryPos = 0u;
+        auto* diskSlot = &*diskSlotIterator;
+        // Merge slot from local storage to existing slot
+        do {
+            for (auto entryPos = 0u; entryPos < getSlotCapacity<T>(); entryPos++) {
+                if (localSlot.slot->header.isEntryValid(entryPos)) {
```
ray6080, line 512: This piece of code presents an optimization opportunity that we should add a fast path for the case that all valid $k$ entries are stored as first $k$ ones in the slot. ( In many cases, there is no deletions to entries in local slots). so we can skip the redundant if check on validity bitmap, and also we don't need to loop until slot capacity, which may not make much difference.

src/storage/index/hash_index.cpp
```diff
@@ -363,6 +418,198 @@ void HashIndex<T>::copyEntryToSlot(slot_id_t slotId, const T& entry, uint8_t fin
...
+            // There should be no entries in the overflow slots, as we never leave gaps in the
+            // builder
+            continue;
+        }
+        diskSlotIterator.seek(slotId);
+        slot_id_t diskEntryPos = 0u;
+        auto* diskSlot = &*diskSlotIterator;
+        // Merge slot from local storage to existing slot
+        do {
+            for (auto entryPos = 0u; entryPos < getSlotCapacity<T>(); entryPos++) {
```
ray6080, line 511: Can you wrap the logic of this for loop to a function? like `mergeSlot`?

src/storage/index/hash_index_builder.cpp
```diff
@@ -17,53 +20,54 @@ namespace kuzu {
...
-        this->indexHeader->nextSplitSlotId = numRequiredSlots - numSlotsOfCurrentLevel;
-    }
-    // The next slot to split should always be within the slots covered by the current level
-    // The level should be increased if it goes beyond that point
-    KU_ASSERT(this->indexHeader->nextSplitSlotId <= this->indexHeader->levelHashMask);
-    auto existingSlots = pSlots->getNumElements();
-    if (numRequiredSlots > existingSlots) {
-        allocatePSlots(numRequiredSlots - existingSlots);
+    if (indexHeader.numEntries == 0) {
+        auto numSlotsOfCurrentLevel = 1u << this->indexHeader.currentLevel;
```
ray6080, line 41: Can you wrap this into a separate function?

src/include/storage/storage_structure/disk_array.h
```diff
@@ -137,7 +139,113 @@ class BaseDiskArrayInternal {
...
+        // Constructs Iterator in an invalid state. Seek must be called before accessing data
+        Iterator(uint32_t valueSize, BaseDiskArrayInternal& diskArray,
+            std::unique_lock<std::shared_mutex>&& lock)
+            : diskArray(diskArray), apCursor(),
+              valueSize(valueSize), walPageIdxAndFrame{common::INVALID_PAGE_IDX,
+                                        common::INVALID_PAGE_IDX, nullptr},
+              idx(0), originalNumElements(diskArray.getNumElementsNoLock(TRX_TYPE)),
+              lock(std::move(lock)) {}
+
+        inline Iterator& seek(size_t newIdx) {
```
ray6080, line 170: Should this function be inline? ditto on `pushBack`, `unpin` and `getPage`.

src/include/storage/storage_structure/disk_array.h
```diff
@@ -137,7 +139,113 @@ class BaseDiskArrayInternal {
...
+        // will be modified, so it may be faster this way.
+        WALPageIdxAndFrame walPageIdxAndFrame;
+        static const transaction::TransactionType TRX_TYPE = transaction::TransactionType::WRITE;
+        uint64_t idx;
+        uint64_t originalNumElements;
+        std::unique_lock<std::shared_mutex> lock;
+        DEFAULT_MOVE_CONSTRUCT(Iterator);
+
+        // Constructs Iterator in an invalid state. Seek must be called before accessing data
+        Iterator(uint32_t valueSize, BaseDiskArrayInternal& diskArray,
```
ray6080, line 162: `BatchWriteIterator` looks a better name to me.

ray6080 (approved):
LGTM!

I can see there are still some TODOs left for following PRs I assume. I list them here as how I understand (some from TODOs you left and some from my comments):
- Move HashIndexBuilder away from DiskArray.
- Allow constructing a PrimaryHashIndex without preparing the index file beforehand. Then we should get rid of `WALReplayerUtils::createEmptyHashIndexFiles`.
- Optimize local hash index to not allocate same number of slots as the persistent one to reduce memory footprint.
- Simplify local storage to replace `unordered_map` based local storage for small insertions/deletions with `HashIndexBuilder`. (if the performance doesn't decrease too much. I think it's fine to sacrifice some performance here for simplicity)
- Store hash in local in memory slots to avoid re-computing the hash during merging.
- Fast path for avoid checking validity bitmap during merging local slot into persistent one.
- Avoid re-hashing a slot multiple times in `reserve`.

src/include/storage/index/hash_index.h
```diff
@@ -2,11 +2,18 @@
...
+#include "common/file_system/virtual_file_system.h"
 #include "common/type_utils.h"
+#include "common/types/internal_id_t.h"
 #include "common/types/ku_string.h"
 #include "common/types/types.h"
 #include "hash_index_header.h"
 #include "hash_index_slot.h"
+#include "storage/buffer_manager/bm_file_handle.h"
+#include "storage/buffer_manager/buffer_manager.h"
+#include "storage/file_handle.h"
```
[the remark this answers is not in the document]
benjaminwinger: It was pulled in by using FileHandle in `PrimaryKeyIndex::createEmptyHashIndexFiles`, but I've moved that implementation to the cpp file now.

src/include/storage/index/hash_index.h
```diff
@@ -25,6 +32,7 @@ class OnDiskHashIndex {
     virtual void prepareRollback() = 0;
     virtual void checkpointInMemory() = 0;
     virtual void rollbackInMemory() = 0;
+    virtual void bulkReserve(uint64_t numValuesToAppend) = 0;
```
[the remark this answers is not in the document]
benjaminwinger: Yes, it's being used so that the HashIndex<T> only needs a type when it is accessed, and can be stored as a `unique_ptr<OnDiskHashIndex>` (and it simplifies a few functions that don't need type information).

src/include/storage/index/hash_index.h
```diff
@@ -184,7 +219,8 @@ class HashIndex final : public OnDiskHashIndex {
     std::unique_ptr<BaseDiskArray<Slot<T>>> pSlots;
     std::unique_ptr<BaseDiskArray<Slot<T>>> oSlots;
     OverflowFileHandle* overflowFileHandle;
-    std::unique_ptr<HashIndexLocalStorage<LocalStorageType>> localStorage;
+    std::unique_ptr<HashIndexLocalStorage<BufferKeyType>> localStorage;
+    HashIndexBuilder<T> bulkInsertLocalStorage;
```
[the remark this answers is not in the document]
benjaminwinger: When I'd benchmarked it before it was somewhat slower, but I'm also planning on trying to optimize it somewhat now that we've removed the constraint that it must be able to be directly written to disk. I imagine that after removing the constraint on the number of primary slots for `bulkInsertLocalStorage` it will be close enough that the difference is dominated by the cost of writing to the WAL anyway.

src/include/storage/index/hash_index.h
```diff
@@ -65,6 +73,18 @@ class HashIndex final : public OnDiskHashIndex {
     void deleteInternal(Key key) const;
     bool insertInternal(Key key, common::offset_t value);
 
+    using BufferKeyType =
+        typename std::conditional<std::same_as<T, common::ku_string_t>, std::string, T>::type;
+    // Appends the buffer to the index. Returns the number of values successfully inserted.
+    // I.e. if a key fails to insert, its index will be the return value
```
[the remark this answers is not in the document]
benjaminwinger: Good idea. I've updated the description

src/storage/index/hash_index.cpp
```diff
@@ -99,10 +105,11 @@ HashIndex<T>::HashIndex(const DBFileIDAndName& dbFileIDAndName,
     const std::shared_ptr<BMFileHandle>& fileHandle, OverflowFileHandle* overflowFileHandle,
     uint64_t indexPos, BufferManager& bufferManager, WAL* wal)
     : dbFileIDAndName{dbFileIDAndName}, bm{bufferManager}, wal{wal}, fileHandle(fileHandle),
-      overflowFileHandle(overflowFileHandle) {
+      overflowFileHandle(overflowFileHandle), bulkInsertLocalStorage{overflowFileHandle} {
+    // TODO: Handle data not existing
     headerArray = std::make_unique<BaseDiskArray<HashIndexHeader>>(*fileHandle,
```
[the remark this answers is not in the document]
benjaminwinger: Yes. I was thinking that we could probably store the headers in the PrimaryKeyIndex instead of in the HashIndex, so that we could write them as a single page (each header will need to be <=16 bytes). Most of the fields in the header can be calculated when they are read from disk.

I think all that's needed is:
- 1 byte for currrentLevel
- 8 bytes for numEntries

The masks are already just being calculated from `currentLevel`. nextSplitSlotId can be calculated from the number of primary slots, but that admittedly makes things more complicated as it's stored elsewhere. Alternatively I suppose they could be stored in two pages, which might be a little simpler.

src/include/storage/index/hash_index.h
```diff
@@ -251,14 +312,43 @@ class PrimaryKeyIndex {
...
+    }
+
 private:
+    // When doing batch inserts, prepareCommit needs to be run before the COPY TABLE record is
+    // logged to the WAL file, since the index is reloaded when that record is replayed. However
+    // prepareCommit will also be run later, and the local storage can't cleared from the
+    // HashIndices until checkpointing is done, and entries will get added twice if
+    // HashIndex::prepareCommit is run twice. It seems simplest to just track whether or not
+    // prepareCommit has been run.
+    bool hasRunPrepareCommit;
```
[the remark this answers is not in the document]
benjaminwinger: Yes, I had tried removing that and ran into some errors, but I'll take another look at it.

src/include/storage/storage_structure/disk_array.h
benjaminwinger: How about just WriteIterator? The fact that it's an iterator implies you'd usually use it for batch updates, though I doubt it would be noticeably slower than `BaseDiskArray::update` for single updates.

src/storage/index/hash_index.cpp
benjaminwinger: Yes, that's a good idea. I think when the builder/`InMemHashIndex` gets restructured the validity mask can be removed and we can just store the number of entries instead.

src/storage/index/hash_index.cpp
```diff
@@ -255,14 +268,54 @@ template<typename T>
...
+    }
+    headerArray->prepareCommit();
+    pSlots->prepareCommit();
+    oSlots->prepareCommit();
+}
+
+template<typename T>
+// Checks that everything in the bulkInsertLocalStorage can be found in the index
+// Only meaningful if called after mergeBulkInserts and before checkpointing.
+void HashIndex<T>::validateBulkInserts() {
```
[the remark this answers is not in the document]
benjaminwinger: I was using it for testing. It actually seems to be broken at the moment, so I think I'll just remove it along with the other debugging functions.

src/storage/index/hash_index.cpp
benjaminwinger: I seem to recall removing that part of `copyKVOrEntryToSlot`, but I must have reverted it. I've fixed it (and merged the first two arguments since it's only being used with an iterator).

src/storage/index/hash_index.cpp
benjaminwinger: Yes, that was part of what I was planning when I get to re-writing and optimizing the `HashIndexBuilder`, particularly since we need it to calculate the new primary slots if they don't match.

src/include/storage/storage_structure/disk_array.h
ray6080: Sure. `WriteIterator` is clear.

src/include/storage/index/hash_index.h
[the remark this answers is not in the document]
benjaminwinger: I tried again to remove it, but I think the reliance on a separate function to initially create the empty files might have been causing issues, and getting the HashIndex and DiskArray to properly handle the files not existing was taking more time; I'll return to it later.

PR #3191, opened 2024-04-01
Rework multiple query result

hououou:
Rework query results of multiple query statements. We use the list chain of query results for multiple query statements. To get each query result, we could loop to get each query result. Implement the following interfaces:

- `toString()` : return the string of the first query result. 
- `hasNextQueryResult()`: return true if we have more query results otherwise false
- `getNextQueryResult()`: return the next query result

Note the query result is the result for each individual query statement.
TODO: We should expose `hasNextQueryResult()``getNextQueryResult()` in java/nodejs/rust/python apis.

### DIFF
--- a/src/include/main/query_result.h
+++ b/src/include/main/query_result.h
@@ -97,15 +97,22 @@ public:
      * @return whether there are more tuples to read.
      */
     KUZU_API bool hasNext() const;
-    std::unique_ptr<QueryResult> nextQueryResult;
+    /**
+     * @return whether there are more query results to read.
+     */
+    KUZU_API bool hasNextQueryResult() const;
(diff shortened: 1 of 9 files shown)


src/include/main/query_result.h
```diff
@@ -158,6 +165,9 @@ class QueryResult {
 
     // execution statistics
     std::unique_ptr<QuerySummary> querySummary;
+
+    // query iterator
+    std::unique_ptr<QueryResultIterator> queryResultIterator;
```
andyfengHKU, line 170: No need to use unique_ptr here.

andyfengHKU (approved):

PR #3192, opened 2024-04-01
Added progress for aggregate scan and order by scan

MSebanc:

### DIFF
--- a/src/include/processor/operator/order_by/order_by_scan.h
+++ b/src/include/processor/operator/order_by/order_by_scan.h
@@ -12,12 +12,11 @@ struct OrderByScanLocalState {
     uint64_t numTuples;
     uint64_t numTuplesRead;
 
-
     void init(
         std::vector<DataPos>& outVectorPos, SortSharedState& sharedState, ResultSet& resultSet);
 
     // NOLINTNEXTLINE(readability-make-member-function-const): Updates vectorsToRead.
-    inline uint64_t scan() { 
+    inline uint64_t scan() {
         uint64_t tuplesRead = payloadScanner->scan(vectorsToRead);
         numTuplesRead += tuplesRead;
         return tuplesRead;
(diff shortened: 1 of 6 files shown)


src/processor/operator/aggregate/hash_aggregate_scan.cpp
```diff
@@ -37,5 +37,15 @@ bool HashAggregateScan::getNextTuplesInternal(ExecutionContext* /*context*/) {
     return true;
 }
 
+double HashAggregateScan::getProgress(ExecutionContext* /*context*/) const {
+    uint64_t totalNumFlatTuples = sharedState->getFactorizedTable()->getTotalNumFlatTuples();
```
andyfengHKU, line 41: Let's use `getNumTuples`. `getTotalNumFlatTuples` has an overhead and is not meant to be called at runtime repeatedly. Double check if you have used this function in somewhere else.

andyfengHKU (changes requested):
As we discussed, order by scan can only be executed with single thread so move the progress to local state.

src/processor/operator/order_by/order_by_scan.cpp
```diff
@@ -12,6 +12,10 @@ void OrderByScanLocalState::init(
     }
     payloadScanner = std::make_unique<PayloadScanner>(
         sharedState.getMergedKeyBlock(), sharedState.getPayloadTables());
+    for (auto& table : sharedState.getPayloadTables()) {
+        numTuples += table->getNumTuples();
```
andyfengHKU, line 16: Let's also initialize numTuples to 0.

src/include/processor/operator/order_by/order_by_scan.h
```diff
@@ -9,12 +9,18 @@ namespace processor {
...
     std::unique_ptr<PayloadScanner> payloadScanner;
+    uint64_t numTuples;
+    uint64_t numTuplesRead;
 
     void init(
         std::vector<DataPos>& outVectorPos, SortSharedState& sharedState, ResultSet& resultSet);
 
     // NOLINTNEXTLINE(readability-make-member-function-const): Updates vectorsToRead.
-    inline uint64_t scan() { return payloadScanner->scan(vectorsToRead); }
+    inline uint64_t scan() {
```
andyfengHKU, line 19: remove inline

andyfengHKU (approved):

PR #3198, opened 2024-04-02
Rework offset column of LIST

hououou:
see #3093 

rework offset column


ray6080 (changes requested):
The main problems is you are still allocating buffer/scanning/committing for main column (chunk) after moving offsets out to a offsetColumn(Chunk). so the pr is quite confusing as it is trying to move offsets out of main column, while not fully done.

I think we need a bit more work on this. let's double see if we really need a separate offsetColumn.

ray6080:
see #3219.

PR #3209, opened 2024-04-04
change LogicalType.toString for nested types

mxwli:
The problem:
The result of LogicalType::toString was not compatible with LogicalType::dataTypeIDFromString. For example,
 toString would give `STRUCT(A:INT32)` while dataTypeIDFromString would expect `STRUCT(A INT32)`. This PR changes that, and updates corresponding tests.


andyfengHKU (approved):
LGTM

PR #3211, opened 2024-04-04
Fix constant compression in-place check for bools

benjaminwinger:
Bools are stored packed in column chunks, so indexing into the data using the size in bytes of the type will not work.

This is done as a quick fix, however I think it would be a good idea to move away from passing around raw `uint8_t*` when passing around type-generic data. Instead, I think we should create abstractions for storing and accessing the raw buffers. That would let us more easily distinguish between, for example, a ValueVector buffer and a ColumnChunk buffer (e.g. the former stores 1 bool per byte, and the latter 1 bool per bit), and also make sure we handle these edge cases.

### DIFF
--- a/src/include/storage/store/null_column.h
+++ b/src/include/storage/store/null_column.h
@@ -43,20 +43,12 @@ public:
     void write(common::node_group_idx_t nodeGroupIdx, common::offset_t offsetInChunk,
         ColumnChunk* data, common::offset_t dataOffset, common::length_t numValues) override;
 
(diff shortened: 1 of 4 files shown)


andyfengHKU (approved):

benjaminwinger:
Apparently this issue was already handled separately for Nulls (which this change initially broke). I don't really like that the implementations diverged.

bigluck:
Thanks @benjaminwinger

PR #3213, opened 2024-04-04
Fix pyarrow segfaulting on fedora 39

### LINKED ISSUE #3202: Python test fails with SEGFAULT on Fedora 39
https://github.com/kuzudb/kuzu/actions/runs/8535225089/job/23381190328#step:14:284

mxwli:
Closes #3202


acquamarin (approved):
can you add a short description about what is the bug and why it appears and how you are solving it?

mxwli:
There are two possible causes of the bug, both of which have been fixed:
1. `auto pyTableName = py::str(objectName);`, when run without a GIL can cause a segfault.
2. `py::reinterpret_borrow` still manages the refcount, so I've elected to not hold a reference to the dataframe inside `PyArrowTableScanFunctionData` anymore.

PR #3220, opened 2024-04-05
Add DATE TO DATE and TIMESTAMP TO DATE casting functions.

### LINKED ISSUE #3218: Unsupported Casting function from DATE to DATE and TIMESTAMP to DATE
Date cannot cast to itself. 

```
CREATE NODE TABLE tab(id SERIAL, col1 STRUCT(a DATE, b INT32), PRIMARY KEY(id));
CREATE (t:tab {col1: {a: date('2024-04-05'), b: 12}});
```
gives `Unsupported casting function from DATE to DATE.`
This is likely because of a missing switch case in `CastFunction::bindCastFunction`

As a result, this is the same for timestamp -> date casts

mxwli:
Closes #3218 

DATE TO DATE casting is called when casting nested datatypes, eg LISTS and STRUCTS, and so registering it is necessary. See the linked issue for an example. 

TIMESTAMP to DATE casting uses `Timestamp::getDate`.


src/include/function/cast/functions/cast_functions.h
```diff
@@ -60,6 +60,39 @@ inline void CastDateToTimestamp::operation(common::date_t& input, common::timest
...
 }
 
+struct CastToDate {
+    template<typename T>
+    static inline void operation(T& input, common::date_t& result);
+};
+
+template<>
+inline void CastToDate::operation(common::date_t& input, common::date_t& result) {
+    result = input;
```
andyfengHKU, line 70: Just a quick comment, why do we perform data to date casting? Should this be a noop(no operation)?

mxwli: When casting between nested types, a cast is performed on the children regardless of whether or not they're the same.
```
CREATE NODE TABLE tab(id SERIAL, col1 STRUCT(a DATE, b INT32), PRIMARY KEY(id));
CREATE (t:tab {col1: {a: date('2024-04-05'), b: 12}});
```
An alternative solution to this is checking whether or not children are identical in nested casting.

andyfengHKU: I tend to check check instead adding a X to X cast if checking is not tricky.

mxwli:
note: in copying the ValueVector in resolving nested types, an assert for `resultVector.state.selVector.isUnfiltered` is not made. This is in line with CastChildFunctionExecutor::executeSwitch, which is the function that would have run previously.

test/test_files/common/todatecast.test
```diff
@@ -0,0 +1,34 @@
...
+
+-STATEMENT RETURN CAST(date('2024-04-05'), 'date')
+---- 1
+2024-04-05
+
+-STATEMENT RETURN CAST(timestamp('2024-04-05'), 'date')
+---- 1
+2024-04-05
+
+-STATEMENT RETURN CAST(CAST('2024-04-05', 'TIMESTAMP_MS'), 'date')
```
andyfengHKU, line 16: What will happen if I cast a `timestamp` with hour, minute and second to `date`?

src/function/vector_cast_functions.cpp
```diff
@@ -69,12 +69,18 @@ static void resolveNestedVector(std::shared_ptr<ValueVector> inputVector, ValueV
     }
 
     // non-nested types
-    scalar_func_exec_t func = CastFunction::bindCastFunction<CastChildFunctionExecutor>(
-        "CAST", inputType->getLogicalTypeID(), resultType->getLogicalTypeID())
-                                  ->execFunc;
-    std::vector<std::shared_ptr<ValueVector>> childParams{inputVector};
-    dataPtr->numOfEntries = numOfEntries;
-    func(childParams, *resultVector, (void*)dataPtr);
+    if (inputType->getLogicalTypeID() != resultType->getLogicalTypeID()) {
```
andyfengHKU, line 72: assert inputVector is `unFiltered` before line 72.

andyfengHKU (approved):

src/function/vector_cast_functions.cpp
mxwli: When run locally, the assertion works just fine, but when run in the CI, the code segfaults on tests. It's possible that the assertion fails on different platforms. Inspecting the previous code, there was no unfiltered assertion previously. It seems like this code was written without the filter in mind, so I removed the assert.

test/test_files/common/todatecast.test
mxwli: It gives the date part of the timestamp without the hour:minute:second. For example,
`RETURN CAST(timestamp('2024-04-05 23:59:59.999'), 'date')`
gives
`2024-04-05`

PR #3221, opened 2024-04-05
Add demo graph algorithm as a table function.

andyfengHKU:
This PR a proof concept to demonstrate how to

- Add graph class as an abstract layer and pass it as the input of an algorithm function.
- Register algorithm function as a table function and query it with `CALL`


src/include/graph/function.h
```diff
@@ -0,0 +1,19 @@
...
+#include "function/function.h"
+#include "function/table/call_functions.h"
+
+namespace kuzu {
+namespace graph {
+
+// CallFunction has the assumption that number of output is known before execution.
+// This does NOT hold for all graph algorithms.
+// So each algorithm need to decide its own shared state to control when to terminate.
+struct DemoAlgorithm : public function::CallFunction {
```
semihsalihoglu-uw, line 12: This is the class that I don't understand. What's "function/function.h vs graph/function.h"? I was expecting to see an algorithm.h class that forms an interface for algorithms, with a generic compute function? I understand Algorithms will be registered as TableFunctions but people implementing new algorithms in the system (us or eventual other developers who write extensions) should not know about the function infrastructure. So it looks like we need a higher-level abstraction over functions to develop algorithms. For example, it's confusing to me that the Graph field is part of DemoAlg's shared state (I am referring to the field in the `struct DemoAlgoSharedState : public CallFuncSharedState` in function.cpp. 

I also need to understand how to pass arguments to functions, e.g., if we needed to pass a start source node to a ShortestPaths algorithm, how do we pass it and how much of that could be simplified. For example, can we abstract away from developers the details of binding expressions or writing any or minimal binding code? Similarly for output expressions of algorithms, I was imagining there is a `Algorithm<AlgorithmResult>` interface and AlgorithmResult has a factorized table as output, from which the output expressions can be read. Or something along those lines.

PR #3226, opened 2024-04-06
Always build rust integration with release runtime library on Windows

### LINKED ISSUE #3225: Linking error when building rust crate in debug mode on Windows
The rust crate fails to build on Windows because the C++ libraries are trying to link to the debug vc runtime (`/MDd`), whereas Rust always links against the release runtime (`/MD`), even when building debug configuration.

Release builds fine.

zaddach:
This fixes #3225

### DIFF
--- a/.github/workflows/ci-workflow.yml
+++ b/.github/workflows/ci-workflow.yml
@@ -273,8 +273,8 @@ jobs:
         run: |
           call "C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Auxiliary\Build\vcvars64.bat"
           set KUZU_TESTING=1
-          set CFLAGS=/MDd
-          set CXXFLAGS=/MDd /std:c++20
+          set CFLAGS=/MD
+          set CXXFLAGS=/MD /std:c++20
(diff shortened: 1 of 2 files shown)


benjaminwinger (approved):
See https://github.com/kuzudb/kuzu/issues/3225#issuecomment-2042630355, but I don't have any objections to doing this. As far as I understand it just means we don't get debug symbols for the MSVC runtime library, but would still get them for kuzu itself in debug mode, and I think this is probably a cleaner way of handling it than having to set environment variables.

benjaminwinger:
The CI pipeline will need to be updated to remove the environment variables overriding with `/MDd` ([here](https://github.com/kuzudb/kuzu/blob/c6e62e9e08818b6ba178ebb341c8c35edfefcdf3/.github/workflows/ci-workflow.yml#L262-L263)).

Additionally, can you add `I have read and agree to the terms under CLA.md` to the description in this PR (I guess you just did in  #3228, but it's easier to keep track if it's included in each PR).

zaddach:
I have read and agree to the terms under CLA.md

PR #3228, opened 2024-04-06
CMAKE_CXX_FLAGS handling fails when variable is empty

### LINKED ISSUE #3227: CMAKE_CXX_FLAGS handling fails when empty string
When I pass an empty string (I was actually trying to pass `/EHsc`, which gets removed in the line before), cmake will trip here, telling me that `STRING(REPLACE ... )` didn't get enough arguments: https://github.com/kuzudb/kuzu/blob/d946982e77af0f8a0bf41e7fea701c61db6f098f/CMakeLists.txt#L108

zaddach:
Fixes #3227

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -104,8 +104,8 @@ if(MSVC)
     # Enables support for custom hardware exception handling
     add_compile_options("/EHa")
     # Remove the default to avoid warnings
-    STRING(REPLACE "/EHsc" "" CMAKE_CXX_FLAGS ${CMAKE_CXX_FLAGS})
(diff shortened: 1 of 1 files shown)


ray6080:
Hi @zaddach , thanks for your fix!
Could you please take a look at our [CLA](https://github.com/kuzudb/kuzu/blob/master/CLA.md) and add “I have read and agree to the terms under CLA.md” to your PR description if you agree with the CLA?

zaddach:
I have read and agree to the terms under CLA.md

ray6080 (approved):

PR #3229, opened 2024-04-07
Ensure that pdb is generated in release mode for win debugging

neeraj9:
I have read and agree to the terms under [CLA.md](https://github.com/kuzudb/kuzu/blob/master/CLA.md)

Noticed that when kuzu_shared.dll is built in debug mode then somehow client app crashes at randomly at time during new or delete. It is not clear what is going wrong there. This change will at least generate PDB (debugging symbols) for release mode, while keeping the release mode DLL same as before. This will allow better debugging on windows.

- https://www.atmosera.com/blog/correctly-creating-native-c-release-build-pdbs/
- https://discourse.cmake.org/t/use-cmake-to-generate-pdb/2583
- https://stackoverflow.com/questions/28178978/how-to-generate-pdb-files-for-release-build-with-cmake-flags

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -112,7 +112,12 @@ if(MSVC)
 endif()
 if(CMAKE_BUILD_TYPE MATCHES Release)
     if(MSVC)
-        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /O2 /utf-8")
+        # use /Zi for debug symbols even in release mode for pdb debugging
+        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /O2 /utf-8 /Zi")
+        # /DEBUG is required for pdb debugging to work and /OPT:REF /OPT:ICF are required to enable COMDAT folding
+        # CMAKE_SHARED_LINKER_FLAGS is used for shared libraries and CMAKE_EXE_LINKER_FLAGS is used for executables
+        set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} /DEBUG /OPT:REF /OPT:ICF")
+        set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} /DEBUG /OPT:REF /OPT:ICF")
         string(REGEX REPLACE "/W[3|4]" "/w" CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS}")
         add_compile_options(/W0)
     else()


benjaminwinger:
> Noticed that when kuzu_shared.dll is built in debug mode then somehow client app crashes at randomly at time during new or delete. It is not clear what is going wrong there.

Are you also building the client app in debug mode? MSVC enables standard library data structure debugging for debug builds which adds extra fields to some classes such as `std::string` (see #2521), which means you can't safely pass some data in either direction without using the same debug definitions (and this actually includes the `Database` and `QueryResult` classes since they each contain a `std::string`).

Is the idea that you would like us to distribute debug symbols with the release builds, or is this for personal debugging use in release mode?
If it's the former, It would be useful to know how the size compares to the current release binaries.
If it's the latter, how does this compare to building using using the `RelWithDebInfo` configuration? I'm not sure if the data structure debug definitions are used in that mode, but that seems like it would be the best way of handling an optional release build with debug symbols (and if you want we could add an easier way of running such builds to the `Makefile`, but I think just using `cmake -B <builddir> -DCMAKE_BUILD_TYPE=RelWithDebInfo && cmake --build <builddir> --config RelWithDebInfo` works).

From what I understand from [the docs](https://learn.microsoft.com/en-us/cpp/build/reference/z7-zi-zi-debug-information-format?view=msvc-170), `-Zi` would require us to distribute the PDB files with the release dll:

> If you create a library from objects that were compiled using /Zi, the associated PDB file must be available when the library is linked to a program. That means, if you distribute the library, you must also distribute the PDB file.

neeraj9:
One of my dependencies is built with `/MTd` for Debug and `/MT` for Release configuration. I have to include them as shared libraries and not building from source.

When I forced `/MTd` to all my code and Kuzu, then I noticed the following warning when configuring `kuzu` (at configure stage and **not** at build).

```
Run Build Command(s): C:/PROGRA~1/Meson/ninja.EXE -v cmTC_544df
[1/2] C:\PROGRA~1\MICROS~3\2022\ENTERP~1\VC\Tools\MSVC\1439~1.335\bin\Hostx64\x64\cl.exe  /nologo /TP   /Zi  /MTd -MDd /showIncludes /FoCMakeFiles\cmTC_544d
f.dir\CMakeCXXCompilerABI.cpp.obj /FdCMakeFiles\cmTC_544df.dir\ /FS -c G:\tools\cmake-3.29.0-windows-x86_64\share\cmake-3.29\Modules\CMakeCXXCompilerABI.cpp
^M
cl : Command line warning D9025 : overriding '/MTd' with '/MDd
```

I also inspected `build/compile_commands.json` for kuzu and noticed that `-MDd` is automatically injected although I dont remember setting it. Is there a cmake configuration in kuzu which depends on `/MDd` for Debug and `/MD` for Release?

neeraj9:
**Conclusion 1:**
I found out that MSVC compiler automatically defaults to `/MD` or `/MDd`, so I do get warnigs about overriding `/MDd` with `/MTd` if I set `CMAKE_CXX_FLAGS` to `/MTd` explicitly. For MSVC it is better to set `CMAKE_MSVC_RUNTIME_LIBRARY` directly to "MultiThreaded$<$<CONFIG:Debug>:Debug>". This will alow correct usage of `/MT` or `/MTd`.

**Conclusion 2:**
I tested with `/MTd` build of `kuzu_shared.dll` and still memory deallocation in my application when allocation was done in `kuzu_shared.dll` fails in debug mode. This is despite `/MTd` used at both places. In the mean time I have switched to using `C-API` of kuzu and explicitly using the `*_destroy` functions provided by Kuzu `C-API` to free allocated memory. This is a bit cumbersome and has potential issues of missing this call, but for now relying on careful code review and creating a wrapper works. I will revisiting the memory allocation / free issue even with aligned C-Runtime on windows bit a little later.

PR #3234, opened 2024-04-08
Implement Implicit Casting of Nested Types and Type Combination with MaxLogicalType

mxwli:
Nested Implicit Casting Rules:
If the source type and destination type have different LogicalTypeIDs, no implicit cast is possible
If both types are Lists, an implicit cast is possible if their children have an implicit cast from src to dest. Arrays follow the same rule, but an additional check is made to make sure the arrays are of the same length
If both types are Structs, check if the field names match and then check if the matching children have an implicit cast from src to dest.
If both types are Maps, check that the key and value types have implicit casts from src to dest
Union casting seems to be completely unsupported right now, so no implicit casting is allowed. 

Update: ARRAY -> LIST implicit casting has been implemented. It goes just the way you would expect it to go. 

Max Logical Type Combination Rules:
Case 1:
If at least one type is nonnested, 
1. if one type is ANY/RDF_VARIANT/STRING, cast to the other
2. if one can cast to the other, take the one with the lowest cost
3. if both are integral and one is signed & the other is unsigned, combine to lowest width signed integral type
4. otherwise cast to the value with a larger `internalTypeOrder`, as defined in `types.cpp`
Case 2:
If both types are nested, the rules are identical to implicit casting rules, but instead of checking "if the src children can implicit cast to the dst children", we check "if the src children can combine with the dst children".


src/common/types/types.cpp
```diff
@@ -958,5 +1028,324 @@ std::unique_ptr<LogicalType> LogicalType::ARRAY(std::unique_ptr<LogicalType> chi
         std::make_unique<ArrayTypeInfo>(std::move(childType), numElements)));
 }
 
+static LogicalTypeID joinToWiderType(const LogicalTypeID& left, const LogicalTypeID& right) {
```
andyfengHKU, line 1031: This function seem to assume left and right are integers. If so, let's assert.

src/common/types/types.cpp
```diff
@@ -958,5 +1028,324 @@ std::unique_ptr<LogicalType> LogicalType::ARRAY(std::unique_ptr<LogicalType> chi
...
+    LogicalType result;
+    if (!tryGetMaxLogicalType(left, right, result)) {
+        throw common::BinderException(stringFormat("Cannot combine logical types {} and {}",
+            left.toString(), right.toString()));
+    }
+    return result;
+}
+
+// if we can combine the child types, then we can combine the list
+bool LogicalTypeUtils::tryCombineListTypes(const LogicalType& left, const LogicalType& right,
```
andyfengHKU, line 1244: if -> If. Check other comments

src/common/types/types.cpp
```diff
@@ -958,5 +1028,324 @@ std::unique_ptr<LogicalType> LogicalType::ARRAY(std::unique_ptr<LogicalType> chi
...
+    if (leftFields.size() != rightFields.size()) {
+        return false;
+    }
+    /*
+    auto comp = [](const StructField* const& a, const StructField* const& b) -> bool {
+        return a->getName() < b->getName();
+    };
+    std::sort(leftFields.begin(), leftFields.end(), comp);
+    std::sort(rightFields.begin(), rightFields.end(), comp);
+    // in general, kuzu seems to treat different element orderings as different types alltogether.
```
andyfengHKU, line 1283: Is this the case that we have discussed? If so, remove the comment. Otherwise let's discuss this part.

src/binder/bind_expression/bind_comparison_expression.cpp
```diff
@@ -26,35 +25,31 @@ std::shared_ptr<Expression> ExpressionBinder::bindComparisonExpression(
...
-    std::vector<LogicalType> childrenTypes;
+    LogicalType combinedType(LogicalTypeID::ANY);
+    bool canCombine = true;
     for (auto& child : children) {
-        childrenTypes.push_back(child->dataType);
+        if (!LogicalTypeUtils::tryGetMaxLogicalType(combinedType, child->dataType, combinedType)) {
+            canCombine = false;
+        }
+    }
+    if (!canCombine) {
```
andyfengHKU, line 35: Is this the expect behaviour? I thought it would be dangerous if we always succeed in binding comparison. Throw exception seems to be more reasonable here.

src/binder/expression_binder.cpp
```diff
@@ -126,6 +126,15 @@ std::shared_ptr<Expression> ExpressionBinder::implicitCast(
     }
 }
 
+// cast without implicit checking.
+std::shared_ptr<Expression> ExpressionBinder::forceCast(
```
andyfengHKU, line 130: well this is exactly implicit cast. User didn't write a `CAST` function but we are casting for them. So `forceCast` is not a good name here. Why are we not reusing implicit cast instead?

src/include/function/cast/vector_cast_functions.h
```diff
@@ -19,6 +19,18 @@ struct CastFunction {
     template<typename EXECUTOR = UnaryFunctionExecutor>
     static std::unique_ptr<ScalarFunction> bindCastFunction(const std::string& functionName,
         common::LogicalTypeID sourceTypeID, common::LogicalTypeID targetTypeID);
+
+private:
+    static bool hasImplicitCastList(const common::LogicalType& srcType,
```
andyfengHKU, line 24: move to cpp.

src/include/common/types/types.h
```diff
@@ -579,15 +579,40 @@ class LogicalTypeUtils {
...
     static std::vector<LogicalTypeID> getAllValidLogicTypes();
+    static bool tryGetMaxLogicalTypeID(const LogicalTypeID& left, const LogicalTypeID& right,
+        LogicalTypeID& result);
+    static bool tryGetMaxLogicalType(const LogicalType& left, const LogicalType& right,
+        LogicalType& result);
+    static LogicalTypeID getMaxLogicalTypeID(const LogicalTypeID& left, const LogicalTypeID& right);
+    static LogicalType getMaxLogicalType(const LogicalType& left, const LogicalType& right);
 
 private:
+    static bool tryCombineListTypes(const LogicalType& left, const LogicalType& right,
```
andyfengHKU, line 606: move to cpp for private static functions.

src/function/vector_cast_functions.cpp
```diff
@@ -104,12 +104,73 @@ static void nestedTypesCastExecFunction(const std::vector<std::shared_ptr<ValueV
...
+bool CastFunction::hasImplicitCastMap(const LogicalType& srcType, const LogicalType& dstType) {
+    auto srcKeyType = *MapType::getKeyType(&srcType);
+    auto srcValueType = *MapType::getValueType(&srcType);
+    auto dstKeyType = *MapType::getKeyType(&dstType);
+    auto dstValueType = *MapType::getValueType(&dstType);
+    return hasImplicitCast(srcKeyType, dstKeyType) && hasImplicitCast(srcValueType, dstValueType);
+}
+
+bool CastFunction::hasImplicitCast(const LogicalType& srcType, const LogicalType& dstType) {
+    if ((LogicalTypeUtils::isNested(srcType) &&
```
andyfengHKU, line 150: Can't we just reuse `tryGetMaxLogicalType`?

src/binder/expression_binder.cpp
mxwli: Implicit cast checks for whether or not the cast follows implicit casting rules, while forceCast does not. Here, forceCast is used for casts that are outside of the rules of implicit casting.

src/function/vector_cast_functions.cpp
mxwli: I wouldn't do that here; max logical type is defined for combining two separate types and doesn't exactly follow implicit casting rules.

src/common/types/types.cpp
```diff
@@ -958,5 +1028,324 @@ std::unique_ptr<LogicalType> LogicalType::ARRAY(std::unique_ptr<LogicalType> chi
...
+    if (left == right || (left == LogicalTypeID::ANY || left == LogicalTypeID::RDF_VARIANT ||
+                             left == LogicalTypeID::STRING)) {
+        result = right;
+        return true;
+    } else if (right == LogicalTypeID::ANY || right == LogicalTypeID::RDF_VARIANT ||
+               right == LogicalTypeID::STRING) {
+        result = left;
+        return true;
+    }
+    auto leftToRight = BuiltInFunctionsUtils::getCastCost(left, right);
```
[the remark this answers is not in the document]
mxwli: If so, it would be a better fit for a different PR

andyfengHKU (approved):

PR #3235, opened 2024-04-08
Add coalesce function

manh9203:
This PR added [coalesce](https://duckdb.org/docs/sql/functions/utility#coalesceexpr-) and [ifnull](https://duckdb.org/docs/sql/functions/utility#ifnullexpr-other) functions.


src/function/utility/coalesce.cpp
```diff
@@ -0,0 +1,79 @@
...
+        }
+        result.setNull(resultPos, isNull);
+    }
+}
+
+function_set CoalesceFunction::getFunctionSet() {
+    function_set functionSet;
+    functionSet.push_back(
+        std::make_unique<ScalarFunction>(name, std::vector<LogicalTypeID>{LogicalTypeID::ANY},
+            LogicalTypeID::ANY, execFunc, nullptr, bindFunc, true /*  isVarLength */));
```
andyfengHKU, line 74: You need a selectFunc for coalesce function as well. Take a look at `case_evaluator` and see this can be down through a minimal modification over execFunc.

Also add test cases where coalesce is being used in WHERE predicate.

andyfengHKU (changes requested):

test/test_files/function/hash/utility/coalesce.test
```diff
@@ -0,0 +1,172 @@
+-GROUP CoalesceFunctionTests
+-DATASET CSV empty
+
+--
+
+-CASE ConstantTests
+
+-LOG NoArgumentError
```
andyfengHKU, line 8: I need to see a few more tests. On tinysnb, run `COALESCE` after a `MATCH (a:person)-[:knows]->(b:person)` and maybe also add some more query edges to it. This is to test different combination of parameter states.

Test coalesce of nested data types, LIST, STRUCT, MAP NODE, REL.

Test `RETURN ifNULL`

andyfengHKU (changes requested):
I need to take another look for selectFunc and some additional testing.

andyfengHKU (approved):
LGTM

PR #3243, opened 2024-04-10
Support allocations of larger-than-256KB memory buffers

ray6080:
Removed the hard limit of memory buffer no larger than 256KB. Now we can store strings larger than 256KB.
This is temporarily done through `malloc`. Should be optimized later.

Fixed two bugs:
- incorrect reset of aux buffer in `NodeRelExpressionEvaluator`.
- incorrect strlen in `MD5Update`.

Left todos for later PRs:
- [ ] factorized table still requires the table schema row size less than 256KB. we can remove this too.
- [ ] larger than 256KB memory buffers are not tracked through BM, should fix this.


src/processor/result/factorized_table.cpp
```diff
@@ -335,6 +335,7 @@ uint64_t FactorizedTable::computeNumTuplesToAppend(
 std::vector<BlockAppendingInfo> FactorizedTable::allocateFlatTupleBlocks(
     uint64_t numTuplesToAppend) {
     auto numBytesPerTuple = tableSchema->getNumBytesPerTuple();
+    // TODO(Guodong): Remove this restriction.
```
andyfengHKU, line 338: open an issue

andyfengHKU (approved):

PR #3247, opened 2024-04-10
Pass nullMask to setValuesFromUncompressed

### LINKED ISSUE #3143: Rel with string property insertion error
The attached test file triggers assertion
```
C++ exception with description "Assertion failed in file "/Users/guodong/Developer/kuzu/src/storage/compression/compression.cpp" on line 408: canUpdateInPlace(value, header)" thrown in the test body.
```
[insert_rel_str_1000_1000_1.txt](https://github.com/kuzudb/kuzu/files/14757647/insert_rel_str_1000_1000_1.txt)

benjaminwinger:
Fixes #3143 by making the compression code aware of null values which don't need to be able to be read later (0 is not necessarily compressible when doing in-place updates to compressed data).

I've made a few modifications to NullMask to clean up its initialization and to allow it to reference existing data in NullColumnChunks. I think eventually it will be useful to separate management of data buffers from the ColumnChunks, and at that point the NullColumnChunk could more or less just be a ColumnChunk that uses a NullMask to store its data, but that will take more work.

`NullMask::resize` was also broken since memcpy takes the number of bytes to copy, not the number of (in this case) 64-bit values, so it was only copying the beginning of the data. I've added a unit test since we obviously aren't hitting that in the integration tests (it's only being used in one spot in the ValueVector's AuxiliaryBuffer).

### DIFF
--- a/src/storage/store/dictionary_column.cpp
+++ b/src/storage/store/dictionary_column.cpp
@@ -101,9 +101,9 @@ void DictionaryColumn::scan(Transaction* transaction, node_group_idx_t nodeGroup
 
 string_index_t DictionaryColumn::append(node_group_idx_t nodeGroupIdx, std::string_view val) {
     auto startOffset = dataColumn->appendValues(nodeGroupIdx,
-        reinterpret_cast<const uint8_t*>(val.data()), val.size());
+        reinterpret_cast<const uint8_t*>(val.data()), nullptr /*nullChunkData*/, val.size());
     return offsetColumn->appendValues(nodeGroupIdx, reinterpret_cast<const uint8_t*>(&startOffset),
-        1 /*numValues*/);
+        nullptr /*nullChunkData*/, 1 /*numValues*/);
 }
 
 void DictionaryColumn::prepareCommit() {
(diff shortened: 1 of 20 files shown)


src/storage/store/dictionary_column.cpp
```diff
@@ -101,9 +101,9 @@ void DictionaryColumn::scan(Transaction* transaction, node_group_idx_t nodeGroup
 
 string_index_t DictionaryColumn::append(node_group_idx_t nodeGroupIdx, std::string_view val) {
     auto startOffset = dataColumn->appendValues(nodeGroupIdx,
-        reinterpret_cast<const uint8_t*>(val.data()), val.size());
+        reinterpret_cast<const uint8_t*>(val.data()), nullptr /*nullChunkData=*/, val.size());
```
ray6080, line 104: Let's remove the `=`?

ray6080 (approved):

src/storage/store/dictionary_column.cpp
benjaminwinger: I'm not sure how that got there.

PR #3279, opened 2024-04-15
Fix err msg in failed rust test

ray6080:


andyfengHKU (approved):

tools/rust_api/src/database.rs
```diff
@@ -159,7 +160,7 @@ mod tests {
     #[test]
     fn create_database_failure() {
         let result: Error = Database::new("", SystemConfig::default())
-            .expect_err("An empty string should not be a valid database path!")
+            .expect_err("IO exception: An empty string should not be a valid database path!")
```
benjaminwinger, line 163: The error here is a description in case there wasn't an error message. [expect_err](https://doc.rust-lang.org/std/result/enum.Result.html#method.expect_err) panics if the result is not an error, but it doesn't care what the error message says.

ray6080: Interesting to learn that "it doesn't care what the error message says."

I'm still confused here on this test. It should still throw an exception. I don't get why it fails now if it doesn't care about what the error message says. And I don't get the different of keeping or removing `let result: Error =` part.

Maybe better you take this PR over?

benjaminwinger: This test doesn't appear to be failing (The rust bridge code catches the exception produced when you create a database with an empty path and turns it into an `Error`, and then this test is just checking that it is actually an error and will only fail if it's not).

```
 test database::tests::create_database ... ok
test database::tests::create_database_failure ... ok
test database::tests::create_database_no_compression ... ok
test database::tests::test_database_max_size ... ok
error: test failed, to rerun pass `--lib`

Caused by:
  process didn't exit successfully: `C:\actions-runner\3\_work\kuzu\kuzu/rs\debug\deps\kuzu-da29f6567caf136d.exe --test-threads=1` (exit code: 3)
test database::tests::test_database_read_only ... 
Error: Process completed with exit code 3.
```

The log says it passed `test database::tests::create_database_failure ... ok`; I think it must be test_database_read_only which is failing.

ray6080: yeah. seems to be `test_database_read_only` failing.

ray6080:
Close as not relevant.

PR #3283, opened 2024-04-15
Handle exceptions when flushing WAL

benjaminwinger:
Unhandled exceptions in the Database destructor can't be easily handled in the rust API, and even in the C++ API it may cause terminate to be called, which is not very user-friendly and will bypass other error reporting mechanisms.

I'm not entirely sure why this is only occurring in the one rust test on windows.
Perhaps `fsync` doesn't produce an error if run on a file opened read-only, while `FlushFileBuffers` does, explaining why it only happens on Windows, but there is a similar test for the C API: `CAPIDatabaseTest.CreationReadOnly`, which doesn't have the same problem. But at the least, this will fix CI and make the issue easier to debug and we can open an issue for the rest.

I also removed some unnecessary code from a different rust test which had been producing warnings.


ray6080 (approved):
Thanks! @benjaminwinger

PR #3287, opened 2024-04-16
String utf8 test

hououou:


src/include/function/string/functions/ltrim_function.h
```diff
@@ -21,7 +21,7 @@ struct Ltrim {
                 break;
             }
         }
-        memcpy(data, data + counter, len - counter);
+        memmove(data, data + counter, len - counter);
```
andyfengHKU, line 24: Comment why using memmove here. Memmove is generally slower than memcpy.

test/test_runner/test_parser.cpp
```diff
@@ -422,8 +422,20 @@ void TestParser::openFile() {
     fileStream.open(path);
 }
 
+std::vector<std::string> TestParser::splitString() {
```
andyfengHKU, line 425: Change this function name because it's ambiguous. Also it's rather weird seeing a splitString function not taking any input parameter. 

By design, these type of functions should be a static function that not being exposed in header file.

andyfengHKU (approved):

PR #3296, opened 2024-04-16
Add Missing get_as_arrow and get_as_df Types

mxwli:
Adds the following types to arrow export: `RECURSIVE_REL`, `SERIAL`, `BLOB`

`RDF_VARIANT` is not added since there is no close equivalent in arrow.

Adds 'SERIAL' to DataFrame export.

Makes progress on #3264 

The way we resolve buffer capacities has been rewritten. Now there are dedicated functions to resize buffers for different supported columnar formats. Previously we were prone to allocating incorrect buffer lengths, and the way child buffers of variable length list arrays were resized was prone to causing segfaults.


andyfengHKU (approved):
LGTM

PR #3297, opened 2024-04-16
Fix cleaning up the test directory on windows CI

benjaminwinger:
The windows and macos CI run multiple tests in different build directories as the same user. Since #3290 this means that the tests use the same temporary directory, and the cleanup_test may remove tests from other runs. This makes the test suite use the [RUNNER_TEMP](https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables) environment variable for the test directory if available.

### DIFF
--- a/.github/workflows/ci-workflow.yml
+++ b/.github/workflows/ci-workflow.yml
@@ -370,7 +370,7 @@ jobs:
 
     steps:
       - uses: actions/checkout@v3
-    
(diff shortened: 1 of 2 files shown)


benjaminwinger:
I've updated this to do the same for the macos runner.

andyfengHKU (approved):

PR #3302, opened 2024-04-17
CALL to be readonly=true

OTooleMichael:
Call which is used to set options is only possible in Read/Write connections and does not persist for the Db.

The means that settings cannot be set by a readonly connection. It seems like the majority of the time this doesn't make sense.

eg. in readonly mode we cannot use the followings

1. CALL enable_semi_mask=false;
2. CALL recursive_pattern_semantic='trail';
3. Any of the S3 settings (can't read from S3);
4. THREADS
5. TIMEOUT
6. VAR_LENGTH_EXTEND_MAX_DEPTH

All of these currently seem safe for read instances, and most seem completely for read operations.

I read and agree to the terms under [CLA.md](https://github.com/kuzudb/kuzu/blob/master/CLA.md)

### DIFF
--- a/src/include/parser/visitor/statement_read_write_analyzer.h
+++ b/src/include/parser/visitor/statement_read_write_analyzer.h
@@ -16,7 +16,7 @@ private:
     inline void visitDropTable(const Statement& /*statement*/) override { readOnly = false; }
     inline void visitAlter(const Statement& /*statement*/) override { readOnly = false; }
     inline void visitCopyFrom(const Statement& /*statement*/) override { readOnly = false; }
-    inline void visitStandaloneCall(const Statement& /*statement*/) override { readOnly = false; }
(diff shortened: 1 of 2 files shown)


OTooleMichael:
@andyfengHKU  Is it not the whole test block that needs to go. Seems like the "READ ONLY TRANSACTION" is only set up in order to force this error to occur?

andyfengHKU:
> @andyfengHKU Is it not the whole test block that needs to go. Seems like the "READ ONLY TRANSACTION" is only set up in order to force this error to occur?

You are right. The entire test case should be removed.

andyfengHKU (approved):
LGTM. Also can u add the following to the PR description.

I read and agree to the terms under [CLA.md](https://github.com/kuzudb/kuzu/blob/master/CLA.md)

Issue #3292, opened 2024-04-16
Lock file does not prevent opening two databases from the same process


mewim:
For example:
```python
db_path = tmp_path / "test_database_close.kuzu"
db1 = kuzu.Database(database_path = db_path, read_only = False)
assert not db1.is_closed
assert db1._database is not None

db2 = kuzu.Database(database_path = db_path, read_only = False)
assert not db2.is_closed
assert db2._database is not None
```

this does not throw any error and created two instances of read-write databases in the same Python process. This is likely due to that only one [`F_WRLCK` lock](https://man7.org/linux/man-pages/man2/fcntl.2.html) is held per process:
> A single process can hold only one type of lock on a file region; if a new lock is applied to an already-locked region, then the existing lock is converted to the new lock type.  (Such conversions may involve splitting, shrinking, or coalescing with an existing lock if the byte range specified by the new lock does not precisely coincide with the range of the existing lock.)

Such behavior may not be desirable. We should decide whether we need to do additional work to prevent this and make this behavior clear to the user in our docs.

PR #3306, opened 2024-04-17
Fix lockInfo for access mode in the same process

hououou:
see #3292 

`fcntl` can not get the existing lock type in the same process. To solve it, I give the following solution. But I do not think it is a good idea.

Overall, we write the `pid` and `lockType` for each database object in an additional `lock.info` file. Every time when we open a database, either in the same process or not, we first require `read lock` for the `lock.info` file, checking if there exists the same process and it does not have conflicts with the existing lock type. Then, after opening the database successfully, we require the `write lock` for the `lock.info` file and then write the corresponding `pid` and `lockType` . Once we close the database, we require the   `write lock` for the `lock.info` file and then delete the corresponding `pid` and `lockType` . 

issue: do not know the best time to delete the information of lock info. Considering the following scenario. After doing `db1.reset()`, we will clear all corresponding `pid` and `lockType` in the `lock.info` file.

> SystemConfig systemConfig;
> systemConfig.readOnly=true;
> auto db1 = std::make_unique<Database>("test", systemConfig);
> auto db2 = std::make_unique<Database>("test", systemConfig);
> auto db3 = std::make_unique<Database>("test", systemConfig);
> db1.reset();
> systemConfig.readOnly=false;
> auto db4 = std::make_unique<Database>("test", systemConfig);

### DIFF
--- a/src/common/file_system/virtual_file_system.cpp
+++ b/src/common/file_system/virtual_file_system.cpp
@@ -1,8 +1,15 @@
 #include "common/file_system/virtual_file_system.h"
 
 #include "common/assert.h"
+#include "common/exception/io.h"
 #include "common/file_system/local_file_system.h"
 #include "main/client_context.h"
+
+#if defined(_WIN32)
+#include <windows.h>
+#else
+#include <unistd.h>
+#endif
 
 namespace kuzu {
 namespace common {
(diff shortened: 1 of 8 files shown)


ray6080:
Close this for now as this is not a ideal solution.

PR #3309, opened 2024-04-18
Optimize concurrent query performance

ted-wq-x:
I have read and agree to the terms under [CLA.md](https://github.com/kuzudb/kuzu/blob/master/CLA.md)

In my test environment , sf100 database ,`match (m:Person)-[r:knows]-(n:Person) where m.__uid="${uid}" optional match (d:Person)-[f:knows]->(n)  return size(m.lastName) as n1, size(n.lastName) as n2, f.creationDate as n3 order by n1, n2, n3 limit 10;
` , 50 concurrency ,QPS from 20 to 40

### DIFF
--- a/src/storage/buffer_manager/memory_manager.cpp
+++ b/src/storage/buffer_manager/memory_manager.cpp
@@ -38,14 +38,15 @@ std::unique_ptr<MemoryBuffer> MemoryAllocator::allocateBuffer(bool initializeToZ
             reinterpret_cast<uint8_t*>(buffer), size);
     }
     page_idx_t pageIdx;
-    std::unique_lock<std::mutex> lock(allocatorLock);
-    if (freePages.empty()) {
-        pageIdx = fh->addNewPage();
-    } else {
-        pageIdx = freePages.top();
-        freePages.pop();
+    {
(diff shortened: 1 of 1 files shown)


src/storage/buffer_manager/memory_manager.cpp
```diff
@@ -39,12 +38,14 @@ std::unique_ptr<MemoryBuffer> MemoryAllocator::allocateBuffer(bool initializeToZ
             reinterpret_cast<uint8_t*>(buffer), size);
     }
     page_idx_t pageIdx;
+    std::unique_lock<std::mutex> lock(allocatorLock);
```
ray6080, line 41: Can u rewrite this as a scoped lock?

```
    page_idx_t pageIdx;
    {
        std::unique_lock<std::mutex> lock(allocatorLock);
        if (freePages.empty()) {
            pageIdx = fh->addNewPage();
        } else {
            pageIdx = freePages.top();
            freePages.pop();
        }
    }
```

ray6080 (approved):
Nice catch! Thanks!

src/storage/buffer_manager/memory_manager.cpp
ted-wq-x: Done

Issue #2647, opened 2024-01-10
Metadata isn't always accurate when scanning


benjaminwinger:
I noticed this when adding an assertion in `string_column.cpp` checking that scans don't go past the number of values as stored in the metadata, which caused `EmptyCreateNodeTest.CreateStructWithWriteTransaction` to fail. The issue isn't causing test failures by itself as the data is intact, and in this case the missing metadata isn't a problem (evidently the metadata for the string data column is correct since the number of values is used to determine the end of last string).

```diff
diff --git a/src/storage/store/string_column.cpp b/src/storage/store/string_column.cpp
index 1c4815113..3da1409de 100644
--- a/src/storage/store/string_column.cpp
+++ b/src/storage/store/string_column.cpp
@@ -28,6 +28,7 @@ StringColumn::StringColumn(std::string name, std::unique_ptr<LogicalType> dataTy
 
 void StringColumn::scanOffsets(Transaction* transaction, const ReadState& state,
     string_offset_t* offsets, uint64_t index, uint64_t numValues, uint64_t dataSize) {
+    KU_ASSERT(index + numValues <= state.metadata.numValues);
     // We either need to read the next value, or store the maximum string offset at the end.
     // Otherwise we won't know what the length of the last string is.
     if (index + numValues < state.metadata.numValues) {
@@ -215,6 +216,7 @@ void StringColumn::scanFiltered(transaction::Transaction* transaction,
         auto pos = nodeIDVector->state->selVector->selectedPositions[i];
         if (!resultVector->isNull(pos)) {
             // TODO(bmwinger): optimize index scans by grouping them when adjacent
+            KU_ASSERT(pos < indexState.metadata.numValues);
             auto offsetInGroup = startOffsetInGroup + pos;
             string_index_t index;
             Column::scan(
```

Additionally, more failures occur when adding bounds checks to the primitive Column code. I haven't looked at those failures as closely, but as one example: `DeleteRelTest.deleteRelsFromOneToOneTableRollbackNormalExecution` is failing when querying after rolling back the previous transaction and beginning a new one (but in this case no changes have been made in the new transaction):
```diff
diff --git a/src/storage/store/column.cpp b/src/storage/store/column.cpp
index 9be02ceed..c8de29315 100644
--- a/src/storage/store/column.cpp
+++ b/src/storage/store/column.cpp
@@ -460,6 +460,7 @@ void Column::scanInternal(
[...]

PR #3313, opened 2024-04-18
Node table read state

### LINKED ISSUE #2890: Creating 2 nodes and an edge between them in a single transaction fails.
Here is how to reproduce:

DDL:

```
create node table SpotTableNode(model_key STRING, technical_name STRING, name STRING, row_count INT64, PRIMARY KEY (model_key))

create node table SpotColumnNode(model_key STRING, technical_name STRING, name STRING, p_unique DOUBLE, p_distinct DOUBLE, p_missing DOUBLE, common_values STRING[], PRIMARY KEY (model_key))

CREATE REL TABLE SpotTableToCol(FROM SpotTableNode TO SpotColumnNode)
```

snippet:

```
BEGIN TRANSACTION
CREATE(t:SpotTableNode { model_key : 'tableSchema::tpcds_1.tpcds.store_returns',technical_name : 'tpcds.store_returns',name : 'Store Returns',row_count : 288279 })

CREATE(t:SpotColumnNode { model_key : 'columnSchema::tpcds_1.tpcds.store_returns.sr_return_time_sk',technical_name : 'sr_return_time_sk',name : 'Return Time',p_unique : 0.055061,p_distinct : 0.318826,p_missing : 0.000000,common_values : ['33413', '31810', '54900', '54638', '45302'] })


MATCH (l:SpotTableNode), (r:SpotColumnNode)
              WHERE l.model_key = 'tableSchema::tpcds_1.tpcds.store_returns' and
                    r.model_key = 'columnSchema::tpcds_1.tpcds.store_returns.sr_return_time_sk'
CREATE (l)-[:SpotTableToCol {  }]->(r)
COMMIT
```

With the  `BEGIN TRANSACTION` and` COMMIT` in place
The relation/edge create fails .  Error is  `RuntimeError: Runtime exception: idx: 0 of the DiskArray to be accessed is >= numElements in DiskArray0. `

If I remove the transaction commands(`BEGIN TRANSACTION`, `COMMIT`)  the snippet works

ray6080:
Introduce `NodeTableReadState`, which is always initialized before `read`.
The state caches column read states of the same node group, and local node group to read.

Fix #2890 and #2647.

#2257 is still not fixed. will leave that for a later PR that to correctly figure out nodeOffset (numValues) to read from persistent storage.

### DIFF
--- a/src/include/processor/operator/scan/scan_node_table.h
+++ b/src/include/processor/operator/scan/scan_node_table.h
@@ -34,7 +34,7 @@ public:
         ExecutionContext* executionContext) final {
         ScanTable::initLocalStateInternal(resultSet, executionContext);
         readState =
-            std::make_unique<storage::TableReadState>(*inVector, info->columnIDs, outVectors);
+            std::make_unique<storage::NodeTableReadState>(*inVector, info->columnIDs, outVectors);
     }
 
     bool getNextTuplesInternal(ExecutionContext* context) override;
--- a/src/include/storage/buffer_manager/buffer_manager.h
+++ b/src/include/storage/buffer_manager/buffer_manager.h
@@ -1,6 +1,5 @@
 #pragma once
 
-#include <functional>
 #include <vector>
 
 #include "storage/buffer_manager/bm_file_handle.h"
@@ -212,7 +211,7 @@ private:
 
     inline uint64_t reserveUsedMemory(uint64_t size) { return usedMemory.fetch_add(size); }
     inline uint64_t freeUsedMemory(uint64_t size) {
-        //        KU_ASSERT(usedMemory.load() >= size);
+        KU_ASSERT(usedMemory.load() >= size);
         return usedMemory.fetch_sub(size);
     }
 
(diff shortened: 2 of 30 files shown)


benjaminwinger (approved):
What's the life-cycle of the read state like? It seems to me like `initializeReadState`/`initReadState`, as well as initializing it the first time and re-initializing it for reads on different tables with a different layout, is also updating it for the next read, which re-uses the same object when possible.

That is, I see it being re-used for multiple tables in `RelTableCollectionScanner` and `ScanMultiNodeTables`, but in `ScanSingleNodeTable` it looks to me like it's being called with more or less the same inputs, but is the idea that the `inVector` is changing between calls of `getNextTuplesInternal` and that information is being updated by the call to `initializeReadState`?

That seems like a lot of functionality that isn't clearly explained by the name `initializeReadState`. Some comments explaining what the function does in the header declaring it might be helpful, and maybe also where it's being used.

ray6080:
The read state is owned by scan node table operators, thus the life cycle is managed by the operator. Each state tries to re-use same state for the same node group.

Each table keeps its own read state inside `ScanMultiNodeTables`. `ScanMultiNodeTables` will switch between different states dynamically based on `inVector`.

> Some comments explaining what the function does in the header declaring it might be helpful, and maybe also where it's being used.

Yep. will add more detailed comments altogether in the next PR on rel table read state.

PR #3323, opened 2024-04-19
Fix multi query with filter result error

ted-wq-x:
query like `match (a:Person) match (b:Person) where a.firstName=b.firstName and a.lastName="Chen" return a limit 1;` result error, filter `a.lastName='Chen' ` is not effective

### DIFF
--- a/src/include/optimizer/filter_push_down_optimizer.h
+++ b/src/include/optimizer/filter_push_down_optimizer.h
@@ -9,6 +9,11 @@ class FilterPushDownOptimizer {
 public:
     FilterPushDownOptimizer() { predicateSet = std::make_unique<PredicateSet>(); }
 
+    explicit FilterPushDownOptimizer(binder::expression_vector equalityPredicates,
+        binder::expression_vector nonEqualityPredicates) {
+        predicateSet = std::make_unique<PredicateSet>(equalityPredicates, nonEqualityPredicates);
+    }
+
     void rewrite(planner::LogicalPlan* plan);
 
 private:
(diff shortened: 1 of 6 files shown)


andyfengHKU (approved):
Hi @ted-wq-x, thanks for the fix. I made a few changes to simplify the code a bit. Once CI passes we are good to go.

ted-wq-x:
> Hi @ted-wq-x, thanks for the fix. I made a few changes to simplify the code a bit. Once CI passes we are good to go.

my fault, ut CrossJoinWithFilterPushDownTest query should be `MATCH (a:person) MATCH (b:person) where a.fName=b.fName and a.fName is null RETURN a.gender` ,in this case ,filter should be pushed down

andyfengHKU:
> > Hi @ted-wq-x, thanks for the fix. I made a few changes to simplify the code a bit. Once CI passes we are good to go.
> 
> my fault, ut CrossJoinWithFilterPushDownTest query should be `MATCH (a:person) MATCH (b:person) where a.fName=b.fName and a.fName is null RETURN a.gender` ,in this case ,filter should be pushed down

Whoops I see. I'll do with a separate PR.

andyfengHKU:
> > > Hi @ted-wq-x, thanks for the fix. I made a few changes to simplify the code a bit. Once CI passes we are good to go.
> > 
> > 
> > my fault, ut CrossJoinWithFilterPushDownTest query should be `MATCH (a:person) MATCH (b:person) where a.fName=b.fName and a.fName is null RETURN a.gender` ,in this case ,filter should be pushed down
> 
> Whoops I see. I'll do with a separate PR.

Fix should be available in #3336

PR #3325, opened 2024-04-19
Optimize Hash Index slot splitting

benjaminwinger:
As referenced in #3298 , the `rehashSlots` method (now merged with `splitSlot` into `splitSlots`) had not been optimized for bulk splitting, which was bottlenecking the performance of multiple copies. This removes that bottleneck, improving the performance of the second copy to ~2.8x the cost of the first (when they are the same size that is).

It was necessary to remove the locking on the disk array `WriteIterator` so that multiple can be used on different pages at the same time (the original slot and new slot in this case). Overflow slots are more complicated, and since we can generally assume there are relatively few of them I made it cache new overflow slots in memory and append them at the end so that it's not necessary to juggle two different slot pointers and iterators which can't overlap at the risk of invalidating the other slot pointer and needing to track which iterator is used for which slot at any given time.
To avoid a similar issue with primary slots, I made the minimum number of primary slots allocated the first time be 16 (one full page of slots) so that the original slot and the new slot are never on the same page when splitting.

### DIFF
--- a/src/storage/index/in_mem_hash_index.cpp
+++ b/src/storage/index/in_mem_hash_index.cpp
@@ -29,7 +29,9 @@ InMemHashIndex<T>::InMemHashIndex(OverflowFileHandle* overflowFileHandle)
       pSlots{std::make_unique<InMemDiskArrayBuilder<Slot<T>>>(dummy, 0, 0, true)},
       oSlots{std::make_unique<InMemDiskArrayBuilder<Slot<T>>>(dummy, 0, 1, true)},
       indexHeader{TypeUtils::getPhysicalTypeIDForType<T>()} {
-    allocatePSlots(1u << this->indexHeader.currentLevel);
+    // Match HashIndex in allocating at least one page of slots so that we don't split within the
+    // same page
+    allocateSlots(BufferPoolConstants::PAGE_4KB_SIZE / pSlots->getAlignedElementSize());
 }
 
 template<typename T>
--- a/src/storage/storage_structure/disk_array.cpp
+++ b/src/storage/storage_structure/disk_array.cpp
@@ -22,7 +22,9 @@ DiskArrayHeader::DiskArrayHeader(uint64_t elementSize)
     : alignedElementSizeLog2{(uint64_t)ceil(log2(elementSize))},
       numElementsPerPageLog2{BufferPoolConstants::PAGE_4KB_SIZE_LOG2 - alignedElementSizeLog2},
       elementPageOffsetMask{BitmaskUtils::all1sMaskForLeastSignificantBits(numElementsPerPageLog2)},
-      firstPIPPageIdx{DBFileUtils::NULL_PAGE_IDX}, numElements{0}, numAPs{0} {}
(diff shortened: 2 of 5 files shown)


src/storage/index/in_mem_hash_index.cpp
```diff
@@ -29,7 +29,9 @@ InMemHashIndex<T>::InMemHashIndex(OverflowFileHandle* overflowFileHandle)
       pSlots{std::make_unique<InMemDiskArrayBuilder<Slot<T>>>(dummy, 0, 0, true)},
       oSlots{std::make_unique<InMemDiskArrayBuilder<Slot<T>>>(dummy, 0, 1, true)},
       indexHeader{TypeUtils::getPhysicalTypeIDForType<T>()} {
-    allocatePSlots(1u << this->indexHeader.currentLevel);
+    // Match HashIndex in allocating at least one page of slots so that we don't split within the
+    // same page
+    allocateSlots(BufferPoolConstants::PAGE_4KB_SIZE / sizeof(Slot<T>));
```
ray6080, line 34: Should we assert that `sizeof(Slot<T>)` is power of 2 and less than `PAGE_4KB_SIZE`?

src/storage/storage_structure/disk_array.cpp
```diff
@@ -165,6 +165,7 @@ uint64_t BaseDiskArrayInternal::pushBack(std::span<uint8_t> val) {
 }
 
 uint64_t BaseDiskArrayInternal::resize(uint64_t newNumElements, std::span<uint8_t> defaultVal) {
```
ray6080, line 167: Do we also need to lock in `BaseDiskArrayInternal::pushBack(std::span<uint8_t> val)`? (L160)

ray6080 (approved):

src/storage/index/in_mem_hash_index.cpp
benjaminwinger: We should assert that it's less than 4KB, but that's something the disk array should assert for any type stored in it.

But some slots are actually not powers of two, e.g. `Slot<uint8_t>` is actually only 212 bytes (limited to 20 entries by the fingerprint capacity), `Slot<uint16_t>` is 232, `Slot<uint32_t>` and `Slot<ku_string_t>` are 248, and then `Slot<uint64_t>` is 256.
I guess that means that this calculation doesn't work for exactly one page of slots as the disk array aligns elements to powers of two, meaning that all of them store 16 slots per page, but this reserves 20 slots for a `HashIndex<uint8_t>`.

src/storage/storage_structure/disk_array.cpp
benjaminwinger: Yes, thanks for catching that.

src/storage/index/in_mem_hash_index.cpp
benjaminwinger: I've updated it to use the size rounded up to the nearest power of two from the disk array, and added the maximum size assertion to the disk array.

PR #3336, opened 2024-04-22
Optimize filter push down

andyfengHKU:
Fix a push down case as mentioned in #3323


acquamarin (approved):

ted-wq-x:
maybe need some uts for testing query like filters with implicit type conversion can also be pushed down

PR #3344, opened 2024-04-22
Fix #3089

### LINKED ISSUE #3089: MAP = MAP comparison is not registered, but the comparison can still be executed.
Code: (only works on the `3054-fix` branch right now)
```py
conn.execute("CREATE NODE TABLE tab(id int64, mp MAP(double, int64), mp2 MAP(int64, double), mp3 MAP(string, string), primary key(id))")
conn.execute(
        "MERGE (t:tab {id: 0, mp: $1, mp2: $2, mp3: $3}) RETURN t.*",
        {"1": {1.0: 5, 2: 3, 2.2: -1},
         "2": {5: -0.5, 4: -0.0, 0: 2.2},
         "3": {'a': '1', 'b': '2', 'c': '3'}})
conn.execute(
        "MATCH (t:tab) WHERE t.mp = map([1.0, 2.0, 2.2], [5, 3, -1]) RETURN t.*")
```
The comparison works, but MAP = MAP is not registered as a valid comparison. Interestingly, if we swap out the second map for a parameter in the last query
```py
result = conn.execute(
        "MATCH (t:tab) WHERE t.mp = $1 RETURN t.*",
        {"1": {1.0: 5, 2: 3, 2.2: -1}})
```
The query throws the following exception:
```
RuntimeError: Parameter 1 has data type MAP(DOUBLE: INT64) but expects STRING.
```
indicating that there is likely a conversion to STRING before the map comparison is made.

manh9203:
This PR resolved #3089 by register all valid logical types for comparison.


src/include/function/comparison/vector_comparison_functions.h
```diff
@@ -13,17 +13,10 @@ struct ComparisonFunction {
     template<typename OP>
     static function_set getFunctionSet(const std::string& name) {
         function_set functionSet;
-        for (auto& comparableType : common::LogicalTypeUtils::getAllValidComparableLogicalTypes()) {
+        // We can only check whether two internal ids are equal or not.
+        for (auto& comparableType : common::LogicalTypeUtils::getAllValidLogicTypes()) {
```
andyfengHKU, line 17: If we switch to all logical types, can u test the following

- compare between `RDF_VARAINT`
- compare between `INTERNAL_ID` especially for non equality comparison

src/include/function/comparison/vector_comparison_functions.h
```diff
@@ -13,17 +13,10 @@ struct ComparisonFunction {
     template<typename OP>
     static function_set getFunctionSet(const std::string& name) {
         function_set functionSet;
-        for (auto& comparableType : common::LogicalTypeUtils::getAllValidComparableLogicalTypes()) {
+        // We can only check whether two internal ids are equal or not.
```
andyfengHKU, line 16: This comment is no longer true so remove it.

andyfengHKU (approved):

Issue #3248, opened 2024-04-10
Implicit casting and resolving literal types for ARRAY when doing similarity search


prrao87:
Consider this similarity search case where I want to return the node whose `vector` property is nearest to a given query vector provided by the user. I'll use the cosine similarity to demonstrate this.

```py
import os
import shutil
import kuzu

if os.path.exists("./db"):
    shutil.rmtree("./db")

# Create database
db = kuzu.Database("./db")
conn = kuzu.Connection(db)

# Define schema
conn.execute("CREATE NODE TABLE Item(id UINT64, item STRING, price DOUBLE, vector DOUBLE[2], PRIMARY KEY (id))");

# Add data
conn.execute("MERGE (a:Item {id: 1, item: 'apple', price: 2.0, vector: cast([3.1, 4.1], 'DOUBLE[2]')})")
conn.execute("MERGE (b:Item {id: 2, item: 'banana', price: 1.0, vector: cast([5.9, 26.5], 'DOUBLE[2]')})");

# Run similarity search
res = conn.execute("MATCH (a:Item) RETURN a.item, a.price, array_cosine_similarity(a.vector, cast([6.0, 25.0], 'DOUBLE[2]')) AS sim ORDER BY sim DESC")
while res.has_next():
    row = res.get_next()
    print(row)
```

I want to return banana as the most similar node, based on the provided vector `[6.0, 25.0]`, which should be the closest to the banana vector. I was able to get the above example code to work after some massaging, to return this:

```
['banana', 1.0, 0.9998642653091405]
['apple', 2.0, 0.9163829638139936]
```

### Verbosity and scope for errors

The fixed-list/array type is currently a bit inconvenient and hard to use for similarity search.
It would be a lot easier if we could simply define this instead (without explicitly performing the cast)

```py
# Add data
conn.execute("MERGE (a:Item {id: 1, item: 'apple', price: 2.0, vector: [3.1, 4.1]})")
conn.execute("MERGE (b:Item {id: 2, item: 'banana', price: 1.0, vector: [5.9, 26.5]})")

# Run similarity search
res = conn.execute("MATCH (a:Item) RETURN a.item, a.price, array_cosine_similarity(a.vector, [6.0, 25.0]) AS sim ORDER BY sim DESC")
```

Can this be incorporated without any breaking changes to other functionality?

mxwli:
[...]

PR #3375, opened 2024-04-24
Add Implicit Casting from List to Array

mxwli:
Progress on #3248 

We've decided not to resolve list literals to arrays, since it's much simpler to just support list -> array implicit casting. As a consequence, implicit casts are now no longer guaranteed to always work.

The mentioned `array_cosine_similarity(a.vector, [6.0, 25.0])` call still doesn't work due to a binding error. The corresponding tests are disabled until this is fixed.


andyfengHKU (approved):
LGTM

PR #3378, opened 2024-04-24
Optimize InMemoryHashIndex lookups

benjaminwinger:
In lookup and appendInternal, it's not actually necessary to check the validity mask for anything other than determining the number of valid entries.
I think this makes a slight difference, but the variance between test runs is large enough in comparison to any performance improvement that it's very hard to tell.

### DIFF
--- a/src/storage/index/in_mem_hash_index.cpp
+++ b/src/storage/index/in_mem_hash_index.cpp
@@ -150,28 +150,43 @@ size_t InMemHashIndex<T>::append(const IndexBuffer<BufferKeyType>& buffer) {
 }
 
 template<typename T>
+entry_pos_t InMemHashIndex<T>::findEntry(SlotIterator& iter, Key key, uint8_t fingerprint) {
+    do {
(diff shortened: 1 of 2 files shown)


src/storage/index/in_mem_hash_index.cpp
```diff
@@ -155,21 +155,22 @@ bool InMemHashIndex<T>::appendInternal(Key key, common::offset_t value, common::
     auto slotID = HashIndexUtils::getPrimarySlotIdForHash(this->indexHeader, hash);
     SlotIterator iter(slotID, this);
     do {
```
ray6080, line 157: By looking at `appendInternal` and `lookup` now, I feel you should be able to refactor the `do while` loop into a single function that can be shared inside both `appendInternal` and `lookup`?

ray6080 (approved):

src/storage/index/in_mem_hash_index.cpp
benjaminwinger: Yes, that's probably a good idea. I've done so.

Issue #3196, opened 2024-04-02
[CI] Use Windows Docker to host self-hosted runners


mewim:

mewim:
Resolved in https://github.com/kuzudb/ci-deployment-compose/pull/9

PR #3389, opened 2024-04-26
Fix USERPROFILE access conflicts in Windows CI

benjaminwinger:
Some tests have been failing when multiple jobs are trying to read/write to the same extension file. E.g. https://github.com/kuzudb/kuzu/actions/runs/8853455160/job/24314357568?pr=3387

The path of that file is derived from `ClientConfig.homeDirectory`, which is set using the `USERPROFILE` environment variable on Windows. This overrides that in CI to refer to a directory unique to the CI job.


benjaminwinger:
Closing for now. This hasn't actually been showing up recently.

benjaminwinger:
Still an issue: https://github.com/kuzudb/kuzu/actions/runs/9522563438/job/26252460308?pr=3642

mewim:
If we do #3196, each CI runner will be isolated, so it should no longer be a problem.

benjaminwinger:
> If we do https://github.com/kuzudb/kuzu/issues/3196, each CI runner will be isolated, so it should no longer be a problem.

That's true. I guess since this hasn't been a huge issue I'll just close this again in favour of that solution.

PR #3404, opened 2024-04-29
Fix issue 3166

andyfengHKU:
I need to check this against CI first.


src/include/storage/store/rel_table.h
```diff
@@ -1,18 +1,25 @@
...
 #include "storage/store/table.h"
 
 namespace kuzu {
 namespace storage {
 
 struct RelTableReadState : public TableReadState {
+    // TODO(Xiyang): rename me
     common::RelDataDirection direction;
 
+    // This direction is not
```
ray6080, line 15: Complete the sentence here.

src/function/struct/struct_pack_function.cpp
```diff
@@ -65,6 +76,36 @@ void StructPackFunctions::execFunc(const std::vector<std::shared_ptr<ValueVector
     }
 }
 
+void StructPackFunctions::undirectedRelPackExecFunc(
+    const std::vector<std::shared_ptr<ValueVector>>& parameters, ValueVector& result, void*) {
+    KU_ASSERT(parameters.size() > 1);
+    // Force copy of
```
ray6080, line 82: Complete the sentence.

src/function/struct/struct_pack_function.cpp
```diff
@@ -33,6 +33,17 @@ void StructPackFunctions::compileFunc(FunctionBindData* /*bindData*/,
     }
 }
```
ray6080, line 35: Better clearly state the implicit assumption of `src` being 0 and `dst` being 1 here.

src/include/expression_evaluator/node_rel_evaluator.h
```diff
@@ -4,42 +4,54 @@
...
 
 namespace kuzu {
-namespace main {
-class ClientContext;
-}
-
 namespace evaluator {
 
-class NodeRelExpressionEvaluator final : public ExpressionEvaluator {
+class PatternExpressionEvaluator : public ExpressionEvaluator {
```
ray6080, line 9: Can u keep `final`?

src/include/expression_evaluator/node_rel_evaluator.h
ray6080, line 1: Let's also rename the file to `pattern_evaluator.h`.

ray6080 (approved):
I don't think it's a good design to fill the flip vector inside rel table scan. the reason is simply because that info is passed down from operator, while not relying on anything inside storage. Better move the filling of flip vector into operators.

src/function/struct/struct_pack_function.cpp
andyfengHKU: I'll try to do this in a more principled way.

PR #3417, opened 2024-04-30
Support asserting RETURN result column names in testing framework

yiyun-sj:
**Column Name Assertion**
- added new optional property to check for column names `-CHECK_COLUMN_NAMES`
- column names form a tuple in the first row of tuples (need to +1 tuple count)
- the same behaviour applies to csv answers and hash compare (column names treated as a tuple)
- if checking order, column names always first in result order

**Example**
```
-CHECK_COLUMN_NAMES
-STATEMENT MATCH (a:User) RETURN a.name, COUNT { MATCH (a)<-[:Follows]-(b:User) } AS num_follower ORDER BY num_follower;
---- 5
a.name|num_follower
Adam|0
Karissa|1
Noura|1
Zhang|2
```


ray6080 (commented):
LGTM! Can u add a test case for this feature?

yiyun-sj:
> LGTM! Can u add a test case for this feature?

Added a test case for basic usage. Let me know if I should add tests for CSV answers and hashing as well.

test/test_files/common/columnname.test
```diff
@@ -0,0 +1,25 @@
...
+-CASE ColumnNameOrdered
+-CHECK_ORDER
+-CHECK_COLUMN_NAMES
+-STATEMENT MATCH (u:User) RETURN u.name, u.age ORDER BY u.age;
+---- 5
+u.name|u.age
+Noura|25
+Adam|30
+Karissa|40
+Zhang|50
```
ray6080, line 25: Add a new line here.

ray6080 (approved):

PR #3418, opened 2024-04-30
Infer test group name directly from test file path

yiyun-sj:
**Test Group Name Generation**
- uses the relative file path as test group name
- delimiter choice is '~' ('.' is reserved by gtest, '/' is reserved by file path)
- current `-GROUP` property is not removed for backward compatibility (is this necessary?)

**Note:** some characters such as : and - result in tests not being actually run (some current test cases are like this)
image

**Example**
the file `test/test_files/tck/match/match8` now has test group name `tck~match~match8`

**Todo**
delete `-GROUP` from all existing test files (do we want to do this?)


test/test_runner/test_parser.cpp
```diff
@@ -33,6 +34,17 @@ std::unique_ptr<TestGroup> TestParser::parseTestFile() {
     return std::move(testGroup);
 }
 
+void TestParser::genGroupName() {
+    std::size_t subStart =
+        TestHelper::appendKuzuRootPath(std::string(TestHelper::E2E_TEST_FILES_DIRECTORY)).length() +
+        1;
+    std::size_t subEnd = path.find_last_of('.') - 1;
+    std::string relPath = path.substr(subStart, subEnd - subStart + 1);
+    std::replace(relPath.begin(), relPath.end(), '-', '_');
```
ray6080, line 43: I don't think we should replace `-` with `_`. Users of the testing framework should follow the convention. themselves.

ray6080 (approved):
LGTM. Let's bring this to the team for some feedback too. We shall make this as the new convention.

test/test_runner/test_parser.cpp
```diff
@@ -67,8 +79,6 @@ void TestParser::parseHeader() {
             break;
         }
         case TokenType::GROUP: {
```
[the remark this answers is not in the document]
yiyun-sj: I kept it in for the sake of backward compatibility. We can remove it if we want to.

test/test_runner/test_parser.cpp
yiyun-sj: I did this as a workaround as a bunch of files/directories had `-` in the codebase which results in a false positive during the test runs. I could also go and change those files and remove this line.

ray6080: What do u mean "false positive" here?

yiyun-sj: Removed this line and fixed the naming of existing tests

test/test_runner/test_parser.cpp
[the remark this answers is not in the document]
ray6080: I think we can just remove it then.

ray6080 (approved):
Can u also update the docs now regarding to the changes on test group name?

ray6080:
Also, can u revisit the false positive issue? We'd better understand what's acceptable directory and file naming conventions under `test_files`.

PR #3428, opened 2024-05-01
Support initialize test case on existing binary db directory

yiyun-sj:
**Binary db Test Init**
- added new dataset type `-KUZU` for binary db directories
- added an example test case and dataset based on demo_db

**Example**
Available at `dataset/binary-demo` and `test/test_files/binary_demo`
These should be deleted before merge.


ray6080 (commented):
Generally LGTM. Let's fix the failed test first before merging this.

ray6080 (approved):

PR #3432, opened 2024-05-01
Allow fuzzy matching on test result

yiyun-sj:
**Fuzzy (regex) matching on errors**
- changed one old test case (the only one that needed regex)
- changed the existing `---- error(regex)` handling (unused by any test cases) to be recognized as a pattern

**Example**
we should now be able to run the example from issue (ExtensionErrorTest) but the behaviour of that test changed due to a code update.


ray6080 (commented):
Can u take a look at the following two tests and rewrite them with regex pattern too?
```
TinySnbExceptionTest.DeleteNodeWithEdgeErrorTest
TinySnbDeleteNodeTest.DeleteNodeWithConnectedRels
```

yiyun-sj:
> Can u take a look at the following two tests and rewrite them with regex pattern too?
> 
> ```
> TinySnbExceptionTest.DeleteNodeWithEdgeErrorTest
> TinySnbDeleteNodeTest.DeleteNodeWithConnectedRels
> ```

done!

ray6080 (approved):

PR #3437, opened 2024-05-02
Support multiple query statements in e2e test framework

yiyun-sj:
**Testing Multiple Query Statements**
- This implementation supports multiple query statements to multiple results

**Example**

```
// this now works
-STATEMENT INSTALL httpfs; // assumed non-error statement
           LOAD EXTENSION httpfs; // assumed non-error statement
           LOAD FROM 'http://extension.kuzudb.com/dataset/test/city.csv' return *;
---- ok
---- ok
---- 3
Guelph|75000
Kitchener|200000
Waterloo|150000
```

### DIFF
--- a/test/include/test_runner/test_group.h
+++ b/test/include/test_runner/test_group.h
@@ -14,11 +14,11 @@ namespace testing {
 
 enum class ResultType {
     OK,
-    ERROR,
-    ERROR_REGEX,
     HASH,
     TUPLES,
     TUPLES_CSV,
+    ERROR_MSG,
+    ERROR_REGEX,
 };
 
 struct TestQueryResult {
--- a/test/include/test_runner/test_group.h
+++ b/test/include/test_runner/test_group.h
@@ -25,7 +25,7 @@ struct TestQueryResult {
     ResultType type;
     uint64_t numTuples = 0;
     // errorMsg, CSVFile, hashValue uses first element
-    std::vector<std::string> expectedResult; 
+    std::vector<std::string> expectedResult;
 };
 
 struct TestStatement {
--- a/test/include/test_runner/test_group.h
+++ b/test/include/test_runner/test_group.h
@@ -16,16 +16,16 @@ enum class ResultType {
     OK,
     HASH,
     TUPLES,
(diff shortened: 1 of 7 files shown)


test/include/test_runner/test_parser.h
```diff
@@ -94,6 +94,7 @@ class TestParser {
     void genGroupName();
     void parseHeader();
     void parseBody();
+    void extractAllResults(TestStatement* statement);
```
ray6080, line 97: `extractAllResults` -> `extractExpectedResults`.

test/test_runner/test_parser.cpp
```diff
@@ -122,45 +122,63 @@ void TestParser::replaceVariables(std::string& str) {
...
+        }
+        if (currentToken.type != TokenType::RESULT) {
+            setCursorToPreviousLine();
+            return;
+        }
+        extractExpectedResult(statement);
+    } while (nextLine());
+}
+
 void TestParser::extractExpectedResult(TestStatement* statement) {
```
ray6080, line 139:
```
TestQueryResult TestParser:: extractExpectedResultFromToken()
```
I would rename the function to `extractExpectedResultFromToken`, and return `TestQueryResult` without passing in `TestStatement`, which should be handled by `extractAllResults` to push in query result. This looks cleaner to me.

ray6080 (approved):

test/include/test_runner/test_group.h
```diff
@@ -12,28 +12,36 @@
...
+    TUPLES_CSV,
+    ERROR_MSG,
+    ERROR_REGEX,
+};
+
+struct TestQueryResult {
+    ResultType type;
+    std::vector<std::string> expectedTuples;
+    uint64_t numTuples = 0;
+    std::string expectedMessage; // errorMsg || CSVFile || hashValue
```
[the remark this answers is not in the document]
yiyun-sj: Sure, we will just use `expectedResult[0]` for all the single string results then.
(I originally wanted to use union to make things more clear but that has some constructor/destructor issues for compile)

PR #3453, opened 2024-05-06
Allow numeric value comparison with precision

yiyun-sj:
**Precision Numeric Compare**
- for `float` and `double` types, compare (epsilon) precision instead of strcmp
- uses the suggested method from [cppreference](https://en.cppreference.com/w/cpp/types/numeric_limits/epsilon)
- 

**Example**
changed `tinysnb~agg~simple.AggSimple` test case to check precision in two of the subcases.


ray6080 (commented):
~~Actually I think we should just always check precision for floating points, thus saving users the efforts to manually specify `CHECK_PRECISION`, as it makes sense to always check precisions.~~

~~Internally, we need to keep a slow path whenever there are any floating point values in the query result, we split the expected tuples into values for each column, and compare each value one at a time as you currently do when `CHECK_PRECISION` is enabled.~~

ray6080 (approved):
As discussed, there can be edge cases when `CHECK_ORDER` is not specified when comparing with precision. Let's simplify our life first with the limitation.
Also, as discussed, add a check to make sure that we throw errors when `CHECK_ORDER` is not enabled for precision checks.

PR #3454, opened 2024-05-06
Fix CSV file answers tuple count bug

yiyun-sj:
Fixes the bug where tuple count is not verified when using the `<FILE>:` answers format for tuples.

### DIFF
--- a/test/test_runner/test_runner.cpp
+++ b/test/test_runner/test_runner.cpp
@@ -150,6 +150,12 @@ bool TestRunner::checkPlanResult(std::unique_ptr<QueryResult>& result, TestState
         while (std::getline(expectedTuplesFile, line)) {
             testAnswer.expectedResult.push_back(line);
         }
+        if (testAnswer.expectedResult.size() != testAnswer.numTuples) {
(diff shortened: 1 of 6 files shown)


test/test_runner/test_runner.cpp
```diff
@@ -150,6 +150,12 @@ bool TestRunner::checkPlanResult(std::unique_ptr<QueryResult>& result, TestState
         while (std::getline(expectedTuplesFile, line)) {
             testAnswer.expectedResult.push_back(line);
         }
+        if (testAnswer.expectedResult.size() != testAnswer.numTuples) {
+            spdlog::error("PLAN{} NOT PASSED.", planIdx);
+            spdlog::info("PLAN: \n{}", planStr);
+            spdlog::info("TUPLE COUNT NOT MATCHING.");
```
ray6080, line 156: Can u provide a bit more info here? such as expected num of tuples, and actual num of tuples. though we can find these through test file and answer csv files, but better provide some convenience when possible here

ray6080 (approved):

PR #3456, opened 2024-05-07
Backtraces

### LINKED ISSUE #3053: [PROPOSAL] Backtrace
The goal is to provide more debugging info when Kuzu throws exceptions, errors or segfaults.

Backtrace can provide a list of the function calls that are currently active in a thread. Usually we need to use an external debugger such as gdb/lldb to get this kind of information.
However, under some cases, using an external debugger is not trivial. for example, when user reports seg faults, we have to reproduce with a user-provided minimal example; sometimes a bug cannot be stably reproduced locally but on CI pipelines, which makes it harder to debug. In these cases, printing stack traces can be helpful.

Expected behaviour is when `DEBUG_STACKTRACE` is enabled:
1. Print stack traces when Kuzu throws `kuAssertFailureInternal` exceptions.
2. Use signal handling to catch seg faults and print stack traces.

We can start with Linux and MacOS platforms.
GlibC provides a set of functions for backtracing. https://www.gnu.org/software/libc/manual/html_node/Backtraces.html
Also see here for signal handling. https://stackoverflow.com/questions/77005/how-to-automatically-generate-a-stacktrace-when-my-program-crashes

This is an initial draft, more comments are welcome :)

benjaminwinger:
Fixes #3053 ~(currently missing backtrace on segfaults).~ Everything is working except backtraces for segfaults on windows.

Adds backtracing using [cpptrace](https://github.com/jeremy-rifkin/cpptrace), and makes the default configuration for tests `RelWithDebInfo`(otherwise the backtraces aren't very useful).
Backtraces are being printed in the `kuzu::common::Exception` constructor since including them in the message will break the exception message tests. Unfortunately this does mean that the stack trace is printed before the exception, but is printed backwards with the expectation that it will be printed after the exception. It also means that exceptions which are handled and not re-thrown will also have their stack traces printed.

Safe signal handling using a separate executable as described [here](https://github.com/jeremy-rifkin/cpptrace/blob/main/docs/signal-safe-tracing.md) was working, except that when it couldn't find the executable it would hang (in a suspiciously similar way to #3459, where only the scheduling thread can be interrupted in gdb). I thought that it might be more reliable just to use the unsafe `cpptrace::generate_trace().print()`, and deal with the possibility of a CI job hanging on rare occasions (mostly it seems fine, but as far as I understand, major memory issues could cause various problems including hangs if we try and do complex stuff like memory allocation when handling the signal).


ray6080 (approved):
Nice one!

PR #3457, opened 2024-05-07
C Api Enhancements

MSebanc:


src/c_api/prepared_statement.cpp
```diff
@@ -25,138 +25,268 @@ void kuzu_prepared_statement_destroy(kuzu_prepared_statement* prepared_statement
...
+kuzu_state kuzu_prepared_statement_get_error_message(kuzu_prepared_statement* prepared_statement,
+    char** out_error_message) {
+    if (prepared_statement == nullptr || prepared_statement->_prepared_statement == nullptr) {
+        return KuzuError;
+    }
+    try {
+        auto error_message =
+            static_cast<PreparedStatement*>(prepared_statement->_prepared_statement)
+                ->getErrorMessage();
+        if (error_message.empty()) {
```
mewim, line 43: The empty error message should not be treated as an error. I think it makes more sense to return it as-is.

src/c_api/query_result.cpp
```diff
@@ -16,55 +16,69 @@ void kuzu_query_result_destroy(kuzu_query_result* query_result) {
...
+kuzu_state kuzu_query_result_get_error_message(kuzu_query_result* query_result,
+    char** out_error_message) {
+    if (out_error_message == nullptr || query_result == nullptr ||
+        query_result->_query_result == nullptr) {
+        return KuzuError;
+    }
+    try {
+        auto error_message =
+            static_cast<QueryResult*>(query_result->_query_result)->getErrorMessage();
+        if (error_message.empty()) {
```
mewim, line 34: Ditto. Empty error message should not be treated as error.

src/c_api/query_result.cpp
```diff
@@ -75,39 +89,63 @@ bool kuzu_query_result_has_next_query_result(kuzu_query_result* query_result) {
     return static_cast<QueryResult*>(query_result->_query_result)->hasNextQueryResult();
 }
 
-kuzu_query_result* kuzu_query_result_get_next_query_result(kuzu_query_result* query_result) {
+kuzu_state kuzu_query_result_get_next_query_result(kuzu_query_result* query_result,
```
mewim, line 92: It might be helpful to check `hasNextQueryResult` and return an error if no next.

src/c_api/value.cpp
```diff
@@ -182,388 +182,883 @@ kuzu_value* kuzu_value_clone(kuzu_value* value) {
     return c_value;
 }
 
-void kuzu_value_copy(kuzu_value* value, kuzu_value* other) {
+kuzu_state kuzu_value_copy(kuzu_value* value, kuzu_value* other) {
```
mewim, line 185: I would double check if `copyValueFrom` can thorw error? I think this cannot fail? If not, it would be better to make this `void`.

src/c_api/value.cpp
```diff
@@ -182,388 +182,883 @@ kuzu_value* kuzu_value_clone(kuzu_value* value) {
...
+        out_value->_value = val;
+        out_value->_is_owned_by_cpp = true;
+    } catch (Exception& e) {
+        return KuzuError;
+    }
+    return KuzuSuccess;
 }
 
-uint64_t kuzu_value_get_struct_num_fields(kuzu_value* value) {
+kuzu_state kuzu_value_get_struct_num_fields(kuzu_value* value, uint64_t* out_result) {
```
mewim, line 236: It might be helpful to check if the value is truly a `struct`.

src/c_api/value.cpp
mewim, line 1: One note in this file in general is that we should check the type for all `kuzu_value_get_*` functions and return `KuzuError` if the function call does not match the expected type. The templated `getValue<>` functions in C++ has an assertion `KU_ASSERT(dataType->getLogicalTypeID() == *);` However, in release mode, these assertions are ignored. Therefore, wrapping the calls with `try-catch` can still resulting in getting the wrong value. So this should be handled at binding side.

mewim:
I did not check the test cases very carefully. But in general, for functions that can return `KuzuError`, it would be nice to also check the error cases, as long as they are easy to produce.

mewim (changes requested):
Nice PR overall, with a few imprvements needed (see my comments).

mewim (approved):

PR #3472, opened 2024-05-10
Fixed issue with commit error not showing in shell

MSebanc:


tools/shell/embedded_shell.cpp
```diff
@@ -362,21 +362,16 @@ void EmbeddedShell::run() {
...
-                    if (lineStrTrimmed.find_first_of(" \t\n\r\f\v") == std::string::npos &&
-                        lineStrTrimmed.length() > 1) {
-                        printf("Error: \"%s\" is not a valid Cypher query. Did you mean to issue a "
-                               "CLI command, e.g., \"%s\"?\n",
-                            lineStr.c_str(), findClosestCommand(lineStrTrimmed).c_str());
-                    } else {
-                        printf("Error: %s\n", queryResult->getErrorMessage().c_str());
+                    std::string errMsg = queryResult->getErrorMessage();
+                    printf("Error: %s\n", errMsg.c_str());
+                    if (errMsg.find("Parser exception:") == 0) {
```
ray6080, line 367: Better abstract `"Parser exception:"` into a constant in `ParserException`? So when somehow the prefix is changed inside `ParserException`, it will also auto reflect here?

ray6080 (approved):

PR #3473, opened 2024-05-10
Fix string hash and add more hash tests

benjaminwinger:
Should fix kuzudb/explorer#129. Not only was the type of the value being used for the bit shift not wide enough (the characters are `char`s, but the result of a bitshift seems to always be at least `int`), but it was signed, and signed bit shifts may add new bits as 0 or 1 depending on the implementation, which I think is causing the difference among the platforms.

The existing string tests were fairly limited so I added some more which better cover the corner cases.

This will break storage since is changes the behaviour of the hash function used in the hash index.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,6 @@
 cmake_minimum_required(VERSION 3.15)
 
-project(Kuzu VERSION 0.4.1.1 LANGUAGES CXX C)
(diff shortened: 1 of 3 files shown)


ray6080 (approved):

ray6080:
> This will break storage since is changes the behaviour of the hash function used in the hash index.

We should bump cmake version then?

benjaminwinger:
> We should bump cmake version then?

Done

PR #3494, opened 2024-05-14
Enable -Werror for GCC Build & Test Job

mxwli:
This will probably break a lot of PR CIs


src/storage/compression/compression.cpp
```diff
@@ -358,7 +358,7 @@ template<typename T>
 bool IntegerBitpacking<T>::canUpdateInPlace(T value, const BitpackHeader& header) {
     T adjustedValue = value - (T)header.offset;
     // If there are negatives, the effective bit width is smaller
-    auto valueSize = std::bit_width((U)abs<T>(adjustedValue));
+    auto valueSize = (int32_t)std::bit_width((U)abs<T>(adjustedValue));
```
benjaminwinger, line 361: If we're typing it explicitly, it would probably make more sense for it to be a `uint8_t`, since that's the type of what it's being compared to (`header.bitWidth`).
I think it would also be more readable to write `uint8_t valueSize = ...` instead of casting.

PR #3512, opened 2024-05-17
Read primary key for delete

andyfengHKU:
- Avoid performing primary key lookup during node deletion. Instead we scan primary key before deleting.
- Avoid having duplicate node delete and rel delete logical operators. Replaced with a single logical delete.


src/processor/map/map_delete.cpp
```diff
@@ -12,107 +14,98 @@ using namespace kuzu::storage;
...
 
-static std::unique_ptr<RelDeleteExecutor> getRelDeleteExecutor(
-    storage::StorageManager& storageManager, const binder::RelExpression& rel,
-    const Schema& inSchema) {
-    auto srcNodePos = DataPos(inSchema.getExpressionPos(*rel.getSrcNode()->getInternalID()));
-    auto dstNodePos = DataPos(inSchema.getExpressionPos(*rel.getDstNode()->getInternalID()));
-    auto relIDPos = DataPos(inSchema.getExpressionPos(*rel.getInternalIDProperty()));
+std::unique_ptr<RelDeleteExecutor> PlanMapper::getRelDeleteExecutor(const BoundDeleteInfo& info,
+    const Schema& schema) const {
+    auto sm = clientContext->getStorageManager();
```
ray6080, line 82: hmm is abbr. necessary for `storageManager`? this doesn't like a common abbr to me. I'd rather stay with `storageManager`.

ray6080 (approved):

PR #3521, opened 2024-05-21
Preliminary Decimal Datatype

mxwli:
Just adds decimal storage & arrow import/export. ~90% of the lines changed are in cypher_lexer.cpp, which is why this PR modifies so much code.

Everything else is nonfunctional, but there exists a lot of partially implemented features.


tools/python_api/test/test_scan_pandas_pyarrow.py
```diff
@@ -632,3 +633,23 @@ def test_pyarrow_map_offset(conn_db_readonly: ConnDB) -> None:
         idx += 1
 
     assert idx == 48
+
+def test_pyarrow_decimal(conn_db_readwrite: ConnDB) -> None:
```
andyfengHKU, line 637: Can u test the following cases

- scan from pandas decimal
- export decimal to python object, pandas
- in cpp side, create a table with decimal column. Insert decimal into it and scan out of the table.

Open the issue about supporting decimal in Java, NodeJS, Rust

src/common/types/types.cpp
```diff
@@ -391,6 +443,9 @@ static std::string getIncompleteTypeErrMsg(LogicalTypeID id) {
 }
 
 LogicalType::LogicalType(LogicalTypeID typeID) : typeID{typeID}, extraTypeInfo{nullptr} {
+    if (typeID == LogicalTypeID::DECIMAL) {
```
andyfengHKU, line 446: move to the switch case below. Maybe we should switch LogicalTypeID now.

src/common/types/types.cpp
```diff
@@ -416,7 +471,12 @@ LogicalType::LogicalType(LogicalTypeID typeID) : typeID{typeID}, extraTypeInfo{n
 
 LogicalType::LogicalType(LogicalTypeID typeID, std::unique_ptr<ExtraTypeInfo> extraTypeInfo)
     : typeID{typeID}, extraTypeInfo{std::move(extraTypeInfo)} {
-    physicalType = getPhysicalType(typeID);
+    if (typeID == LogicalTypeID::DECIMAL) {
+        // special case, since decimal physical type is determined by its extratypeinfo
+        physicalType = DecimalType::getPhysicalTypeID(this->extraTypeInfo);
```
andyfengHKU, line 476: Then we should change the getPhysicalType interface to also take extraTypeInfo

src/common/types/types.cpp
```diff
@@ -555,12 +622,17 @@ LogicalType LogicalType::fromString(const std::string& str) {
...
+               upperDataTypeString.starts_with("NUMERIC")) {
+        dataType = *parseDecimalType(trimmedStr);
     } else if (upperDataTypeString == "RDF_VARIANT") {
         dataType = *LogicalType::RDF_VARIANT();
     } else {
         dataType.typeID = strToLogicalTypeID(upperDataTypeString);
     }
-    dataType.physicalType = LogicalType::getPhysicalType(dataType.typeID);
+    if (dataType.typeID != LogicalTypeID::DECIMAL) {
+        dataType.physicalType = LogicalType::getPhysicalType(dataType.typeID);
```
andyfengHKU, line 634: ditto just change getPhysicalType interface

src/common/types/types.cpp
```diff
@@ -572,7 +644,11 @@ void LogicalType::serialize(Serializer& serializer) const {
     case PhysicalTypeID::ARRAY:
     case PhysicalTypeID::STRUCT:
         extraTypeInfo->serialize(serializer);
+        break;
     default:
+        if (typeID == LogicalTypeID::DECIMAL) {
```
andyfengHKU, line 649: I wonder if we can change this to
```
if extraTypeInfo != nullptr
    extraTypeInfo.serialize(serializer)
```

src/common/types/types.cpp
```diff
@@ -594,7 +670,11 @@ std::unique_ptr<LogicalType> LogicalType::deserialize(Deserializer& deserializer
         extraTypeInfo = StructTypeInfo::deserialize(deserializer);
     } break;
     default:
-        extraTypeInfo = nullptr;
+        if (typeID == LogicalTypeID::DECIMAL) {
```
andyfengHKU, line 673: ditto

andyfengHKU (changes requested):
Overall LGTM. But I need to take a look again.

src/common/types/types.cpp
mxwli: getPhysicalType takes a LogicalTypeID, which means that either the if case has to be above getPhysicalType, or we have to throw inside getPhysicalType when the ID is DECIMAL

src/common/types/types.cpp
mxwli: As a whole LogicalType or just (LogicalTypeID, extraTypeInfo)?

src/common/types/types.cpp
mxwli: `make test` passes when you change the whole switch case to `if extraTypeInfo != nullptr ...`

src/common/types/types.cpp
mxwli: This might not be possible with deserialization, because extraTypeInfo is initially nullptr.

src/include/common/types/types.h
```diff
@@ -142,6 +142,7 @@ enum class LogicalTypeID : uint8_t {
     TIMESTAMP_NS = 38,
     TIMESTAMP_TZ = 39,
     INTERVAL = 40,
+    DECIMAL = 41,
```
andyfengHKU, line 145: put it under INT128

src/include/common/type_utils.h
```diff
@@ -189,6 +189,7 @@ class TypeUtils {
         case LogicalTypeID::UNION:
             return func(union_entry_t());
         /* NOLINTEND(bugprone-branch-clone)*/
+        case LogicalTypeID::DECIMAL:
```
andyfengHKU, line 192: ditto

andyfengHKU (commented):
Don't forget to rebase.

andyfengHKU (approved):

PR #3523, opened 2024-05-21
Add default value to CREATE

yiyun-sj:
# Description

Allows default values when creating and updating tables


dataset/binary-demo/version.txt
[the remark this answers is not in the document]
yiyun-sj: sure. I just regened this db and replaced the original. I will double check later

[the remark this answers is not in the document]
yiyun-sj: Not sure why all my generated dbs don't have this file...

ray6080 (approved):
LGTM! I will let @andyfengHKU to do the final around.

src/include/binder/ddl/bound_create_table_info.h
```diff
@@ -46,13 +46,20 @@ struct BoundCreateTableInfo {
...
+        const std::unique_ptr<parser::ParsedExpression>& defaultValue)
+        : name{std::move(name)}, type{std::move(type)},
+          defaultValue{defaultValue ? defaultValue->copy() : nullptr} {}
     EXPLICIT_COPY_DEFAULT_MOVE(PropertyInfo);
 
 private:
-    PropertyInfo(const PropertyInfo& other) : name{other.name}, type{other.type} {}
+    PropertyInfo(const PropertyInfo& other)
+        : name{other.name}, type{other.type},
+          defaultValue{other.defaultValue ? other.defaultValue->copy() : nullptr} {}
```
andyfengHKU, line 62: Ditto try to enforce defaultValue to not be a nullptr

src/include/binder/ddl/bound_alter_info.h
```diff
@@ -49,15 +50,17 @@ struct BoundExtraRenameTableInfo : public BoundExtraAlterInfo {
 struct BoundExtraAddPropertyInfo : public BoundExtraAlterInfo {
     std::string propertyName;
     common::LogicalType dataType;
-    std::shared_ptr<Expression> defaultValue;
+    std::unique_ptr<parser::ParsedExpression> defaultValue;
```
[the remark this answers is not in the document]
yiyun-sj: I wanted to use binding information originally, but setting the serialization and deserialization for all the expression classes did not go very smooth so I scrapped that. Also, duckdb uses parsed expressions rather than bound expressions, which probably has some reason behind their choice to do that

src/include/catalog/catalog_entry/node_table_catalog_entry.h
```diff
@@ -21,6 +21,7 @@ class NodeTableCatalogEntry final : public TableCatalogEntry {
     bool isParent(common::table_id_t /*tableID*/) override { return false; }
     common::TableType getTableType() const override { return common::TableType::NODE; }
     const Property* getPrimaryKey() const { return getProperty(primaryKeyPID); }
+    size_t getPrimaryKeyPos() const { return getPropertyPos(primaryKeyPID); }
```
andyfengHKU, line 24: don't use size_t. This will cause a problem/warning when we compile to WASM.

src/include/catalog/catalog_entry/table_catalog_entry.h
```diff
@@ -49,9 +49,11 @@ class KUZU_API TableCatalogEntry : public CatalogEntry {
     bool containProperty(const std::string& propertyName) const;
     common::property_id_t getPropertyID(const std::string& propertyName) const;
     const Property* getProperty(common::property_id_t propertyID) const;
+    size_t getPropertyPos(common::property_id_t propertyID) const;
```
andyfengHKU, line 52: ditto

src/binder/bind/bind_updating_clause.cpp
```diff
@@ -152,7 +153,10 @@ static void validatePrimaryKeyExistence(const NodeTableCatalogEntry* nodeTableEn
         }
         return; // No input needed for SERIAL primary key.
     }
-    if (!node.hasPropertyDataExpr(primaryKey->getName())) {
+    auto pkeyDefault = defaultExprs.at(nodeTableEntry->getPrimaryKeyPos());
+    if (!node.hasPropertyDataExpr(primaryKey->getName()) &&
+        pkeyDefault->expressionType == ExpressionType::LITERAL &&
```
andyfengHKU, line 158: I would add a function in ExpressionUtil to check if a given expression is NULL literal or not

PR #3546, opened 2024-05-27
Fix disabled test for 3524

andyfengHKU:
# Description

This PR turns some of the disabled test due to #3524. Tests related to join order and RDF are still failing.

Fixes # (issue)

# Contributor agreement


src/include/processor/operator/scan/scan_node_table.h
```diff
@@ -29,26 +29,27 @@ struct ScanNodeTableInfo {
...
     ScanNodeTableInfo(const ScanNodeTableInfo& other)
         : table{other.table}, columnIDs{other.columnIDs} {}
-
-    std::unique_ptr<ScanNodeTableInfo> copy() const {
-        return std::make_unique<ScanNodeTableInfo>(*this);
-    }
 };
 
 class ScanNodeTable final : public ScanTable {
+    static constexpr PhysicalOperatorType type_ = PhysicalOperatorType::SCAN_NODE_TABLE;
```
ray6080, line 44: Why make this a constexpr? Is this used anywhere else other than the constructor?

src/include/processor/operator/scan/scan_node_table.h
```diff
@@ -71,9 +71,8 @@ class ScanNodeTable final : public ScanTable {
 private:
     common::vector_idx_t currentTableIdx;
     // TODO(Guodong): Refactor following three fields into a vector of structs.
-    std::vector<std::unique_ptr<ScanNodeTableInfo>> infos;
+    std::vector<ScanNodeTableInfo> nodeInfos;
```
ray6080, line 74: Aren't we following the convention of keeping info and states (local and shared) separately? Why move the local state into info?

src/include/storage/store/table.h
```diff
@@ -6,7 +6,8 @@
 namespace kuzu {
 namespace storage {
 
-enum class TableScanSource : uint8_t { COMMITTED = 0, UNCOMMITTED = 1, NONE = 2 };
+enum class TableScanSource : uint8_t { COMMITTED = 0, UNCOMMITTED = 1, FRONTIER = 2, NONE = 3 };
```
ray6080, line 9: Do we need `FRONTIER` here?

src/processor/operator/scan/lookup_node_table.cpp
```diff
@@ -0,0 +1,40 @@
...
+bool LookupNodeTable::getNextTuplesInternal(ExecutionContext* context) {
+    if (hasExecuted) {
+        return false;
+    }
+    hasExecuted = true;
+    auto transaction = context->clientContext->getTx();
+    auto nodeID = nodeIDVector->getValue<nodeID_t>(0);
+    KU_ASSERT(tableIDToNodeInfo.contains(nodeID.tableID));
+    auto& nodeInfo = tableIDToNodeInfo.at(nodeID.tableID);
+    // TODO(Guodong): The following lines are probably incorrect.
```
ray6080, line 31: Can u leave a todo to me to fix the scan of uncommitted part?

ray6080 (approved):
Thanks for the fix!

src/processor/operator/scan/lookup_node_table.cpp
andyfengHKU: I might just fix this in my following PR

src/include/processor/operator/scan/scan_node_table.h
andyfengHKU: I'm doing this for simplicity. I'll try to make a decision once I figured out how to turn on IndexScan properly.

src/include/processor/operator/scan/scan_node_table.h
andyfengHKU: This is a new style I'm trying to enforce for constant values that is local to a class.

PR #3554, opened 2024-05-28
Support populating DEFAULT values in COPY FROM statements

yiyun-sj:
# Description

Support populating DEFAULT values in COPY FROM statements. Currently only on node tables.


src/binder/bind/copy/bind_copy_from.cpp
```diff
@@ -83,10 +83,14 @@ std::unique_ptr<BoundStatement> Binder::bindCopyNodeFrom(const Statement& statem
                 FileTypeUtils::toString(bindData->config.fileType)));
         }
     }
+    expression_vector defaultExprs;
```
andyfengHKU, line 86: This is a fine approach but I want to suggest a conceptually more clear solution.

Consider the following example

An input CSV with 2 columns `a, b`.
A table schema with 3 columns `x, y, z`. And with default value `1, 2, 3` respectively.
And the cypher query is `COPY <table> (z, y) FROM <csv>`

============

After `bindScanSource`, we will obtain two expressions `expr(a), expr(b)` corresponding to the two columns from CSV file. 

What we are doing in this PR it to generate another vector of default expressions align with table schema, i.e. `expr(1), expr(2), expr(3)`

Then in `NodeBatchInsert`, we generate a mapping (i.e. `ColumnPositions`) to align `x` with `expr(3)`, `y` with `expr(b)` and `z` with `expr(a)`.

=============

I suggest we resolve this mapping at binder stage. So after obtaining `expr(a), expr(b)` we generate another vector expressions by reordering & padding the expressions we get from `bindScanSource`. So we expect `expr(1), expr(b), expr(a)` as the input to `NodeBatchInsert`.

This is very similar logic as what we have done in map_copy.cpp. If we do it in binder, we can remove those code in mapper. And I think we no longer need to maintain ColumnPositions in `NodeBatchInsert`

src/processor/operator/persistent/node_batch_insert.cpp
```diff
@@ -124,8 +128,21 @@ void NodeBatchInsert::executeInternal(ExecutionContext* context) {
...
     }
 
+    auto nodeInfo = ku_dynamic_cast<BatchInsertInfo*, NodeBatchInsertInfo*>(info.get());
     while (children[0]->getNextTuple(context)) {
         auto originalSelVector = nodeLocalState->columnState->getSelVectorShared();
+        auto numTuples = originalSelVector->getSelSize();
+        for (auto i = 0u; i < nodeInfo->columnPositions.size(); ++i) {
+            auto pos = nodeInfo->columnPositions[i];
+            auto defaultVector = nodeLocalState->columnVectors[i];
+            if (!pos.isValid()) {
```
andyfengHKU, line 138: If we make the above change, you can no longer use invalid position to check for default expressions. You may need to pass additional field to indicate which expression is default and which is input.

andyfengHKU (changes requested):

src/include/binder/copy/bound_copy_from.h
```diff
@@ -19,8 +20,19 @@ struct BoundCopyFromInfo {
     std::unique_ptr<BoundBaseScanSource> source;
     // Row offset of input data to generate internal ID.
     std::shared_ptr<Expression> offset;
+    expression_vector columnExprs;
+    std::vector<common::LogicalType> columnTypes;
```
andyfengHKU, line 24: Do we need this? You can obtain columnType from columnExpr

src/processor/operator/persistent/node_batch_insert.cpp
```diff
@@ -97,23 +92,28 @@ void NodeBatchInsert::initLocalStateInternal(ResultSet* resultSet, ExecutionCont
...
             auto& columnType = nodeInfo->columnTypes[i];
-            auto nullVector = std::make_shared<ValueVector>(columnType);
-            nullVector->setState(state);
-            nullVector->setAllNull();
-            nodeLocalState->nullColumnVectors.push_back(nullVector);
-            nodeLocalState->columnVectors.push_back(nullVector.get());
+            std::shared_ptr<ValueVector> defaultVector = std::make_shared<ValueVector>(columnType);
+            defaultVector->setAllNull();
+            // TODO(Sam): Figure out how to ensure this state is set
+            // defaultVector->setState(state);
```
andyfengHKU, line 103: I think you can still setState here instead of doing at runtime.

andyfengHKU (approved):
LGTM

src/include/binder/copy/bound_copy_from.h
yiyun-sj: Currently we need columnTypes to include all columns whereas columnExprs skips SERIAL column. We can remove this once I'm done the serial refactor PR

src/processor/operator/persistent/node_batch_insert.cpp
```diff
@@ -124,8 +126,20 @@ void NodeBatchInsert::executeInternal(ExecutionContext* context) {
         token = nodeLocalState->localIndexBuilder->getProducerToken();
     }
 
+    auto nodeInfo = ku_dynamic_cast<BatchInsertInfo*, NodeBatchInsertInfo*>(info.get());
     while (children[0]->getNextTuple(context)) {
         auto originalSelVector = nodeLocalState->columnState->getSelVectorShared();
+        auto numTuples = originalSelVector->getSelSize();
```
ray6080, line 132: I'd separate this into a function, which can: 1. better isolate the operation; 2. give it a proper name so easier to understand the flow here inside while loop.

src/processor/operator/persistent/node_batch_insert.cpp
```diff
@@ -97,23 +92,30 @@ void NodeBatchInsert::initLocalStateInternal(ResultSet* resultSet, ExecutionCont
     }
     // NOLINTEND(bugprone-unchecked-optional-access)
 
-    KU_ASSERT(state != nullptr);
-    for (auto i = 0u; i < nodeInfo->columnPositions.size(); ++i) {
-        auto pos = nodeInfo->columnPositions[i];
-        if (pos.isValid()) {
-            nodeLocalState->columnVectors.push_back(resultSet->getValueVector(pos).get());
+    for (auto i = 0u; i < nodeInfo->columnEvaluators.size(); ++i) {
```
ray6080, line 95: Can u separate the handling of default and non-default vectors into different for loop? that looks easier to understand to me.
The first loop: 1. evaluate non-default evaluators and set column vectors; 2. figure out `state`.
The second loop: 1. create new local vector for default columns, set state for them, and add them to column vectors.

ray6080 (approved):

src/include/binder/copy/bound_copy_from.h
yiyun-sj: Currently columnType contains SERIAL columns while columnExpr doesn't. I will change this logic during my SERIAL refactor pr.

PR #3557, opened 2024-05-29
Disk array packed headers

benjaminwinger:
Packs the disk array headers into as few pages as possible (each header page points to the next one to support adding new disk arrays to the disk array collection) and packs the hash index headers into two pages.
I've also fixed some padding issues related to #3484, but there still seems to be a small amount of uninitialized data in the main hash index file (either due to something in the slot, or maybe related to the disk array write iterator).

A new class DiskArrayCollection gets passed around with the MetadataDAHInfo structure since the indices no longer refer to individual pages that can be easily read independently. Instead the DiskArrayCollection reads all the headers when initialized and the MetadataDAHInfo provides the index of the header for a particular chunk and an easy way of constructing the disk array.


benjaminwinger:
I'm not really sure why this is requiring a larger buffer pool size for the multi copy test to succeed. I'm suspicious that it might be caused by a subtle bug.

src/include/storage/storage_structure/disk_array.h
```diff
@@ -104,9 +101,9 @@ struct PIPUpdates {
 class DiskArrayInternal {
 public:
     // Used when loading from file
-    DiskArrayInternal(FileHandle& fileHandle, DBFileID dbFileID, common::page_idx_t headerPageIdx,
-        BufferManager* bufferManager, WAL* wal, transaction::Transaction* transaction,
-        bool bypassWAL = false);
+    DiskArrayInternal(FileHandle& fileHandle, DBFileID dbFileID,
```
ray6080, line 104: Should this now all be `BMFileHandle`?

src/include/storage/storage_structure/disk_array_collection.h
```diff
@@ -0,0 +1,83 @@
...
+    DiskArrayCollection(BMFileHandle& fileHandle, DBFileID dbFileID, BufferManager* bufferManager,
+        WAL* wal, common::page_idx_t firstHeaderPage = 0, bool bypassWAL = false);
+
+    static void writeEmptyHeadersToFile(FileHandle& handle, common::page_idx_t firstHeaderPage,
+        size_t numHeaders);
+
+    void prepareCommit();
+
+    void checkpointInMemory() {
+        for (size_t i = 0; i < headersForWriteTrx.size(); i++) {
```
ray6080, line 43: Should we avoid `size_t`? as I understand, it's not compatible on 32-bit platforms.

ray6080 (approved):

src/include/storage/wal_replayer_utils.h
```diff
@@ -3,6 +3,7 @@
 #include <string>
 
 #include "catalog/catalog_entry/node_table_catalog_entry.h"
+#include "common/types/internal_id_t.h"
```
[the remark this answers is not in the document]
benjaminwinger: I'm not actually sure why that was added; I don't recall modfying that file, but it is where `table_id_t` is declared.

src/include/storage/store/table_data.h
```diff
@@ -1,5 +1,6 @@
 #pragma once
 
+#include "storage/storage_structure/disk_array.h"
```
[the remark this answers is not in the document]
benjaminwinger: Yes, I guess I missed doing that in a few places.

src/include/common/constants.h
```diff
@@ -79,7 +79,7 @@ struct BufferPoolConstants {
     static constexpr uint64_t DEFAULT_VM_REGION_MAX_SIZE = (uint64_t)1 << 43; // (8TB)
 #endif
 
-    static constexpr uint64_t DEFAULT_BUFFER_POOL_SIZE_FOR_TESTING = 1ull << 26; // (64MB)
+    static constexpr uint64_t DEFAULT_BUFFER_POOL_SIZE_FOR_TESTING = 1ull << 27; // (128MB)
```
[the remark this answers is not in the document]
benjaminwinger: It's the multi copy test (I've run into this before; that time it was caused by a bug), which I think writes a gigabyte or so of hash index data. But it doesn't need all that memory at one time, so it should be fine with a small buffer pool.

src/include/storage/storage_structure/disk_array_collection.h
benjaminwinger: It's not incompatible, it's just that we can't expect it to be 64-bit on a 32-bit platform, so it shouldn't be used for things like iterating over the elements in the disk array where we expect to be able to go beyond. But `std::vector::size` returns a `size_t` anyway (or at least usually does, according to cppreference, it's a `size_type` defined in the vector implementation).

src/include/storage/storage_structure/disk_array.h
benjaminwinger: Yes. It probably could have been changed in one of the previous PRs.

PR #3559, opened 2024-05-30
Add required module in pyproject.toml

DylanShang:
Fixes # (issue)
Fix bug in the virtual environment: Can't find module "make"

### DIFF
--- a/tools/python_api/pyproject.toml
+++ b/tools/python_api/pyproject.toml
@@ -93,5 +93,5 @@ strict = true
 docstring-code-format = true
 
 [build-system]
-requires = ["setuptools", "wheel"]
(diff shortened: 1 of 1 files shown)


mewim:
Hi @DylanShang,

Thanks for your contribution. In the build process of kuzu, we treat cmake and C++ compiler as a system-level dependency (i.e. it should be installed by the package manager of the operating system prior to the build process as specified in [our docs](https://docs.kuzudb.com/developer-guide/)). For our own build-and-deploy process, using the system cmake is also more desirable and less error prone than pulling and installing the cmake distribution from pypi because it ensures we use exactly the same toolchain to build all the pre-compiled binaries and API bindings. If you have some specific reasons to add this as a build dependency, feel free to discuss with me.

PR #3561, opened 2024-05-31
Apply zone map to scan

andyfengHKU:
# Description

This PR apply zone map statistics to scan and avoid scanning node group if any column predicate will be evaluated to false. This PR only handles scan node table and comparison predicate with constant values.

Some benchmark result.

Dataset: LDBC 100 comment

| predicate | without zone map | with zone map |
| ---------  | ------------------ | -------------- |
|  length > 2000 |  ~910ms       |    4ms  |
|  length > 1900 |  ~910ms       |    ~240ms  |
|  length > 1000 |  ~910ms       |    ~840ms  |

Apparently how much benefit we get from zone map depends on data distribution. But it might worth to do zone map on a more fine-grained granularity than column chunk.

Fixes # (issue)

# Contributor agreement


test/test_files/tinysnb/filter/zone_map.test
ray6080, line 1: Remove this.

src/include/storage/store/table.h
```diff
@@ -16,6 +17,7 @@ struct TableScanState {
     TableScanSource source = TableScanSource::NONE;
     std::unique_ptr<TableDataScanState> dataScanState;
     common::node_group_idx_t nodeGroupIdx = common::INVALID_NODE_GROUP_IDX;
+    common::ZoneMapCheckResult zoneMapResult;
```
ray6080, line 20: Shouldn't this be in the same class with the predicates?

src/include/common/enums/zone_map_check_result.h
```diff
@@ -0,0 +1,14 @@
+#pragma once
+
+#include <cstdint>
+
+namespace kuzu {
+namespace common {
+
+enum class ZoneMapCheckResult : uint8_t {
+    NONE = 0,
+    ALWAYS_FALSE = 1,
```
ray6080, line 10: Will `ALWAYS_SCAN` AND `SKIP_SCAN` be more explicit?

src/storage/predicate/constant_predicate.cpp
```diff
@@ -0,0 +1,93 @@
...
+            return ZoneMapCheckResult::ALWAYS_FALSE;
+        }
+    }
+    case ExpressionType::LESS_THAN_EQUALS: {
+        if (LessThan::operation<T>(constant, min)) {
+            return ZoneMapCheckResult::ALWAYS_FALSE;
+        }
+    }
+    default:
+        break;
```
ray6080, line 58: KU_UNCREABLE.

ray6080 (approved):

src/storage/predicate/constant_predicate.cpp
```diff
@@ -0,0 +1,92 @@
...
+    default:
+        KU_UNREACHABLE;
+    }
+    return ZoneMapCheckResult::ALWAYS_SCAN;
+}
+
+ZoneMapCheckResult ColumnConstantPredicate::checkZoneMap(
+    const CompressionMetadata& metadata) const {
+    auto physicalType = value.getDataType()->getPhysicalType();
+    switch (physicalType) {
```
benjaminwinger, line 65: Booleans are also supported by the statistics backend.
You should actually be able to use the `StorageValueType` (in compression.h):
E.g.
```cpp
return TypeUtils::visit(..., 
    [&]<StorageValueType T>(T) { return checkZoneMapSwitch<T>(...); },
    [&](auto) { return ZoneMapCheckResult::ALWAYS_SCAN; },
);
```

benjaminwinger (approved):

PR #3564, opened 2024-05-31
Track variable sized memory manager allocations through the buffer manager

benjaminwinger:
To follow up the missing part of #3243, this adds BufferManager tracking for variable-sized memory allocations made through the MemoryManager.

I also noticed that the buffer manager wasn't freeing the memory related to evictions if it fails to allocate the total amount of memory requested when claiming a frame. I don't know how much of our code at the moment can safely recover from failing to claim memory and continue allocations though, so it may not make a difference at the moment. I had noticed subsequent failures which seemed to be related to data not being successfully written in the multi copy test that was failing due to not having enough memory in #3557 (e.g. accessing out of bounds in a hash index disk array, which I think would have been in bounds if the data had been updated correctly), so at the moment it may not always be possible to continue using the database after such a failure.

### DIFF
--- a/src/common/in_mem_overflow_buffer.cpp
+++ b/src/common/in_mem_overflow_buffer.cpp
@@ -7,7 +7,7 @@ uint8_t* InMemOverflowBuffer::allocateSpace(uint64_t size) {
     if (requireNewBlock(size)) {
         allocateNewBlock(size);
     }
-    auto data = currentBlock->block->buffer + currentBlock->currentOffset;
+    auto data = currentBlock->data() + currentBlock->currentOffset;
     currentBlock->currentOffset += size;
     return data;
 }
--- a/src/include/common/in_mem_overflow_buffer.h
+++ b/src/include/common/in_mem_overflow_buffer.h
@@ -11,10 +11,12 @@ namespace common {
 struct BufferBlock {
 public:
     explicit BufferBlock(std::unique_ptr<storage::MemoryBuffer> block)
-        : size{block->size}, currentOffset{0}, block{std::move(block)} {}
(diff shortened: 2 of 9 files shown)


ray6080 (approved):
Thanks! Can we add some tests over this? We should add a new call function `memory_usage()`, which returns currently used memory in bm.

benjaminwinger:
For the tests, do we just want to make sure it's not increasing after queries? In the e2e tests, is there a way of storing and comparing variables? Otherwise it seems like it would be easier to write a C++ test that could check both during and after the memory is allocated, but that wouldn't need a call function (we could have both I guess).

Page caching will also be an issue with that, as unless we also add a way to drop those caches we won't be able to guarantee that the memory usage won't go up or down after a query if new pages are cached, or if old pages are dropped to make space.

ray6080:
> For the tests, do we just want to make sure it's not increasing after queries? In the e2e tests, is there a way of storing and comparing variables? Otherwise it seems like it would be easier to write a C++ test that could check both during and after the memory is allocated, but that wouldn't need a call function (we could have both I guess).
> 
> Page caching will also be an issue with that, as unless we also add a way to drop those caches we won't be able to guarantee that the memory usage won't go up or down after a query if new pages are cached, or if old pages are dropped to make space.

Yeah, I think we just want to make sure that the mem usage is not going up after re-run same small queries. The case could be we record the `mem_usage()` after we open the database, then run a small query should bump the `mem_usage()` a bit higher, but if we run the small query again, the `mem_usage` should not go up assuming the tables the query touched is small to be fully cached already, and MM always releases its block buffers immediately after it's unpinned.

There isn't a way of storing variables now in testing framework, might be easier to do through CPP instead of Cypher.

PR #3565, opened 2024-05-31
Materialize SERIAL

yiyun-sj:
# Description

Rework serial into sequence under the hood.

## Benchmark times ran on AC4:
Threads: 1 Exec time: 34792
Threads: 2 Exec time: 19609.3
Threads: 4 Exec time: 11393.9
Threads: 8 Exec time: 7819.6
Threads: 16 Exec time: 7572.59
Threads: 32 Exec time: 7337.54
Threads: 64 Exec time: 8403.89

## Benchmark times ran on AC4:
Threads: 1 Exec time: 30793.9
Threads: 2 Exec time: 10203.4
Threads: 4 Exec time: 5823.29
Threads: 8 Exec time: 3857.11
Threads: 16 Exec time: 3011.69
Threads: 32 Exec time: 2550.39
Threads: 64 Exec time: 5425.65
Threads: 128 Exec time: 5002.29

Using dataset: comments_0_0 from ldbc, 2.1gb, padded with SERIAL PKEY column


src/catalog/catalog.cpp
```diff
@@ -202,6 +202,17 @@ void Catalog::dropTableSchema(transaction::Transaction* transaction, table_id_t
         // DO NOTHING.
     }
     }
+    for (auto& property : tableEntry->getPropertiesRef()) {
+        if (property.getDataType()->getLogicalTypeID() == LogicalTypeID::SERIAL) {
+            auto seqName = std::string(tableEntry->getName())
```
ray6080, line 207: Can u wrap this into a function? Should reuse it in `bind_Create_rdf_graph.cpp` too.

src/function/table/call/storage_info.cpp
```diff
@@ -160,6 +160,7 @@ static void appendColumnChunkStorageInfo(node_group_idx_t nodeGroupIdx,
         [&]<typename T>(T)
             requires(std::integral<T> || std::floating_point<T>)
         {
+            // TODO(Sam): Check if this needs any change
```
ray6080, line 163: I think so. We shouldn't need to differentiate serial from others now.

ray6080 (commented):
Generally looks good to me. Thanks! @andyfengHKU can take a quick look, too.

Let's fix the lock contention problem and revisit the benchmark again.

src/processor/operator/persistent/writer/parquet/parquet_writer.cpp
```diff
@@ -156,6 +157,10 @@ void ParquetWriter::setSchemaProperties(const LogicalType& type, SchemaElement&
         schemaElement.logicalType.TIMESTAMP.isAdjustedToUTC = false;
         schemaElement.logicalType.TIMESTAMP.unit.__isset.MILLIS = true;
     } break;
+    case LogicalTypeID::SERIAL: {
```
andyfengHKU, line 160: @acquamarin check here too.

src/binder/bind/bind_ddl.cpp
```diff
@@ -38,16 +41,46 @@ static void validateUniquePropertyName(const std::vector<PropertyInfo>& infos) {
     }
 }
 
+static std::unique_ptr<parser::ParsedFunctionExpression> createSerialDefaultExpr(std::string name) {
+    std::string literalRaw = "'" + name + "'";
```
andyfengHKU, line 45: You can leave literalRaw and functionRaw empty. We keep these information to throw the original input when an exception happens. Since this expr is created by us and exception is not expected, we don't care about `xxxRaw`

src/parser/transform/transform_ddl.cpp
```diff
@@ -170,6 +170,8 @@ std::unique_ptr<Statement> Transformer::transformAddProperty(
     auto propertyName = transformPropertyKeyName(*addPropertyCtx->oC_PropertyKeyName());
     auto dataType = transformDataType(*addPropertyCtx->kU_DataType());
     std::unique_ptr<ParsedExpression> defaultValue;
+    // TODO(Sam): Do we want to warn users to not set default on SERIAL (i.e. it will be
```
andyfengHKU, line 173: Yes and we should throw. But do it in binder if you can.

src/binder/bind/copy/bind_copy_rdf_graph.cpp
```diff
@@ -80,10 +81,13 @@ std::unique_ptr<BoundStatement> Binder::bindCopyRdfFrom(const parser::Statement&
     auto lSource = std::make_unique<BoundFileScanSource>(std::move(lFileScanInfo));
     auto lTableID = rdfGraphEntry->getLiteralTableID();
     auto lEntry = catalog->getTableCatalogEntry(clientContext->getTx(), lTableID);
-    auto lCopyInfo = BoundCopyFromInfo(lEntry, std::move(lSource), offset, std::move(lColumns),
-        logical_type_vec_t{*LogicalType::SERIAL(), *LogicalType::RDF_VARIANT(),
-            *LogicalType::STRING()},
-        std::vector<bool>{false, false}, nullptr /* extraInfo */);
+    auto lNodeEntry = lEntry->constPtrCast<NodeTableCatalogEntry>();
+    auto ser = expressionBinder.bindExpression(*lNodeEntry->getPrimaryKey()->getDefaultExpr());
+    auto lCopyInfo =
```
andyfengHKU, line 86: this line is getting a bit long. Can u split it as
```
auto lColumns =
auto lColmnTypes = 
```

Btw, you probably can remove `columnTypes` from `BoundCopyFromInfo` because now you are passing `columnExprs`

src/processor/operator/ddl/create_table.cpp
```diff
@@ -10,6 +11,10 @@ namespace kuzu {
 namespace processor {
 
 void CreateTable::executeDDLInternal(ExecutionContext* context) {
+    for (auto& child : children) {
```
andyfengHKU, line 14: I'm against this change. In this design, `CreateTable` is no longer source operator so it does not match our definition of `SimpleOperator`. Also the right way to call a child operator is to use `getNextTuple()`. This code also make a strong assumption that if `CreateTable` has a child, then it must be `CreateSequence` which is not good for future maintenance .

So I would just do `catalog->createSequence();` here instead of calling child operator. Let me know if this is not doable.

andyfengHKU (approved):
LGTM. Just make sure you fix the operator child comment.

kuzuexport/copy.cypher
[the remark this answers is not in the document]
yiyun-sj: I will remove these

src/binder/bind/copy/bind_copy_rdf_graph.cpp
yiyun-sj: I will come back to the columnTypes removal later as there is some more change to serial which might affect this

src/processor/operator/ddl/create_table.cpp
yiyun-sj: I agree. Going to revert these changes. I think I will handle it in the catalog instead to resolve any nesting (i.e. relTableGroup)

src/processor/operator/ddl/alter.cpp
```diff
@@ -17,7 +17,9 @@ void Alter::executeDDLInternal(ExecutionContext* context) {
             context->clientContext->getTx(), info.tableID);
         auto addedPropID = schema->getPropertyID(boundAddPropInfo.propertyName);
         auto addedProp = schema->getProperty(addedPropID);
-        defaultValueEvaluator->evaluate(context->clientContext);
+        // TODO(Sam/Guodong): We can maybe just fix the add column issue here using this new
```
ray6080, line 20: Yeah I think so.

ray6080:
Can u also add the benchmark numbers we got to PR description? Just to keep it as a record for revisiting later.

src/expression_evaluator/function_evaluator.cpp
```diff
@@ -10,30 +11,33 @@ using namespace kuzu::main;
...
     for (auto& child : children) {
-        child->evaluate(clientContext);
+        child->evaluate();
     }
     auto expr = expression->constPtrCast<binder::ScalarFunctionExpression>();
     if (execFunc != nullptr) {
-        auto bindData = expr->getBindData();
-        bindData->clientContext = clientContext;
-        execFunc(parameters, *resultVector, bindData);
+        auto bindData = expr->getBindData()->copy();
```
andyfengHKU, line 31: Let's make this copy at compile time and save it in the class.

src/processor/operator/ddl/alter.cpp
yiyun-sj: I will make an issue for this and address it in another pr

PR #3577, opened 2024-06-03
Cleanup hash index initialization

benjaminwinger:
Following the rewriting of how hash index and disk array headers are created in #3557, I found that removing createEmptyHashIndexFiles was working with a few tweaks, so this makes the hash index able to initialize itself if the file provided is nonexistent/empty.

I've also removed a couple of unused functions (and functions which were only used in a couple of places and were better off written inline) and added manual zeroing of the padding in `SlotEntry<T>` (which has padding if the type is smaller than 8 bytes).


src/include/storage/index/hash_index_slot.h
```diff
@@ -2,10 +2,13 @@
 
 #include <array>
 #include <cstdint>
+#include <cstring>
+#include <type_traits>
 
 #include "common/constants.h"
 #include "common/types/internal_id_t.h"
 #include <bit>
+#include <common/assert.h>
```
ray6080, line 11: `<common/assert.h>` -> `"common/assert.h"`

src/storage/index/hash_index.cpp
```diff
@@ -659,11 +622,19 @@ PrimaryKeyIndex::PrimaryKeyIndex(const DBFileIDAndName& dbFileIDAndName, bool re
...
+    hashIndexDiskArrays = std::make_unique<DiskArrayCollection>(*fileHandle,
+        dbFileIDAndName.dbFileID, &bufferManager, wal,
+        INDEX_HEADER_PAGES /*firstHeaderPage follows the index header pages*/, true /*bypassWAL*/);
 
     if (keyDataTypeID == PhysicalTypeID::STRING) {
         overflowFile = std::make_unique<OverflowFile>(dbFileIDAndName, &bufferManager, wal,
             readOnly, vfs, context);
     }
+    if (newIndex) {
+        for (size_t i = 0; i < NUM_HASH_INDEXES * 2; i++) {
```
ray6080, line 634: Can u add a comment here on `NUM_HASH_INDEXES * 2`? one for each pSlot and oSlot disk array.

src/storage/storage_structure/overflow_file.cpp
```diff
@@ -199,8 +188,8 @@ void OverflowFile::readFromDisk(transaction::TransactionType trxType, common::pa
 
 void OverflowFile::writePageToDisk(common::page_idx_t pageIdx, uint8_t* data) const {
     if (pageIdx < numPagesOnDisk) {
-        // TODO: updatePage does an unnecessary read + copy. We just want to overwrite
-        DBFileUtils::updatePage(*getBMFileHandle(), dbFileID, pageIdx, false, *bufferManager, *wal,
+        DBFileUtils::updatePage(*getBMFileHandle(), dbFileID, pageIdx,
```
ray6080, line 191: I feel we should rename the param `bool isInsertingNewPage` inside `DBFileUtils::updatePage`. The param is not intuitive. `bool readPageBeforeUpdate` should be better?

ray6080 (approved):

src/storage/storage_structure/overflow_file.cpp
benjaminwinger: Maybe. I'm hesitant to flip the meaning of the boolean though, as that's something easy to overlook. Maybe we should instead use the `BufferManager::PageReadPolicy` to make it more obvious what the parameter means?

PR #3581, opened 2024-06-03
Add create subgraph statement

andyfengHKU:
# Description 

This PR removes previous `graph()` function and replaced with `PROJECT GRAPH` clause.

Below shows an example 
```
PROJECT GRAPH g (person, knows) 
```

### Graph projection

Memgraph has a different projection
```
MATCH p = (a)-[e]->(b) WITH graph(p) AS g RETURN weakly_connected_component(g)
```
which looks cool but hard to implement because
- deduplication is needed 
- schema information is missing, i.e. for multi-label case, we don't know which table a node belongs to until runtime.

So I don't recommend going in this direction in the near future.

# Contributor agreement


src/include/main/client_context.h
```diff
@@ -118,8 +118,8 @@ class KUZU_API ClientContext {
     // only use for test framework
     std::vector<std::shared_ptr<parser::Statement>> parseQuery(std::string_view query);
 
+    // Remote database.
```
semihsalihoglu-uw, line 121: I'd remove the comment or say more. It's not very explanatory as is.

test/test_files/tinysnb/algorithm/basic.test
```diff
@@ -3,7 +3,17 @@
 --
 
 -CASE BasicAlgorithm
--STATEMENT CALL variable_length_path(graph("person", "knows"), 1, 1) RETURN *;
+
+-STATEMENT PROJECT GRAPH PK (person (age), knows) CALL variable_length_path(PK, 1, 1) RETURN *;
```
semihsalihoglu-uw, line 7: Should we change the inner ( ) where we specify the projected properties with { }? It would be better to use different delimiters and may make it closer to the { } we use in node and rel patterns in MATCH clauses, e.g., (a:Person { age: 25}). We also use { } to project intermediate properties in var-len rel patterns: e.g., `(a:User)-[e:Follows*1..2 (r, n | WHERE r.since > 2020 | {r.since}, {n.name})]`.

semihsalihoglu-uw (approved):

PR #3589, opened 2024-06-05
Take null values into account for copy statistics

### LINKED ISSUE #3572: Incorrect statistics on relationship table.
### Kùzu version

master

### What happened?

When running `CALL storage_info('knows') RETURN *` on `tinysnb`. The min & max value on `date` column is incorrect. It reflects the statistics for initial copy but not for the second one.

_No response_

benjaminwinger:
I've added additional tests to the copy rel statistics test, and also fixed a typo in tinysnb ("temprature" -> "temperature").
I also added special handling for bools in `column.cpp:getMinMax` since it was treating bool data as uncompressed instead of bitpacked.

This also uncovered an issue with the list columns, where we're assuming its safe to read null lists, where in reality that's only the case if the size is 0. Unfortunately, compressing data with nulls may result in null values being non-zero after decompression due to frame of reference compression and, well, ignoring null values when determining the range used for compression.

Here I've just made the ListColumnChunk use a size of zero for any null offset, though it may be useful to remove null chunks for columns which don't need them like the list size and column.

Fixes #3572 (again)


src/storage/store/column_chunk.cpp
```diff
@@ -403,67 +403,59 @@ bool ColumnChunk::sanityCheck() {
...
 ColumnChunkMetadata ColumnChunk::getMetadataToFlush() const {
     KU_ASSERT(numValues <= capacity);
-    StorageValue minValue = {}, maxValue = {};
+    std::optional<StorageValue> minValue, maxValue;
     if (capacity > 0) {
+        std::optional<NullMask> nullMask;
+        if (nullChunk) {
+            nullMask = nullChunk->getNullMask();
+        }
+        auto minmax_element = [&]<typename T>(std::span<T> data) {
```
ray6080, line 412: Can we make this as a function instead of a lambda? The former is more readable to me 😅 .

src/storage/store/column_chunk.cpp
```diff
@@ -403,67 +403,59 @@ bool ColumnChunk::sanityCheck() {
 
 ColumnChunkMetadata ColumnChunk::getMetadataToFlush() const {
     KU_ASSERT(numValues <= capacity);
-    StorageValue minValue = {}, maxValue = {};
+    std::optional<StorageValue> minValue, maxValue;
     if (capacity > 0) {
+        std::optional<NullMask> nullMask;
```
ray6080, line 408: Are we able to merge the code of computing min/max here with the one (`getMinMax`) inside `Column`? I feel they duplicate the same logic though very minor variations.

ray6080 (approved):

src/storage/store/column_chunk.cpp
benjaminwinger: Yes, that's a good idea.

PR #3591, opened 2024-06-05
Added query progress callbacks for nodejs api

MSebanc:
# Description

Added support for optional query callbacks that can be added to nodejs execute and query commands. 

# Contributor agreement

### DIFF
--- a/tools/nodejs_api/src_cpp/include/node_progress_bar_display.h
+++ b/tools/nodejs_api/src_cpp/include/node_progress_bar_display.h
@@ -1,7 +1,5 @@
 #pragma once
 
-#include <iostream>
-
 #include "common/task_system/progress_bar_display.h"
 #include <napi.h>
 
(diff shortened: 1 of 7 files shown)


tools/nodejs_api/src_cpp/include/node_progress_bar_display.h
```diff
@@ -0,0 +1,26 @@
+#pragma once
+
+#include <iostream>
```
mewim, line 3: Is `iostream` actually used in the implementation or is it only imported for printing/debugging? If not required, it should be removed.

mewim (approved):
LGTM

tools/nodejs_api/src_cpp/include/node_progress_bar_display.h
MSebanc: Yeah it's not used, accidentally left over from copying the terminal display class

PR #3618, opened 2024-06-10
add decimal datatype to Python and Java API

mxwli:
Also enables decimal comparison by changing a few lines.

[Doesn't add decimals to nodejs or rust API.](https://github.com/kuzudb/kuzu/issues/3619)

# Contributor agreement

### DIFF
--- a/src/include/function/comparison/vector_comparison_functions.h
+++ b/src/include/function/comparison/vector_comparison_functions.h
@@ -14,10 +14,7 @@ struct ComparisonFunction {
     static function_set getFunctionSet(const std::string& name) {
         function_set functionSet;
         for (auto& comparableType : common::LogicalTypeUtils::getAllValidLogicTypeIDs()) {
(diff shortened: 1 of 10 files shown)


acquamarin (commented):
Can you add tests that covers the decimal datatype in python and java?
Also, i think we should support decimal in the following apis as well:
1. Scan from pandas
2. Scan from arrow
3. get_as_df()
and other...

mxwli:
> Can you add tests that covers the decimal datatype in python and java? Also, i think we should support decimal in the following apis as well:
> 
> 1. Scan from pandas
> 2. Scan from arrow
> 3. get_as_df()
>    and other...

I've previously added decimals to arrow related API, but I'm not sure if I should add it for numpy related APIs, since numpy doesn't have support for exact decimals, and its target use isn't in financial systems. However, I can export it as a python object column in `get_as_df`

mewim (approved):

PR #3620, opened 2024-06-10
Use a clock-based Buffer Manager eviction strategy

### LINKED ISSUE #2137: Buffer Manager Eviction Rework

benjaminwinger:
Instead of queueing potential pages to evict when unpinning, all pages currently in memory are kept in the evictionqueue. When more space is needed, pages are evicted if they haven't changed since the last time we scanned them. The eviction queue is now only accessed when fetching pages from disk and has no global lock.

Fixes #2137

Based on the hash table implementation in the VMCache paper, except that we store page states by reference and for simplicity store the eviction candidates in a circular buffer instead of a hash table since lookups aren't necessary.

While this isn't the first place we're using atomics, GCC at least depends on libatomic for 16-byte compare and swap because it's not generating the 16-byte compare exchange instruction (cmpxchg16b) directly (see [the open bug](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=94649), and also the [atomic docs](https://gcc.gnu.org/wiki/Atomic/GCCMM/LIbrary)). Even with libatomic, it won't use cmpxchg16b without custom flags and instead uses a fallback (the `-mcx16` flag is for this specifically, but `-march=native` will also enable it if your CPU supports it or `-march=x86_64-v2` or newer). Benchmarks with `-mcx16` were almost identical, so I don't think this is much of a bottleneck (using 8-byte atomic eviction and insertion cursors should make it rare that two different threads are trying to evict or insert to the same spot anyway).

Performance improvements are minimal when there is plenty of buffer room.
I tried running some large copy benchmarks (two copies of 60M integers) with a restricted buffer, and found that the master version kept throwing buffer manager exceptions if I didn't give it a buffer pool of at least 1.5GB. With this version it works with the minimum buffer pool size of 64MB, and there was also a significant performance improvement when running this version with a 1.5GB buffer pool compared to the original.

Benchmarks are fairly rough; I ran each at least twice and averaged them (times are the first/second copies, each of 60M integers).

|   | master (ba0a6020e610) | this pr (07fcd85f) |
| ------------- | ------------- | -- |
| Unrestricted BufferPoolSize (~38GB)  | 4.4s/8.0 s  | 4.4s/7.8s | 
| BufferPoolSize = 1.5GB | 4.5s/23s  | 4.4s/13s |
| BufferPoolSize = 64MB | exception | 4.9s/24s |


ray6080:
Can u also benchmark on some pure scan queries on large dataset like ldbc100 comment table? I think the cold run should be able to present performance difference.

ray6080 (approved):
Nice PR!

One thing to discuss here: I feel we should start to take the mem usage of the queue into consideration. BM's available memory should be `bufferPoolSize - queueSize`. Though the ratio of queue's memory usage is small, the principle is to try to make memory usage more accurate when possible and convenient. What do you think?

benjaminwinger:
> I feel we should start to take the mem usage of the queue into consideration. BM's available memory should be bufferPoolSize - queueSize. Though the ratio of queue's memory usage is small, the principle is to try to make memory usage more accurate when possible and convenient. What do you think?

I think that's reasonable.

benjaminwinger:
> Can u also benchmark on some pure scan queries on large dataset like ldbc100 comment table? I think the cold run should be able to present performance difference.

There is a difference, but it doesn't appear to be huge.

`MATCH (c:Comment) RETURN c.*;`:
- Before: 11.3 seconds 
- After: 10.7 seconds

`MATCH (c:Comment) RETURN max(c.content);`:
- Before: 4.9 seconds (6.0 seconds with 64MB buffer pool size)
- After: 4.5 seconds (5.5 seconds with 64MB buffer pool size)

(both run cold from a database stored on an ssd).

PR #3621, opened 2024-06-10
Apply node semi mask to gds

andyfengHKU:
# Description

This PR enables passing node as an argument to GDS algorithms (e.g. shortest_path). Instead of explicitly accumulating node ids, we reuse our semi mask which is a lot more light weighted and automatically handles duplicate ids. This approach should work as long as we are passing node as runtime arguments.

There is, however, another semantic issue I realize while implementing this PR, e.g. 
```
MATCH (a:person) WHERE a.ID = 0 PROJECT GRAPH PK (person, knows) CALL shortest_path(PK, a, 2) RETURN a.fName, dst, length
```

There is no way to provide a guarantee that the  matched `a` will exist in projected graph `PK`. We need to make a decision on semantics before approaching this.

@semihsalihoglu-uw u may skip the optimizer code during review.

Fixes # (issue)

# Contributor agreement


semihsalihoglu-uw:
Another question: Is there any performance degradation or improvement of moving the shortest paths to the algorithms package?

src/include/planner/operator/sip/side_way_info_passing.h
```diff
@@ -5,13 +5,58 @@
...
+    NONE = 0,
+    PROBE_TO_BUILD = 1,
+    BUILD_TO_PROBE = 2,
+};
+
+/*
+ * We perform side way information passing in different scenarios
+ * 1. Semi mask
+ *
+ * During join order enumeration, we might disable semi mask if probe cardinality is large to avoid
```
semihsalihoglu-uw, line 38: Say a bit more here. materialization of what?

src/include/planner/operator/sip/side_way_info_passing.h
```diff
@@ -5,13 +5,58 @@
...
+ * */
+enum class SIPDirection {
+    NONE = 0,
+    PROBE_TO_BUILD = 1,
+    BUILD_TO_PROBE = 2,
+};
+
+/*
+ * We perform side way information passing in different scenarios
+ * 1. Semi mask
```
semihsalihoglu-uw, line 36: `1. Semi mask` looks like and incomplete sentence and not even a proper category (what you call scenario I think). Do you want to categorize as "Regular Joins or GDS calls" vs "Unnesting correlated sub-queries"?

src/include/planner/operator/sip/side_way_info_passing.h
```diff
@@ -5,13 +5,58 @@
...
+ *
+ * During join order enumeration, we might disable semi mask if probe cardinality is large to avoid
+ * large materialization.
+ *
+ * During filter push down, we disable semi mask if join condition is not id-based
```
semihsalihoglu-uw, line 41: I'm a bit lost with the comment. Are you enumerating cases when we do not use semi-mask? If so, have a sentence that says something like: "We do not use semi-mask-based sip for every join we make. Below are several cases when semi-mask-based sip is disabled: - .... - ..."

src/include/planner/operator/sip/side_way_info_passing.h
```diff
@@ -5,13 +5,58 @@
...
+ *
+ * If we add semi mask on build side, direction is BUILD_TO_PROBE and dependency is
```
semihsalihoglu-uw, line 43: Maybe put these two comments first because these comments are related to when we use a semi-mask.

src/include/planner/operator/sip/side_way_info_passing.h
```diff
@@ -5,13 +5,58 @@
...
+ * PROBE_DEPENDS_ON_BUILD (because scan need to be generated before semi masker).
+ *
+ * If we add semi mask on probe side, direction is PROBE_TO_BUILD and dependency is
+ * BUILD_DEPENDS_ON_PROBE.
+ *
+ * 2. Unnest corrected subquery
```
semihsalihoglu-uw, line 49: Unnest -> Unnesting

src/include/planner/operator/sip/side_way_info_passing.h
```diff
@@ -5,13 +5,58 @@
...
+ *
+ * When unnest correlated subquery, we first accumulate probe plan and pass information to build
```
semihsalihoglu-uw, line 51: unnest correlated -> unnesting a correlated ...

semihsalihoglu-uw, line 51: acummulate *the* probe plan and ... to *the* build ...

src/include/planner/operator/sip/side_way_info_passing.h
```diff
@@ -5,13 +5,58 @@
...
+ * side. Semi mask is NONE, direction is PROBE_TO_BUILD and dependency is PROBE_DEPENDS_ON_BUILD.
```
semihsalihoglu-uw, line 52: Do you mean SemiMaskPosition is NONE?

src/include/planner/operator/sip/side_way_info_passing.h
```diff
@@ -5,13 +5,58 @@
...
+ *
+ * During join order enumeration, we might disable semi mask if probe cardinality is large to avoid
+ * large materialization.
+ *
+ * During filter push down, we disable semi mask if join condition is not id-based
+ *
+ * If we add semi mask on build side, direction is BUILD_TO_PROBE and dependency is
+ * PROBE_DEPENDS_ON_BUILD (because scan need to be generated before semi masker).
+ *
+ * If we add semi mask on probe side, direction is PROBE_TO_BUILD and dependency is
```
semihsalihoglu-uw, line 46: I would be more explicit and instead of saying "If we add semi mask on the probe side, we set SemiMaskPosition to ON_PROBE, ..." Similarly for the above comment.

src/include/planner/operator/sip/logical_semi_masker.h
```diff
@@ -6,37 +6,50 @@
 namespace kuzu {
 namespace planner {
 
+/*
+ * There are two types of semi mask in the system.
```
semihsalihoglu-uw, line 10: I'm confused about the SemiMaskType notion? Regardless of how the node offsets are "collected", the semi mask is a mask of nodeOffsets, right? So it looks like SemiMaskType is really SemiMaskConstructionType or SemiMaskFillingTyp, meaning a field to identify how the semi mask is constructed (or as you call it collected but constructed or filled is more clear).

src/optimizer/acc_hash_join_optimizer.cpp
```diff
@@ -12,10 +13,101 @@
 using namespace kuzu::common;
 using namespace kuzu::binder;
 using namespace kuzu::planner;
+using namespace kuzu::function;
 
 namespace kuzu {
 namespace optimizer {
 
+static std::shared_ptr<LogicalOperator> appendAccumulate(std::shared_ptr<LogicalOperator> child) {
```
semihsalihoglu-uw, line 21: I'm afraid of reviewing this code, so as you suggest, I'm skipping it.

semihsalihoglu-uw (approved):

anuchak:
@andyfengHKU does this also support passing destinations for shortest path ? I mean like this - 

`MATCH (a:person), (b:person) WHERE a.ID < 10 AND b.ID < 20 PROJECT GRAPH PK (person, knows) CALL shortest_path(PK, a, b, 2) RETURN a.fName, dst, length`

Or is it specifically for 1 src all destinations query ?

src/function/gds/gds.cpp
```diff
@@ -0,0 +1,29 @@
...
+#include "processor/operator/gds_call.h"
+
+using namespace kuzu::binder;
+
+namespace kuzu {
+namespace function {
+
+// By default, create column expressions as variables.
+binder::expression_vector GDSAlgorithm::getResultColumns(Binder* binder) const {
+    auto columnName = getResultColumnNames();
```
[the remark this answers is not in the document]
andyfengHKU: okay I'm moving them into `getResultColumns`

src/function/gds/shortest_paths.cpp
```diff
@@ -100,21 +106,32 @@ class ShortestPath final : public GDSAlgorithm {
...
+        columns.push_back(binder->createVariable("length", *LogicalType::INT64()));
+        return columns;
     }
 
     void bind(const binder::expression_vector& params) override {
-        ExpressionUtil::validateExpressionType(*params[1], ExpressionType::LITERAL);
-        auto upperBound = params[1]->constCast<LiteralExpression>().getValue().getValue<int64_t>();
-        bindData = std::make_unique<ShortestPathBindData>(upperBound);
+        KU_ASSERT(params.size() == 3);
+        auto inputNode = params[1];
```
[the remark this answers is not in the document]
andyfengHKU: This allows multiple source node IDs

src/processor/map/map_gds.cpp
```diff
@@ -12,23 +16,36 @@ namespace processor {
...
         tableSchema->appendColumn(std::move(columnSchema));
     }
     auto table =
         std::make_shared<FactorizedTable>(clientContext->getMemoryManager(), tableSchema->copy());
-    auto sharedState = std::make_shared<GDSCallSharedState>(table);
+    auto graph = std::make_unique<OnDiskGraph>(clientContext, graphEntry.nodeTableIDs[0],
+        graphEntry.relTableIDs[0]);
+    std::unique_ptr<NodeOffsetSemiMask> mask = nullptr;
+    if (call.getInfo().getBindData()->hasNodeInput()) {
+        auto& node = call.getInfo().getBindData()->nodeInput->constCast<binder::NodeExpression>();
```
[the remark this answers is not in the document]
andyfengHKU: As per our discussion. We do have a way to pass destination IDs. It is just not in this PR.

andyfengHKU:
> @andyfengHKU does this also support passing destinations for shortest path ? I mean like this -
> 
> `MATCH (a:person), (b:person) WHERE a.ID < 10 AND b.ID < 20 PROJECT GRAPH PK (person, knows) CALL shortest_path(PK, a, b, 2) RETURN a.fName, dst, length`
> 
> Or is it specifically for 1 src all destinations query ?

It works for any number of src node input

PR #3625, opened 2024-06-11
Support DEFAULT for REL tables

yiyun-sj:
Allows DEFAULT for rel tables

### DIFF
--- a/src/binder/bind/copy/bind_copy_rdf_graph.cpp
+++ b/src/binder/bind/copy/bind_copy_rdf_graph.cpp
@@ -106,10 +106,10 @@ std::unique_ptr<BoundStatement> Binder::bindCopyRdfFrom(const parser::Statement&
     rrrExtraInfo->infos.push_back(sLookUp.copy());
     rrrExtraInfo->infos.push_back(pLookUp.copy());
     rrrExtraInfo->infos.push_back(oLookUp.copy());
-    expression_vector rrrCopyColumns {sOffset, oOffset, offset, pOffsetInternal};
-    std::vector<bool> rrrDefaults {false, false, false, false};
-    auto rrrCopyInfo = BoundCopyFromInfo(rrrEntry, std::move(rrrSource), offset, 
-        rrrCopyColumns, rrrDefaults, std::move(rrrExtraInfo));
+    expression_vector rrrCopyColumns{sOffset, oOffset, offset, pOffsetInternal};
+    std::vector<bool> rrrDefaults{false, false, false, false};
+    auto rrrCopyInfo = BoundCopyFromInfo(rrrEntry, std::move(rrrSource), offset, rrrCopyColumns,
+        rrrDefaults, std::move(rrrExtraInfo));
     // Bind copy literal triples
     func = inMemory ? BuiltInFunctionsUtils::matchFunction(clientContext->getTx(),
                           RdfLiteralTripleInMemScan::name, functions) :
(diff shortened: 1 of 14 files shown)


src/processor/map/map_copy_from.cpp
```diff
@@ -133,26 +100,26 @@ std::unique_ptr<PhysicalOperator> PlanMapper::mapPartitioner(LogicalOperator* lo
...
-        columnTypes.push_back(info->columnTypes);
-    }
-    auto sharedState = std::make_shared<PartitionerSharedState>(std::move(columnTypes));
+    std::vector<LogicalType> columnTypes;
+    std::vector<std::unique_ptr<evaluator::ExpressionEvaluator>> columnEvaluators;
+    getColumnEvaluators(copyFromInfo.columnExprs, outFSchema, columnEvaluators, columnTypes);
+    // Manually set columnTypes for _ID columns
+    columnTypes[0] = *LogicalType::INTERNAL_ID();
+    columnTypes[1] = *LogicalType::INTERNAL_ID();
+    columnTypes[2] = *LogicalType::INTERNAL_ID();
```
yiyun-sj, line 116: I dont really like how this looks even though previously we did something similar. However, the alternative is to manually create copies of the INTERNAL_ID version of offsets in the binding stage similar to the pid one I did in rdf copy bind, which I also don't think looks clean either ...

src/binder/bind/copy/bind_copy_rdf_graph.cpp
```diff
@@ -48,6 +49,8 @@ std::unique_ptr<BoundStatement> Binder::bindCopyRdfFrom(const parser::Statement&
     auto pOffset = expressionBinder.createVariableExpression(*LogicalType::INT64(), rdf::PID);
     auto oOffset = expressionBinder.createVariableExpression(*LogicalType::INT64(),
         InternalKeyword::DST_OFFSET);
+    auto pOffsetInternal = std::make_shared<VariableExpression>(*LogicalType::INTERNAL_ID(),
+        pOffset->getUniqueName(), std::string(rdf::PID));
```
yiyun-sj, line 53: (the manual creation mentioned in comment below)

src/processor/operator/partitioner.cpp
```diff
@@ -82,43 +82,46 @@ void PartitioningBuffer::merge(std::unique_ptr<PartitioningBuffer> localPartitio
...
-        if (pos.isValid()) {
-            dataChunk.insert(i, resultSet.getValueVector(pos));
-        } else {
-            auto nullVector = std::make_shared<ValueVector>(columnTypes[i]);
-            nullVector->setAllNull();
-            dataChunk.insert(i, nullVector);
-        }
+void Partitioner::evaluateData(const common::sel_t& numTuples) {
+    for (auto& evaluator : dataInfo->columnEvaluators) {
+        evaluator->getLocalStateRef().count = numTuples;
```
andyfengHKU, line 109: I wanna remove the `getXXXRef` naming and use `getXXXUnsafe` instead.

src/processor/map/map_copy_from.cpp
andyfengHKU, line 116: Why do we need to do this manually?

src/planner/plan/plan_copy.cpp
```diff
@@ -24,31 +24,19 @@ static void appendIndexScan(std::vector<IndexLookupInfo> infos, LogicalPlan& pla
...
-    for (auto& lookupInfo : extraInfo->infos) {
-        payloads.push_back(lookupInfo.offset);
-    }
     std::vector<std::unique_ptr<LogicalPartitionerInfo>> infos;
-    // Partitioner for FWD direction rel data.
     auto relTableEntry =
         ku_dynamic_cast<TableCatalogEntry*, RelTableCatalogEntry*>(copyFromInfo.tableEntry);
-    infos.push_back(std::make_unique<LogicalPartitionerInfo>(extraInfo->fromOffset, payloads,
+    // Partitioner for FWD direction rel data.
+    infos.push_back(std::make_unique<LogicalPartitionerInfo>(0 /* idx of src offset */,
```
andyfengHKU, line 31: Let's try to make 0, 1 constexpr.

src/include/processor/operator/partitioner.h
```diff
@@ -69,28 +67,40 @@ struct PartitionerLocalState {
...
+    std::vector<common::LogicalType> columnTypes;
+    std::vector<std::unique_ptr<evaluator::ExpressionEvaluator>> columnEvaluators;
+    std::vector<bool> defaultColumns;
+
+    PartitionerDataInfo(std::vector<common::LogicalType> columnTypes,
+        std::vector<std::unique_ptr<evaluator::ExpressionEvaluator>> columnEvaluators,
+        std::vector<bool> defaultColumns)
+        : columnTypes{std::move(columnTypes)}, columnEvaluators{std::move(columnEvaluators)},
+          defaultColumns{std::move(defaultColumns)} {}
+    inline std::unique_ptr<PartitionerDataInfo> copy() {
```
andyfengHKU, line 93: I think you can use object rather than unique_ptr. If so use `EXPLICIT_COPY_DEFAULT_MOVE` macro.

src/include/processor/operator/partitioner.h
```diff
@@ -102,21 +112,21 @@ class Partitioner : public Sink {
...
+    void evaluateData(const common::sel_t& numTuples);
+    common::DataChunk constructDataChunk(const std::shared_ptr<common::DataChunkState>& state);
     // TODO: For now, RelBatchInsert will guarantee all data are inside one data chunk. Should be
     //  generalized to resultSet later if needed.
     void copyDataToPartitions(common::partition_idx_t partitioningIdx,
         common::DataChunk chunkToCopyFrom);
 
 private:
     std::vector<std::unique_ptr<PartitioningInfo>> infos;
+    std::unique_ptr<PartitionerDataInfo> dataInfo;
```
andyfengHKU, line 129: Not related to this PR but I guess you could also change unique_ptr of `PartitioningInfo` to object?

andyfengHKU (approved):

src/processor/map/map_copy_from.cpp
yiyun-sj: When we make the variable expressions, we do it with an INT64 datatype for the sake of the index scans. So we need to change these back to their actual datatype. I left a comment above about the alternative

PR #3626, opened 2024-06-11
Add PostgresSQL Non Reserved Keywords

mxwli:
Adds the following to Kuzu's list of non reserved keywords:
```
ADD
ALTER
ATTACH
BY
CALL
COMMENT
COMMIT
COPY
DELETE
DETACH
DROP
EXPLAIN
EXTENSION
IF
KEY
LOAD
MATCH
MERGE
READ
RENAME
RETURN
ROLLBACK
SET
TRANSACTION
WRITE
```
All of which are listed as non-reserved in [Postgres](https://www.postgresql.org/docs/current/sql-keywords-appendix.html)
Removes the following keywords from Kuzu's list of Reserved Keywords:
```
DEFAULT
END
IN
```
All of which are listed as reserved from the same source.
Note that `AS` is also reserved in PG, but remains non-reserved in Kuzu for this PR.

# Contributor agreement


andyfengHKU (approved):
As we discussed, let's add a table to our website. In this table, we should list

keyword, reserved or not, comply to Postgres (Yes/No/NA, na meaning this keyword is not in PG)

PR #3635, opened 2024-06-12
Multi label graph interface

andyfengHKU:
# Description

This PR supports multi label graph in `Graph` interface.

# Contributor agreement

### DIFF
--- a/src/graph/on_disk_graph.cpp
+++ b/src/graph/on_disk_graph.cpp
@@ -10,52 +10,126 @@ using namespace kuzu::common;
 namespace kuzu {
 namespace graph {
 
-NbrScanState::NbrScanState(MemoryManager* mm) {
+static std::unique_ptr<RelTableScanState> getRelScanState(RelDataDirection direction, ValueVector* srcVector, ValueVector* dstVector) {
+    // Empty columnIDs since we do not scan any rel property.
+    auto columnIDs = std::vector<column_id_t>{};
+    auto scanState = std::make_unique<RelTableScanState>(columnIDs, direction);
+    scanState->nodeIDVector = srcVector;
+    scanState->outputVectors.push_back(dstVector);
+    return scanState;
+}
+
+OnDiskGraphScanState::OnDiskGraphScanState(MemoryManager* mm) {
     srcNodeIDVectorState = DataChunkState::getSingleValueDataChunkState();
     dstNodeIDVectorState = std::make_shared<DataChunkState>();
     srcNodeIDVector = std::make_unique<ValueVector>(*LogicalType::INTERNAL_ID(), mm);
     srcNodeIDVector->state = srcNodeIDVectorState;
     dstNodeIDVector = std::make_unique<ValueVector>(*LogicalType::INTERNAL_ID(), mm);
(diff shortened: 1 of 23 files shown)


src/graph/on_disk_graph.cpp
```diff
@@ -10,52 +10,132 @@ using namespace kuzu::common;
...
+            }
+            fwdRelTables.insert({relTableID, storage->getTable(relTableID)->ptrCast<RelTable>()});
+        }
+        nodeTableIDToFwdRelTables.insert({nodeTableID, std::move(fwdRelTables)});
+        common::table_id_map_t<storage::RelTable*> bwdRelTables;
+        for (auto& relTableID : nodeTableEntry.getBwdRelTableIDSet()) {
+            if (!graphEntry.containsRelTableID(relTableID)) {
+                continue;
+            }
+            bwdRelTables.insert({relTableID, storage->getTable(relTableID)->ptrCast<RelTable>()});
```
semihsalihoglu-uw, line 59: When we develop the on-disk-graph, do we construct a version with both forward and backward edges? Many graph algorithms only need one direction.

If we have a TODO list for algorithms package, can you add "Support for constructing graphs (on-disk and in-mem) that only contain one directional relationships."

src/planner/plan/plan_read.cpp
```diff
@@ -146,6 +146,18 @@ void Planner::planGDSCall(const BoundReadingClause& readingClause,
             planReadOp(getGDSCall(readingClause), predicatesToPush, *plan);
         }
     }
+    if (bindData->hasNodeOutput()) {
```
semihsalihoglu-uw, line 149: What's this change about? Is it connected to multi-label graphs or something independent?

src/processor/map/map_semi_masker.cpp
```diff
@@ -44,9 +44,11 @@ std::unique_ptr<PhysicalOperator> PlanMapper::mapSemiMasker(LogicalOperator* log
             KU_ASSERT(physicalOp->getChild(0)->getOperatorType() == PhysicalOperatorType::GDS_CALL);
             auto gds = physicalOp->getChild(0)->ptrCast<GDSCall>();
             KU_ASSERT(gds->hasSemiMask());
-            auto mask = gds->getSemiMask();
-            auto tableID = mask->getNodeTable()->getTableID();
-            masksPerTable.at(tableID).emplace_back(mask, 0);
+            // TODO(Xiyang): we can refactor this logic with RECURSIVE JOIN above.
```
semihsalihoglu-uw, line 47: Do you want to address this TODO while you're at it? You might leave algorithms anyways.

semihsalihoglu-uw (approved):

src/planner/plan/plan_read.cpp
andyfengHKU: I'll soon open a PR to make algorithm not only returns node_id but also node properties.

src/processor/map/map_semi_masker.cpp
andyfengHKU: I'll do this in a separate PR.

PR #3641, opened 2024-06-13
Add SEQUENCE to import/export framework

yiyun-sj:


src/processor/operator/ddl/create_sequence.cpp
```diff
@@ -11,11 +11,27 @@ namespace processor {
...
+    }
     catalog->createSequence(context->clientContext->getTx(), info);
 }
 
 std::string CreateSequence::getOutputMsg() {
-    return stringFormat("Sequence {} has been created.", info.sequenceName);
+    switch (info.onConflict) {
+    case common::ConflictAction::ON_CONFLICT_THROW:
+        return stringFormat("Sequence {} has been created.", info.sequenceName);
+    case common::ConflictAction::ON_CONFLICT_DO_NOTHING:
```
ray6080, line 30: The output message here is a bit confusing to me. When the conflict action is `ON_CONFLICT_DO_NOTHING`, it is also possible that "sequence has been created".

ray6080 (approved):
- Can u also add a test for `CREATE SEQUENCE IF NOT EXISTS`?
- I think we should keep exported sequences inside `schema.cypher` file, instead of a separate `sequence.cypher` file.

PR #3642, opened 2024-06-13
Make canUpdateInPlace operate on multiple values at once

benjaminwinger:
Originally all checks were single value, but there's now one place where we can take advantage of processing multiple values at once. I also noticed that in that situation null values weren't being handled correctly, which could lead to unnecessary out of place updates.

### DIFF
--- a/src/storage/compression/compression.cpp
+++ b/src/storage/compression/compression.cpp
@@ -1,5 +1,6 @@
 #include "storage/compression/compression.h"
 
+#include <algorithm>
 #include <cstdint>
 #include <limits>
 #include <string>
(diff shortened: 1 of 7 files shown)


src/storage/store/column.cpp
```diff
@@ -663,11 +662,10 @@ bool Column::canCommitInPlace(const ChunkState& state, const std::vector<offset_
     if (state.metadata.compMeta.canAlwaysUpdateInPlace()) {
         return true;
     }
-    for (auto i = 0u; i < dstOffsets.size(); i++) {
-        if (!state.metadata.compMeta.canUpdateInPlace(chunk->getData(), srcOffset + i,
-                dataType.getPhysicalType())) {
-            return false;
-        }
+    std::optional<NullMask> nullMask = chunk->getNullChunk()->getNullMask();
+    if (!state.metadata.compMeta.canUpdateInPlace(chunk->getData(), srcOffset, dstOffsets.size(),
```
ray6080, line 666: Should we just pass in the `std::optional<NullMask>` into `state.metadata.compMeta.canUpdateInPlace`?

src/storage/compression/compression.cpp
```diff
@@ -340,10 +382,19 @@ BitpackInfo<T> IntegerBitpacking<T>::getPackingInfo(const CompressionMetadata& m
...
     auto info = getPackingInfo(metadata);
-    auto newmetadata = CompressionMetadata(StorageValue(std::min(metadata.min.get<T>(), value)),
-        StorageValue(std::max(metadata.max.get<T>(), value)), metadata.compression);
+    auto [min, max] = getTypedMinMax<T>(values, nullMask, nullMaskOffset);
+    KU_ASSERT((min && max) || (!min && !max));
+    // If all values are null update can trivially be done in-place
+    if (!min) {
+        return true;
+    }
+    auto newmetadata =
```
ray6080, line 394: newmetadata -> newMetadata.

ray6080 (approved):

src/storage/store/column.cpp
benjaminwinger: Oh, yes that would be simpler. I was thinking along the lines of the update code which takes a pointer since it may point to an existing nullmask, but here we don't need that.

PR #3645, opened 2024-06-13
Disable support for non-constant default values on add column

yiyun-sj:


ray6080 (commented):
Are we able to only disable this on rel table?

ray6080 (approved):

PR #3649, opened 2024-06-16
Algorithm return node properties

andyfengHKU:
# Description

This PR adds additional argument to all algorithm function to output node instead of node ID. 

E.g.  we can now return node properties from weakly_connected_component
```
CALL weakly_connected_component
RETURN _node.fName, group_id
```

On the side, this PR fixes a bug when reusing the same scan state object to scan table. Some intermediate states were not reset correctly.

Fixes # (issue)

# Contributor agreement

### DIFF
--- a/src/binder/bind/read/bind_in_query_call.cpp
+++ b/src/binder/bind/read/bind_in_query_call.cpp
@@ -90,7 +90,7 @@ std::unique_ptr<BoundReadingClause> Binder::bindInQueryCall(const ReadingClause&
         }
         auto func = BuiltInFunctionsUtils::matchFunction(functionName, childrenTypes, entry);
         auto gdsFunc = *func->constPtrCast<GDSFunction>();
-        gdsFunc.gds->bind(children);
+        gdsFunc.gds->bind(children, this, graphEntry);
         outExprs = gdsFunc.gds->getResultColumns(this);
         auto info = BoundGDSCallInfo(gdsFunc.copy(), graphEntry.copy(), std::move(outExprs));
         boundReadingClause = std::make_unique<BoundGDSCall>(std::move(info));
--- a/src/function/gds/page_rank.cpp
+++ b/src/function/gds/page_rank.cpp
@@ -26,7 +26,8 @@ struct PageRankBindData final : public GDSBindData {
     PageRankBindData(std::shared_ptr<binder::Expression> nodeOutput, bool outputAsNode)
         : GDSBindData{std::move(nodeOutput), outputAsNode} {};
     PageRankBindData(const PageRankBindData& other)
-        : GDSBindData{other}, dampingFactor{other.dampingFactor}, maxIteration{other.maxIteration}, delta{other.delta} {}
+        : GDSBindData{other}, dampingFactor{other.dampingFactor}, maxIteration{other.maxIteration},
+          delta{other.delta} {}
 
     std::unique_ptr<GDSBindData> copy() const override {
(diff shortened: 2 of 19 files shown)


src/binder/bind/read/bind_in_query_call.cpp
```diff
@@ -90,7 +90,7 @@ std::unique_ptr<BoundReadingClause> Binder::bindInQueryCall(const ReadingClause&
         }
         auto func = BuiltInFunctionsUtils::matchFunction(functionName, childrenTypes, entry);
         auto gdsFunc = *func->constPtrCast<GDSFunction>();
-        gdsFunc.gds->bind(children);
+        gdsFunc.gds->bind(children, this, graphEntry);
```
semihsalihoglu-uw, line 93: Why is the 2nd argument passed?

src/function/gds/page_rank.cpp
```diff
@@ -67,9 +75,10 @@ class PageRank final : public GDSAlgorithm {
      * Inputs are
      *
      * graph::ANY
+     * outputProperty::BOOL
```
semihsalihoglu-uw, line 78: what is `outputProperty`? What property? PageRank property or node properties? Can you provide some documentation?

src/function/gds/shortest_paths.cpp
```diff
@@ -104,33 +109,37 @@ class ShortestPath final : public GDSAlgorithm {
      * graph::ANY
      * srcNode::NODE
      * upperBound::INT64
+     * outputProperty::BOOL
```
semihsalihoglu-uw, line 112: Ditto.

src/include/function/gds/gds.h
```diff
@@ -17,13 +18,23 @@ namespace function {
 
 // Struct maintaining GDS specific information that needs to be obtained at compile time.
 struct GDSBindData {
+    std::shared_ptr<binder::Expression> nodeOutput;
+    // If outputAsNode is true, we will scan all properties of the node.
+    // Otherwise, we return node ID only.
+    // This can be extended by allowing user specifify what properties to return in the arguments.
```
semihsalihoglu-uw, line 24: Remove this "This can be extended by allowing user specifify what properties to return in the arguments." I think we should make it an item in an issue called "Algorithms Package TODOs".

src/include/storage/store/column.h
```diff
@@ -45,19 +45,29 @@ class Column {
...
         ChunkState& getChildState(common::idx_t child);
         const ChunkState& getChildState(common::idx_t child) const;
 
-        ColumnChunkMetadata metadata;
-        uint64_t numValuesPerPage = UINT64_MAX;
-        common::node_group_idx_t nodeGroupIdx = common::INVALID_NODE_GROUP_IDX;
-        std::unique_ptr<ChunkState> nullState = nullptr;
-        // Used for struct/list/string columns.
-        std::vector<ChunkState> childrenStates;
+        void resetState() {
```
semihsalihoglu-uw, line 62: Why is this change part of this PR?

semihsalihoglu-uw (approved):

src/binder/bind/read/bind_in_query_call.cpp
andyfengHKU: binder class itself

src/function/gds/page_rank.cpp
andyfengHKU: Comment is in gds.h

PR #3651, opened 2024-06-17
Enable semi mask

### LINKED ISSUE #3541: Optimization: Re-enable semi mask
### Description

Semi mask is disabled in #3524 to simplify the rework. See https://github.com/kuzudb/kuzu/pull/3524#discussion_r1612126572.

andyfengHKU:
# Description

This PR re-enables semi mask after our refactor on node table scan.

Fixes #3541 

# Contributor agreement

### DIFF
--- a/src/include/storage/store/node_table.h
+++ b/src/include/storage/store/node_table.h
@@ -2,12 +2,12 @@
 
 #include "common/cast.h"
 #include "common/types/types.h"
+#include "processor/operator/mask.h"
 #include "storage/index/hash_index.h"
 #include "storage/stats/nodes_store_statistics.h"
 #include "storage/store/chunked_node_group.h"
 #include "storage/store/node_table_data.h"
 #include "storage/store/table.h"
-#include "processor/operator/mask.h"
 
 namespace kuzu {
 namespace evaluator {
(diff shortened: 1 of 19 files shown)


src/include/storage/store/table.h
```diff
@@ -22,6 +22,7 @@ struct TableScanState {
     std::unique_ptr<TableDataScanState> dataScanState;
     common::node_group_idx_t nodeGroupIdx = common::INVALID_NODE_GROUP_IDX;
 
+    // Zone map
```
ray6080, line 25: Either remove the comment or be more explicit to explain the usage of `columnPredicateSets` and `zoneMapResult`.

src/include/storage/store/node_table.h
```diff
@@ -22,14 +23,16 @@ class LocalNodeTable;
 struct NodeTableScanState final : TableScanState {
     // Local storage node group.
     LocalNodeNG* localNodeGroup = nullptr;
+    // Semi mask
```
ray6080, line 26: Remove the comment here.

ray6080 (approved):
Can u also add into the PR description that now node scan applies vector-level semi mask instead of node-level? Also, we should run some benchmarks over the comparison of node-level vs. vector-level. maybe we can grab some reported numbers from our benchmark server

andyfengHKU:
> Can u also add into the PR description that now node scan applies vector-level semi mask instead of node-level? Also, we should run some benchmarks over the comparison of node-level vs. vector-level. maybe we can grab some reported numbers from our benchmark server

I was going to do some benchmark. But then I realize our benchmark report already shows improvement.

Issue #3532, opened 2024-05-23
Feature: Compression support for more data types


ray6080:
### API

Other

### Description

We don't support compression for some data types: `INT128`, `UUID`, `DOUBLE`, `FLOAT`, `INTERVAL`. This leads to unexpected file size reported in #3529.

ray6080:
We now have `INT128`, `FLOAT` and `DOUBLE` covered.

PR #3658, opened 2024-06-17
Add compression for int128

royi-luo:
# Description

#3532

Add support for compression of int128_t using a bitpacking algorithm from [duckdb](https://github.com/duckdb/duckdb/blob/main/src/storage/compression/bitpacking_hugeint.cpp).

Additionally, expanded the `int128_t` type + some utilities to support this.

# Testing

We have existing integration tests on the tinysnb dataset that use int128 or UUIDs (which are encoded with int128). There is no regression on these tests.

Additionally, unit tests were added (for example in `test/storage/compression_test.cpp`).

# Benchmarks

The INT128 benchmarks were conducted on two datasets, both with identical single-column schema `create node table person (ID INT128, PRIMARY KEY (ID));`. The results of the benchmarks will be compared to uncompressed + INT64 implementations.

The INT64 benchmarks were conducted on similar datasets but with column type INT64.

To conduct the benchmarks, a copy of the dataset is run and then a query.

```
PROFILE COPY person FROM "dataset/int128-db/vPerson.csv";
PROFILE MATCH (a:person) RETURN a.ID;
```

For the 1st dataset (constructed as **favourable for FOR + bitpacking**) the data column is `2^127 - NUM_NODES, ... 2^127 - 2, 2^127 - 1` so the bitpacking bitwidth should be appx. `log2(NUM_NODES)`

For the 2nd dataset (constructed as **unfavourable for FOR + bitpacking**) the data column is `1, 2^125 - 2, 3, 2^125 - 4...` so the bitpacking bitwidth should be 125 (For INT64, it will be 61).

Results are below, the logs referenced for the results are attached. I'll also summarize it in the comments.
[compressed64-favourable.log](https://github.com/user-attachments/files/15890625/compressed64-favourable.log)
[compressed64-unfavourable.log](https://github.com/user-attachments/files/15890626/compressed64-unfavourable.log)
[uncompressed128-favourable.log](https://github.com/user-attachments/files/15890627/uncompressed128-favourable.log)
[uncompressed128-unfavourable.log](https://github.com/user-attachments/files/15890628/uncompressed128-unfavourable.log)
[compressed128-favourable.log](https://github.com/user-attachments/files/15890692/compressed128-favourable.log)
[compressed128-unfavourable.log](https://github.com/user-attachments/files/15890693/compressed128-unfavourable.log)

# Contributor agreement


royi-luo:
## Compression + Favourable

### 10M nodes

**COPY**: 16270.112000 (BATCH_INSERT) + 2165.257000 (IN_QUERY_CALL) ms

**QUERY**: 195.067000 (RESULT_COLLECTOR) + 0.712000 (PROJECTION) + 887.819000 (SCAN_MULTI_NODE_TABLE) ms

**DATA SIZE**: 28155904 bytes ~= 28MB

### 100M nodes

**COPY**: 154751.880000 (BATCH_INSERT) + 22658.890000 (IN_QUERY_CALL) ms

**QUERY**: 2169.660000 (RESULT_COLLECTOR) + 13.473000 (PROJECTION) + 10107.456000 (SCAN_MULTI_NODE_TABLE) ms

**DATA SIZE**: 280801280 bytes ~= 280MB

## Compression + Unfavourable

### 10M nodes

**COPY**: 16700.908000 (BATCH_INSERT) + 3404.738000 (IN_QUERY_CALL) ms

**QUERY**: 1609.878000 (RESULT_COLLECTOR) + 1.461000 (PROJECTION) + 986.156000 (SCAN_MULTI_NODE_TABLE) ms

**DATA SIZE**: 162742272 bytes ~= 163MB

### 100M nodes

**COPY**: 152409.754000 (BATCH_INSERT) + 21385.475000 (IN_QUERY_CALL) ms

**QUERY**: 5279.330000 (RESULT_COLLECTOR) + 20.080000 (PROJECTION) + 4675.911000 (SCAN_MULTI_NODE_TABLE) ms

**DATA SIZE**: 1612627968 bytes ~= 1612MB

## Uncompressed + Favourable

### 10M nodes

**COPY**: 36144.830000 (BATCH_INSERT) + 2126.818000 (IN_QUERY_CALL) ms

**QUERY**: 447.879000 (RESULT_COLLECTOR) + 1.559000 (PROJECTION) + 449.387000 (SCAN_MULTI_NODE_TABLE) ms

**DATA SIZE**: 162742272 bytes ~= 163MB

### 100M nodes

**COPY**: 157139.965000 (BATCH_INSERT) + 19755.251000 (IN_QUERY_CALL) ms

**QUERY**: 5389.821000 (RESULT_COLLECTOR) + 23.956000 (PROJECTION) + 5400.968000 (SCAN_MULTI_NODE_TABLE) ms

**DATA SIZE**: 1612627968 bytes ~= 1612MB
# Benchmark Results

## Uncompressed + Unfavourable

Skipping as the results are similar to the favourable dataset and we are not very interested in comparing uncompressed performance. The log is attached though.

royi-luo:
In the favourable scenario, the size of the database when compressing 128-bit and 64-bit integers is almost identical. In this scenario, the maximum offset is appx. `NUM_NODES` regardless of the width of the physical data type and so this is as expected. I would expect the compression ratio will be `128 / log2(NUM_NODES)` so for 100M nodes this is ~4.8x. However, the observed compression ratio is actually slightly better (~5.7x) but I'm not familiar enough with the system to be sure as to why this is the case.

As for performance, copy performance appears to be similar, and query performance seems to trade RESULT_COLLECTOR performance (~2x faster with compression) with SCAN_MULTI_NODE_TABLE performance (~2x faster without compression).

When comparing with INT64 compression, even when keeping in mind that our data type is twice is wide, the IN_QUERY_CALL and SCAN_MULTI_NODE_TABLE steps are still considerable slower. One potential cause of this (although we can't be sure without profiling) is mentioned in the [comments on the duckdb implementation](https://github.com/duckdb/duckdb/pull/8277#pullrequestreview-1539056128) as we also have additional overflow checks in our int128 implementation which could be a source of extra overhead.

src/include/common/numeric_utils.h
```diff
@@ -0,0 +1,76 @@
...
+constexpr int BitWidth(T x) {
+    return std::bit_width(x);
+}
+
+template<>
+constexpr int BitWidth<int128_t>(int128_t x) {
+    if (x.high != 0) {
+        constexpr size_t BITS_PER_BYTE = 8;
+        return sizeof(x.low) * BITS_PER_BYTE +
+               std::bit_width(static_cast<MakeUnSignedT<decltype(x.high)>>(x.high));
```
royi-luo, line 70: use `MakeValueUnsigned`

src/include/storage/compression/bitpacking_int128.h
```diff
@@ -0,0 +1,14 @@
+#pragma once
+
+#include "common/types/int128_t.h"
+
```
royi-luo, line 4: add comment with source in header file as well

src/include/storage/compression/sign_extend.h
```diff
@@ -38,16 +38,19 @@ T Load(const uint8_t* ptr) {
...
+template<class T, class T_U = typename common::NumericUtils::MakeUnSignedT<T>, uint64_t CHUNK_SIZE>
 static void SignExtend(uint8_t* dst, uint8_t width) {
     KU_ASSERT(width < sizeof(T) * 8);
     T const mask = T_U(1) << (width - 1);
     for (uint64_t i = 0; i < CHUNK_SIZE; ++i) {
         T value = Load<T>(dst + i * sizeof(T));
-        value = value & ((T_U(1) << width) - T_U(1));
+        static constexpr uint8_t num_bits_in_t_u = sizeof(T_U) * 8;
+        const T_U and_mask = (~(T_U(1) << (num_bits_in_t_u - 1))) >> (num_bits_in_t_u - 1 - width);
+        value = value & and_mask;
```
royi-luo, line 49: clean this up (maybe move to private function)

src/storage/compression/bitpacking_int128.cpp
```diff
@@ -0,0 +1,315 @@
...
+        out[i] |= static_cast<common::int128_t>(in[OFFSET + 3]) << 96;
+    }
+}
+
+//===--------------------------------------------------------------------===//
+// Packing
+//===--------------------------------------------------------------------===//
+
+static void packSingle(const common::int128_t in, uint32_t* __restrict& out, uint16_t delta,
+    uint16_t shl, common::int128_t mask) {
```
royi-luo, line 129: consider renaming `shl` and `shr` to `shiftLeft` and `shiftRight`

test/common/int128_test.cpp
```diff
@@ -0,0 +1,66 @@
+#include <string>
+
+#include "common/types/int128_t.h"
+#include "gtest/gtest.h"
+
+using namespace kuzu::common;
+
+TEST(Int128Tests, LeftShiftTest) {
+    {
+        int128_t original_value{1ULL << 63, 0};
```
royi-luo, line 10: clean up tests

test/storage/bitpacking_test.cpp
```diff
@@ -0,0 +1,60 @@
+#include <algorithm>
```
royi-luo, line 1: can probably remove as this is redundant with compression_test but try to add more tests to compression_test

src/storage/compression/bitpacking_int128.cpp
```diff
@@ -0,0 +1,315 @@
...
+    case 96:
+        packDelta96(in, out);
+        break;
+    case 128:
+        packDelta128(in, out);
+        break;
+    default:
+        static constexpr uint8_t num_bits_in_t_u = sizeof(common::int128_t) * 8;
+        const auto and_mask =
+            (~(common::int128_t(1) << (num_bits_in_t_u - 1))) >> (num_bits_in_t_u - 1 - width);
```
royi-luo, line 276: clean this up

src/storage/compression/bitpacking_int128.cpp
```diff
@@ -0,0 +1,315 @@
...
+        for (common::idx_t oindex = 0; oindex < IntegerBitpacking<uint32_t>::CHUNK_SIZE - 1;
```
royi-luo, line 277:
```suggestion
        for (common::idx_t oindex = 0; oindex < IntegerBitpacking<uint128_t>::CHUNK_SIZE - 1;
```

src/include/common/numeric_utils.h
```diff
@@ -0,0 +1,85 @@
...
+
+// produces bitmasks in the form ...111111
+// can produce masks with # of bits up to sizeof(T) * 8 - 1
+template<IsIntegral T>
+inline T getSaturatedBitmask(size_t width) {
+    static constexpr uint8_t numBitsInT = sizeof(T) * 8;
+    const T fullMask = ~(T(1) << (numBitsInT - 1));
+    const size_t numBitsToDiscard = (numBitsInT - 1 - width);
+    return (fullMask >> numBitsToDiscard);
+}
```
royi-luo, line 82: duplicate of BitmaskUtils::all1sMaskForLeastSignificantBits

benjaminwinger (approved):
Generally looks good!

src/include/common/numeric_utils.h
benjaminwinger: Not quite identical, as that function supports `numBits == 64`, but they should probably be unified.

src/include/common/numeric_utils.h
```diff
@@ -0,0 +1,85 @@
...
+
+// produces bitmasks in the form ...111111
+// can produce masks with # of bits up to sizeof(T) * 8 - 1
+template<IsIntegral T>
+inline T getSaturatedBitmask(size_t width) {
+    static constexpr uint8_t numBitsInT = sizeof(T) * 8;
+    const T fullMask = ~(T(1) << (numBitsInT - 1));
+    const size_t numBitsToDiscard = (numBitsInT - 1 - width);
+    return (fullMask >> numBitsToDiscard);
```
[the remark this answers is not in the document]
royi-luo: Yea this was basically a hack to get around the overflow checks in `int128_t`. We don't have an unsigned version and if the shift value is 127 it'll complain about subtracting 1 from `INT128_MIN`.
Not using `~T(0) >> ...` was to avoid the sign bit filling that comes with right shifting :smile: 
I'll think a bit harder about seeing if I can find a cleaner approach I guess worst case I can have this as a specialization for `int128_t` only so it doesn't mess with the primitive integer types.

src/include/storage/compression/compression.h
```diff
@@ -23,20 +24,24 @@ class ColumnChunkData;
...
     template<typename T>
-        requires std::integral<T> && std::numeric_limits<T>::is_signed
+        requires std::same_as<std::remove_cvref_t<T>, common::int128_t>
+    explicit StorageValue(T value) : signedInt128(value) {}
+    template<typename T>
+        requires std::integral<T> && common::NumericUtils::IsSigned<T>
     explicit StorageValue(T value) : signedInt(value) {}
     template<typename T>
-        requires std::integral<T> && (!std::numeric_limits<T>::is_signed)
+        requires std::integral<T> && (!common::NumericUtils::IsSigned<T>)
```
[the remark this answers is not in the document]
royi-luo: will fix, this was probably a result of me find + replacing (then realising in some cases I wanted to handle primitives and `int128_t` separately and only changing the `std::integral` part hehe)

src/storage/compression/compression.cpp
```diff
@@ -326,15 +329,25 @@ void ConstantCompression::copyFromPage(const uint8_t* srcBuffer, uint64_t srcOff
 }
 
 template<typename T>
-static inline T abs(typename std::enable_if<std::is_unsigned<T>::value, T>::type value) {
+inline T abs(T value);
```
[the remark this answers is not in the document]
royi-luo: My compiler was complaining about giving the `static` specifier to an explicit specialization and I wasn't sure if I could remove the `static` keyword from just the specialization so I removed it from the original template functions as well. Do you know if the above is ok?

src/include/common/numeric_utils.h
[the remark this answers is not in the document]
benjaminwinger: >  Not using ~T(0) >> ... was to avoid the sign bit filling that comes with right shifting 😄

Good point, I wasn't thinking about int128 being signed when I wrote that. If this works, then I guess leave it, though maybe leave a comment indicating that it has to avoid overflows since it needs to work for int128.

This sounds like a good argument for a uint128 type, even just for internal use. Signed overflow may be undefined, but a lot of stuff relies on unsigned overflow being well-defined. That may be more work than it's worth though; I don't think we'll need to do stuff like implementing an int128 version of our hash function.
https://github.com/kuzudb/kuzu/blob/21dfb5b72d26e92a29ee87b8aaff03d2ce34be77/src/include/function/hash/hash_functions.h#L20-L24

src/include/common/numeric_utils.h
royi-luo: consolidated

src/include/storage/compression/compression.h
```diff
@@ -23,20 +24,24 @@ class ColumnChunkData;
...
-concept StorageValueType = (std::integral<T> || std::floating_point<T>);
+concept StorageValueType = (common::NumericUtils::IsIntegral<T> || std::floating_point<T>);
 // Type storing values in the column chunk statistics
-// Only supports integers (up to 64bit), floats and bools
+// Only supports integers (up to 128bit), floats and bools
 union StorageValue {
     int64_t signedInt;
     uint64_t unsignedInt;
     double floatVal;
+    common::int128_t signedInt128;
```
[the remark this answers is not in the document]
royi-luo: updated constructors

src/include/storage/compression/compression.h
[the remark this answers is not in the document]
royi-luo: updated

src/include/storage/compression/compression.h
```diff
@@ -55,40 +60,19 @@ union StorageValue {
     template<StorageValueType T>
     T get() const {
         if constexpr (std::integral<T>) {
-            if constexpr (std::numeric_limits<T>::is_signed) {
+            if constexpr (common::NumericUtils::IsSigned<T>) {
```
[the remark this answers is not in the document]
royi-luo: updated

test/common/CMakeLists.txt
```diff
@@ -1,7 +1,2 @@
 add_kuzu_test(types_test
-        date_test.cpp
-        interval_test.cpp
-        null_mask_test.cpp
-        string_test.cpp
-        time_test.cpp
-        timestamp_test.cpp)
+        int128_test.cpp)
```
royi-luo, line 2: restore

src/include/common/numeric_utils.h
```diff
@@ -0,0 +1,75 @@
...
+#include <type_traits>
+
+#include "common/types/int128_t.h"
+#include "common/types/types.h"
+#include <bit>
+#include <concepts>
+
+namespace kuzu {
+namespace common {
+namespace NumericUtils {
```
ray6080, line 12: Instead of nesting the namespace `NumericUtils` here, Is it possible to add a utility struct here. Also, for namespace, we always use lower case characters.

ray6080 (approved):

ray6080:
can u refresh the PR title a bit? `Royi/3532` doesn't seem necessary for the PR title.
Also, I wonder if now the tinysnb dataset becomes a bit smaller (https://github.com/kuzudb/kuzu/issues/3529). If so, can u comment there with some new numbers?

src/storage/compression/compression.cpp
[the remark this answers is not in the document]
royi-luo: keeping it inline as quick googling showed keeping the `static` may not be a good idea

src/include/common/numeric_utils.h
royi-luo: Updated namespace name, I'd prefer keeping it as a namespace over a struct as there is no state being stored

royi-luo:
> can u refresh the PR title a bit? `Royi/3532` doesn't seem necessary for the PR title. Also, I wonder if now the tinysnb dataset becomes a bit smaller (#3529). If so, can u comment there with some new numbers?

image

New tinysnb database looks like this

```
> ll
total 19M
-rw-r--r-- 1 kuzu kuzu 6.7K Jun 19 15:44 catalog.kz
-rw-r--r-- 1 kuzu kuzu  25M Jun 19 15:44 data.kz
-rw-r--r-- 1 kuzu kuzu 4.3M Jun 19 15:44 metadata.kz
-rw-r--r-- 1 kuzu kuzu  84K Jun 19 15:44 n-0.hindex
-rw-r--r-- 1 kuzu kuzu  44K Jun 19 15:44 n-1.hindex
-rw-r--r-- 1 kuzu kuzu  44K Jun 19 15:44 n-2.hindex
-rw-r--r-- 1 kuzu kuzu  12K Jun 19 15:44 n-2.hindex.ovf
-rw-r--r-- 1 kuzu kuzu 2.0K Jun 19 15:44 nodes.statistics_and_deleted.ids
-rw-r--r-- 1 kuzu kuzu 3.7K Jun 19 15:44 rels.statistics
```

PR #3665, opened 2024-06-18
Updated progress bar for asynchronous queries

MSebanc:
# Description

Improved progress bar handling of asynchronous query calls

# Contributor agreement

### DIFF
--- a/src/include/common/task_system/progress_bar.h
+++ b/src/include/common/task_system/progress_bar.h
@@ -21,7 +21,7 @@ public:
 
     static std::shared_ptr<ProgressBarDisplay> DefaultProgressBarDisplay();
 
-    void addPipeline(std::string id);
+    void addPipeline();
 
     void finishPipeline(std::string id);
 
--- a/src/include/common/task_system/progress_bar.h
+++ b/src/include/common/task_system/progress_bar.h
@@ -23,17 +23,17 @@ public:
 
     void addPipeline();
 
-    void finishPipeline(std::string id);
+    void finishPipeline(uint64_t queryID);
 
-    void endProgress(std::string id);
+    void endProgress(uint64_t queryID);
 
-    void startProgress(std::string id);
+    void startProgress(uint64_t queryID);
 
     void toggleProgressBarPrinting(bool enable);
 
     void setShowProgressAfter(uint64_t showProgressAfter);
 
-    void updateProgress(std::string id, double curPipelineProgress);
+    void updateProgress(uint64_t queryID, double curPipelineProgress);
 
     void setDisplay(std::shared_ptr<ProgressBarDisplay> progressBarDipslay);
 
@@ -42,9 +42,9 @@ public:
     bool getProgressBarPrinting() const { return trackProgress; }
 
 private:
-    void resetProgressBar(std::string id);
+    void resetProgressBar(uint64_t queryID);
 
-    void updateDisplay(std::string id, double curPipelineProgress);
+    void updateDisplay(uint64_t queryID, double curPipelineProgress);
 
     bool shouldUpdateProgress() const;
 
--- a/src/include/common/task_system/progress_bar.h
+++ b/src/include/common/task_system/progress_bar.h
@@ -21,32 +21,30 @@ public:
 
     static std::shared_ptr<ProgressBarDisplay> DefaultProgressBarDisplay();
 
-    void addPipeline();
+    void addPipeline(std::string id);
 
-    void finishPipeline();
+    void finishPipeline(std::string id);
 
-    void endProgress();
+    void endProgress(std::string id);
 
-    void addJobsToPipeline(int jobs);
-
-    void finishJobsInPipeline(int jobs);
-
-    void startProgress();
+    void startProgress(std::string id);
 
     void toggleProgressBarPrinting(bool enable);
 
     void setShowProgressAfter(uint64_t showProgressAfter);
 
-    void updateProgress(double curPipelineProgress);
+    void updateProgress(std::string id, double curPipelineProgress);
 
     void setDisplay(std::shared_ptr<ProgressBarDisplay> progressBarDipslay);
+
+    std::shared_ptr<ProgressBarDisplay> getDisplay() { return display; }
 
     bool getProgressBarPrinting() const { return trackProgress; }
 
 private:
-    void resetProgressBar();
+    void resetProgressBar(std::string id);
 
-    void updateDisplay(double curPipelineProgress);
+    void updateDisplay(std::string id, double curPipelineProgress);
 
     bool shouldUpdateProgress() const;
 
--- a/src/include/common/task_system/progress_bar_display.h
+++ b/src/include/common/task_system/progress_bar_display.h
@@ -15,11 +15,12 @@ public:
     virtual ~ProgressBarDisplay() = default;
 
     // Update the progress of the pipeline and the number of finished pipelines. queryID is used to
-    // identify the query
+    // identify the query when we track progress of multiple queries asynchronously
     virtual void updateProgress(uint64_t queryID, double newPipelineProgress,
         uint32_t newNumPipelinesFinished) = 0;
 
-    // Finish the progress display. queryID is used to identify the query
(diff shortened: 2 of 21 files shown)


src/main/connection.cpp
```diff
@@ -71,6 +76,12 @@ std::unique_ptr<QueryResult> Connection::executeWithParams(PreparedStatement* pr
     return clientContext->executeWithParams(preparedStatement, std::move(inputParams));
 }
 
+std::unique_ptr<QueryResult> Connection::executeWithParamsWithId(
```
andyfengHKU, line 79: Do we need this? I'm pretty sure if we prepare, we cannot support multiple statements. You can add a test case on this.

src/include/common/task_system/progress_bar.h
```diff
@@ -23,30 +23,28 @@ class ProgressBar {
 
     void addPipeline();
 
-    void finishPipeline();
+    void finishPipeline(std::string id);
```
andyfengHKU, line 26: better to rename to queryID and comment when this is needed.

src/include/main/connection.h
```diff
@@ -74,6 +76,10 @@ class Connection {
      */
     KUZU_API std::unique_ptr<QueryResult> executeWithParams(PreparedStatement* preparedStatement,
         std::unordered_map<std::string, std::unique_ptr<common::Value>> inputParams);
+
+    std::unique_ptr<QueryResult> executeWithParamsWithId(PreparedStatement* preparedStatement,
```
andyfengHKU, line 80: I would remove this.

src/include/processor/execution_context.h
```diff
@@ -1,17 +1,26 @@
...
 
 struct ExecutionContext {
+    std::string queryId;
     common::Profiler* profiler;
     main::ClientContext* clientContext;
 
-    ExecutionContext(common::Profiler* profiler, main::ClientContext* clientContext)
-        : profiler{profiler}, clientContext{clientContext} {}
+    ExecutionContext(common::Profiler* profiler, main::ClientContext* clientContext, std::string id)
+        : profiler{profiler}, clientContext{clientContext} {
```
andyfengHKU, line 16: I think @ray6080 will need a system level query id soon. So let's make this id generated from database which you will need a counter and a lock. Then ExecutionContext should always be created with an id and does not need to generate one in the constructor.


Also I would suggest using a integer based id.

src/main/client_context.cpp
```diff
@@ -482,7 +483,7 @@ std::unique_ptr<QueryResult> ClientContext::executeAndAutoCommitIfNecessaryNoLoc
     }
     auto queryResult = std::make_unique<QueryResult>(preparedStatement->preparedSummary);
     auto profiler = std::make_unique<Profiler>();
-    auto executionContext = std::make_unique<ExecutionContext>(profiler.get(), this);
+    auto executionContext = std::make_unique<ExecutionContext>(profiler.get(), this, id);
```
andyfengHKU, line 486: So here try to get a system level id from database.

src/main/client_context.cpp
```diff
@@ -503,7 +504,7 @@ std::unique_ptr<QueryResult> ClientContext::executeAndAutoCommitIfNecessaryNoLoc
         }
     } catch (std::exception& e) {
         transactionContext->rollback();
-        progressBar->endProgress();
+        progressBar->endProgress(executionContext->queryId);
```
andyfengHKU, line 507: We prefer ID over Id as a convention I think.

tools/nodejs_api/src_cpp/include/node_connection.h
```diff
@@ -78,16 +79,21 @@ class ConnectionExecuteAsyncWorker : public Napi::AsyncWorker {
     ~ConnectionExecuteAsyncWorker() override = default;
 
     void Execute() override {
+        std::string id = common::UUID::toString(
```
andyfengHKU, line 82: ditto here try to get a system level id from database.

src/include/main/client_context.h
```diff
@@ -109,8 +109,9 @@ class KUZU_API ClientContext {
     // Query.
     std::unique_ptr<PreparedStatement> prepare(std::string_view query);
     std::unique_ptr<QueryResult> executeWithParams(PreparedStatement* preparedStatement,
-        std::unordered_map<std::string, std::unique_ptr<common::Value>> inputParams);
-    std::unique_ptr<QueryResult> query(std::string_view queryStatement);
+        std::unordered_map<std::string, std::unique_ptr<common::Value>> inputParams,
```
andyfengHKU, line 112: ditto I would not change this API given we don't have multi query in a prepared statement.

andyfengHKU (changes requested):
Overall LGTM but I need to take another look for system level query id.

Also @mewim should take a look on the js part.

src/include/common/task_system/progress_bar_display.h
```diff
@@ -14,11 +16,13 @@ class ProgressBarDisplay {
 
     virtual ~ProgressBarDisplay() = default;
 
-    // Update the progress of the pipeline and the number of finished pipelines
-    virtual void updateProgress(double newPipelineProgress, uint32_t newNumPipelinesFinished) = 0;
+    // Update the progress of the pipeline and the number of finished pipelines. queryID is used to
+    // identify the query
```
andyfengHKU, line 18: Say queryID is needed when we track progress of multiple queries asynchronously.

src/include/processor/execution_context.h
```diff
@@ -1,17 +1,20 @@
 #pragma once
 
 #include "common/profiler.h"
+#include "common/types/uuid.h"
```
andyfengHKU: this should be no longer needed.

mewim:
I do not think we should print anything from the Node.js API side. If it is for debug only, you should delete the printing.

src/include/main/database.h
```diff
@@ -139,6 +142,8 @@ class Database {
     std::unique_ptr<extension::ExtensionOptions> extensionOptions;
     std::unique_ptr<DatabaseManager> databaseManager;
     common::case_insensitive_map_t<std::unique_ptr<storage::StorageExtension>> storageExtensions;
+    uint64_t queryID;
```
andyfengHKU, line 145: Let's wrap this into a struct called QueryIDGenerator

src/include/main/database.h
```diff
@@ -121,6 +122,8 @@ class Database {
     common::case_insensitive_map_t<std::unique_ptr<storage::StorageExtension>>&
     getStorageExtensions();
 
+    uint64_t getNextQueryID();
```
andyfengHKU, line 125: ditto move to QueryIDGenerator.

tools/nodejs_api/src_cpp/include/node_progress_bar_display.h
```diff
@@ -1,6 +1,10 @@
 #pragma once
 
+#include <optional>
+#include <unordered_set>
+
 #include "common/task_system/progress_bar_display.h"
+#include "common/task_system/terminal_progress_bar_display.h"
```
andyfengHKU, line 7: include terminal_progress_bar does not sound correct to me.

tools/nodejs_api/src_cpp/node_connection.cpp
```diff
@@ -41,6 +41,8 @@ Napi::Value NodeConnection::InitAsync(const Napi::CallbackInfo& info) {
 
 void NodeConnection::InitCppConnection() {
     this->connection = std::make_shared<Connection>(database.get());
+    connection->getClientContext()->getProgressBar()->setDisplay(
+        std::make_shared<NodeProgressBarDisplay>());
```
andyfengHKU, line 45: not related to this PR. But I think you might be able to use unique_ptr for display in ProgressBar class.

andyfengHKU (approved):

tools/nodejs_api/src_cpp/include/node_connection.h
```diff
@@ -2,6 +2,7 @@
 
 #include <utility>
 
+#include "common/types/uuid.h"
```
andyfengHKU, line 5: remove this

PR #3668, opened 2024-06-18
Fix 3652

### LINKED ISSUE #3652: Bug: Kuzu CLI hangs with `iri` property
### Kùzu version

v0.4.2

Fedora 39

### What happened?

```cypher
CREATE NODE TABLE V (id INT64, PRIMARY KEY(id));
CREATE REL TABLE E (FROM V TO V, iri STRING);
MATCH (v:V)-[e:E]->(v2:V) RETURN *;
```
Last statement will hang. Replacing `iri` by `p` works as usual.

_No response_

andyfengHKU:
# Description

The problem is that user use a reserved keyword as column name. This PR performs a check over column name in DDL.

Fixes #3652

# Contributor agreement

### DIFF
--- a/src/binder/binder.cpp
+++ b/src/binder/binder.cpp
@@ -172,15 +172,44 @@ std::string Binder::getUniqueExpressionName(const std::string& name) {
     return "_" + std::to_string(lastExpressionId++) + "_" + name;
 }
 
-bool Binder::isReservedPropertyName(const std::string& name) {
+struct ReservedNames {
+    // Column name that might conflict with internal names.
+    static std::unordered_set<std::string> getColumnNames() {
+        return {
+            InternalKeyword::ID,
+            InternalKeyword::LABEL,
+            InternalKeyword::SRC,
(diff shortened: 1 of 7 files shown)


src/binder/binder.cpp
```diff
@@ -172,15 +172,36 @@ std::string Binder::getUniqueExpressionName(const std::string& name) {
...
-        return true;
-    }
-    if (normalizedName == StringUtils::getUpper(std::string(rdf::PID))) {
-        return true;
-    }
-    return false;
+    return reservedColumnNames.contains(normalizedName);
+}
+
+static std::unordered_set<std::string> reservedPropertyLookup = {
```
ray6080, line 197: What's the difference between `reservedPropertyLookup` and `reservedColumnNames`? Is the latter one used by ddl statements and the former one used by read clauses? I don't think the name is clear here. better rename it or add comments to explain it a bit more.

ray6080 (approved):

PR #3670, opened 2024-06-19
Remove redundant CSRListEntries data

benjaminwinger:
The CSR offsets/lengths are also cached in the csrHeaderChunks. Pre-calculating them may have a slight benefit, as getCSRLength/getStartCSROffset handle a few edge cases which make them slower than just a memory lookup, but the differences are minor and pre-calculating them slows down small scans.

This reduces the cost of small rel table scans by more than half. E.g. in one query from the LDBC fintech sf1 dataset which was taking ~60s, largely as a result of doing many small rel table scans as part of a recursive join, this reduced it to ~30s in total (more improvements are possible; this was the easiest of them).

### DIFF
--- a/src/storage/store/rel_table_data.cpp
+++ b/src/storage/store/rel_table_data.cpp
@@ -17,9 +17,7 @@ namespace storage {
 RelDataReadState::RelDataReadState(const std::vector<column_id_t>& columnIDs)
     : TableDataScanState{columnIDs}, nodeGroupIdx{INVALID_NODE_GROUP_IDX}, numNodes{0},
       currentNodeOffset{0}, posInCurrentCSR{0}, readFromPersistentStorage{false},
-      readFromLocalStorage{false}, localNodeGroup{nullptr} {
-    resetState();
-}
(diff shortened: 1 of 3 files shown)


src/storage/store/rel_table_data.cpp
```diff
@@ -88,7 +78,6 @@ void RelDataReadState::resetState() {
     numNodes = 0;
     currentNodeOffset = 0;
     posInCurrentCSR = 0;
-    csrListEntries.resize(StorageConstants::NODE_GROUP_SIZE, {0, 0});
     csrHeaderChunks = ChunkedCSRHeader(false);
```
ray6080, line 81: Now revisit this, I think we can just reset offset and length chunks by `resetEmpty()`, instead of re-creating new chunks?

benjaminwinger: Done. I also removed the resetState call in the constructor. The only thing that the constructor wasn't doing was resizing the csrListEntries, so now it's entirely redundant.

ray6080 (approved):

PR #3672, opened 2024-06-19
Fix Python Tests

mxwli:
# Description

Change numpy version to 1.26; 2.0.0 has compatibility issues with our current api

# Contributor agreement

### DIFF
--- a/tools/python_api/requirements_dev.txt
+++ b/tools/python_api/requirements_dev.txt
@@ -1,6 +1,6 @@
 # required for tests
 networkx~=3.0
-numpy
+numpy==1.26.4
 pandas
 polars
 pyarrow>=15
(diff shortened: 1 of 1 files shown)


tools/python_api/requirements_dev.txt
```diff
@@ -1,6 +1,6 @@
 # required for tests
 networkx~=3.0
-numpy
+numpy=1.26.4
```
mewim, line 3: Let's use `~=1.26`, so we still get auto-upgrade for minor releases.

mewim (changes requested):
See my comment

mewim (approved):

royi-luo (approved):

PR #3680, opened 2024-06-20
Fix zero column node COPY

yiyun-sj:
this fixes the case where no columns are actually copied and only implicit serial tuples are filled. 
this only occurs for node tables since rel will always have from/to columns

### DIFF
--- a/src/processor/operator/persistent/node_batch_insert.cpp
+++ b/src/processor/operator/persistent/node_batch_insert.cpp
@@ -75,7 +75,6 @@ void NodeBatchInsert::initGlobalStateInternal(ExecutionContext* context) {
 }
 
 void NodeBatchInsert::initLocalStateInternal(ResultSet* resultSet, ExecutionContext* context) {
-    std::shared_ptr<DataChunkState> state;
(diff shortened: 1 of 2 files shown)


src/processor/operator/persistent/node_batch_insert.cpp
```diff
@@ -111,7 +108,8 @@ void NodeBatchInsert::initLocalStateInternal(ResultSet* resultSet, ExecutionCont
 
     nodeLocalState->nodeGroup = NodeGroupFactory::createNodeGroup(ColumnDataFormat::REGULAR,
         nodeInfo->columnTypes, info->compressionEnabled);
-    nodeLocalState->columnState = std::move(state);
+    KU_ASSERT(resultSet->dataChunks[0]);
```
ray6080, line 111: Can u also assert that there is only one dataChunk in the resultSet? This should be an implicit assumption we hold for now.

ray6080 (approved):
thanks!

src/processor/operator/persistent/node_batch_insert.cpp
yiyun-sj: When i gdb'd the example case, if saw there were 2 datachunks in the resultSet. I'm not sure if that implicit assumption is being upheld currently...

PR #3690, opened 2024-06-21
Rust multithreaded tests

benjaminwinger:
Since we've removed the logger (which did some non-thread-safe global operations) it's possible to run the rust tests in parallel. I also added a relwithdebinfo build profile for use in CI after noticing that the tests on the macos runner were running significantly slower than in release mode (on Linux its not as big of a difference, ~4s in release mode vs ~15s in debug mode, but on macos it was ~4s in release mode and ~600s in debug mode).

I also noticed that we weren't explicitly setting the build parallelism for the bundled cmake build in a few places (and auto-detection doesn't seem to be working on macos at least, so that was slowing down some builds as they were running single-threaded).

### DIFF
--- a/.github/workflows/ci-workflow.yml
+++ b/.github/workflows/ci-workflow.yml
@@ -54,6 +54,7 @@ jobs:
     runs-on: kuzu-self-hosted-testing
     env:
       NUM_THREADS: 32
+      CMAKE_BUILD_PARALLEL_LEVEL: 32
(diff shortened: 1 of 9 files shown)


benjaminwinger:
I think the multithreaded part is running into process limits for virtual memory, making mmap fail.

andyfengHKU (approved):

PR #3704, opened 2024-06-25
Reduce initial capacity of ChunkedCSRHeader

### LINKED ISSUE #3695: Optimization: Optimizing the scheduling overhead of the short queries
### Description

In some short quries,for example, ldbc sf100 database,` match (a:vertex)-[]-(b) where a.__uid='10010832' return count(b);` 
image

ScanMultiRelTable::initLocalStateInternal wasted a lot of time

image

benjaminwinger:
Small scans, particularly ones across many rel tables with many threads, end up spending much more time initializing the scan state's header chunks than actually scanning. Starting them small and letting them grow as necessary speeds up small scans without much extra cost for large ones

Fixes #3695


ray6080 (approved):
I feel we should cover more queries in our benchmark set. Should we add some multi rel table scan queries into our benchmark?

PR #3723, opened 2024-06-27
Add PyArrow.lib.Table Scanning

mxwli:
# Contributor agreement

### DIFF
--- a/tools/python_api/src_cpp/py_connection.cpp
+++ b/tools/python_api/src_cpp/py_connection.cpp
@@ -62,6 +62,24 @@ static std::unique_ptr<function::ScanReplacementData> tryReplacePolars(py::dict&
     }
 }
 
+static std::unique_ptr<function::ScanReplacementData> tryReplacePyArrow(py::dict& dict,
+    py::str& objectName) {
+    if (!dict.contains(objectName)) {
+        return nullptr;
+    }
+    auto entry = dict[objectName];
+    if (PyConnection::isPyArrowTable(entry)) {
+        auto scanReplacementData = std::make_unique<function::ScanReplacementData>();
+        scanReplacementData->func = PyArrowTableScanFunction::getFunction();
+        auto bindInput = function::TableFuncBindInput();
+        bindInput.inputs.push_back(Value::createValue(reinterpret_cast<uint8_t*>(entry.ptr())));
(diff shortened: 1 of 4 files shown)


andyfengHKU (approved):

tools/python_api/src_cpp/py_connection.cpp
```diff
@@ -100,7 +124,7 @@ static std::unique_ptr<function::ScanReplacementData> replacePythonObject(
     }
     if (nameMatchFound) {
         throw BinderException(
-            stringFormat("Variable {} found but no matches were DataFrames", objectName));
+            stringFormat("Variable {} found but no matches were scannable", objectName));
```
prrao87, line 127: Could this error message be made a bit more helpful? Something like `Unable to scan {}. Can only scan from Pandas/Polars DataFrames or Arrow tables`. I feel like users should be immediately made aware of what works here rather than the fact that it doesn't work.

tools/python_api/test/test_scan_pyarrow.py
```diff
@@ -0,0 +1,20 @@
...
+            [1, 2, 3],
+            ['a', 'b', 'c'],
+            [True, False, None],
+        ],
+        names=["col1", "col2", "col3"]
+    )
+    result = conn.execute("LOAD FROM tab RETURN *")
+    assert (result.get_next() == [1, 'a', True])
+    assert (result.get_next() == [2, 'b', False])
+    assert (result.get_next() == [3, 'c', None])
```
prrao87, line 20: Do we need to test a broader variety of PyArrow types like dates, 32-bit integers, floats/doubles, etc.? Or do the other tests for Polars already cover this implicitly?

mxwli: Pandas tests covers datatype scanning quite thoroughly

tools/python_api/src_cpp/py_connection.cpp
mxwli: I initially didn't do it because I thought it wouldn't scale well in the future, but it would probably work given that we already have much larger error messages. Perhaps we could tell the user the type of the variable that was found as well.

prrao87: Yup, I recommend something like `Found {} of type {} which is not supported for scanning`.

Issue #3698, opened 2024-06-24
Optimization: Rel batch insert memory usage improvement


ray6080:
### Description

`COPY` into a rel table can consume crazy amount of memory for some large num of rels dataset. For example, https://www.cs.cornell.edu/~arb/data/colisten-Spotify/ and also the one mentioned in #3660.

The memory usage should mainly come from the design of partitioning and materializing all rel tuples first before constructing node groups in parallel, though more detailed profiling into memory usage is needed.

To ease the peak memory usage, we should address/make several changes regarding to the rel batch insert pipeline:
- Make memory allocation of ColumnChunk go through MM, thus the memory usage can be bound by BM (#3660).
- Allow spilling to disk during partitioning, so we can sequentially read back each partition when constructing node groups in parallel. (Note that for the case a node group is very large, in the extreme case only one node group exists, we should optimize this further later with the segmenting design to chunk node group into more fine-grained storage units).
- Lightweight compression for in-memory materialized ColumnChunks. @benjaminwinger can fill more details here.

Issue #3660, opened 2024-06-17
Feature: make copy respect buffer pool size setting 


min-mwei:
### API

Other

### Description

See some details in the discussion
[(https://discord.com/channels/1196510116388806837/1196510116871159909/1252368122556448779)]

Issue #3459, opened 2024-05-07
Hang during multi-copy


benjaminwinger:
After a failed copy (due to duplicate primary keys), I've been encountering a hang. Sometimes it triggers an IOException from reading off the end of the metadata file, which appears to be caused by the `numExistingNodeGroups` in `NodeBatchInsert::initGlobalStateInternal` being much higher than it should be (I think as many as if the first copy had succeeded; I'd reproduced this with just a single tuple in the initial copy, and it went from 1 existing node group in the second failed copy, to 456 in the third one which threw the `IOException`). I'm not sure exactly where it's hanging though; interrupting the process in gdb just interrupts one thread in the task scheduler waiting on other threads to finish.

Initially reproduced in a custom branch, but reproduce-able on master (commit 68b59367dd99e4).

To reproduce: run the same COPY statement three times in the same process (first succeeds, second fails, third hangs; I've only actually tried this in the shell).

benjaminwinger:
Yes, I've tried and it seems to be working now.

PR #3743, opened 2024-07-03
Track ColumnChunk allocations through the BufferManager

benjaminwinger:
Implements #3660 (also see #3698)

I've merged the MemoryAllocator and MemoryManager; it wasn't obvious the purpose of the distinction (though it's occurred to me that we could have separate allocators for stuff using malloc like the ColumnChunks and our own allocator). Mainly I just wanted to avoid having to store separate references to the MemoryManager everywhere. The MemoryManager seemed better suited than the memory allocator to pass around to initialize column chunk buffers, however there wasn't an way of getting a MemoryManager from an existing buffer (just a MemoryAllocator). While I could have changed that, merging MemoryManager and MemoryAllocator seemed simpler.

Note that it was necessary to raise the buffer pool size used for the tests; that should be able to be lowered again once spilling to disk is implemented as part of testing that behaviour.

## Issues
I was having some issues with hanging instead of throwing exceptions when doing large rel copies with a restricted buffer pool size (looking at the behaviour when there is not sufficient memory). 
As far as I can tell, sometimes, with a large buffer pool of ~2000-3000 MB the hanging seems to be it continuously swapping out pages in the hash index. It may be that it will eventually throw an exception, but there sometimes is enough free memory for it to almost continuously try to do a small amount of work.

What's more concerning is that restricting the buffer pool size to the minimum 64MB on the same tests it will hang, and running it with backtraces will show a backtrace from a buffer manager exception (when looking up in the hash index; a couple of times it was the assert in freeUsedMemory), but that exception isn't stopping execution and it continues running with high CPU usage (possibly related to #3459).

I suspect that spilling to disk (#3698) will more or less solve these issues, as we generally shouldn't need to operate in such a memory restricted environment, and with spilling to disk ideally won't need to actually fail unless both disk and memory are full, but I'm concerned that we don't seem to be failing reliably in those circumstances.


src/storage/store/list_chunk_data.cpp
```diff
@@ -112,9 +116,13 @@ void ListChunkData::resetToEmpty() {
     ColumnChunkData::resetToEmpty();
     sizeColumnChunk->resetToEmpty();
     offsetColumnChunk->resetToEmpty();
-    dataColumnChunk =
-        ColumnChunkFactory::createColumnChunkData(ListType::getChildType(this->dataType).copy(),
-            enableCompression, 0 /* capacity */, ResidencyState::IN_MEMORY);
+    // TODO(bmwinger): can this just be made resetToEmpty?
```
ray6080, line 119: I think so.

src/storage/store/rel_table.cpp
```diff
@@ -7,6 +7,9 @@
 #include "storage/storage_manager.h"
 #include "storage/store/node_table.h"
 #include "storage/store/rel_table_data.h"
+#include "transaction/transaction.h"
+#include <main/client_context.h>
+#include <storage/store/node_table.h>
```
ray6080, line 12:
```suggestion
#include "main/client_context.h"
#include "storage/store/node_table.h"
```

ray6080 (approved):
Is the spilling gonna be in another PR?

src/storage/store/list_chunk_data.cpp
benjaminwinger: I initially wasn't sure as it was added in #2706 and I thought there might have been some subtle issue with resetToEmpty that caused a bug.
But actually the behaviour replaced in #2706 was calling `unique_ptr::reset` (where the default resets to nullptr), not `resetToEmpty`; so this is probably fine.

PR #3745, opened 2024-07-03
MVCC Changes Port ADD COLUMN

yiyun-sj:


src/storage/store/chunked_node_group.cpp
```diff
@@ -243,6 +243,16 @@ bool ChunkedNodeGroup::delete_(const Transaction* transaction, row_idx_t rowIdxI
     return versionInfo->delete_(transaction, rowIdxInChunk);
 }
 
+void ChunkedNodeGroup::addColumn(Transaction* transaction, TableAddColumnState& addColumnState,
+    bool enableCompression) {
+    auto numRows = getNumRows();
+    auto chunkData = 
+        ColumnChunkFactory::createColumnChunkData(addColumnState.property.getDataType().copy(),
+            enableCompression, capacity, ResidencyState::IN_MEMORY);
```
ray6080, line 251: Should also handle the case that the chunked node group is persistent, and we should also flush the appended chunkData to disk.

src/storage/store/chunked_node_group.cpp
```diff
@@ -243,6 +243,16 @@ bool ChunkedNodeGroup::delete_(const Transaction* transaction, row_idx_t rowIdxI
     return versionInfo->delete_(transaction, rowIdxInChunk);
 }
 
+void ChunkedNodeGroup::addColumn(Transaction* transaction, TableAddColumnState& addColumnState,
+    bool enableCompression) {
+    auto numRows = getNumRows();
+    auto chunkData =
```
ray6080, line 249: I think you can directly construct a ColumnChunk here, which will create the ColumnChunkData for you.

src/storage/store/node_table.cpp
```diff
@@ -189,17 +189,12 @@ bool NodeTable::delete_(Transaction* transaction, TableDeleteState& deleteState)
...
-    // defaultEvaluator);
-    // TODO(Guodong): addColumn is not going through localStorage design for now. So it needs to add
-    // tableID into the wal's updated table set separately, as it won't trigger prepareCommit.
-    // wal->addToUpdatedTables(tableID);
+void NodeTable::addColumn(Transaction* transaction, TableAddColumnState& addColumnState) {
+    auto& property = addColumnState.property;
+    KU_ASSERT(property.getColumnID() == columns.size());
+    columns.push_back(ColumnFactory::createColumn(property.getName(),
+        property.getDataType().copy(), dataFH, bufferManager, wal, enableCompression));
+    nodeGroups->addColumn(transaction, addColumnState);
```
ray6080, line 197: Should also handle localNodeTable, which should add a new column too.

src/expression_evaluator/literal_evaluator.cpp
```diff
@@ -10,6 +10,7 @@ namespace kuzu {
 namespace evaluator {
 
 void LiteralExpressionEvaluator::evaluate() {
+    resultVector->setState(flatState);
```
andyfengHKU, line 13: Why is this needed if u already set it in `resolveResultVector` l36?

src/include/storage/store/table.h
```diff
@@ -113,6 +113,25 @@ struct TableDeleteState {
...
+struct TableAddColumnState {
+    const catalog::Property& property;
+    evaluator::ExpressionEvaluator& defaultEvaluator;
+
+    TableAddColumnState(const catalog::Property& property, 
+        evaluator::ExpressionEvaluator& defaultEvaluator)
+        : property{property}, defaultEvaluator{defaultEvaluator} {}
+    virtual ~TableAddColumnState() = default;
+
+    // template<typename T>
```
ray6080, line 125: Can u clean these up?

src/expression_evaluator/literal_evaluator.cpp
yiyun-sj: when we populateDefaultVal in storage during add column, we call multiple evaluates but we init only once. For example, if u evaluate 2048, 2048, 1, then the last one will fail since the state remains the 2048 selection size

src/storage/store/column_chunk_data.cpp
```diff
@@ -285,26 +285,6 @@ ColumnChunkMetadata ColumnChunkData::getMetadataToFlush() const {
     return getMetadataFunction(buffer.get(), bufferSize, capacity, numValues, minValue, maxValue);
 }
 
-template<ResidencyState SCAN_RESIDENCY_STATE>
```
ray6080, line 288: Can u revert the changes here?

ray6080 (approved):
Thanks!

Issue #3664, opened 2024-06-18
Feature: Operator Printing


andyfengHKU:
### Description

As we approach the benchmark stage, plan printing is being used much more frequent. We need to make the following change to plan printing.

### Stage 1: Print static information

The first step is to print operator static information correctly and thoroughly, e.g. what table is being scanned and what columns are being scanned ...

A complete list is as follow

- [x] SimpleAggregate
    -  aggregate expressions 
- [x] SimpleAggregateScan
    -  N/A 
- [x] HashAggregate
    -  hash keys, hash payloads, aggregate expressions
- [x] HashAggregateScan
    - N/A
- [x] Alter
    - alter information, e.g. rename, add column, ...  
- [x] Attach
    - database name 
- [x] CopyFrom
    -  table name and source, e.g. file name, subquery, ...
- [x] CopyTo
    -  table name to file name
- [x] CreateMacro
    - macro name  
- [x] CreateSequence
    - sequence name 
- [x] CreateTable
    - table name and possible config 
- [x] CreateType
    - type name
- [x] CrossProduct
    - N/A
- [x] Delete
    - expressions and possible config
- [x] Detach
    - database name
- [x] Distinct
    - expressions 
- [x] DropSequence
    - sequence name
- [x] DropTable
    - table name
- [x] DummyScan
    - N/A
- [x] EmptyResult
    - N/A
- [x] Explain
    - Profile/Explain
- [x] ExpressionsScan
    - expressions
- [x] Extension
    - action and extension name
- [x] ExportDatabase
    - N/A
- [ ] Flatten
    - ideally we should print one expression to indicate which data chunk are we flattening
- [x] Filter
    - expression name 
- [x] GDSCall
    - algorithm name 
- [x] HashJoinBuild
    - key & payloads 
- [x] HashJoinProbe
    - key 
- [x] ImportDatabase
    - N/A 
- [x] IndexLookup
    - index name (maybe table name)
- [x] Insert
    - expressions & action (e.g. ON CONFLICT DO NOTHING)
- [x] IntersectBuild
    - same as hash join build
- [x] Intersect
    - same as hash join probe 
- [x] InstallExtension
    - extension name
- [x] Limit
    - number
- [x] LoadExtension
    - extension name 
- [x] Merge
    - the pattern to merge
- [x] MultplicityReducer
    - N/A 
- [x] OffsetScanNodeTable
     - N/A 
- [x] Partitioner
     - ask @ray6080  
[...]

PR #3781, opened 2024-07-08
Completed UseDatabase and TableFunctionCall Operators

HamzaAmmarPython:
# Description

Completes the UseDatabase and TableFunctionCall Operators from #3664 part 1

UseDatabase:
image

TableFunctionCall:
image

# Contributor agreement

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -64,3 +64,7 @@ scripts/generate-cpp-docs/c/docs
 # Generated datasets
 dataset/binary-demo
 dataset/databases/tinysnb
+
+#Database
+newDatabase
+history.txt
--- a/.gitignore
+++ b/.gitignore
@@ -64,7 +64,3 @@ scripts/generate-cpp-docs/c/docs
 # Generated datasets
 dataset/binary-demo
 dataset/databases/tinysnb
-
-#Database
-newDatabase
-history.txt
--- a/src/include/processor/operator/simple/use_database.h
+++ b/src/include/processor/operator/simple/use_database.h
@@ -10,14 +10,16 @@ struct UseDatabasePrintInfo final : OPPrintInfo{
 
     UseDatabasePrintInfo(std::string dbName)
         : dbName(std::move(dbName)) {}
-    UseDatabasePrintInfo(const UseDatabasePrintInfo& other)
-        : OPPrintInfo(other), dbName(other.dbName) {}
     
     std::string toString() const override;
 
     std::unique_ptr<OPPrintInfo> copy() const override{
-        return std::make_unique<UseDatabasePrintInfo>(*this);
+        return std::unique_ptr<UseDatabasePrintInfo>(new UseDatabasePrintInfo(*this));
     }
+
+    private:
+        UseDatabasePrintInfo(const UseDatabasePrintInfo& other)
+            : OPPrintInfo(other), dbName(other.dbName) {}
 };
 
 class UseDatabase final : public Simple {
--- a/src/include/processor/operator/simple/use_database.h
+++ b/src/include/processor/operator/simple/use_database.h
@@ -8,7 +8,7 @@ namespace processor {
 struct UseDatabasePrintInfo final : OPPrintInfo {
     std::string dbName;
 
-    UseDatabasePrintInfo(std::string dbName) : dbName(std::move(dbName)) {}
+    explicit UseDatabasePrintInfo(std::string dbName) : dbName(std::move(dbName)) {}
 
     std::string toString() const override;
 
--- a/src/include/processor/operator/simple/use_database.h
+++ b/src/include/processor/operator/simple/use_database.h
@@ -5,21 +5,20 @@
 namespace kuzu {
 namespace processor {
 
-struct UseDatabasePrintInfo final : OPPrintInfo{
+struct UseDatabasePrintInfo final : OPPrintInfo {
     std::string dbName;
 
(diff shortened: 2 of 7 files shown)


.gitignore
```diff
@@ -63,4 +63,8 @@ scripts/generate-cpp-docs/c/docs
 
 # Generated datasets
 dataset/binary-demo
-dataset/databases/tinysnb
\ No newline at end of file
+dataset/databases/tinysnb
+
+#Database
+newDatabase
+history.txt
```
MSebanc, line 70: Let's remove these changes to .gitignore. If you are testing via the shell, run the shell from outside your kuzu folder and the files won't show up in git.

src/include/processor/operator/simple/use_database.h
```diff
@@ -5,6 +5,21 @@
 namespace kuzu {
 namespace processor {
 
+struct UseDatabasePrintInfo final : OPPrintInfo{
+    std::string dbName;
+
+    UseDatabasePrintInfo(std::string dbName)
+        : dbName(std::move(dbName)) {}
+    UseDatabasePrintInfo(const UseDatabasePrintInfo& other)
```
MSebanc, line 13: Since the copy constructor doesn't need to be public, we have changed it to be private. Here is an example of how to make it private. 
https://github.com/kuzudb/kuzu/blob/bfb2480ffacbdf0520ba2846ba293c80f1f7486e/src/include/processor/operator/ddl/alter.h#L21-L29
Move the copy constructor to be private and switch `make_unique` in the copy function to `unique_ptr`.

src/include/processor/operator/table_function_call.h
```diff
@@ -7,6 +7,21 @@
 namespace kuzu {
 namespace processor {
 
+struct TableFunctionCallPrintInfo final : OPPrintInfo{
+    std::string funcName;
+
+    TableFunctionCallPrintInfo(std::string funcName)
+        : funcName(std::move(funcName)) {}
+    TableFunctionCallPrintInfo(const TableFunctionCallPrintInfo& other)
```
MSebanc, line 15: Ditto

src/processor/map/map_simple.cpp
```diff
@@ -31,7 +31,7 @@ static DataPos getOutputPos(const LogicalSimple* logicalSimple) {
 std::unique_ptr<PhysicalOperator> PlanMapper::mapUseDatabase(
     planner::LogicalOperator* logicalOperator) {
     auto useDatabase = logicalOperator->constPtrCast<LogicalUseDatabase>();
-    auto printInfo = std::make_unique<OPPrintInfo>(useDatabase->getExpressionsForPrinting());
+    auto printInfo = std::make_unique<UseDatabasePrintInfo>(useDatabase->getExpressionsForPrinting());
```
MSebanc, line 34: Eventually, later on in the issue, we plan to remove `getExpressionsForPrinting`. So, to make that later part simpler, find a way to pass the information needed for the print info without calling `getExpressionsForPrinting`.  There is most likely a 'getDBName()` function.

src/processor/map/map_table_function_call.cpp
```diff
@@ -24,7 +24,7 @@ std::unique_ptr<PhysicalOperator> PlanMapper::mapTableFunctionCall(
     info.outputType =
         outPosV.empty() ? TableScanOutputType::EMPTY : TableScanOutputType::SINGLE_DATA_CHUNK;
     auto sharedState = std::make_shared<TableFunctionCallSharedState>();
-    auto printInfo = std::make_unique<OPPrintInfo>(call.getExpressionsForPrinting());
+    auto printInfo = std::make_unique<TableFunctionCallPrintInfo>(call.getExpressionsForPrinting());
```
MSebanc, line 27: Ditto

MSebanc (changes requested):
Could you add some example printings in the PR description so we can see how the changes look.

MSebanc (changes requested):
Could you run clang format to fix the formatting issues on your pr. https://github.com/kuzudb/kuzu/blob/master/scripts/run-clang-format.py

We use `clang-format 18`, make sure you use the same version

src/processor/operator/simple/use_database.cpp
```diff
@@ -14,5 +14,11 @@ std::string UseDatabase::getOutputMsg() {
     return "Used database successfully.";
 }
 
+std::string UseDatabasePrintInfo::toString() const{
+    std::string result = "Database Name: ";
```
MSebanc, line 18: Let's change it to say `"Database: "` instead of `"Database Name: "` to match the other database physical operators

HamzaAmmarPython: How do I run the clang format? From the error message on the test it seems like it's trying to push to my repository but failing

MSebanc: You can run clang format locally using the script I linked. If you are having trouble using the script you can ask Chang on discord for help figuring out how to run the script 👍

MSebanc:
Don't forget to paste an example of TableFunctionCall 👍

HamzaAmmarPython:
Everything is done but the tests still fail, seems it's the extension tests though, what should I do about that?

MSebanc (approved):
LGTM! The extension tests will fail since they don't run on pr's from forks so don't worry about those, you didn't change anything that would affect the extensions.

PR #3791, opened 2024-07-10
Fix creation of new overflow slots when reserving space in the hash index

benjaminwinger:
When splitting slots to reserve space in the hash index it's possible (but unlikely) that the newly added slot will need an overflow slot, but adding such overflow slots was broken and keys put into them would be lost since the previous slot was not being set to point to them.
Unfortunately this is a quite rare occurrence and not easy to test.

*Should be testable using a bunch of keys (>1 slot) that collide for the hash index size used and rehash into a new slot. E.g. keys which hash into slot 0 with the default hash index size of 16 slots, and then all hash to slot 16 once enough other keys are added, forcing multiple slots to be created when slot 0 is split. My initial attempt was not failing on the master branch, so I suspect I made a mistake somewhere (Edit: initially forgot to make sure they all went into the same index)*.


src/storage/index/hash_index.cpp
```diff
@@ -131,6 +131,14 @@ void HashIndex<T>::splitSlots(HashIndexHeader& header, slot_id_t numSlotsToSplit
     // So instead buffer new overflow slots here and append them at the end
     std::vector<Slot<T>> newOverflowSlots;
 
+    auto getNextOvfSlot = [&](slot_id_t nextOvfSlotId) {
+        if (nextOvfSlotId >= oSlots->getNumElements()) {
```
ray6080, line 135: Better assert here `nextOvfSlotId - oSlots->getNumElements()` is within the range of `newOverflowSlots.size()`

ray6080 (approved):

ray6080:
@benjaminwinger , I'm merging this now so it can be in our next dev build.

Issue #3717, opened 2024-06-27
GDS Features and Optimizations


semihsalihoglu-uw:
### Description

I'm writing this to keep track of optimizations and new features we should be adding to Kuzu's GDS component. There is a lot of pieces to this, but this should capture the general scope
- [ ] Graph Interface Improvements:
   - [ ] In-memory graph instance based on CSR
   - [ ] Improve on-disk to scan a set of vertex's neighbors
   - [ ] Change Graph to return iterators instead of vectors when scanning lists
- [ ] Experiment with the performance improvement of templating FrontierUpdateFn instead of virtual classes for VertexUpdate functions. This means templating functions that takes FrontierUpdateFn as input, specifically those in GDSUtils.
- [ ] Fix https://github.com/kuzudb/kuzu/issues/3716
- [ ] Shortest Paths Optimizations
   - [ ] Materializing results in parallel
   - [ ] Write Paths: with BFS-tree-based compact implementation (using EdgeUpdateLists)
   - [ ] Weighted Shortest Paths
   - [ ] If possible, skip writing compact BFS-tree-based storage to first to a FactorizedTable for further operators and then to ValueVectors. Instead write it as a compact data type directly to FactorizedTable and have a separate flatten operator that directly writes to ValueVectors.
   - [ ] Bidirectional BFS
- [ ] Rewrite existing PageRank and WCC
- [ ] Implement other algorithms: Identify a suite of algorithms and implement.
- SHORTEST: Optimizations:
	- Make masks memory manager backed
	- Disk-based vs in-mem graphs
	- Single directional vs double-directional graphs
- SHORTEST: Write paths.
- Further optimizations: 
	- Optimize disk-based scans
	- Sparse-dense switch with a configurable option
	- Remove the 256 limit
	- Optimize for single-label graphs
	- Optimize making update function templates instead of virtual calls
	- Parallelize writing to factorized table

PR #3792, opened 2024-07-10
Making up for the lost worker thread in GDSTasks

semihsalihoglu-uw:
# Description

Addresses the "- [ ] Spawn new threads when executing a GDSAlgorithm in task scheduler." item of: https://github.com/kuzudb/kuzu/issues/3717. This is a simple change that adds a new boolean `launchNewWorkerThread` argument to TaskScheduler::scheduleTaskAndWaitOrError function to spawn a new thread or not. By default this argument is false. GDSUtils sets it to true when launching GDSTasks.

I can't think of tests for this but wrote the following program to test that the change behaves as expected in several ways. 
```
int main(int argc, char *argv[]) {
    usleep(10000000); // 1st Usleep: expected num threads = 1

    auto kuzuDBDir = "/kuzu-dbs/test";
    auto numThreads = argc == 3 && std::atoi(argv[1]) > 0 ? std::atoi(argv[1]) : 1;
    auto distance = argc == 3 && std::atoi(argv[2]) > 0 ? std::atoi(argv[2]) : 4;
    std::cout << "numThreads: " << numThreads << std::endl;
    filesystem::remove_all(kuzuDBDir);
    auto systemConfig = SystemConfig(-1u, numThreads, true, false, -1u);
    auto database = std::make_unique<Database>(kuzuDBDir, systemConfig);
    auto connection = std::make_unique<Connection>(database.get());

    usleep(10000000);  // 2nd usleep: expected num threads = numThreads +1. 
    
    // Create schema.
    connection->query("CREATE NODE TABLE Person(ID INT64, PRIMARY KEY(ID));");
    connection->query("CREATE REL TABLE Knows(FROM Person TO Person);");

    // Create nodes.
    connection->query("COPY Person FROM /test_shortest_paths/6-hop-k-10-person.csv';");
    connection->query("COPY Knows FROM '/test_shortest_paths/6-hop-k-10-knows.csv' (DELIM=\" \");");

    std::chrono::steady_clock::time_point begin = std::chrono::steady_clock::now();
    // The query starts executing and takes ~15s. During this time we expect to see numThreads as numThreads + 2.
    auto result = connection->query("PROJECT GRAPH PK (person, knows)   MATCH (a:Person) WHERE a.ID = 1   CALL shortest_paths(PK, a, " +
                                        std::to_string(distance) + ",  true) RETURN length, count(*);");
    std::chrono::steady_clock::time_point end = std::chrono::steady_clock::now();

    // Print query result.
    std::cout << result->toString();
    std::cout << "Time difference = " << std::chrono::duration_cast<std::chrono::milliseconds>(end - begin).count() << "[ms]" << std::endl;
    usleep(10000000); // Here, num threads should go back to numThreads + 1.
}
```
First with several 10s usleep calls I tested that when I launch the database with k threads, i.e., pass the first argument to the program as k, then we see the following behavior. First usleep is when the program starts and number of threads should be 1. I check this in Activity Monitor and find this program (sp_tester was its name). Then there is a 2nd usleep after we create the database object with k threads. Here the number threads should be k + 1. Then once the query starts executing, which takes ~15s, number of threads must be k+2 because the task scheduler must have launched a new thread. Finally, after the query finishes, there is a 3rd usleep, during which time the number of threads must come down to k+1. I tried with multiple k and observed this behavior. 

Second, I tested that the program doesn't halt when k = 1, where as it halts when k = 1 in the master branch.

Third, I ran some runtime experiments both on master and on my branch to see that the runtime behavior with k = i on my branch is similar to the runtime behavior for k = i + 1 on master. Here were the results, which demonstrate the expected behavior.
```
This branch           Master
1 Thread: 14009    Hangs
2 Thread: 7510     13586
3 Thread: 4348     7170
4 Thread: 3925     5317
7 Thread: 3011     3276
8 Thread: 3926     3420
9 Thread: ----     3317
```

# Contributor agreement

- [ x] I have read and agree to the [Contributor Agreement](https://github.com/kuzudb/kuzu/blob/master/CLA.md).


semihsalihoglu-uw:
> Add a test to reproduce the hanging problem. This should be doable by
> 
> ```
> CALL THREADS=1;
> // run gds
> ```

There is hanging problem any more. You mean add a test that ensures there is no hanging?

semihsalihoglu-uw:
> > Add a test to reproduce the hanging problem. This should be doable by
> > ```
> > CALL THREADS=1;
> > // run gds
> > ```
> 
> There is no hanging problem any more. You mean add a test that ensures there is no hanging?

andyfengHKU:
> > > Add a test to reproduce the hanging problem. This should be doable by
> > > ```
> > > CALL THREADS=1;
> > > // run gds
> > > ```
> > 
> > 
> > There is no hanging problem any more. You mean add a test that ensures there is no hanging?

Yes. Add a test that will hang before u make this change. And succeed after.

andyfengHKU (approved):

PR #3795, opened 2024-07-10
Mvcc data chunk scan

yiyun-sj:


src/storage/store/list_chunk_data.cpp
```diff
@@ -157,6 +157,25 @@ void ListChunkData::appendNullList() {
...
 }
 
+void ListChunkData::scan(ValueVector& output, offset_t offset, length_t length) const {
+    KU_ASSERT(offset + length <= numValues);
+    if (nullData) {
+        output.setNullFromBits(nullData->getNullData()->getNullMask().getData(), offset,
+            0 /*dstOffset*/, length);
+    }
+    auto dataOffset = getListStartOffset(offset);
+    auto dataSize = 0ul;
```
ray6080, line 167: I think there is one missing part: list start offsets can be un-sorted. this is due to updating a list in the middle of the chunk (say the chunk has 100 values, dataColumnChunk has 1000 values. we update a list value at pos 10, we append new list data to the end of the dataColumnChunk, and update the offset at pos 10 pointing to new pos at dataColumnChunk, which would be 1000). If unsorted, we should fall back to read the list value one at a time. The logic should roughly be same as here  https://github.com/kuzudb/kuzu/blob/96e9f94b8a8fc398974084b2e91afdd0fca3ff27/src/storage/store/list_column.cpp#L238

src/storage/store/string_chunk_data.cpp
```diff
@@ -88,6 +88,21 @@ void StringChunkData::append(ColumnChunkData* other, offset_t startPosInOtherChu
     }
 }
 
+void StringChunkData::scan(ValueVector& output, offset_t offset, length_t length) const {
+    KU_ASSERT(offset + length <= numValues);
+    if (nullData) {
```
ray6080, line 93: You can assert that nullData always exists for StringChunkData.

src/storage/store/string_chunk_data.cpp
```diff
@@ -88,6 +88,21 @@ void StringChunkData::append(ColumnChunkData* other, offset_t startPosInOtherChu
...
+        output.setNullFromBits(nullData->getNullData()->getNullMask().getData(), offset,
+            0 /*dstOffset*/, length);
+    }
+    for (auto i = 0u; i < length; i++) {
```
ray6080, line 97: We can have a fast path when there is no nulls (`hasNoNullsGuarantee`), so we can avoid the `if` check inside a for loop.
See a reference here, though we don't need the check on `isUnfiltered()` here.
https://github.com/kuzudb/kuzu/blob/96e9f94b8a8fc398974084b2e91afdd0fca3ff27/src/include/function/unary_function_executor.h#L127

src/storage/store/column_chunk_data.cpp
```diff
@@ -360,10 +361,11 @@ void ColumnChunkData::initializeScanState(ChunkState& state) const {
 
 void ColumnChunkData::scan(ValueVector& output, offset_t offset, length_t length) const {
     KU_ASSERT(offset + length <= numValues);
-    // TODO(Guodong): Rework properly as scan.
-    for (auto i = 0u; i < length; i++) {
-        lookup(offset + i, output, i);
+    if (nullData) {
+        output.setNullFromBits(nullData->getNullData()->getNullMask().getData(), offset,
```
ray6080, line 365: Can we wrap this into NullColumnData::scan()?

ray6080 (approved):
Thanks!

Can we add some tests on the `scan` interface?
Generally it would be good if we have more tests coverage on the case we do a bunch of insertions (10K or larger depending on how long the test takes) and then scan over the table. Our existing test cases are very small and not cover enough data types.

PR #3805, opened 2024-07-11
Operator printing

HamzaAmmarPython:
# Description

Adds to #3664 (setProperty, Skip, StandaloneCall, TableFunctionCall, TopK, Transaction, OrderBy, UnionAllScan, Unwind)

NOTE: Most of these pictures do expand longer than shown, I just couldn't fit them into one screenshot
setProperty:
image

Skip:
image

StandaloneCall:
image

TableFunctionCall:
image

TopK:
image

Transaction:
image

OrderBy:
image

UnionAllScan:
image

Unwind:
image

# Contributor agreement


src/processor/operator/order_by/order_by.cpp
```diff
@@ -1,10 +1,20 @@
...
 
+#include "binder/expression/expression_util.h"
+
 using namespace kuzu::common;
 
 namespace kuzu {
 namespace processor {
 
+std::string OrderByPrintInfo::toString() const {
+    std::string result = "Keys: ";
```
MSebanc, line 11: Let have `Order By: ` instead of `Keys: `

src/processor/operator/order_by/order_by.cpp
```diff
@@ -1,10 +1,20 @@
...
+    result += binder::ExpressionUtil::toString(keys);
+    result += ", Payloads: ";
```
MSebanc, line 13: I think let's call this `Expressions: `

src/processor/operator/order_by/top_k.cpp
```diff
@@ -8,6 +9,16 @@ using namespace kuzu::common;
 namespace kuzu {
 namespace processor {
 
+std::string TopKPrintInfo::toString() const {
+    std::string result = "Keys: ";
```
MSebanc, line 13: Ditto

src/processor/operator/order_by/top_k.cpp
```diff
@@ -8,6 +9,16 @@ using namespace kuzu::common;
...
+    result += binder::ExpressionUtil::toString(keys);
+    result += ", Payloads: ";
```
MSebanc, line 15: Ditto

src/processor/operator/persistent/set.cpp
```diff
@@ -1,5 +1,9 @@
 #include "processor/operator/persistent/set.h"
 
+#include "binder/expression/expression_util.h"
+
+using namespace kuzu::binder;
```
MSebanc, line 5: I don't think this is needed

src/processor/operator/persistent/set.cpp
```diff
@@ -45,5 +49,10 @@ std::unique_ptr<PhysicalOperator> SetRelProperty::clone() {
         id, printInfo->copy());
 }
 
+std::string SetPropertyPrintInfo::toString() const {
+    std::string result = "Operation: ";
```
MSebanc, line 53: Let us call this `Properties: `

src/processor/operator/skip.cpp
```diff
@@ -7,6 +7,12 @@ void Skip::initLocalStateInternal(ResultSet* resultSet, ExecutionContext* /*cont
     dataChunkToSelect = resultSet->dataChunks[dataChunkToSelectPos];
 }
 
+std::string skipPrintInfo::toString() const {
+    std::string result = "Number: ";
```
MSebanc, line 11: Should be `Skip: `

MSebanc (changes requested):
Update the screenshots after you make these changes. Also, for consistency lets have every PrintInfo struct directly above the Physical Operator class.

src/processor/operator/order_by/top_k.cpp
HamzaAmmarPython: Is this a duplicate from the comment above?

src/processor/map/map_set.cpp
```diff
@@ -85,7 +86,12 @@ std::unique_ptr<PhysicalOperator> PlanMapper::mapSetNodeProperty(LogicalOperator
     for (auto& info : set->getInfos()) {
         executors.push_back(getNodeSetExecutor(info, *inSchema));
     }
-    auto printInfo = std::make_unique<OPPrintInfo>(set->getExpressionsForPrinting());
+    auto& info = set->getInfos();
+    std::string result = ExpressionUtil::toString(info[0].setItem);
```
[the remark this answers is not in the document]
HamzaAmmarPython: The problem with this is that info has a private type, only accessible through certain files. So I had to convert it to string in the map_set.cpp before I could bring it into the set.h and set.cpp files

src/processor/operator/order_by/top_k.cpp
MSebanc: Should be called Expressions: like the payload comment above

src/include/processor/operator/persistent/set.h
```diff
@@ -1,11 +1,28 @@
...
 
+#include "binder/query/updating_clause/bound_set_info.h"
 #include "processor/operator/physical_operator.h"
 #include "set_executor.h"
 
 namespace kuzu {
 namespace processor {
 
+struct SetPropertyPrintInfo final : OPPrintInfo {
+    std::string operation;
```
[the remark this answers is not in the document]
HamzaAmmarPython: It allows for two, it just looks kind of bad: 
image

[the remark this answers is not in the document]
MSebanc: Have it be `Properties: ` and have a comma between each property so it goes down to the next line

src/processor/map/map_set.cpp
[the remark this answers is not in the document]
MSebanc: Which part do you need that's private? You can send infos over to PrintInfo and would still be able to access setItem. Or, you can create a vector of expression_pair similar to the for loop you have right now and send that. It is better to do the actual string conversions in toString.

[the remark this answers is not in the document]
HamzaAmmarPython: sending info through printInfo requires the other side to have public access to the     std::vector<std::unique_ptr<RelSetExecutor>> type. However since it is private in set.h, I cannot access it. That's why I convert it first in map_set.cpp and then send it over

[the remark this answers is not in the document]
MSebanc: Oh yes I see the issue. 
https://github.com/kuzudb/kuzu/blob/433c8d303a1face982a62d4f69c0a51b5b6879e5/src/processor/map/map_index_scan_node.cpp#L27-L31

Try doing something similar to what I do in `map_index_scan_node` where you make your own vector with what you need and send that and do the string conversions in toString. That way you avoid the issue with `BoundSetPropertyInfo`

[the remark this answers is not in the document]
HamzaAmmarPython: Ok perfect, I will do that this weekend and hopefully have it done by monday

src/processor/map/map_set.cpp
```diff
@@ -85,7 +86,11 @@ std::unique_ptr<PhysicalOperator> PlanMapper::mapSetNodeProperty(LogicalOperator
     for (auto& info : set->getInfos()) {
         executors.push_back(getNodeSetExecutor(info, *inSchema));
     }
-    auto printInfo = std::make_unique<OPPrintInfo>(set->getExpressionsForPrinting());
+    std::vector<binder::expression_pair> expressions;
+    for (auto i = 0u; i < set->getInfos().size(); ++i) {
```
MSebanc, line 89: If we use a for loop like 
```
for (auto& info : set->getInfos) {
    expressions.push_back(info.setItem);
}
```
then we can avoid creating the getInfo(index) function

src/processor/map/map_set.cpp
```diff
@@ -140,7 +145,11 @@ std::unique_ptr<PhysicalOperator> PlanMapper::mapSetRelProperty(LogicalOperator*
     for (auto& info : set->getInfos()) {
         executors.push_back(getRelSetExecutor(info, *inSchema));
     }
-    auto printInfo = std::make_unique<OPPrintInfo>(set->getExpressionsForPrinting());
+    std::vector<binder::expression_pair> expressions;
+    for (auto i = 0u; i < set->getInfos().size(); ++i) {
```
MSebanc, line 148: ditto

src/processor/operator/order_by/top_k.cpp
```diff
@@ -8,6 +9,18 @@ using namespace kuzu::common;
 namespace kuzu {
 namespace processor {
 
+std::string TopKPrintInfo::toString() const {
+    std::string result = "Keys: ";
```
MSebanc, line 13: Let's call this `Order By: `

src/processor/operator/persistent/set.cpp
MSebanc, line 5: Remove this namespace since you're still using `binder::` in this file

src/processor/operator/order_by/top_k.cpp
```diff
@@ -8,6 +9,18 @@ using namespace kuzu::common;
...
+    result += binder::ExpressionUtil::toString(keys);
+    result += ", Expressions: ";
+    result += binder::ExpressionUtil::toString(payloads);
+    result += ", Skip Value: ";
```
MSebanc, line 17: Could you double check that if you don't include a skip in the query that it shows up as 0

MSebanc (approved):
LGTM, just a couple more changes are needed and then you can merge it in when the tests all pass 👍👍

src/processor/operator/order_by/top_k.cpp
HamzaAmmarPython: I Updated the pictures, is it correct?

MSebanc: Yeah works correctly! But come to think of it can we remove value and number and have it say `Skip: ` and `Limit: ` instead of `Skip Value: ` and `Limit Number: `

PR #3809, opened 2024-07-11
Mvcc Detach Delete

yiyun-sj:


src/include/storage/store/table.h
```diff
@@ -141,6 +141,7 @@ class Table {
 
     common::TableType getTableType() const { return tableType; }
     common::table_id_t getTableID() const { return tableID; }
+    std::string getTableName() const { return tableName; }
```
ray6080, line 144: Is this used anywhere?

src/storage/store/rel_table.cpp
```diff
@@ -164,63 +155,72 @@ bool RelTable::delete_(Transaction* transaction, TableDeleteState& deleteState)
...
+    RelTableDeleteState* deleteState) {
+    KU_ASSERT(deleteState->srcNodeIDVector.state->getSelVector().getSelSize() == 1);
     const auto tableData =
         direction == RelDataDirection::FWD ? fwdRelTableData.get() : bwdRelTableData.get();
     const auto reverseTableData =
         direction == RelDataDirection::FWD ? bwdRelTableData.get() : fwdRelTableData.get();
     std::vector<column_id_t> columnsToScan = {NBR_ID_COLUMN_ID, REL_ID_COLUMN_ID};
-    const auto relIDVectors = std::vector<ValueVector*>{deleteState->dstNodeIDVector.get(),
-        deleteState->relIDVector.get()};
+    const auto relIDVectors = std::vector<ValueVector*>{&deleteState->dstNodeIDVector,
```
ray6080, line 166: `relIDVectors` doesn't seem to be an accurate name. I think you can directly initialize vectors into `relReadState->outputVectors`?

ray6080 (approved):
Thanks!

src/include/storage/store/table.h
yiyun-sj: yup, to get tableName for localTable during exception message

src/storage/store/rel_table.cpp
yiyun-sj: sure

PR #3813, opened 2024-07-12
Optimize small rel table lookups

benjaminwinger:
Adds a random access version of `OnDiskGraph::scanFwd` which reads only the CSR header information necessary for the requested nodes.

I profiled `RelScanTestTwitter.TestRandomAccessScanMany` and compared it to a version using the other scan function, and `scanFwdRandom` was ~50 times faster `scanFwd` (I left the test since it covers some edge cases such as >2048 rels for a single node).

The underlying code in RelTableData still scans a single node at a time, and I think making it scan multiple could improve performance by an additional 50% or so when scanning multiple nodes within a single node group and there are few edges for the nodes scanned (currently initializeScanState is still taking >2x the time of the actual scan, and close to half of that is initialization which would be redundant for nodes in the same group). However that will be more complicated and the relative performance improvement is less significant than what this already achieves.

I've added a couple of simple unit tests that test OnDiskGraph directly, as none of the algorithms make use of the new function yet.


src/graph/on_disk_graph.cpp
```diff
@@ -135,22 +180,25 @@ std::vector<nodeID_t> OnDiskGraph::scanBwd(nodeID_t nodeID, table_id_t relTableI
...
-    relTable->initializeScanState(context->getTx(), relTableScanState);
-    auto& dstSelVector = scanState.dstNodeIDVector->state->getSelVector();
-    while (relTableScanState.hasMoreToRead(context->getTx())) {
-        relTable->scan(context->getTx(), relTableScanState);
-        KU_ASSERT(dstSelVector.isUnfiltered());
-        for (auto i = 0u; i < dstSelVector.getSelSize(); ++i) {
-            auto nbrID = scanState.dstNodeIDVector->getValue<nodeID_t>(i);
-            nbrNodeIDs.push_back(nbrID);
+    // TODO(bmwinger): Partition node IDs by nodeGroupIdx and pass all node IDs together to the scan
+    for (size_t i = 0; i < nodeIDs.size(); i++) {
```
ray6080, line 190: I'm not sure if this is what GDS actually wants. @semihsalihoglu-uw should take a look at the change here. Don't you need to track which nbrIDs belong to which srcNodeID? (this is a question to @semihsalihoglu-uw )

ray6080 (approved):
The change on the storage side looks good to me.

src/graph/on_disk_graph.cpp
semihsalihoglu-uw: This function currently will only be called to scan 1 node. We don't have any code that calls this function to scan multiple nodes.

Soon what I will need however is an implementation where I can write programs that look like this:
```
GDSAlgorithm pseudocode:
vector<nodeIDs> nodeIDsToScan;
// populate nodeIDsToScan with up to 64 or 128 nodes, whose adjacency lists will be scanned. These will be 
// sorted but not necessarily in the same node group, though because this is a small number it is more likely 
// that they will be in the same node group.
auto someStateInfo = onDiskGraph->prepareToScan(nodeIDsToScan);
for (nodeID_t nodeID : nodeIDs) {
    iterator<nodeID, relID> itr = onDiskGraph->scan(nodeID, someStateInfo);
}
```
That's the ideal interface I would like. I don't want onDiskGraph to return a vector. I want it to return an iterator. The reason is that for the InMemGraph, I think we can just avoid the copies of the lists. We might even be able to do something similar in OnDiskGraph, i.e., avoid copies from BM to some temporary place, from which the GDSAlgorithms read; instead the iterator might be able to directly read and decompress from the BM though that might be too optimistic. Regardless, I still want the ultimate function to read 1 node at a time, but I can pass the information that I'm about to scan 128 nodes. That's why I'm imagining a "prepareScan"-like function.

benjaminwinger: Okay. I think the iterator interface will be a little complicated. I think for now it would probably be best to change this to be like the non-random functions (single node returning a vector) and then we could switch to the iterator interface in a separate PR.

> We might even be able to do something similar in OnDiskGraph, i.e., avoid copies from BM to some temporary place, from which the GDSAlgorithms read; instead the iterator might be able to directly read and decompress from the BM though that might be too optimistic. 

I took a quick look at implementing an iterator over the results, and I think that even if it's too complicated to read directly from the BM, we still avoid the extra copies/allocations for the std::vector and have an iterator that scans into the ValueVector inside the OnDiskGraphScanState as we iterate and points directly to that data. When I was profiling this the cost of creating the vector was about 1/3 of the new runtime, so just removing the std::vector should have a noticeable benefit.

PR #3819, opened 2024-07-14
Adding basic single shortest path that finds paths

semihsalihoglu-uw:
# Description

Adds a simple new GDSAlgorithm that reuses most of the shortest path logic that uses the PathLengthsFrontier and ShortestPathsAlgorithm. The main changes are `SPSinglePathsFrontierCompute`, `SinglePaths`, and `SinglePathsFrontiers`. Currently only writes internal node IDs as a new field called `pathNodeIDs`. Once Graph interface also scans rel IDs, can keep track of and return relIDs as well.

Overall this PR provides the infrastructure for the "Single Shortest Paths" semantics that returns the actual paths. So this code would eventually be triggered in queries that use the `SHORTEST` keyword in recursive relationship patterns and returns the variables bound to those patterns, e.g.:
```
MATCH (a:Person)-[r:Knows* SHORTEST]->(b:Person)
RETURN r
```

# Contributor agreement


test/test_files/function/gds/shortest_paths_large.test
```diff
@@ -15,27 +15,56 @@
...
 
+# The first output tuples 11 comes from the fact that we also output 0-length paths, i.e., the from from source node
+# to source node. Both for that path and the 10 paths of length 1, we output an empty pathNodeIDs, since there are no
+# intermediate nodes.
+-CASE SingleSPPathsLargeSingleLabel
+-STATEMENT PROJECT GRAPH PK (person1, knows11)
+           MATCH (a:person1) WHERE a.ID = 1
+           CALL single_sp_paths(PK, a, 4, true)
+           RETURN size(pathNodeIDs), count(*);
+-PARALLELISM 8
```
andyfengHKU, line 26: Remove -PARALLELISM from this file. Parallelism is enabled by default. And we should deprecate this configuration in test file.

src/function/gds/shortest_paths.cpp
```diff
@@ -19,6 +19,50 @@ using namespace kuzu::graph;
...
+ * "dense" in the sense that it keeps space for one "backwards edge" for each possible node that can
+ * be in the destination. Supports multi-label nodes.
+ */
+class SinglePaths {
+public:
+    explicit SinglePaths(
+        std::vector<std::tuple<common::table_id_t, uint64_t>> nodeTableIDAndNumNodes,
+        MemoryManager* mm) {
+        for (const auto& [tableID, numNodes] : nodeTableIDAndNumNodes) {
+            lastEdges.insert({tableID, mm->allocateBuffer(false, numNodes * sizeof(nodeID_t))});
```
andyfengHKU, line 41: this is fine since we are still prototyping. Though I doubt we will have good performance if we always allocate towards maxNumNodes for each source. In my mind a dynamic data structure is more robust than this static array.

src/function/gds/shortest_paths.cpp
```diff
@@ -310,12 +426,23 @@ class ShortestPathsAlgorithm final : public GDSAlgorithm {
...
-            sharedState->graph.get(), spFrontierCompute,
+        auto sourceState = ShortestPathsSourceState(sharedState->graph.get(), sourceNodeID,
+            outputType, executionContext->clientContext->getMemoryManager());
+
+        std::unique_ptr<Frontiers> frontiers;
+        std::unique_ptr<FrontierCompute> frontierCompute;
+        if (outputType == RecJoinOutputType::LENGTHS) {
+            frontiers = std::make_unique<PathLengthsFrontiers>(sourceState.pathLengths.get());
+            frontierCompute = std::make_unique<SPLengthsFrontierCompute>(
+                reinterpret_cast<PathLengthsFrontiers*>(frontiers.get()));
```
andyfengHKU, line 437: Avoid using reinterpret_cast as long as u can. Search for ptrCast in the code base and mimic the design.

reinterpret_cast is not safe as it will blindly cast without perform any type/memory check. Use reinterpret_cast only when the base type is unknown, e.g. when u declare a unit8_t* buffer.

src/function/gds/shortest_paths.cpp
```diff
@@ -310,12 +426,23 @@ class ShortestPathsAlgorithm final : public GDSAlgorithm {
...
+        } else {
+            frontiers = std::make_unique<SinglePathsFrontiers>(sourceState.pathLengths.get(),
+                sourceState.singlePaths.get());
+            frontierCompute = std::make_unique<SPSinglePathsFrontierCompute>(
+                reinterpret_cast<SinglePathsFrontiers*>(frontiers.get()));
```
andyfengHKU, line 442: ditto

andyfengHKU (approved):
Not a big change. LGTM.

test/test_files/function/gds/shortest_paths_large.test
semihsalihoglu-uw: Done, but I wouldn't deprecate this. We might want to test cases with 1 or very large number of threads.

src/function/gds/shortest_paths.cpp
semihsalihoglu-uw: I think dense structures will be more robust for most cases. I have a todo "Experiment with sparse-vs-dense implementation switching" here: https://github.com/kuzudb/kuzu/issues/3717, which I will get to at some point.

src/function/gds/shortest_paths.cpp
semihsalihoglu-uw: I can't put ptrCast to these functions directly because frontiers is of class Frontiers* and that's the base abstract class. I need to cast to the parent class PathLengthsFrontiers*. But I found a bit more verbose way to achieve the same thing without reinterpret cast.

src/function/gds/shortest_paths.cpp
semihsalihoglu-uw: Ditto.

PR #3822, opened 2024-07-15
Pass transaction in hash index

ray6080:
# Description

Replace a few places which directly use `TransactionType::WRITE` and `TransactionType::READ_ONLY` with transactions passed in.

### DIFF
--- a/src/include/storage/index/hash_index.h
+++ b/src/include/storage/index/hash_index.h
@@ -85,7 +85,8 @@ public:
     // - the key has been marked as deleted in the local storage, return false;
     // - the key is neither deleted nor found in the local storage, lookup in the persistent
     // storage.
-    bool lookupInternal(transaction::Transaction* transaction, Key key, common::offset_t& result) {
+    bool lookupInternal(const transaction::Transaction* transaction, Key key,
+        common::offset_t& result) {
         if (transaction->isReadOnly()) {
             return lookupInPersistentIndex(transaction, key, result);
         }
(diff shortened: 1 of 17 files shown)


src/include/storage/index/hash_index.h
```diff
@@ -150,9 +149,9 @@ class HashIndex final : public OnDiskHashIndex {
     inline BMFileHandle* getFileHandle() const { return fileHandle; }
 
 private:
-    bool lookupInPersistentIndex(transaction::TransactionType trxType, Key key,
+    bool lookupInPersistentIndex(transaction::Transaction* transaction, Key key,
         common::offset_t& result) {
-        auto& header = trxType == transaction::TransactionType::READ_ONLY ?
+        auto& header = transaction->getType() == transaction::TransactionType::READ_ONLY ?
```
benjaminwinger, line 154: It would be more concise and maybe clearer to use isReadOnly here

src/storage/storage_structure/disk_array.cpp
```diff
@@ -83,13 +83,13 @@ bool DiskArrayInternal::checkOutOfBoundAccess(TransactionType trxType, uint64_t
...
+void DiskArrayInternal::get(uint64_t idx, Transaction* transaction, std::span<std::byte> val) {
     std::shared_lock sLck{diskArraySharedMtx};
-    KU_ASSERT(checkOutOfBoundAccess(trxType, idx));
+    KU_ASSERT(checkOutOfBoundAccess(transaction->getType(), idx));
     auto apCursor = getAPIdxAndOffsetInAP(storageInfo, idx);
-    page_idx_t apPageIdx = getAPPageIdxNoLock(apCursor.pageIdx, trxType);
+    page_idx_t apPageIdx = getAPPageIdxNoLock(apCursor.pageIdx, transaction->getType());
     auto& bmFileHandle = fileHandle;
-    if (trxType == TransactionType::READ_ONLY || !hasTransactionalUpdates ||
+    if (transaction->getType() == TransactionType::READ_ONLY || !hasTransactionalUpdates ||
```
benjaminwinger, line 92: We should probably prefer isReadOnly here too

benjaminwinger (commented):
I think all the transaction functions being used are const, so shouldn't the transaction pointers be const too?

benjaminwinger (approved):

PR #3827, opened 2024-07-15
Add decimal support to rust API

benjaminwinger:
I'm not keen on the way Decimal `Value`s have to be created through the C++ API (see `create_value_decimal` in `kuzu_rs.h`) but as there isn't any decimal type, and a constructor with multiple numeric arguments to create a decimal may not be very clear, I think any solution is outside the scope of this PR. Maybe we should move to using static methods to create Values similarly to LogicalType, to make things more explicit, but it still would not be very easy to interact with a Decimal Value.

### DIFF
--- a/tools/rust_api/include/kuzu_rs.h
+++ b/tools/rust_api/include/kuzu_rs.h
@@ -1,10 +1,13 @@
 #pragma once
 
+#include <cstdint>
 #include <memory>
 
 #include "rust/cxx.h"
 #ifdef KUZU_BUNDLED
 #include "common/type_utils.h"
+#include "common/types/int128_t.h"
+#include "common/types/types.h"
 #include "common/types/value/nested.h"
 #include "common/types/value/node.h"
 #include "common/types/value/recursive_rel.h"
@@ -70,6 +73,12 @@ inline std::unique_ptr<kuzu::common::LogicalType> create_logical_type_rdf_varian
     return std::make_unique<kuzu::common::LogicalType>(kuzu::common::LogicalType::RDF_VARIANT());
 }
 
+inline std::unique_ptr<kuzu::common::LogicalType> create_logical_type_decimal(uint32_t precision,
+    uint32_t scale) {
+    return std::make_unique<kuzu::common::LogicalType>(
+        kuzu::common::LogicalType::DECIMAL(precision, scale));
+}
+
 std::unique_ptr<kuzu::common::LogicalType> logical_type_get_list_child_type(
     const kuzu::common::LogicalType& logicalType);
 std::unique_ptr<kuzu::common::LogicalType> logical_type_get_array_child_type(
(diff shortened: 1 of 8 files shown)


tools/rust_api/src/ffi.rs
```diff
@@ -340,6 +376,12 @@ pub(crate) mod ffi {
         fn create_value_int128_t(high: i64, low: u64) -> UniquePtr<Value>;
         fn create_value_uuid_t(high: i64, low: u64) -> UniquePtr<Value>;
         fn create_value_internal_id(offset: u64, table: u64) -> UniquePtr<Value>;
+        fn create_value_decimal(
```
mxwli, line 379: I suggest separating decimal creation into two parts
One part could be a user facing function that takes `value: i128, scale: u32` and determines the high, low, precision, and scale, to call the less user friendly `create_value_decimal`, or some variant.
Internally, a DECIMAL(10, 3) can easily be cast to a DECIMAL(18, 3) so you can resolve the precision to be the minimal precision that fits.

tools/rust_api/include/kuzu_rs.h
```diff
@@ -186,6 +205,21 @@ template<typename T>
 std::unique_ptr<kuzu::common::Value> create_value(const T value) {
     return std::make_unique<kuzu::common::Value>(value);
 }
+inline std::unique_ptr<kuzu::common::Value> create_value_decimal(int64_t high, uint64_t low,
```
mxwli, line 208: I initially wrote decimal thinking of it as a front end for integers, which unfortunately makes it difficult to interact with in some parts of teh code. We could create a decimal type `decimal_t` containing `int128_t val, uint32_t precision, uint32_t scale` to construct decimal values for C++ API users. Changing value creation to that of LogicalType creation works, but it's a lot more work than adding `decimal_t`.

mxwli:
I would also add a way for users to convert from a kuzu::common::Value decimal to a rust_decimal::Decimal or floating point number, but ultimately this decision is up to you. I'm not familiar enough with how people use rust.

mxwli (approved):
I would also add a way for users to convert from a kuzu::common::Value decimal to a rust_decimal::Decimal or floating point number, but ultimately this decision is up to you. I'm not familiar enough with how people use rust.

tools/rust_api/src/ffi.rs
benjaminwinger: From the rust side, you create decimals using the [`Value::Decimal { value: rust_decimal::Decimal, precision: u32 }`](https://github.com/kuzudb/kuzu/pull/3827/files#diff-fba2944519c68ae7315eeeb0cbbeb726e56030ab9f774d5eeb9b8d90502a716fR273-R276) enum. [E.g.](https://github.com/kuzudb/kuzu/pull/3827/files#diff-fba2944519c68ae7315eeeb0cbbeb726e56030ab9f774d5eeb9b8d90502a716fR1276) (note that the `dec!` macro is part of a separate crate, rust_decimal_macros).
The `ffi::create_value_decimal` is just used as part of the internal conversion to a C++ Value which is used when Rust Values are passed to `Connection::execute`.
I'd added the precision explicitly as I was assuming that when passed to `Connection::execute`, the precision needs to match the parameter type. Or will it be cast automatically? If so we could change it to `Value::Decimal(rust_decimal::Decimal)`.

tools/rust_api/include/kuzu_rs.h
benjaminwinger: That sounds reasonable to me.

tools/rust_api/src/ffi.rs
mxwli: Ah I see. I'm fairly certain you don't need to match the precision explicitly. If you're talking about something where a column or function expects a decimal with a higher precision, but as a parameter, you pass a decimal with a lower precision, then there would be an implicit cast.

benjaminwinger: I figured that out through some experimentation. I've since updated the PR to remove the explicit precision.

PR #3830, opened 2024-07-15
Mvcc Concurrent Testing

yiyun-sj:


test/include/graph_test/graph_test.h
```diff
@@ -32,6 +37,35 @@ class EmptyDBTest : public PrivateGraphTest {
     std::string getInputDir() override { KU_UNREACHABLE; }
 };
 
+class ConcurrentTest {
```
ray6080, line 40: `ConcurrentTest` -> `ConcurrentTestExecutor` / `ConcurrentTestRunner`.

test/include/graph_test/graph_test.h
```diff
@@ -32,6 +37,35 @@ class EmptyDBTest : public PrivateGraphTest {
...
+public:
+    ConcurrentTest(bool& connectionsPaused, main::Connection& connection, std::string databasePath):
+      connectionPaused{connectionsPaused}, connection{connection}, databasePath{databasePath} {}
```
ray6080, line 43: Better leave an empty line between constructor and other functions.

test/include/graph_test/graph_test.h
```diff
@@ -32,6 +37,35 @@ class EmptyDBTest : public PrivateGraphTest {
...
+    }
+
+private:
+    void runStatements();
+
+    bool& connectionPaused;
+    main::Connection& connection;
+    std::string databasePath;
+    std::thread connThread;
+    std::vector<TestStatement*> statementsQueue;
```
ray6080, line 66: I think simply `statements` is fine.

test/graph_test/graph_test.cpp
```diff
@@ -47,5 +48,12 @@ void DBTest::createNewDB() {
     createDBAndConn();
 }
 
+void ConcurrentTest::runStatements() {
+    while (connectionPaused) {}
```
ray6080, line 52: hmm isn't it simpler if you only run `runStatements` when `connectionPaused` is false? so no need to busy wait here.

test/include/graph_test/graph_test.h
```diff
@@ -32,6 +37,35 @@ class EmptyDBTest : public PrivateGraphTest {
...
 };
 
+class ConcurrentTest {
+public:
+    ConcurrentTest(bool& connectionsPaused, main::Connection& connection, std::string databasePath):
+      connectionPaused{connectionsPaused}, connection{connection}, databasePath{databasePath} {}
+    void execute() {
+        connThread = std::thread(&ConcurrentTest::runStatements, this);
+    }
+    void join() {
```
ray6080, line 47: Looks to me `join()` is unnecessary, I think you should always join inside `execute`, which makes the interface simpler.

test/include/graph_test/graph_test.h
```diff
@@ -32,6 +37,35 @@ class EmptyDBTest : public PrivateGraphTest {
...
+        statementsQueue.clear();
+    }
+    void addStatement(TestStatement* statement) {
+        statementsQueue.emplace_back(statement);
+    }
+
+private:
+    void runStatements();
+
+    bool& connectionPaused;
```
ray6080, line 62: `connectionPaused` is also not needed if you don't busy wait inside `runStatements`.

test/include/graph_test/graph_test.h
ray6080, line 40: Add a comment here saying that `ConcurrentTest` is not thread safe.

test/include/graph_test/graph_test.h
```diff
@@ -66,21 +103,43 @@ class DBTest : public PrivateGraphTest {
...
     }
 
-    inline void CheckConn(std::string connName) {
-        if (connMap[connName] == nullptr) {
-            connMap[connName] = std::make_unique<main::Connection>(database.get());
-        }
-    }
+protected:
+    bool isConcurrent = false;
+    bool connectionsPaused = false;
```
ray6080, line 140: I wonder if you can replace these two bool fields with the enum class ConcurrentStatusFlag.

yiyun-sj: we probably can just keep the bool fields as they make a little more sense to me than the enum

ray6080 (approved):

Issue #3619, opened 2024-06-10
Feature: Expose Decimal to Rust and (possibly) NodeJS API


mxwli:
### API

Other

### Description

#3618 exposes the datatypes to the Python and Java APIs.

We possibly don't need to do anything for javascript; I couldn't find a decimal equivalent. 

[link to decimal datatype in rust](https://docs.rs/rust_decimal/latest/rust_decimal/struct.Decimal.html#)
Rust and Kuzu use similar physical representations. Rust just fixes the width to 128 bits, while Kuzu chooses from [16, 32, 64, 128] bits

PR #3835, opened 2024-07-16
Add DECIMAL conversion for Node.js API

mewim:
# Description

This PR adds support for DECIMAL type for Node.js API. Due to Node.js does not support decimal type natively, this PR converts decimal for Node.js Number.

Fixes # (issue)
#3619


mxwli (approved):
Looks good to me

PR #3836, opened 2024-07-16
NetworkX multi-edge support

### LINKED ISSUE #3640: Bug: Issue with Kuzu to Networkx
### What happened?

Hi,

I am using the kuzu to networkx functionality:
`res = conn.execute('MATCH (u:User)-[r:Rating]->(m:Movie) RETURN u, r, m LIMIT 250')`
`G = res.get_as_networkx(directed=False)`

When I use this for the graph I created, I noticed that it does not support multiple edges between the same nodes (say A and B have a connection between them with condition X and Y). eg: It removes 
`(A)-[rel: a2b {category:Y}]->(B)` and only keeps `(A)-[rel: a2b {category:X}]->(B)`.

Is there a way to resolve this?

Thanks in advance!

mewim:
# Description

This PR addresses #3640 by changing `DiGraph` to `MultiGraph`.

Fixes #3640

### DIFF
--- a/tools/python_api/src_py/query_result.py
+++ b/tools/python_api/src_py/query_result.py
@@ -233,7 +233,9 @@ class QueryResult:
         self.check_for_query_result_close()
         self._query_result.resetIterator()
 
-    def get_as_networkx(self, directed: bool = True) -> nx.Graph | nx.DiGraph:  # noqa: FBT001
+    def get_as_networkx(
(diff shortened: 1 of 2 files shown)


prrao87 (approved):
Tried this on my existing NetworkX codebases and it works! Nothing breaks. LGTM 👌🏽.

PR #3850, opened 2024-07-17
Mvcc catalog locking

yiyun-sj:


src/include/catalog/catalog_set.h
```diff
@@ -43,6 +48,12 @@ class CatalogSet {
     static std::unique_ptr<CatalogSet> deserialize(common::Deserializer& deserializer);
 
 private:
+    bool containsEntryNoLock(transaction::Transaction* transaction, const std::string& name) const;
+    CatalogEntry* getEntryNoLock(transaction::Transaction* transaction, const std::string& name);
+    void createEntryNoLock(transaction::Transaction* transaction,
+        std::unique_ptr<CatalogEntry> entry);
+    void dropEntryNoLock(transaction::Transaction* transaction, const std::string& name);
+
     void validateExist(transaction::Transaction* transaction, const std::string& name) const;
```
ray6080, line 57: Can u double check and rename private functions not grabbing a lock to with the `NoLock` suffix? e.g., `validateExist`, `validateNotExist`, `emplace`, `erase`, `traverseVersionChainsForTransaction`, `getCommittedEntry`. I think that's easier to read since we have some functions starting to have `NoLock` suffix now.

test/test_files/ddl/concurrency.test
```diff
@@ -23,3 +21,148 @@
...
+9|lastJobDuration|INTERVAL|False
+10|workedHours|INT64[]|False
+11|usedNames|STRING[]|False
+12|courseScoresPerTerm|INT64[][]|False
+13|grades|INT64[4]|False
+14|height|FLOAT|False
+15|u|UUID|False
+16|propA|INT64|False
+
+-CASE ConcurrentSingleWriteCreate
```
ray6080, line 53: ConcurrentSingleWriteCreate, ConcurrentSingleWriteUpdate and ConcurrentSingleWriteDelete don't really belong to ddl, can u reorganize the tests a bit?
I would suggest move the ddl part to `transaction/ddl/` as `ddl_concurrency.test`. and separate the above three into a new test file under `transaction` as `dml_concurrency.test`.

ray6080 (approved):

PR #3863, opened 2024-07-18
Turn on primary key scan while using PreparedStatement

ted-wq-x:
# Description

When PARAMETER exists in the statement, PRIMARY_KEY_SCAN is supported as LogicalScanNodeTableType

# Contributor agreement

### DIFF
--- a/src/optimizer/filter_push_down_optimizer.cpp
+++ b/src/optimizer/filter_push_down_optimizer.cpp
@@ -1,5 +1,6 @@
 #include "optimizer/filter_push_down_optimizer.h"
 
+#include "binder/expression/function_expression.h"
 #include "binder/expression/literal_expression.h"
 #include "binder/expression/property_expression.h"
 #include "main/client_context.h"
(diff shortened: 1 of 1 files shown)


src/optimizer/filter_push_down_optimizer.cpp
```diff
@@ -139,6 +140,26 @@ static ColumnPredicateSet getPropertyPredicateSet(const Expression& property,
     return propertyPredicateSet;
 }
 
+// TODO need determine whether the expression is a constant based on its output type
+static bool isLiteralExpression(const std::shared_ptr<Expression> expression) {
```
andyfengHKU, line 144: I think isConstant might be a more accurate name here?

src/optimizer/filter_push_down_optimizer.cpp
```diff
@@ -139,6 +140,26 @@ static ColumnPredicateSet getPropertyPredicateSet(const Expression& property,
     return propertyPredicateSet;
 }
 
+// TODO need determine whether the expression is a constant based on its output type
```
andyfengHKU, line 143: I don't fully understand this TODO. I thought the expression output data type is guaranteed to be the same as index data type. What do we need to determine here?

src/optimizer/filter_push_down_optimizer.cpp
```diff
@@ -139,6 +140,26 @@ static ColumnPredicateSet getPropertyPredicateSet(const Expression& property,
...
+static bool isLiteralExpression(const std::shared_ptr<Expression> expression) {
+    switch (expression->expressionType) {
+    case ExpressionType::LITERAL:
+    case ExpressionType::PARAMETER: {
+        return true;
+    }
+    case ExpressionType::FUNCTION: {
```
andyfengHKU, line 150: Can u leave a TODO for me here.
```
TODO(Xiyang): fold parameter expression in binder.
```

`CAST($1)` should ideally be folded once we obtain `$1` from user input. I'm fairly certain `CAST("1", INT64)` has already been folded to `1` in current master. We just need to propagate this to parameter and then we don't need to do this check recursively.

andyfengHKU (approved):
LGTM. Thanks for the PR!

ted-wq-x:
done

PR #3864, opened 2024-07-18
Hash index support for MVCC

benjaminwinger:
Access to the hash index has been mostly moved into the node table so that it can handle visibility checks.
I've fixed a few (presumably) gcc-14-related compilation issues introduced by the mvcc-base branch.
I've also removed the dummy transactions in a couple places related to the hash index so that copying and lookups work properly.

I haven't benchmarked the performance impact of these visibility checks on the hash index yet. It should mostly affect lookups, since every positive lookup needs a visibility check, but since negative lookups don't, it shouldn't affect copy performance (except where there are lots of duplicate values from other transactions, but not the initial copy, or even multiple copies without anything in-between).

I also found dealing with access to the lowest level of the NodeTable a little confusing because of how the offset has to change each time to be relative to the beginning of each part, but they all basically take the same integer-based offset types as arguments (NodeTable->NodeGroup->GroupCollection<ChunkedNodeGroup>->ChunkedNodeGroup (and in this case to VersionInfo, but it would be similarly complicated to access the chunks)). In particular, one issue was that NodeGroup gives public access to its data through `getChunkedNodeGroup`, but I wasn't able to find the correct group without using the private `findChunkedGroupFromRowIdx`. I'll have to look in more detail at how everything is accessed, but it seems like having more restricted APIs for accessing the data would make them easier to work with and harder to make mistakes. Maybe using simple struct wrappers for the offsets so that they can't be confused. E.g. `struct GlobalOffset { offset_t offset; };` `struct NodeGroupOffset { offset_t offsetInNodeGroup; };`

### DIFF
--- a/src/include/storage/index/hash_index.h
+++ b/src/include/storage/index/hash_index.h
@@ -86,10 +86,10 @@ public:
     // - the key is neither deleted nor found in the local storage, lookup in the persistent
     // storage.
     bool lookupInternal(const transaction::Transaction* transaction, Key key,
-        common::offset_t& result) {
+        common::offset_t& result, visible_func isVisible) {
         KU_ASSERT(transaction->isWriteTransaction() || transaction->isReadOnly() ||
                   transaction->isDummy());
-        auto localLookupState = localStorage->lookup(key, result);
+        auto localLookupState = localStorage->lookup(key, result, isVisible);
         if (localLookupState == HashIndexLocalLookupState::KEY_DELETED) {
             return false;
         }
@@ -97,7 +97,7 @@ public:
             return true;
         }
         KU_ASSERT(localLookupState == HashIndexLocalLookupState::KEY_NOT_EXIST);
-        return lookupInPersistentIndex(transaction, key, result);
+        return lookupInPersistentIndex(transaction, key, result, isVisible);
     }
 
     // For deletions, we don't check if the deleted keys exist or not. Thus, we don't need to check
@@ -111,18 +111,18 @@ public:
     // index, if
     //   so, return false, else insert the key to the local storage.
     bool insertInternal(const transaction::Transaction* transaction, Key key,
-        common::offset_t value) {
+        common::offset_t value, visible_func isVisible) {
         common::offset_t tmpResult;
-        auto localLookupState = localStorage->lookup(key, tmpResult);
+        auto localLookupState = localStorage->lookup(key, tmpResult, isVisible);
         if (localLookupState == HashIndexLocalLookupState::KEY_FOUND) {
             return false;
         }
-        if (localLookupState == HashIndexLocalLookupState::KEY_NOT_EXIST) {
-            if (lookupInPersistentIndex(transaction, key, tmpResult)) {
+        if (localLookupState != HashIndexLocalLookupState::KEY_DELETED) {
+            if (lookupInPersistentIndex(transaction, key, tmpResult, isVisible)) {
                 return false;
             }
         }
-        return localStorage->insert(key, value);
+        return localStorage->insert(key, value, isVisible);
     }
 
     using BufferKeyType =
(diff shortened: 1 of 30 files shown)


src/storage/store/node_table.cpp
```diff
@@ -313,5 +322,20 @@ uint64_t NodeTable::getEstimatedMemoryUsage() const {
...
+    auto [nodeGroupIdx, offsetInGroup] = StorageUtils::getNodeGroupIdxAndOffsetInChunk(offset);
+    auto* nodeGroup = getNodeGroup(nodeGroupIdx);
+    if (nodeGroup->isDeleted(transaction, offsetInGroup)) {
+        return false;
+    }
+    return nodeGroup->isInserted(transaction, offsetInGroup);
+}
+
+bool NodeTable::lookupPK(transaction::Transaction* transaction, common::ValueVector* keyVector,
+    uint64_t vectorPos, common::offset_t& result) const {
```
ray6080, line 335: Can u leave a todo here for me? We need to check the transaction local hash index as well.

src/storage/store/node_table.cpp
```diff
@@ -148,7 +153,9 @@ void NodeTable::insert(Transaction* transaction, TableInsertState& insertState)
     }
     const auto localTable = transaction->getLocalStorage()->getLocalTable(tableID,
         LocalStorage::NotExistAction::CREATE);
-    localTable->insert(transaction, insertState);
+    ku_dynamic_cast<LocalTable*, LocalNodeTable*>(localTable)
+        ->insert(transaction, insertState,
+            [&](offset_t offset) { return isVisible(transaction, offset); });
```
ray6080, line 158: I think the visibility check here needs to be changed, unless I'm missing something here. For inserting into local node table, the `isVisible` check should be performed on local node table, while it seems to be checking committed part of the node table here through `NodeTable::isVisible()`. We should let LocalNodeTable have its own `isVisible` implementation and not pass in the param?

src/include/storage/index/hash_index.h
```diff
@@ -33,6 +33,8 @@ template<typename T>
 class DiskArray;
 class DiskArrayCollection;
 
+using visible_func = std::function<bool(common::offset_t)>;
```
ray6080, line 36: This is duplicated with the one in `local_hash_index.h`. Can we remove this one?

src/include/storage/local_storage/local_hash_index.h
```diff
@@ -7,6 +7,8 @@
 namespace kuzu {
 namespace storage {
 
+using visible_func = std::function<bool(common::offset_t)>;
```
ray6080, line 10: Ditto on duplicating with `in_mem_hash_index.h`.

src/include/storage/local_storage/local_node_table.h
```diff
@@ -11,14 +11,17 @@ namespace storage {
 class ChunkedNodeGroup;
 struct TableScanState;
 
+using visible_func = std::function<bool(common::offset_t)>;
```
ray6080, line 14: Ditto on duplicating with `local_hash_index.h`.

ray6080 (approved):
Thanks! There is one comment on local node table you can take a look. Otherwise looks good to me.

ray6080:
Can u also uncomment the test `NodeInsertionDeletionTests.DeleteAddMixedTest` under `test/storage/node_insertion_deletion_test.cpp`? That one should work fine now.

src/storage/store/node_table.cpp
benjaminwinger: Okay. I'm suspicious that this case is not being covered by the tests as I'd initially added it using a dummy transaction and the number of test failures was unchanged.

src/include/storage/index/hash_index.h
benjaminwinger: Sure

src/storage/store/node_table.cpp
ray6080: Yeah, it could be not covered yet. I will add tests on this.

PR #3873, opened 2024-07-18
Add pyarrow `COPY FROM` testing

mxwli:
# Description

Adds very basic testing. By my understanding of the COPY FROM function, there is basically zero change in the core scanning functionality. These tests are just to make sure COPY FROM works; advanced scanning is already covered.

# Contributor agreement

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@ bazel-*
 .bazelrc
 **/test/test_files/test_list
 mm-256KB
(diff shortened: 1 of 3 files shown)


history.txt
andyfengHKU, line 1: remove this. actually update gitignore to ignore history.txt

andyfengHKU (approved):

PR #3898, opened 2024-07-22
Add support for passing literals to array functions

mxwli:
# Description

`array_cross_product([1.0, 2.0, 3.0], [1.0, 2.0, 3.0])` and the likes should work now.

# Contributor agreement

### DIFF
--- a/src/function/array/array_functions.cpp
+++ b/src/function/array/array_functions.cpp
@@ -15,7 +15,8 @@ namespace function {
 static LogicalType interpretLogicalType(binder::Expression* expr) {
     if (expr->expressionType == ExpressionType::LITERAL &&
         expr->dataType.getLogicalTypeID() == LogicalTypeID::LIST) {
-        auto numChildren = expr->constPtrCast<binder::LiteralExpression>()->getValue().getChildrenSize();
+        auto numChildren =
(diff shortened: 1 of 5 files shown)


src/function/array/array_functions.cpp
```diff
@@ -139,8 +152,20 @@ scalar_func_exec_t getScalarExecFunc(LogicalType type) {
...
+        auto numChildren =
+            arguments[0]->constPtrCast<binder::LiteralExpression>()->getValue().getChildrenSize();
+        leftType = LogicalType::ARRAY(ListType::getChildType(leftType).copy(), numChildren);
+    }
+    if (arguments[1]->expressionType == ExpressionType::LITERAL &&
+        rightType.getLogicalTypeID() == LogicalTypeID::LIST) {
+        auto numChildren =
+            arguments[1]->constPtrCast<binder::LiteralExpression>()->getValue().getChildrenSize();
+        rightType = LogicalType::ARRAY(ListType::getChildType(rightType).copy(), numChildren);
+    }
```
andyfengHKU, line 168: Apparently this is duplicated code with l17 -l30. Let's refactor them into a function

mxwli: sounds good

PR #3912, opened 2024-07-23
Fix RDF and Rel Group Entries in WAL Recovery

yiyun-sj:


src/catalog/catalog_set.cpp
```diff
@@ -169,7 +169,6 @@ void CatalogSet::alterEntry(Transaction* transaction, const binder::BoundAlterIn
 
 CatalogEntrySet CatalogSet::getEntries(Transaction* transaction) {
     CatalogEntrySet result;
-    std::lock_guard lck{mtx};
```
yiyun-sj, line 172: @ray6080 had to delete this due to deadlock. Wondering if this can stay deleted? since we traverse version chains for the entries the tranactional guarantee should be fine. Just afraid of any undefined data behaviour when another connection emplaces a new entry.

src/transaction/transaction.cpp
```diff
@@ -65,10 +70,16 @@ void Transaction::pushCatalogEntry(CatalogSet& catalogSet, CatalogEntry& catalog
     case CatalogEntryType::TYPE_ENTRY: {
         KU_ASSERT(
             catalogEntry.getType() == CatalogEntryType::DUMMY_ENTRY && catalogEntry.isDeleted());
+        if (newCatalogEntry->hasParent()) {
```
ray6080, line 73: Is this for sequence? I'm not following sequence for now. Maybe you can just walk me through this line of change tomorrow.

ray6080 (approved):
Thanks!

ray6080:
Can u unskip and double check tests related to `dataset/rdf/rdf_variant`? 
- `CApiRdfVariantTest`
- Python api test: https://github.com/kuzudb/kuzu/blob/dcdd534d6f0e2445d7f828d6bf5987bf8f08a556/tools/python_api/test/conftest.py#L149. and check the list of py tests with todo related to rdf.

src/catalog/catalog_set.cpp
```diff
@@ -50,12 +50,24 @@ CatalogEntry* CatalogSet::getEntryNoLock(Transaction* transaction, const std::st
     return entry;
 }
 
+static void LogEntryForTrx(Transaction* transaction, CatalogSet& set, CatalogEntry& entry) {
```
ray6080, line 53: LogEntryForTrx -> logEntryForTrx.

Function names always stay as the first letter lower case regardless of static or non-static

PR #3919, opened 2024-07-23
Fix Sequence WAL and rollback behaviour

yiyun-sj:


src/storage/wal_replayer.cpp
```diff
@@ -362,8 +362,9 @@ void WALReplayer::replayUpdateSequenceRecord(const WALRecord& walRecord) const {
     const auto sequenceID = sequenceEntryRecord.sequenceID;
     const auto entry =
         clientContext.getCatalog()->getSequenceCatalogEntry(clientContext.getTx(), sequenceID);
-    entry->replayVal(sequenceEntryRecord.data.usageCount, sequenceEntryRecord.data.currVal,
-        sequenceEntryRecord.data.nextVal);
+    // TODO(Sam): remove unnecessary valuevector allocation
```
ray6080, line 365: This looks like an important todo. Let's address it properly.

I see two options:
1. expose an extra interface which correctly replays nextKVal.
2. refactor `nextKVal` to not populate `resultVector`. This doesn't look like an easy way, but it would be a good thing to do.

ray6080 (approved):

src/catalog/catalog_entry/sequence_catalog_entry.cpp
```diff
@@ -31,66 +31,76 @@ int64_t SequenceCatalogEntry::currVal() {
     return result;
 }
 
+void SequenceCatalogEntry::nextVal() {
```
ray6080, line 34: `nextVal` -> `nextValNoLock`. or pass in the lock as param.

src/catalog/catalog_entry/sequence_catalog_entry.cpp
```diff
@@ -31,66 +31,76 @@ int64_t SequenceCatalogEntry::currVal() {
...
+
 // referenced from DuckDB
+void SequenceCatalogEntry::nextKVal(transaction::Transaction* transaction, const uint64_t& count) {
+    KU_ASSERT(count > 0);
+    std::lock_guard<std::mutex> lck(mtx);
+    SequenceRollbackData rollbackData{sequenceData.usageCount, sequenceData.currVal};
+    for (auto i = 0ul; i < count; i++) {
+        nextVal();
+    }
+    transaction->pushSequenceChange(this, count, rollbackData);
```
ray6080, line 79: Let's move L79 out of the scope of the lock of `lck`.

ray6080 (approved):

PR #3924, opened 2024-07-24
Fix duplicate primary key update

andyfengHKU:
# Description

Fixes # (issue)

# Contributor agreement

### DIFF
--- a/src/storage/store/node_table.cpp
+++ b/src/storage/store/node_table.cpp
@@ -21,7 +21,6 @@ using namespace kuzu::evaluator;
 namespace kuzu {
 namespace storage {
 
-
 NodeTable::NodeTable(StorageManager* storageManager, const NodeTableCatalogEntry* nodeTableEntry,
     MemoryManager* memoryManager, VirtualFileSystem* vfs, main::ClientContext* context,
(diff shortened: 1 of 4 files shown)


src/storage/store/node_table.cpp
```diff
@@ -147,16 +147,41 @@ offset_t NodeTable::validateUniquenessConstraint(const Transaction* transaction,
...
+    common::offset_t dummyOffset;
+    auto& selVector = pkVector->state->getSelVector();
+    KU_ASSERT(selVector.getSelSize() == 1);
+    if (pkVector->isNull(selVector[0])) {
+        throw RuntimeException(ExceptionMessage::nullPKException());
+    }
+    if (pkIndex->lookup(transaction, pkVector, selVector[0], dummyOffset,
+            [&](offset_t offset) { return isVisible(transaction, offset); })) {
+        throw RuntimeException(
+            ExceptionMessage::duplicatePKException(getPkStr(*pkVector, selVector[0])));
```
ray6080, line 170: Alternative option to get the string here is going through `ValueVector::getAsValue()`, which I think it's a better way.

ray6080 (approved):

PR #3927, opened 2024-07-24
Removed getExpressionsForPrinting() from physicalOperators

hamzakammar:
# Description

Adds to #3664 
Removed all getExpressionsForPrinting() functions from the PhysicalOperators
# Contributor agreement


MSebanc (commented):
There are some `getExpressionsForPrinting()` that you missed so it is currently not building

PR #3929, opened 2024-07-24
Fix manual tx in table lookup

andyfengHKU:
# Description

Fixes # (issue)

# Contributor agreement

### DIFF
--- a/src/include/processor/operator/mask.h
+++ b/src/include/processor/operator/mask.h
@@ -59,7 +59,7 @@ public:
     // if the mask is set to true (mask value is equal to the expected currentMaskValue) or not.
     void incrementMaskValue(common::offset_t offset, uint8_t currentMaskValue) {
         if (offset >= maskData->getSize()) [[unlikely]] { // Handle uncommitted node offsets.
-            return ;
+            return;
         }
         if (maskData->isMasked(offset, currentMaskValue)) {
             maskData->setMask(offset, currentMaskValue + 1);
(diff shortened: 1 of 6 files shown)


src/include/processor/operator/mask.h
```diff
@@ -58,6 +58,9 @@ class MaskCollection {
     // Note: blindly update mask does not parallelize well, so we minimize write by first checking
     // if the mask is set to true (mask value is equal to the expected currentMaskValue) or not.
     void incrementMaskValue(common::offset_t offset, uint8_t currentMaskValue) {
+        if (offset >= maskData->getSize()) { // Handle uncommitted node offsets.
```
ray6080, line 61: Can u remove these lines of code?
https://github.com/kuzudb/kuzu/blob/11b093bda0632b201ae43de07b12fbf074df67dc/src/processor/operator/semi_masker.cpp#L37-L39
I think I only protected it against case of SingleTableSemiMasker. I would also mark `unlikely` here.

test/test_files/issue/issue.test
```diff
@@ -500,3 +500,20 @@ Binder exception: Table rel2 does not exist.
 ---- 2
 {_ID: 0:0, _LABEL: label1, id: 0}
 {_ID: 1:0, _LABEL: label2, id: 0}
+
+-CASE RecursiveJoinManualTransaction
```
ray6080, line 504: Can u also attach an issue id at the end of the case name?

ray6080 (approved):

PR #3944, opened 2024-07-25
Fix more checkpoint-related issues

royi-luo:
# Description

Implement the following fixes:
1. Avoid checkpointing list data chunk if it is empty (assertion failure)
2. Clear indexTable in `StringChunkData` when switching to in-memory residency (as the data is cleared so there is nothing for the indices to refer to)
3. Fix list column chunk's numValues going out of sync with its children

# Contributor agreement


ray6080 (approved):
Thanks!

PR #3946, opened 2024-07-26
Fix local node delete leading to inconsistent node offsets

ray6080:
# Description

When newly inserted nodes are deleted within the same transaction, during commit, deleted nodes are not appended to table's nodeGroups. This shifts offsets of the nodes after the deleted one, leading to connected rels recording inconsistent src/dst node offsets.

### DIFF
--- a/src/include/storage/store/node_group.h
+++ b/src/include/storage/store/node_group.h
@@ -118,7 +118,8 @@ public:
     const std::vector<common::LogicalType>& getDataTypes() const { return dataTypes; }
     NodeGroupDataFormat getFormat() const { return format; }
     common::row_idx_t append(const transaction::Transaction* transaction,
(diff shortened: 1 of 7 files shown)


src/include/storage/store/node_group.h
```diff
@@ -203,6 +205,7 @@ class NodeGroup {
     // `nextRowToAppend` is a cursor to allow us to pre-reserve a set of rows to append before
     // acutally appending data. This is an optimization to reduce lock-contention when appending in
     // parallel.
+    // TODO(Guodong): Remove this field.
```
andyfengHKU, line 208: Are u going to resolve this TODO?

andyfengHKU (approved):

src/include/storage/store/node_group.h
ray6080: Yeah. but I don't want to do it in this PR.

PR #3951, opened 2024-07-26
More checkpoint fixes

royi-luo:
# Description

Fix issues discovered by running test suite with `checkpoint_threshold=0` with `auto_checkpoint=true`

Fix failing tests in:
- `test/test_files/update_rel/create_ldbc_sf01.test`, `test/test_files/transaction/create_rel/insert_rels_to_different_nodes.test` - correctly use old/new csr header offsets in `CSRNodeGroup::checkpointColumn`

- `test/test_files/update_node/set_tinysnb.test` - fix infinite loop in `UpdateInfo::hasUpdates`

- `test/test_files/update_node/set_empty.test` - sync num values for struct columns/chunks

- `test/test_files/transaction/create_rel/insert_rels_to_newly_added_node.test` - ensure values returned by csr header get start/end offset is increasing as nodeOffset increases

- `test/test_files/update_rel/create_batch.test` - (partially fixed, still incorrect query result) - fixed segfault caused by filling gaps in node group collections with nullptrs

# Contributor agreement

### DIFF
--- a/src/include/storage/store/column.h
+++ b/src/include/storage/store/column.h
@@ -90,7 +90,7 @@ public:
     common::offset_t appendValues(ColumnChunkData& persistentChunk, ChunkState& state,
         const uint8_t* data, const common::NullMask* nullChunkData, common::offset_t numValues);
 
-    virtual void checkpointColumnChunk(ColumnCheckpointState& checkpointState);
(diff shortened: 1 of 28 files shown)


ray6080:
These are now fixed in #3963 and #3959 .

PR #3962, opened 2024-07-28
Fix drop table leading to incorrect reload tables

ray6080:
# Description

Checkpoint should respect catalog entries for valid tables.
Also, add some more additional exception messages instead of ku_assert to avoid runtime seg faults.

To reproduce the issue, set `auto_checkpoint=true` and `checkpoint_threshold=0`, run tests under `transaction/ddl`.

### DIFF
--- a/src/storage/storage_manager.cpp
+++ b/src/storage/storage_manager.cpp
@@ -212,11 +212,28 @@ void StorageManager::checkpoint(main::ClientContext& clientContext) {
         O_RDWR | O_CREAT, &clientContext);
     const auto writer = std::make_shared<BufferedFileWriter>(*metadataFileInfo);
     Serializer ser(writer);
+    const auto nodeTableEntries =
+        clientContext.getCatalog()->getNodeTableEntries(&DUMMY_CHECKPOINT_TRANSACTION);
+    const auto relTableEntries =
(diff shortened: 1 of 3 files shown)


src/storage/storage_manager.cpp
```diff
@@ -212,11 +212,28 @@ void StorageManager::checkpoint(main::ClientContext& clientContext) {
...
+    const auto numTables = nodeTableEntries.size() + relTableEntries.size();
     ser.writeDebuggingInfo("num_tables");
-    ser.write<uint64_t>(tables.size());
-    // TODO(Guodong): We should avoid deleted tables.
-    for (auto& [_, table] : tables) {
-        table->checkpoint(ser);
+    ser.write<uint64_t>(numTables);
+    for (const auto tableEntry : nodeTableEntries) {
+        if (!tables.contains(tableEntry->getTableID())) {
+            throw RuntimeException(
```
andyfengHKU, line 224: Protect with LCOV?

andyfengHKU (approved):

src/storage/storage_manager.cpp
ray6080: Yeah. Wanna leave this to later when we go over codecov.

PR #3983, opened 2024-07-30
Add checkpoint statement

andyfengHKU:
# Description

See title.

# Contributor agreement


src/processor/operator/transaction.cpp
```diff
@@ -53,14 +57,17 @@ void Transaction::validateActiveTransaction(const TransactionContext& context) c
...
+            throw TransactionManagerException(stringFormat("No active transaction for {}.",
+                TransactionActionUtils::toString(transactionAction)));
         }
     } break;
-    case TransactionAction::ROLLBACK: {
-        if (!context.hasActiveTransaction()) {
-            throw TransactionManagerException("No active transaction for ROLLBACK.");
+    case TransactionAction::CHECKPOINT: {
+        if (context.hasActiveTransaction()) {
+            throw TransactionManagerException(stringFormat("Found active transaction for {}.",
```
ray6080, line 69: Would it be better to say "Checkpoint is not allowed in an active transaction` here?

ray6080 (approved):

PR #3987, opened 2024-07-30
[DO NOT REVIEW] Logical explain

MSebanc:
# Description

Logical Explain

# Contributor agreement


scripts/antlr4/keywords.txt.copy
```diff
@@ -54,6 +54,7 @@ JOIN
 KEY
 LIMIT
 LOAD
+LOGICAL
```
andyfengHKU, line 57: Let's also make this nonReservedKeyword change in Cypher.g4

src/include/main/plan_printer.h
```diff
@@ -38,22 +38,32 @@ class OpProfileTree {
 public:
     OpProfileTree(processor::PhysicalOperator* opProfileBoxes, common::Profiler& profiler);
 
+    explicit OpProfileTree(std::shared_ptr<planner::LogicalOperator> opProfileBoxes);
```
andyfengHKU, line 41: U can probably just use `LogicalOperator*` instead of shared_ptr

src/planner/operator/ddl/logical_alter.cpp
```diff
@@ -0,0 +1,45 @@
+#include "planner/operator/ddl/logical_alter.h"
+
+namespace kuzu {
+namespace planner {
+
+std::string LogicalAlterPrintInfo::toString() const {
```
andyfengHKU, line 6: User LCOV_EXCL_START & LCOV_EXCL_STOP to escape code coverage checking from l6-45. U might need to apply this to all toString functions.

src/include/planner/operator/ddl/logical_alter.h
```diff
@@ -6,17 +6,41 @@
 namespace kuzu {
 namespace planner {
 
+struct LogicalAlterPrintInfo final : OPPrintInfo {
+    common::AlterType alterType;
+    std::string tableName;
+    binder::BoundExtraAlterInfo* info;
```
andyfengHKU, line 12: If this object points to the info stored in the logical operator, then the life cycle should be fine.

Otherwise, it's better to copy the object here for safety.

src/include/planner/operator/ddl/logical_create_table.h
```diff
@@ -6,18 +6,41 @@
 namespace kuzu {
 namespace planner {
 
+struct LogicalCreateTablePrintInfo final : OPPrintInfo {
+    common::TableType tableType;
+    std::string tableName;
+    binder::BoundExtraCreateCatalogEntryInfo* info;
```
andyfengHKU, line 12: ditto check if u want to copy

src/planner/planner.cpp
```diff
@@ -54,7 +54,7 @@ std::unique_ptr<LogicalPlan> Planner::getBestPlan(const BoundStatement& statemen
     auto plan = std::make_unique<LogicalPlan>();
     switch (statement.getStatementType()) {
     case StatementType::QUERY: {
-        plan = getBestPlan(planQuery(statement));
+        plan = getBestPlan(planQuery(statement)); // not here
```
andyfengHKU, line 57: remove comment

src/planner/operator/simple/logical_create_table.cpp
```diff
@@ -0,0 +1,70 @@
+#include "planner/operator/ddl/logical_create_table.h"
+
+using namespace kuzu::common;
+
+namespace kuzu {
+namespace planner {
+
+std::string LogicalCreateTablePrintInfo::toString() const {
```
andyfengHKU, line 8: This is duplicate code with `create_table.cpp` we should refactor the two

src/planner/operator/ddl/logical_alter.cpp
andyfengHKU, line 6: ditto duplicate with `alter.cpp`

andyfengHKU (changes requested):
The overall framework looks good to me. But a lot of logical operator PrintInfo is duplicated with there physical counter part, e.g.

Both `LogicalAlter` and `Alter` use the same print logic. So `Alter` can potentially reuse the print info in `LogicalAlter`

@MSebanc, u can approach with one of the following

- merge the framework but remove duplicated code, or
- let @hamzakammar fix the duplication and then merge to this PR.

src/optimizer/filter_push_down_optimizer.cpp
```diff
@@ -174,6 +176,10 @@ std::shared_ptr<LogicalOperator> FilterPushDownOptimizer::visitScanNodeTableRepl
         scan.setPropertyPredicates(std::move(propertyPredicateSets));
     }
 
+    // TODO(Guodong): make index scan works under write transaction
+    if (context->getTx()->isWriteTransaction()) {
```
andyfengHKU, line 180: Why adding l179-l182? This should have been removed.

src/include/planner/operator/ddl/logical_alter.h
```diff
@@ -2,21 +2,42 @@
 
 #include "binder/ddl/bound_alter_info.h"
 #include "logical_ddl.h"
+#include "planner/operator/ddl/base_alter.h"
 
 namespace kuzu {
 namespace planner {
 
+struct LogicalAlterPrintInfo final : OPPrintInfo {
```
andyfengHKU, line 10: After reading through the code I suggest do the following.

```
struct LogicalAlterPrintInfo final : OPPrintInfo {
common::AlterType alterType;
std::string tableName;
binder::BoundExtraAlterInfo* info;

std::string toString() ....
}

struct AlterPrintInfo final : OPPrintInfo {
LogicalAlterPrintInfo info

std::string toString() const override {return info->toString();};
```
And remove BaseLogicalAlterPrintInfo.

If this make sense, propagate this change to CreateTable as well

andyfengHKU (changes requested):

andyfengHKU (approved):

src/main/plan_printer.cpp
```diff
@@ -349,5 +413,32 @@ nlohmann::json PlanPrinter::toJson(PhysicalOperator* physicalOperator, Profiler&
...
+    return planner::LogicalOperatorUtils::logicalOperatorTypeToString(
+        logicalOperator->getOperatorType());
+}
+
+std::string LogicalPlanPrinter::getOperatorParams(
+    std::shared_ptr<LogicalOperator> logicalOperator) {
+    return logicalOperator->getPrintInfo()->toString();
+}
+
+nlohmann::json LogicalPlanPrinter::toJson(std::shared_ptr<LogicalOperator> logicalOperator) {
```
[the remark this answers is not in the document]
ray6080: I chose to keep it for now. It should be a useful feature to me later.

PR #3989, opened 2024-07-30
Random Split for Multi Copy Testing

yiyun-sj:


ray6080 (commented):
Generally looks good to me.
I have one comment related to usability:
```
-SEED 3024167388 2272459668
-MULTI_COPY_RANDOM 5 test "${KUZU_ROOT_DIRECTORY}/dataset/large-serial/serialtable_merged.csv"
```
The separation of `SEED` and `MULTI_COPY_RANDOM` doesn't make much sense to me. It implies there is a  random engine being globally set with given seeds to generate random values, and multiple `MULTI_COPY_RANDOM` can reuse the same seeds. I feel that's confusing, and may be unnecessary. While let `MULTI_COPY_RANDOM` take params (if not specified, provide random seeds) are easier to use and understand.

ray6080 (approved):
Thanks! Can u also document this on our dev docs?

PR #3999, opened 2024-07-31
Fix rollback replay

ray6080:
# Description

Fix problems during recovery when there are multiple transaction rollbacks before the database closes.


andyfengHKU (approved):
Check coverage

ray6080:
> Check coverage

Should be covered already.

Issue #1816, opened 2023-07-14
In-memory database


LowLevelMahn:
like memgraph (or as a different example sqlite) have some sort of only-in-memory storage if persistence is not needed at all

ray6080:
Hi @LowLevelMahn thanks for raising the issue. This is a feature we're also looking into. Will make the design public once we get started on it.

stevesimmons:
I am also interested in an in-memory storage model. 

My actual use case is rather than one global graph db, having one small graph db per "case", in S3 blob storage. Then quickly loading the small graph db into either an in-memory db or a temporary on-disk directory. And writing it back to S3 if there were any writes. 

I can do this with Sqlite via the serialize/deserialize commands. It would be great with Kuzu to be able to serialise the catalog/metadata/data/index/wal files (per the physical storage described in https://github.com/kuzudb/kuzu/issues/1474) to a single binary blob, and then deserialise it to reload.

ray6080:
Hi @stevesimmons , thanks! That is a quite interesting use case. It makes sense to serialize/deserialize the whole database into/from a single blob, though we have to think a bit more how this can be done altogether with our on-disk storage changes. I'm adding that into our design for in-memory storage mode.

bigluck:
I'd love this feature too; my use case is very similar to the one described by @stevesimmons .

I know that you're working with an `IMPORT`/`EXPORT` command, which should facilitate importing/dumping the database into S3. 

But when importing an existing database, I would like to keep it only in memory, as it is a temporary database instance used only for reading data from the graph.

ATM I'm forced to initialize the DB using the following code:
```python
with TemporaryDirectory() as tmp_path:
   ...
```

But disk is way slower than memory.

Duckdb for example supports in-memory databases by default: https://duckdb.org/docs/api/python/overview#using-an-in-memory-database

hpvd:
just +1 on on this

dremekie:
+1

prrao87:
@dremekie Could you please upvote the topmost post so that we can use the stats in our sorting? Thanks!

sapalli2989:
In this regard: How could *derived data* be handled in Kuzu database?

For example, given some persistent data `A` stored in DB, and other data `B` derived from `A`, it makes more sense to ad-hoc generate `B` on each app startup and store it only temporarily until DB shutdown. Reason I'd like to store this derived data temporarily for the database session is leveraging Cypher queries on it.

[...]

PR #4012, opened 2024-08-02
In memory mode

ray6080:
# Description

Add the support for in-memory mode.  https://github.com/kuzudb/kuzu/issues/1816.

## Feature
When the database path is omitted, set to empty or set to `:memory:` (follow duck's convention here), the database will be open under in-memory mode.

The main differences between on-disk and in-memory mode are:
Under on-disk mode, all data will be persistent on disk. All transactions are logged in WAL, in which changes will be merged into database files during checkpoint.
While under the in-memory mode, there is no writes to WAL, no data is persistent to disk and `CHECKPOINT` will do nothing. All data are lost when the process finishes.

Restrictions on in-memory mode:
- The database cannot be open as read-only.
- When use Httpfs extension, we don't support file cache.
- Attaching an in-memory database is not allowed. (maybe we should allow?)

## Implementation
After MVCC changes, supporting in-mem mode becomes much easier.
New pages in BMFileHandle are pinned in BM and only unpinned when BMFileHandle gets destructed. Accesses to these pages directly grab frames from BM.

## Benchmark
Did a few simple benchmarks on ldbc-100.
#### COPY
```
| table       | in mem (ms) | on disk (ms) |
| ----------- | ----------- | ------------ |
| Comment     | 14879       | 34696        |
| Person      | 513         | 749          |
| knows       | 1781        | 2123         |
| likeComment | 29246       | 34581        |
```

```
| space usage | in-mem mode | on-disk mode |
| ----------- | ----------- | ------------ |
| mem         | 56.7GB      | 32.2GB       |
| disk        | 0           | 29GB         |
```

#### Scan
```
Q1: match (c:Comment) return min(c.ID), min(c.creationDate), min(c.locationIP), min(c.browserUsed), min(c.content), min(c.length)
Q2: MATCH (a:Person)-[:knows]->(b:Person)-[:knows]->(c:Person) RETURN MIN(a.birthday), MIN(b.birthday), MIN(c.birthday)
Q3: MATCH (a:Person)-[r:knows* 2..2]->(b:Person) RETURN COUNT(*);

| query | in-mem | on-disk (cold) | on-disk (warm) |
| ----- | ------ | -------------- | -------------- |
| Q1    | 1896   | 5330           | 1791           |
| Q2    | 1004   | 948            | 893            |
| Q3    | 72133  | 84419          | 
```
#### Insertions
1M node insertions.

```
CREATE NODE TABLE Person (id INT64, name STRING, age INT64, net_worth FLOAT, PRIMARY KEY (id));

| 1M txn in-mem (ms) | 1M txn on-disk (ms) | 1 txn in-mem (ms) | 1 txn on-disk (ms) |
| ------------------ | ------------------- | ---------------- | ------------------- |
| 47814              | 79313               | 38895            | 40054               |
```
## TODO
- [ ] rewrite copy to / export tests (copy to should not rely on db path)

### DIFF
--- a/src/include/storage/buffer_manager/bm_file_handle.h
+++ b/src/include/storage/buffer_manager/bm_file_handle.h
@@ -9,6 +9,7 @@
 #include "common/copy_constructors.h"
 #include "common/types/types.h"
 #include "storage/buffer_manager/vm_region.h"
+#include "storage/enums/page_read_policy.h"
 #include "storage/file_handle.h"
 
 namespace kuzu {
@@ -100,6 +101,12 @@ class BMFileHandle final : public FileHandle {
     friend class BufferManager;
 
 public:
+    uint8_t* pinPage(common::page_idx_t pageIdx, PageReadPolicy readPolicy);
+    void optimisticReadPage(common::page_idx_t pageIdx,
+        const std::function<void(uint8_t*)>& readOp);
+    // The function assumes that the requested page is already pinned.
+    void unpinPage(common::page_idx_t pageIdx);
+
     // This function assumes the page is already LOCKED.
     void setLockedPageDirty(common::page_idx_t pageIdx) {
         KU_ASSERT(pageIdx < numPages);
@@ -111,10 +118,12 @@ public:
     // does not hold any of the pages of the file.
     void resetToZeroPagesAndPageCapacity();
     void removePageIdxAndTruncateIfNecessary(common::page_idx_t pageIdx);
+    void removePageFromFrameIfNecessary(common::page_idx_t pageIdx);
+    void flushAllDirtyPagesInFrames();
+    void flushPageIfDirtyWithoutLock(common::page_idx_t pageIdx);
 
     common::file_idx_t getFileIndex() const { return fileIndex; }
 
-    BufferManager* getBM() { return bm; }
     uint8_t* getFrame(common::page_idx_t pageIdx);
 
     PageState* getPageState(common::page_idx_t pageIdx) {
--- a/src/include/storage/buffer_manager/bm_file_handle.h
+++ b/src/include/storage/buffer_manager/bm_file_handle.h
@@ -3,6 +3,7 @@
 #include <atomic>
 #include <cmath>
 #include <cstdint>
+#include <functional>
 
 #include "common/concurrent_vector.h"
 #include "common/constants.h"
--- a/src/include/storage/buffer_manager/bm_file_handle.h
+++ b/src/include/storage/buffer_manager/bm_file_handle.h
@@ -32,63 +32,59 @@ public:
 
     PageState() { stateAndVersion.store(EVICTED << NUM_BITS_TO_SHIFT_FOR_STATE); }
 
-    inline uint64_t getState() const { return getState(stateAndVersion.load()); }
-    inline static uint64_t getState(uint64_t stateAndVersion) {
+    uint64_t getState() const { return getState(stateAndVersion.load()); }
+    static uint64_t getState(uint64_t stateAndVersion) {
         return (stateAndVersion & STATE_MASK) >> NUM_BITS_TO_SHIFT_FOR_STATE;
     }
-    inline static uint64_t getVersion(uint64_t stateAndVersion) {
-        return stateAndVersion & VERSION_MASK;
-    }
-    inline static uint64_t updateStateWithSameVersion(uint64_t oldStateAndVersion,
-        uint64_t newState) {
+    static uint64_t getVersion(uint64_t stateAndVersion) { return stateAndVersion & VERSION_MASK; }
+    static uint64_t updateStateWithSameVersion(uint64_t oldStateAndVersion, uint64_t newState) {
         return ((oldStateAndVersion << 8) >> 8) | (newState << NUM_BITS_TO_SHIFT_FOR_STATE);
     }
-    inline static uint64_t updateStateAndIncrementVersion(uint64_t oldStateAndVersion,
-        uint64_t newState) {
+    static uint64_t updateStateAndIncrementVersion(uint64_t oldStateAndVersion, uint64_t newState) {
         return (((oldStateAndVersion << 8) >> 8) + 1) | (newState << NUM_BITS_TO_SHIFT_FOR_STATE);
     }
-    inline void spinLock(uint64_t oldStateAndVersion) {
+    void spinLock(uint64_t oldStateAndVersion) {
         while (true) {
             if (tryLock(oldStateAndVersion)) {
                 return;
             }
         }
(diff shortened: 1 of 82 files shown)


src/include/storage/storage_structure/db_file_utils.h
```diff
@@ -33,6 +34,11 @@ class DBFileUtils {
 public:
     constexpr static common::page_idx_t NULL_PAGE_IDX = common::INVALID_PAGE_IDX;
 
+    static uint8_t* pinPage(BMFileHandle& fileHandle, common::page_idx_t pageIdx,
```
benjaminwinger, line 37: I'm a little concerned that this adds new ways of doing pinning/optimistic reads without making it obvious what the differences are at a glance.
Maybe `BufferManager::pin`/`BufferManager::optimisticRead` should be private, or maybe we shouldn't even be passing around the BufferManager at all for these file operations. The BMFileHandle already stores a pointer to the BufferManager, so we could have a `BMFileHandle::pin(page_idx_t, PageReadPolicy)` function instead that handles the memory mode. While that doesn't really make the differences any clearer, it at least would make the BMFileHandle functions the more obvious choice, and we could remove the BufferManager arguments from DBFileUtils and avoid passing it around as much and make the BufferManager more of an internal detail that you shouldn't need to interact with when working with file handles (though #3743 will still mean that the MemoryManager needs to get passed around, but with this removed maybe we can even remove `MemoryManager::getBufferManager`).

src/storage/store/column_chunk_data.cpp
```diff
@@ -28,8 +29,13 @@ namespace storage {
 ColumnChunkMetadata uncompressedFlushBuffer(const uint8_t* buffer, uint64_t bufferSize,
     BMFileHandle* dataFH, page_idx_t startPageIdx, const ColumnChunkMetadata& metadata) {
     KU_ASSERT(dataFH->getNumPages() >= startPageIdx + metadata.numPages);
-    dataFH->getFileInfo()->writeFile(buffer, bufferSize,
-        startPageIdx * BufferPoolConstants::PAGE_4KB_SIZE);
+    if (dataFH->isInMemoryMode()) {
+        const auto frame = dataFH->getBM()->getFrame(*dataFH, startPageIdx);
+        memcpy(frame, buffer, bufferSize);
+    } else {
+        dataFH->getFileInfo()->writeFile(buffer, bufferSize,
```
benjaminwinger, line 36: Maybe we should move away from exposing the FileInfo through the file handle and add a write function that more or less is equivalent to this block (i.e. memcpy in in-memory-mode and use the FileInfo otherwise).

ray6080: Separating this to another PR for refactoring.

src/include/storage/wal/wal.h
```diff
@@ -25,16 +24,13 @@ struct SequenceRollbackData;
...
-using lock_t = std::unique_lock<std::mutex>;
-
 class WALReplayer;
 class WAL {
     friend class WALReplayer;
 
 public:
-    WAL(const std::string& directory, bool readOnly, BufferManager& bufferManager,
-        common::VirtualFileSystem* vfs, main::ClientContext* context);
+    WAL(const std::string& directory, bool readOnly, common::VirtualFileSystem* vfs,
```
andyfengHKU, line 32: if u are passing around clientContext, u probably don't need to pass around vfs

src/main/database.cpp
```diff
@@ -139,6 +140,12 @@ void Database::openLockFile() {
 }
 
 void Database::initAndLockDBDir() {
+    if (DBConfig::isDBPathInMemory(databasePath)) {
+        if (dbConfig.readOnly) {
+            throw Exception("Cannot open an in-memory database under READ ONLY mode.");
```
andyfengHKU, line 145: cover or escape. See if u want to escape across other places in this PR

src/main/database.cpp
```diff
@@ -118,7 +119,7 @@ void Database::addExtensionOption(std::string name, LogicalTypeID type, Value de
     extensionOptions->addExtensionOption(name, type, std::move(defaultValue));
 }
 
-ExtensionOption* Database::getExtensionOption(std::string name) {
+ExtensionOption* Database::getExtensionOption(std::string name) const {
```
andyfengHKU, line 122: ditto

src/storage/wal/wal.cpp
```diff
@@ -34,13 +38,13 @@ WAL::WAL(const std::string& directory, bool readOnly, BufferManager& bufferManag
 WAL::~WAL() {}
 
 void WAL::logBeginTransaction() {
-    lock_t lck{mtx};
+    std::unique_lock<std::mutex> lck{mtx};
```
andyfengHKU, line 41: why are we changing this?

andyfengHKU (approved):

src/storage/wal/wal.cpp
ray6080: I removed the typedef `lock_t` in header file, which I don't think it's necessary.

src/main/database.cpp
ray6080: it's covered in tests. but codecov is not running under in-mem mode, I guess that's why codecov is not reporting them as covered here.

src/include/storage/wal/wal.h
ray6080: This is something I need to discuss with @acquamarin regarding to how vfs should make use of ClientContext. We are passing both in several different places, which should refactored altogether if possible.

src/include/storage/buffer_manager/bm_file_handle.h
```diff
@@ -118,27 +114,32 @@ class BMFileHandle final : public FileHandle {
 
     common::file_idx_t getFileIndex() const { return fileIndex; }
 
+    BufferManager* getBM() { return bm; }
```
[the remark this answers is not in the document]
ray6080: Removed already.

src/include/storage/buffer_manager/bm_file_handle.h
```diff
@@ -32,63 +32,59 @@ class PageState {
 
     PageState() { stateAndVersion.store(EVICTED << NUM_BITS_TO_SHIFT_FOR_STATE); }
 
-    inline uint64_t getState() const { return getState(stateAndVersion.load()); }
-    inline static uint64_t getState(uint64_t stateAndVersion) {
+    uint64_t getState() const { return getState(stateAndVersion.load()); }
```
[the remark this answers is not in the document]
ray6080: Will do this in a separate refactoring PR on BMFileHandle along with this comment https://github.com/kuzudb/kuzu/pull/4012#discussion_r1706376779.

PR #4018, opened 2024-08-03
Fix overflow of `sel_t`

### LINKED ISSUE #4017: Bug: Embedding vectors are not the same as source DataFrame when importing from Pandas
### Kùzu version

master

macOS

### What happened?

When copying from a Pandas DataFrame to a Kùzu node table, the `DOUBLE` or `FLOAT` arrays that represent vector embeddings aren't the same for the entire DataFrame. When a `MATCH` query is run to return the embeddings from the Kùzu table and these are compared with the original DataFrame, in a dataset of 985 rows, 42 rows have the same values, but 943 do not, and their cosine similarity values are < 1, meaning that they are represented by different vectors inside Kùzu.

Code to reproduce is below:

I'm running this code from the latest master as follows:

```py
import shutil
import sys
sys.path.append("/Users/prrao/code/kuzu/tools/python_api/build")

import kuzu
import numpy as np
import pandas as pd
from faker import Faker

def cosine_similarity(a, b):
    return np.dot(a, b) / (np.linalg.norm(a) * np.linalg.norm(b))

database_path = "test_db"

shutil.rmtree(database_path, ignore_errors=True)

db = kuzu.Database(database_path)
conn = kuzu.Connection(db)

try:
    conn.execute("CREATE NODE TABLE User(id INT64, name STRING, embedding DOUBLE[1536], PRIMARY KEY (name))")
except Exception as e:
    print(e)

faker = Faker()
Faker.seed(1)

data = []
for i in range(1000):
    data.append({
        "id": i,
        "name": faker.name(),
        "embedding": np.random.rand(1536).tolist()
    })

df = pd.DataFrame(data).drop_duplicates(subset=["name"])
print("Source DataFrame:", df.shape)

conn.execute("COPY User FROM df")

db_df = conn.execute("MATCH (u:User) RETURN u.*").get_as_df()
print("Database DataFrame:", db_df.shape)

df_merged = pd.merge(df, db_df, left_on="id", right_on="u.id", how="inner")
print("Merged DataFrame:", df_merged.shape)
print(df_merged[["id", "u.id", "embedding", "u.embedding"]].tail(10))

# Check similarity between source and database embeddings
df_merged["sim_score"] = df_merged.apply(lambda row: cosine_similarity(row["embedding"], row["u.embedding"]), axis=1)
print(df_merged[["id", "sim_score"]])
print((df_merged["sim_score"] > 0.99).value_counts())
```
The following result is obtained:

```
Source DataFrame: (985, 3)
Database DataFrame: (985, 3)
Merged DataFrame: (985, 6)
      id  u.id                                          embedding                                        u.embedding
975  989   989  [0.11469857666759731, 0.25736647465744755, 0.7...  [0.7384855615260035, 0.6159779637028786, 0.552...
976  990   990  [0.9801812407704692, 0.6031128989247752, 0.327...  [0.4768895629545835, 0.5446360810116926, 0.819...
977  991   991  [0.28722128218939313, 0.1273117820700359, 0.39...  [0.5659100456824389, 0.2538021641845758, 0.442...
978  992   992  [0.11626981031889738, 0.3826308186698233, 0.90...  [0.8708881007265271, 0.7411757231754165, 0.616...
979  993   993  [0.2141375113923989, 0.10755055124676727, 0.86...  [0.5969850682903762, 0.17366890564670012, 0.19...
980  994   994  [0.441755211182427, 0.12157388640024913, 0.029...  [0.13273759997367962, 0.22715805561243552, 0.7...
981  995   995  [0.06854590302160812, 0.28852409101988175, 0.2...  [0.8896470204647744, 0.5822630661044171, 0.002...
982  996   996  [0.11458441032033362, 0.040105556362980854, 0....  [0.8606024293410687, 0.40924329014364136, 0.49...
983  998   998  [0.41789522987161287, 0.44325303671634675, 0.7...  [0.07072541087142692, 0.5153843838447705, 0.00...
984  999   999  [0.420093420093298, 0.8931212980330067, 0.7289...  [0.5156046624467546, 0.23097723970954975, 0.44...
      id  sim_score
0      0   1.000000
1      1   1.000000
2      2   1.000000
3      3   1.000000
4      4   1.000000
..   ...        ...
980  994   0.754558
981  995   0.753057
982  996   0.755798
983  998   0.748541
984  999   0.759478

[985 rows x 2 columns]
sim_score
False    943
True      42
Name: count, dtype: int64
```

The `sim_score` column that represents the cosine similarity between each `FLOAT[1536]` array in Kùzu and the corresponding numpy array in Pandas, should be 1, as the two columns should be identical.

However, as shown, there are 42 cases for which the two underlying arrays are the same, but 943 cases where they are different from one another (see the indices 975 to 985 in the dataset, as an example).

## Expected result

We should be seeing the `sim_score` output as follows:
```
sim_score
False    0
True     985
```
i.e., the number of entries for which the similarity score is > 0.99 should be 100% of the dataset, because the same underlying arrays should exist in either store.

ray6080:
# Description

In nested types, such as LIST and ARRAY, `uint16_t` is not enough to select from child data vector. 

Close #4017.


andyfengHKU (approved):

tools/python_api/test/test_datatype.py
```diff
@@ -393,3 +396,22 @@ def test_rdf_variant(conn_db_readwrite: ConnDB) -> None:
...
+    conn, db = conn_db_readwrite
+
+    data = []
+    for i in range(1000):
+        data.append({
+            "id": i,
+            "embedding": np.random.rand(1670).tolist()
+        })
+
+    df = pd.DataFrame(data)
```
prrao87, line 411: I think the same test should also be done for PyArrow/Polars too, what do you think @mxwli ?

Issue #4021, opened 2024-08-04
HINT (a join b) join e does not work


semihsalihoglu-uw:
I tried this:
```
explain MATCH (a)-[e]->(b) HINT (a JOIN b) JOIN e RETURN *;
Error: Binder exception: Cannot resolve join condition.
```
Two things: 1) We should either support this or give a better error message; 2) We should explain what the exact conditions we expect from the HINT in the documentation.

andyfengHKU:
This is fixed in #4026

PR #4026, opened 2024-08-04
Update join hint error message

### LINKED ISSUE #4019: Join hint with implicit variable names
I have Person and Likes tables. I tried the following query (which I know won't work but I am testing what the system does):
```
explain match (a:Person)-[:Likes]->(b:Person)-[:Likes]->(c:Person), (a)-[:Likes]->(c) HINT a JOIN b JOIN c return *;
Error: Binder exception: Cannot find  in join hint.
```

We should give a better error message. Something like: "HINT clause requires every node and relationship table variable to be named. You have a relationship pattern with label Likes that does not have a named variable, e.g., -[:Likes]-> should be something like -[e:Likes]->." If the pattern does not have a label, you can change that part as: "...you have an unlabeled relationship pattern, ..., e.g., -[]->". 

If the problem is due to an implicit node variable, then the error message can be something like: "HINT clause requires every node and relationship table variable to be named. You have a node pattern with label Person that does not have a named variable, e.g., (:Person) should be something like (a:Person)." Similarly if the pattern does not have a label, you can change the error message accordingly.

In general it's very useful to give clear examples in error message that point users to the mistake. If you think there are many edge cases and don't want to give an error, you can just give a generic example. For example for the first one you can say: "..., e.g., -[:Foo]-> should be something like -[e:Foo]->." and similarly for the second one: "..., e.g., (:Person) should be something like (a:Person)".

andyfengHKU:
# Description

Fixes #4019 and #4021 

# Contributor agreement

### DIFF
--- a/src/binder/bind/read/bind_match.cpp
+++ b/src/binder/bind/read/bind_match.cpp
@@ -23,6 +23,9 @@ static void validateHintCompleteness(const BoundJoinHintNode& root, const QueryG
     binder::expression_set set;
     collectHintPattern(root, set);
     for (auto& nodeOrRel : queryGraph.getAllPatterns()) {
+        if (nodeOrRel->getVariableName().empty()) {
+            throw BinderException("Cannot hint join order in a match patter with anonymous node or relationship.");
+        }
         if (!set.contains(nodeOrRel)) {
             throw BinderException(
                 stringFormat("Cannot find {} in join hint.", nodeOrRel->toString()));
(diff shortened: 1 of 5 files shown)


ray6080 (approved):

src/binder/bind/read/bind_match.cpp
```diff
@@ -23,6 +23,10 @@ static void validateHintCompleteness(const BoundJoinHintNode& root, const QueryG
     binder::expression_set set;
     collectHintPattern(root, set);
     for (auto& nodeOrRel : queryGraph.getAllPatterns()) {
+        if (nodeOrRel->getVariableName().empty()) {
+            throw BinderException(
+                "Cannot hint join order in a match patter with anonymous node or relationship.");
```
semihsalihoglu-uw, line 28: patter -> pattern

PR #4038, opened 2024-08-06
Fix delete on empty database

### LINKED ISSUE #3897: Bug: Unexpected segmentation fault on empty table
### Kùzu version

v0.4.2

Ubuntu 22.04

### What happened?

Hi! I found that when the table is empty, directly running `MATCH (n) DETACH DELETE n;` will cause an immediate segmentation fault and core dump without any notification. 
image

After compiling with debug info, I found that this query triggered an assertion error `Assertion failed in file "/home/auroraeth/Project/1/kuzu/kuzu/src/binder/expression/node_rel_expression.cpp" on line 18: tableIDs.size() == 1`.
Perhaps it would be better to do some exception handling? Thank you!

_No response_

andyfengHKU:
# Description

Fixes #3897. Query compilation fails when we try to delete from an empty database.

Add test cases for #3807 

# Contributor agreement

### DIFF
--- a/test/test_files/transaction/delete_rel/delete_empty.test
+++ b/test/test_files/transaction/delete_rel/delete_empty.test
@@ -1,6 +1,10 @@
 -DATASET CSV empty
 --
 
+-CASE DeleteRelEmptyDB
+-STATEMENT MATCH ()-[e]->() DELETE e;
+---- ok
+
 -CASE RelDeleteCheckpoint
 -STATEMENT CREATE NODE TABLE person (ID INT64, PRIMARY KEY (ID));
 ---- ok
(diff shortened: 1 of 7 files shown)


test/test_files/transaction/delete_rel/delete_empty.test
```diff
@@ -1,41 +1,45 @@
...
--STATEMENT MATCH (:person)-[e:knows]->(:person) RETURN COUNT(e);
----- 1
-0
--STATEMENT CHECKPOINT;
----- ok
--STATEMENT MATCH (:person)-[e:knows]->(:person) RETURN COUNT(e);
----- 1
-0
+
+#-CASE RelDeleteCheckpoint
```
ray6080, line 8: Uncomment this test.

ray6080 (approved):
Can u change the PR title to a more descriptive one? I think mostly this PR is fixing the issue of delete under empty tables, instead of adding test cases.

PR #4039, opened 2024-08-07
Make property case insensitive

andyfengHKU:
# Description

This PR contains the following changes

### Make property case insensitive

We have already make table name case insensitive, it does not make sense for our property to remain case sensitive. So we start applying case-insensitive-map everywhere when binding properties.

### Remove property_id_t

We use to have an `o_id` like identifier for property, i.e. `property_id_t`. Though I start to realize this is unnecessary. We could just use property name as the identifier since property should be unique within a table.

Fixes # (issue)

# Contributor agreement


test/test_files/rdf/tinysnb.test
```diff
@@ -9,12 +9,9 @@
 -STATEMENT CREATE (a:R_r {iri: "a"})-[e:R_rt {iri: "b"}]->(b:R_r {iri: "c"}) RETURN a.iri, e.iri, b.iri;
 ---- 1
 a|b|c
--STATEMENT MATCH (a)-[e:knows|:R]->(b) RETURN *;
+-STATEMENT MATCH (a:organisation)-[e:knows|:R]->(b:organisation) RETURN *;
```
ray6080, line 12: Why change the test case here?

test/test_files/rdf/tinysnb.test
```diff
@@ -9,12 +9,9 @@
...
 ---- error
 Binder exception: Relationship pattern e contains both PropertyGraph relationship label knows and RDFGraph label R. Mixing relationships tables from an RDFGraph and PropertyGraph in one pattern is currently not supported.
--STATEMENT MATCH (a)-[:marries|:R*]->(b) RETURN *;
----- error
-Binder exception: Relationship pattern  contains both PropertyGraph relationship label marries and RDFGraph label R. Mixing relationships tables from an RDFGraph and PropertyGraph in one pattern is currently not supported.
```
ray6080, line 17: Ditto. why remove these?

ray6080 (approved):
I'm expecting more test cases that explicitly test case insensitive. some may already be covered in some level, but I just list high-level categories here:
- DDL. alter table rename, add and drop property; create table with duplicated property names; 
- DML. create/merge/set
- Match return properties.
- Extensions: match on duckdb/sqlite/postgres external tables.

Does this automatically apply to LOAD FROM? If so, we should also add tests on that.

test/test_files/rdf/tinysnb.test
andyfengHKU: Because resource iri table has `id` as `SERIAL`, and person table has `ID` as `INT64`, we now consider them as the same property with different type.

test/test_files/rdf/tinysnb.test
andyfengHKU: same reason

scripts/antlr4/Cypher.g4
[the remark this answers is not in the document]
andyfengHKU: yes

src/catalog/catalog_entry/table_catalog_entry.cpp
```diff
@@ -47,89 +44,115 @@ std::unique_ptr<TableCatalogEntry> TableCatalogEntry::alter(const BoundAlterInfo
...
+void TableCatalogEntry::vacuumColumnIDs() {
+    propertyCollection.vacuumColumnIDs();
 }
 
-const Property* TableCatalogEntry::getProperty(property_id_t propertyID) const {
-    auto it = std::find_if(properties.begin(), properties.end(),
-        [&propertyID](const auto& property) { return property.getPropertyID() == propertyID; });
-    KU_ASSERT(it != properties.end());
-    return &(*it);
+//<<<<<<< HEAD
```
ray6080, line 55: Clean this.

src/catalog/catalog_entry/table_catalog_entry.cpp
```diff
@@ -47,89 +44,115 @@ std::unique_ptr<TableCatalogEntry> TableCatalogEntry::alter(const BoundAlterInfo
...
-    return it->getColumnID();
+common::idx_t TableCatalogEntry::getPropertyIdx(const std::string& propertyName) const {
+    return propertyCollection.getIdx(propertyName);
 }
 
-void TableCatalogEntry::addProperty(std::string propertyName, LogicalType dataType,
-    std::unique_ptr<parser::ParsedExpression> defaultExpr) {
-    properties.emplace_back(std::move(propertyName), std::move(dataType), std::move(defaultExpr),
-        nextPID++, nextColumnID++, tableID);
+//<<<<<<< HEAD
```
ray6080, line 81: Ditto. Clean stuff there.

src/catalog/catalog_entry/table_catalog_entry.cpp
```diff
@@ -47,89 +44,115 @@ std::unique_ptr<TableCatalogEntry> TableCatalogEntry::alter(const BoundAlterInfo
...
+void TableCatalogEntry::dropProperty(const std::string& propertyName) {
+    propertyCollection.drop(propertyName);
 }
 
+void TableCatalogEntry::renameProperty(const std::string& propertyName,
+    const std::string& newName) {
+    propertyCollection.rename(propertyName, newName);
+}
+
+// void TableCatalogEntry::resetColumnIDs() {
```
ray6080, line 134: Ditto clean up.

PR #4045, opened 2024-08-08
Merge BMFileHandle and FileHandle

ray6080:
# Description

Merge `BMFileHandle` and `FileHandle`.


benjaminwinger (approved):
Generally looks good to me.

I do wonder a little if the FileHandle in its current state is doing too much (even before the merge). ColumnChunkData doesn't really need the buffer manager side of things, it just has to write to disk while keeping the state information up to date, while similarly Column should just be updating data through the buffer manager and doesn't need the ability to write directly to the file.

Maybe we need to change how the functionality is divided. E.g. BMFileHandle could handle writing through the buffer manager, and FileHandle could handle direct writes through `writePagesToFile`, etc. Then they could both internally share some metadata (the FileInfo, number of pages, page states, etc.).
I'm not suggesting that this should be changed now (that division would have no relationship to the current one, so there would be no disadvantage if we merge this now and change them again later), but I think it's worth considering.

ray6080:
> I do wonder a little if the FileHandle in its current state is doing too much (even before the merge).

Yeah. I have the same concern too. My previous thought is to decouple the read/write to file to a util class that will take in FileHandle instead to get the fileInfo for reads and writes. While FileHandle will appear to be manly interact with BufferManager.
`BMFileHandle could handle writing through the buffer manager, and FileHandle could handle direct writes through writePagesToFile, etc. Then they could both internally share some metadata (the FileInfo, number of pages, page states, etc.).` This makes sense to me. Let me think a bit how this can be done in a separate PR.

PR #4052, opened 2024-08-09
Refactor list auxiliary buffer

andyfengHKU:
# Description

This PR solves a bug when we try to refer a list child vector from another list child vector. The bug was because we only set the vector itself but did not copy the corresponding auxiliary buffer metadata, e.g. size & capacity. This PR fixes the bug and also refactor related APIs.

Fixes # (issue)

# Contributor agreement

### DIFF
--- a/src/include/common/vector/value_vector.h
+++ b/src/include/common/vector/value_vector.h
@@ -146,61 +146,54 @@ struct KUZU_API BlobVector {
     }
 };
 
-// Currently, ListVector is used for both VAR_LIST and ARRAY physical type
+// ListVector is used for both LIST and ARRAY physical type
 class KUZU_API ListVector {
 public:
-    static void setDataVector(const ValueVector* vector, std::shared_ptr<ValueVector> dataVector) {
-        KU_ASSERT(vector->dataType.getPhysicalType() == PhysicalTypeID::LIST ||
-                  vector->dataType.getPhysicalType() == PhysicalTypeID::ARRAY);
-        auto listBuffer =
-            ku_dynamic_cast<AuxiliaryBuffer*, ListAuxiliaryBuffer*>(vector->auxiliaryBuffer.get());
-        listBuffer->setDataVector(std::move(dataVector));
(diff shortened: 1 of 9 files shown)


src/include/common/vector/value_vector.h
```diff
@@ -146,61 +146,54 @@ struct KUZU_API BlobVector {
...
 public:
+    static const ListAuxiliaryBuffer& getBuffer(const ValueVector& vector) {
+        return vector.auxiliaryBuffer->constCast<ListAuxiliaryBuffer>();
+    }
+    static ListAuxiliaryBuffer& getBufferUnsafe(const ValueVector& vector) {
+        return vector.auxiliaryBuffer->cast<ListAuxiliaryBuffer>();
+    }
+    // If you call setDataVector during initialize, there must be a followed up
+    // copyListEntryAndBufferMetaData at runtime.
+    // TODO(Xiyang): try to merge setDataVector & copyListEntryAndBufferMetaData
```
ray6080, line 160: Can we fix this TODO in this PR? I feel it weird to have separate `setDataVector` and `copyListEntryAndBufferMetaData`.

ray6080 (approved):

src/include/common/vector/value_vector.h
andyfengHKU: I'm leaving this TODO because I cannot fix it now. For `properties()` the idx of data vector is only available at compile time. I cannot do it at runtime.

src/planner/operator/factorization/flatten_resolver.cpp
```diff
@@ -192,7 +192,8 @@ void GroupDependencyAnalyzer::visitNodeOrRel(std::shared_ptr<binder::Expression>
         auto& node = expr->constCast<NodeExpression>();
         visit(node.getInternalID());
     } break;
-    case LogicalTypeID::REL: {
+    case LogicalTypeID::REL:
+    case LogicalTypeID::RECURSIVE_REL: {
```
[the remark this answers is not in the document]
andyfengHKU: yes

PR #4058, opened 2024-08-12
Optimize catalog access

andyfengHKU:
# Description

`getTableCatalogEntry` API in the catalog is introducing noticeable overhead when binding large query graph since we now need to check different versions of entry under MVCC.

This PR directly save `TableCatalogEntry*` in node/rel expression instead of `tableID` to avoid repeated access of catalog entries.

Fixes # (issue)

# Contributor agreement

### DIFF
--- a/src/binder/expression/node_rel_expression.cpp
+++ b/src/binder/expression/node_rel_expression.cpp
@@ -1,7 +1,7 @@
 #include "binder/expression/node_rel_expression.h"
 
-#include "common/exception/runtime.h"
 #include "catalog/catalog_entry/table_catalog_entry.h"
+#include "common/exception/runtime.h"
 
 using namespace kuzu::catalog;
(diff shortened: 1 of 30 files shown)


src/binder/expression/node_rel_expression.cpp
```diff
@@ -1,29 +1,47 @@
...
 #include "common/exception/runtime.h"
 
+using namespace kuzu::catalog;
 using namespace kuzu::common;
 
 namespace kuzu {
 namespace binder {
 
-void NodeOrRelExpression::addTableIDs(const table_id_vector_t& tableIDsToAdd) {
+table_id_vector_t NodeOrRelExpression::getTableIDsX() const {
```
ray6080, line 12: What's `X` here? You meant `getTableIDsVec`?

ray6080 (approved):

PR #4060, opened 2024-08-12
Enabled progress bar by default

### LINKED ISSUE #4025: Enable progress bar by default
I think the progress bar in CLI and Explorer are very good features and they should be enabled by default. I remember that there was an argument that if someone pipes their CLI results in a shell script that the progress bar pollutes the logs. This seems to be an edge case enough that it shouldn't determine our choice. We should go for better usability when possible.

hamzakammar:
# Description

Fixes #4025 

Enabled Progress Bar by Default

# Contributor agreement


andyfengHKU (changes requested):
We shouldn't turn on progress bar by default in the kernel. Let the shell/explorer turn on progress bar on the time of start. 

I'll let @MSebanc and @mewim handle this.

PR #4065, opened 2024-08-13
Enabled progress bar by default for shell

### LINKED ISSUE #4025: Enable progress bar by default

MSebanc:
# Description

Set progress bar to be enabled by default for shell

Fixes #4025 

# Contributor agreement


andyfengHKU (commented):
So does this mean we won't have a config file for the shell?

MSebanc:
> So does this mean we won't have a config file for the shell?

I'll still implement the config file. The config file does more than just setting the commands at the start, it allows you to run queries during start up too. I can hold off on merging this pr and we can see if it makes sense to have the commands as flags as well. 

This pr was mainly to have the progress bar turned on by default and I added the other couple commands in as flags because it was simple.

andyfengHKU (approved):

PR #4073, opened 2024-08-13
Refactor generation of grammar files. 

### LINKED ISSUE #3683: Remove Cypher.g4.copy and keywords.txt.copy file from repo
### Description

We currently use `Cypher.g4.copy` and `keywords.txt.copy` under `script/anltr4` to check if the grammar file has any changes. These copied files are not necessary, since we can store a hash value to serve the same purpose.

mxwli:
Uses `md5sum` to generate hash of cypher.g4 and keywords.txt
Hash replaces file comparison.
Closes #3683


ray6080 (approved):
Thanks!

PR #4074, opened 2024-08-13
Vacuum dropped columns during checkpoint

ray6080:
# Description

The PR is to fix a bug due to not vacuum dropped columns during checkpoint. The bug happens when we recover after dropped a column and performed checkpoint. During the recovery, we will have some columns as nullptr, which lead to seg faults when we collect data types from table columns.

This PR solves the issue by always vacuuming dropped columns from storage during checkpoint. Note that the column ID inside property will be reset due to the vacuuming, that's why the checkpoint of catalog now happens after the checkpoint of tables. Letting storage modifies catalog may not be a good practice, alternatively, we can let `TableCatalogEntry` reset its column ids during checkpoint independently, which sounds like a better idea to me.

In the future, we should also reclaim disk pages of persistent column chunks being dropped during checkpoint.

A minor refactoring:
- aligned columnID for rel table properties (`nextColumnID` starts from 1 to skip the anonymous NBR_ID column inside `RelTableCatalogEntry`). thus removed the special implementation of `RelTableCatalogEntry::getColumnID`.

### DIFF
--- a/src/include/catalog/catalog_entry/table_catalog_entry.h
+++ b/src/include/catalog/catalog_entry/table_catalog_entry.h
@@ -42,7 +42,6 @@ public:
     std::string getComment() const { return comment; }
     void setComment(std::string newComment) { comment = std::move(newComment); }
     virtual bool isParent(common::table_id_t /*tableID*/) { return false; };
-    // TODO(Guodong/Ziyi): This function should be removed. Instead we should use CatalogEntryType.
     virtual common::TableType getTableType() const = 0;
     virtual function::TableFunction getScanFunction() { KU_UNREACHABLE; }
     binder::BoundAlterInfo* getAlterInfo() const { return alterInfo.get(); }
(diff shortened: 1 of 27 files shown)


src/include/catalog/catalog_entry/table_catalog_entry.h
```diff
@@ -56,9 +55,11 @@ class KUZU_API TableCatalogEntry : public CatalogEntry {
     //===--------------------------------------------------------------------===//
     uint32_t getNumProperties() const { return properties.size(); }
     const std::vector<Property>& getPropertiesRef() const { return properties; }
+    std::vector<Property>& getPropertiesRef() { return properties; }
```
andyfengHKU, line 58: Let's call this getPropertiesUnsafe();

src/include/catalog/catalog_entry/table_catalog_entry.h
```diff
@@ -56,9 +55,11 @@ class KUZU_API TableCatalogEntry : public CatalogEntry {
...
     bool containProperty(const std::string& propertyName) const;
     common::property_id_t getPropertyID(const std::string& propertyName) const;
     const Property* getProperty(common::property_id_t propertyID) const;
+    Property* getProperty(common::property_id_t propertyID);
```
andyfengHKU, line 62: double check if this is needed or not.

andyfengHKU (approved):

PR #4118, opened 2024-08-21
Fix JSON null handling

mxwli:
# Description

JSON assigned NULL values to int8 as a placeholder. This doesn't work out too good if you try to put it through maxlogicaltype. 
```json
[
null,
{"col1": 1, "col2": 2, "col3": "asdf"}
]
```
In the above case, it would assign the first object to INT8, then the second object to STRUCT. It'd then combine the two to deduce that we're reading from a one column table and interpret the whole thing as string.

It now assigns nulls to be ANY, and then purges the result to turn any instances of ANY into STRING. The above example now works as expected.

Fixes # (issue)

# Contributor agreement


mxwli:
Feel free to merge this if you don't have any feedback; it's a tiny PR.

PR #4134, opened 2024-08-23
Added cli support for single line comments and added multiline mode

MSebanc:
# Description

The shell now supports single line comments when writing a query over multiple lines. As well, there is a new multiline mode that allows going back to previous lines and making changes. However, there is a little more work that needs to be done for that mode, so it is not enabled by default. It is functional though and good to add in, just minor additions are needed over time. Can be enabled by typing `:multiline`.

```
kuzu> :multiline
Multi line mode enabled
kuzu> CREATE NODE TABLE
    · Person(name STRING,
    · age INT64,
    ‣ PRIMARY KEY (name) // a comment here too
    · );
```
The `‣` indicates that my cursor is currently on that line.

# Contributor agreement

### DIFF
--- a/tools/shell/embedded_shell.cpp
+++ b/tools/shell/embedded_shell.cpp
@@ -329,13 +329,13 @@ std::vector<std::unique_ptr<QueryResult>> EmbeddedShell::processInput(std::strin
     if (!continueLine && input[0] == ':') {
         processShellCommands(input);
         // process queries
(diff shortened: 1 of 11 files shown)


andyfengHKU (approved):
Cool. I would skip reviewing linenoise as I'm no longer familiar with it. I would try to understand it on-demand (i.e. when bug happens)

PR #4158, opened 2024-08-27
Ignore node batch insert failures

royi-luo:
# Description

Partial addresses https://github.com/kuzudb/kuzu/issues/1872

Add the ability to ignore copy errors related to primary-key constraints (duplicate/null PK). All copy errors (which we wish to be skippable) in node batch insert should now be ignoreable.

Some additional changes:

# Contributor agreement


src/function/cast_from_string_functions.cpp
```diff
@@ -781,7 +781,7 @@ static bool tryCastUnionField(ValueVector* vector, uint64_t rowToAdd, const char
     } break;
     case LogicalTypeID::STRING: {
         if (!utf8proc::Utf8Proc::isValid(input, len)) {
-            throw CopyException{"Invalid UTF8-encoded string."};
+            throw ConversionException{"Invalid UTF8-encoded string."};
```
royi-luo, line 784: Add tests for skipping this

src/include/processor/operator/persistent/index_builder.h
```diff
@@ -30,21 +31,21 @@ class IndexBuilderGlobalQueues {
...
     template<typename T>
-    void insert(size_t index, storage::IndexBuffer<T> elem) {
+    void insert(size_t index, storage::IndexBuffer<T> elem, IndexBuilderErrorHandler& errors) {
         auto& typedQueues = std::get<Queue<T>>(queues).array;
         typedQueues[index].push(std::move(elem));
         if (typedQueues[index].approxSize() < SHOULD_FLUSH_QUEUE_SIZE) {
             return;
         }
-        maybeConsumeIndex(index);
+        return maybeConsumeIndex(index, errors);
```
royi-luo, line 40:
```suggestion
        maybeConsumeIndex(index, errors);
```

src/include/processor/operator/persistent/node_batch_insert.h
```diff
@@ -114,16 +116,16 @@ class NodeBatchInsert final : public BatchInsert {
     // written
     void writeAndResetNodeGroup(transaction::Transaction* transaction,
         std::unique_ptr<storage::ChunkedNodeGroup>& nodeGroup,
-        std::optional<IndexBuilder>& indexBuilder) const;
+        std::optional<IndexBuilder>& indexBuilder);
```
royi-luo, line 119: add back const

src/processor/operator/persistent/index_builder.cpp
```diff
@@ -92,51 +92,68 @@ void IndexBuilderSharedState::quitProducer() {
...
-                auto value = stringColumnChunk.getValue<std::string>(i);
-                localBuffers.insert(std::move(value), nodeOffset + i);
+                if (checkNonNullConstraint(chunk, nodeOffset, i, errors)) {
+                    auto value = stringColumnChunk.getValue<std::string>(i);
+                    localBuffers.insert(std::move(value), nodeOffset + i, errors);
+                }
             }
         },
-        [&](auto) {
+        [&]<typename T>(T) {
```
royi-luo, line 117: revert

src/processor/operator/persistent/index_builder.cpp
```diff
@@ -92,51 +92,68 @@ void IndexBuilderSharedState::quitProducer() {
...
             throw CopyException(ExceptionMessage::invalidPKType(chunk.getDataType().toString()));
```
royi-luo, line 118: Don't think ignoring this error makes sense (since it applies to an entire column) so didn't make it skippable

src/processor/operator/persistent/index_builder.cpp
```diff
@@ -92,51 +92,68 @@ void IndexBuilderSharedState::quitProducer() {
...
+bool IndexBuilder::checkNonNullConstraint(const ColumnChunkData& chunk, offset_t nodeOffset,
+    offset_t chunkOffset, IndexBuilderErrorHandler& errors) {
+    const auto& nullChunk = chunk.getNullData();
+    if (nullChunk.isNull(chunkOffset)) {
+        TypeUtils::visit(
+            chunk.getDataType().getPhysicalType(),
+            [&](struct_entry_t) {
+                // primary key cannot be struct
+                KU_UNREACHABLE;
+            },
```
royi-luo, line 147: added since `handleOrStoreError<struct_entry_t>` doesn't compile

src/processor/operator/persistent/index_builder_error_handler.cpp
```diff
@@ -0,0 +1,61 @@
...
+      context(context), pkType(pkType), nodeTable(nodeTable), queryID(queryID) {}
+
+void IndexBuilderErrorHandler::addNewVectors() {
+    offsetVector.push_back(std::make_shared<ValueVector>(LogicalTypeID::INTERNAL_ID,
+        context->clientContext->getMemoryManager()));
+    offsetVector.back()->state = DataChunkState::getSingleValueDataChunkState();
+    keyVector.push_back(
+        std::make_shared<ValueVector>(pkType, context->clientContext->getMemoryManager()));
+    keyVector.back()->state = DataChunkState::getSingleValueDataChunkState();
+}
```
royi-luo, line 24: Made vectors size 1 because `NodeTable::delete_()` requires the vector to be size 1. If this assertion is safe to remove I can update to make vectors larger

src/processor/operator/persistent/index_builder_error_handler.cpp
```diff
@@ -0,0 +1,61 @@
...
+            *offsetVector[i],
+            *keyVector[i],
+        };
+        nodeTable->delete_(context->clientContext->getTx(), deleteState);
+
+        populatedErrors.push_back({
+            .message = errorMessages[i],
+            .filePath = "",
+            .skippedLine = "",
+            .lineNumber = 0,
```
royi-luo, line 40: These fields are to be populated in a future PR

src/processor/operator/persistent/node_batch_insert.cpp
```diff
@@ -198,10 +207,18 @@ void NodeBatchInsert::finalizeInternal(ExecutionContext* context) {
         }
     }
     if (nodeSharedState->globalIndexBuilder) {
-        nodeSharedState->globalIndexBuilder->finalize(context);
+        auto* localNodeState =
+            ku_dynamic_cast<BatchInsertLocalState*, NodeBatchInsertLocalState*>(localState.get());
+        KU_ASSERT(localNodeState->errorHandler.has_value());
+        nodeSharedState->globalIndexBuilder->finalize(context,
+            localNodeState->errorHandler.value());
+        localNodeState->errorHandler->flushStoredErrors();
```
royi-luo, line 215: Just to double-check accessing the local node state during `finalize` shouldn't be a problem right? Otherwise I'll just create a new error handler instance.

src/include/processor/operator/persistent/index_builder_error_handler.h
```diff
@@ -0,0 +1,65 @@
...
+    static constexpr common::idx_t DELETE_VECTOR_SIZE = 1;
+    static constexpr uint64_t LOCAL_WARNING_LIMIT = 1024;
+
+    bool ignoreErrors;
+    uint64_t warningLimit;
+    ExecutionContext* context;
+    common::LogicalTypeID pkType;
+    storage::NodeTable* nodeTable;
+    uint64_t queryID;
+    std::atomic<common::row_idx_t>* sharedDeletedRowCounter;
```
ray6080, line 54: I don't enjoy passing atomic pointer around like this. Can u make pass the shared_ptr? Also I don't think we need `atomic` here, grabbing the lock if you can seems better to me.

src/include/processor/operator/persistent/index_builder_error_handler.h
```diff
@@ -0,0 +1,65 @@
...
+
+namespace processor {
+template<typename T>
+struct IndexBuilderError {
+    std::string message;
+    T key;
+    common::nodeID_t nodeID;
+};
+
+class IndexBuilderErrorHandler {
```
ray6080, line 20: Can u rename this to `NodeBatchInsertErrorHandler`?

ray6080 (commented):
Let me take another quick Look later.

src/include/main/client_config.h
```diff
@@ -33,6 +33,8 @@ struct ClientConfig {
     uint32_t recursivePatternCardinalityScaleFactor;
     // maximum number of cached warnings
     uint64_t warningLimit;
+    // If ignoring copy errors
```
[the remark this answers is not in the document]
royi-luo: removed

src/processor/operator/persistent/index_builder.cpp
```diff
@@ -30,13 +29,13 @@ PhysicalTypeID IndexBuilderGlobalQueues::pkTypeID() const {
...
-void IndexBuilderGlobalQueues::consume() {
+void IndexBuilderGlobalQueues::consume(IndexBuilderErrorHandler& errors) {
     for (auto index = 0u; index < NUM_HASH_INDEXES; index++) {
-        maybeConsumeIndex(index);
+        maybeConsumeIndex(index, errors);
     }
 }
 
-void IndexBuilderGlobalQueues::maybeConsumeIndex(size_t index) {
+void IndexBuilderGlobalQueues::maybeConsumeIndex(size_t index, IndexBuilderErrorHandler& errors) {
```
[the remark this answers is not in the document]
royi-luo: updated

src/include/processor/operator/persistent/index_builder_error_handler.h
```diff
@@ -0,0 +1,65 @@
...
+    common::LogicalTypeID pkType;
+    storage::NodeTable* nodeTable;
+    uint64_t queryID;
+    std::atomic<common::row_idx_t>* sharedDeletedRowCounter;
+
+    std::vector<std::shared_ptr<common::ValueVector>> keyVector;
+    std::vector<std::shared_ptr<common::ValueVector>> offsetVector;
+    std::vector<std::string> errorMessages;
+
+    uint64_t getNumErrors() const;
```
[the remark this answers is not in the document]
royi-luo: updated

src/include/processor/operator/persistent/index_builder_error_handler.h
royi-luo: updated

src/include/processor/operator/persistent/index_builder_error_handler.h
royi-luo: updated

src/processor/map/map_copy_from.cpp
```diff
@@ -87,9 +87,23 @@ std::unique_ptr<PhysicalOperator> PlanMapper::mapCopyNodeFrom(LogicalOperator* l
         columnTypes.push_back(expr->getDataType().copy());
         columnEvaluators.push_back(exprMapper.getEvaluator(expr));
     }
-    auto info =
-        std::make_unique<NodeBatchInsertInfo>(nodeTableEntry, storageManager->compressionEnabled(),
-            std::move(columnTypes), std::move(columnEvaluators), copyFromInfo->columnEvaluateTypes);
+    bool ignoreErrors = CopyConstants::DEFAULT_IGNORE_ERRORS;
+    if (copyFromInfo->source->type == common::ScanSourceType::FILE) {
```
royi-luo, line 91: I'm assuming `IGNORE_ERRORS` only applies when copying from a file right? I don't think we have options when copying from a subquery.

ray6080: Nope. We might need to add it later.

src/processor/operator/persistent/node_batch_insert.cpp
ray6080: Should be fine here.

src/processor/operator/persistent/node_batch_insert.cpp
```diff
@@ -151,11 +158,14 @@ void NodeBatchInsert::writeAndResetNodeGroup(transaction::Transaction* transacti
     std::unique_ptr<ChunkedNodeGroup>& nodeGroup, std::optional<IndexBuilder>& indexBuilder) const {
     const auto nodeSharedState =
         ku_dynamic_cast<BatchInsertSharedState*, NodeBatchInsertSharedState*>(sharedState.get());
+    const auto nodeLocalState =
+        ku_dynamic_cast<BatchInsertLocalState*, NodeBatchInsertLocalState*>(localState.get());
```
ray6080, line 162: Use `ptrCast` instead of `ku_dynamic_cast` here.

ray6080 (approved):
LGTM. Can u do a quick benchmark too, just similar to what we did with LOAD FROM?

src/processor/operator/persistent/node_batch_insert.cpp
royi-luo: updated

royi-luo:
Some quick benchmarks:
Query (each dataset has 10M nodes):
```
CREATE NODE TABLE person (value INT32, PRIMARY KEY(value));
COPY person FROM "{file}.csv"(IGNORE_ERRORS=true);
```
No errors (ignore_errors=false): 648ms
No errors (ignore_errors=true): 647ms
10% duplicate PK (ignore errors=true): 982ms 
10% duplicate PK (ignore errors=true): 1358ms 
All duplicate PK (ignore errors=true): 5054ms

src/include/storage/index/in_mem_hash_index.h
```diff
@@ -80,21 +80,20 @@ class InMemHashIndex final {
     using Key = std::conditional_t<std::same_as<T, common::ku_string_t>, std::string_view, T>;
     // Appends the buffer to the index. Returns the number of values successfully inserted.
     // I.e. if a key fails to insert, its index will be the return value
-    size_t append(const IndexBuffer<BufferKeyType>& buffer, visible_func isVisible) {
-        reserve(indexHeader.numEntries + buffer.size());
+    size_t append(const IndexBuffer<BufferKeyType>& buffer, uint64_t bufferOffset,
+        visible_func isVisible) {
+        reserve(indexHeader.numEntries + buffer.size() - bufferOffset);
         // Do both searches after splitting. Returning early if the key already exists isn't a
         // particular concern and doing both after splitting allows the slotID to be reused
```
royi-luo, line 87: @ray6080 do my changes have any implications on this comment? I'm not too sure how the slotID reuse works but it does mention returning early

src/include/storage/index/hash_index.h
```diff
@@ -125,21 +125,20 @@ class HashIndex final : public OnDiskHashIndex {
...
             common::offset_t result;
-            for (size_t i = 0; i < buffer.size(); i++) {
+            for (size_t i = bufferOffset; i < buffer.size(); i++) {
                 const auto& [key, value] = buffer[i];
                 if (lookupInPersistentIndex(transaction, key, result, isVisible)) {
-                    return i;
+                    return i - bufferOffset;
                 }
             }
         }
```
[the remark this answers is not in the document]
royi-luo: k never mind the amount of duplication is a bit egregious so I'll fix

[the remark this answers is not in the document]
royi-luo: updated

src/storage/store/node_table.cpp
```diff
@@ -392,7 +392,7 @@ void NodeTable::serialize(Serializer& serializer) const {
 
 bool NodeTable::isVisible(const Transaction* transaction, offset_t offset) const {
     auto [nodeGroupIdx, offsetInGroup] = StorageUtils::getNodeGroupIdxAndOffsetInChunk(offset);
-    auto* nodeGroup = getNodeGroupNoLock(nodeGroupIdx);
+    auto* nodeGroup = getNodeGroup(nodeGroupIdx);
```
royi-luo, line 395: Ran into occasional segfaults due to nodeGroup because sometimes NULL (when running with many threads). Let me know if anyone has any more insight on this but I think there might have previously been some data races here (maybe on reallocating the nodeGroup vector, etc.)

ray6080: Nice catch! You're right, there can be race conditions during node batch insert.
However, the change here can cause performance regression with index lookups, especially during copy rel pipeline, we need to do lookups to figure out bound/dst node offsets first, under which case, locking is not required.

Can u add `isVisibleNoLock` and let `lookupPK` go through that?

src/include/storage/index/in_mem_hash_index.h
benjaminwinger: The comment should be removed at this point. This was the original context, but I guess it wasn't updated after changes were made.
https://github.com/kuzudb/kuzu/blob/e5eda4c60d2b4ee8acb1962724c8518395bafff9/src/storage/index/hash_index_builder.cpp#L125-L152
The idea was that checking if the key already exists is better done after we reserve space for it since returning early at that time was only when we're failing and we would have to re-compute the slot ID in case it changed after splitting some of the slots. 
Of course in the context of this PR we may now care about the performance costs of trying to insert values which already exist. However I think the current way of doing things, where values that already exist have the same cost as ones that don't, is probably fine. Even if every single key already exists, the amount of space reserved will be small since it will never be filled, so the extra cost of doing that before checking if the key exists should be minimal.

royi-luo: Cool, thanks Ben!

src/storage/store/node_table.cpp
royi-luo: updated

src/include/storage/index/in_mem_hash_index.h
```diff
@@ -80,21 +80,20 @@ class InMemHashIndex final {
...
+        for (size_t i = bufferOffset; i < buffer.size(); i++) {
             hashes[i] = HashIndexUtils::hash(buffer[i].first);
-        }
-        for (size_t i = 0; i < buffer.size(); i++) {
             auto& [key, value] = buffer[i];
             if (!appendInternal(key, value, hashes[i], isVisible)) {
-                return i;
+                return i - bufferOffset;
             }
         }
```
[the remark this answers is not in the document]
benjaminwinger: A little late to comment on this, but if you're combining the loops then you may as well get rid of the hashes array and pass the results directly to appendInternal.

I'd seen similar results when I'd written it initially. It's worth noting that it doesn't help that we're targeting the default x86-64-v1 microarchitecture (from ~2003), so any any such tests should also try with newer instruction sets which could speed up auto-vectorized code (though I'd tried a while ago and didn't see a noticeable improvement yet from the newer micro-architectures; if we get to the point that there is then we might want to look at either targeting a newer version by default, or providing multiple builds).
I think it would require a significant amount of time to look into, but it might be beneficial at some point to try and get some vectorization working properly.

I have a suspicion that in the two loop version, the bulk of the CPU time is being spent in the second one, which isn't well suited to running multiple instructions in parallel (with all the branching), so in the combined version the hash calculations may actually be able to be done simultaneously with the more serial branching code (I don't know how much of this the CPU can pipeline in reality, but while the results of the hash function are needed immediately in appendInternal, the calculation of the next hash would be independent of the end of the previous call to appendInternal, and can just be discarded if the function has to return early). That's not to say there will be an improvement with the combined one; the pipelining may still slow things down slightly, but it's a hypothesis as to why there is little difference even assuming that the vectorized hashing is working properly in the two loop version.
This probably just isn't a very good place for vectorization.

PR #4162, opened 2024-08-29
Make page size be flexible configurable

ray6080:
# Description

- Fixes the overflow of PageCursor due to tweaking of default page size.
- Add ci workflow to run under 64KB, 256KB page sizes.


src/storage/store/compression_flush_buffer.cpp
```diff
@@ -26,48 +25,45 @@ ColumnChunkMetadata CompressedFlushBuffer::operator()(const uint8_t* buffer,
...
-            memcpy(frame, compressedBuffer.get(), BufferPoolConstants::PAGE_4KB_SIZE);
+            memcpy(frame, compressedBuffer.get(), PAGE_SIZE);
         } else {
-            dataFH->getFileInfo()->writeFile(compressedBuffer.get(),
-                BufferPoolConstants::PAGE_4KB_SIZE,
-                (startPageIdx + metadata.numPages - 1) * BufferPoolConstants::PAGE_4KB_SIZE);
+            dataFH->getFileInfo()->writeFile(compressedBuffer.get(), PAGE_SIZE,
+                (startPageIdx + metadata.numPages - 1) * PAGE_SIZE);
         }
+        numPages++;
```
benjaminwinger, line 66: `numPages` isn't being used in the loop. It was just running once since writing to the last page will extend the file to that length. We shouldn't need to write to each individual page (and making this a loop just means that it writes the last page multiple times). I don't know if there is a noticeable performance benefit, but writing just the last page avoids the need for multiple system calls and the file system can handle the resizing internally.

I also have a suspicion that this code block is entirely unnecessary in in-memory-mode since the frames should default to zeroed and we've already reserved the pages in memory.

benjaminwinger (approved):

src/storage/store/compression_flush_buffer.cpp
ray6080: Yeah make sense, only writing to the last page is needed. I got it wrong here.

> I also have a suspicion that this code block is entirely unnecessary in in-memory-mode since the frames should default to zeroed and we've already reserved the pages in memory.
Nice catch on this. I think so. Added an extra check to avoid the write under in-mem mode.

PR #4163, opened 2024-08-30
list lambda body support use parameter more than once

excaliburwyj:
# Description

lambda body support use parameter more than once
for list_filter, list_transform:
the lambda only has one parameter name support: x->x+x, x->x>1 and x<5
for list_reduce:
the lambda has two parameters, the parameter on the left indicates the first element in the list or the result of the previous reduce operation, the right one is just the subsequent element in the list.
e.g. RETURN LIST_REDUCE(['a:', 'b:', 'c:', 'd:'], (x, y) -> y+x) 
result is d:c:b:a:

I have read and agree to the terms under [CLA.md](https://github.com/kuzudb/kuzu/blob/master/CLA.md).

### DIFF
--- a/src/function/list/list_reduce.cpp
+++ b/src/function/list/list_reduce.cpp
@@ -22,12 +22,28 @@ static std::unique_ptr<FunctionBindData> bindFunc(ScalarBindFuncInput input) {
         ListType::getChildType(input.arguments[0]->getDataType()).copy());
 }
 
+static uint32_t findIndex(const std::vector<std::string>& strings, const std::string& str) {
+    auto it = std::find(strings.begin(), strings.end(), str);
+    if (it != strings.end()) {
+        return it - strings.begin();
+    } else {
+        throw common::RuntimeException{"list_reduce lambda var name cannot found."};
+    }
+}
+
 static void reduceList(const list_entry_t& listEntry, uint64_t pos, common::ValueVector& result,
     common::ValueVector& inputDataVector, const common::ValueVector& tmpResultVector,
     evaluator::ListLambdaBindData& bindData) {
-    auto xParam = bindData.lambdaParamEvaluators[0]->resultVector.get();
-    auto yParam = bindData.lambdaParamEvaluators[1]->resultVector.get();
-    auto paramPos = xParam->state->getSelVector()[0];
+    std::vector<ValueVector*> params(bindData.lambdaParamEvaluators.size());
+    std::vector<uint32_t> indexInVars(bindData.lambdaParamEvaluators.size());
+    for (uint32_t i = 0; i < bindData.lambdaParamEvaluators.size(); i++) {
+        auto param = bindData.lambdaParamEvaluators[i]->resultVector.get();
+        params[i] = param;
+        indexInVars[i] =
+            findIndex(bindData.varNames, bindData.lambdaParamEvaluators[i]->getVarName());
+    }
+
+    auto paramPos = params[0]->state->getSelVector()[0];
     auto tmpResultPos = tmpResultVector.state->getSelVector()[0];
     if (listEntry.size == 0) {
         throw common::RuntimeException{"Cannot execute list_reduce on an empty list."};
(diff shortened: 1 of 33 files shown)


src/function/list/list_reduce.cpp
```diff
@@ -22,12 +22,28 @@ static std::unique_ptr<FunctionBindData> bindFunc(ScalarBindFuncInput input) {
...
+
 static void reduceList(const list_entry_t& listEntry, uint64_t pos, common::ValueVector& result,
     common::ValueVector& inputDataVector, const common::ValueVector& tmpResultVector,
     evaluator::ListLambdaBindData& bindData) {
-    auto xParam = bindData.lambdaParamEvaluators[0]->resultVector.get();
-    auto yParam = bindData.lambdaParamEvaluators[1]->resultVector.get();
-    auto paramPos = xParam->state->getSelVector()[0];
+    std::vector<ValueVector*> params(bindData.lambdaParamEvaluators.size());
+    std::vector<uint32_t> indexInVars(bindData.lambdaParamEvaluators.size());
+    for (uint32_t i = 0; i < bindData.lambdaParamEvaluators.size(); i++) {
```
andyfengHKU, line 39: The idx of parameter should be resolved at compile time because it will never change.

I would either move this logic to `bindFunc` and add `indexInVars` in `ListLambdaBindData`. 

Alternatively you can also consider making this a general logic for ListLambdaEvaluator, meaning not only `list_reduce`, `list_transform` and `list_filter` should also try to resolve a idx based on parameter name. Either way sounds reasonable to me.

src/function/list/list_reduce.cpp
```diff
@@ -22,12 +22,28 @@ static std::unique_ptr<FunctionBindData> bindFunc(ScalarBindFuncInput input) {
         ListType::getChildType(input.arguments[0]->getDataType()).copy());
 }
 
+static uint32_t findIndex(const std::vector<std::string>& strings, const std::string& str) {
```
andyfengHKU, line 25: use common::idx_t instead of uint32_t. Double check other places in this PR as well.

src/include/expression_evaluator/lambda_evaluator.h
```diff
@@ -21,12 +29,16 @@ class LambdaParamEvaluator : public ExpressionEvaluator {
         return std::make_unique<LambdaParamEvaluator>(expression);
     }
 
+    const std::string& getVarName() { return varName; }
```
andyfengHKU, line 32: Let's not save `varName` as a field but always do 
```
const std::string& getVarName() { return getExpression()->toString(); }
```

src/include/expression_evaluator/lambda_evaluator.h
```diff
@@ -71,6 +89,8 @@ class ListLambdaEvaluator : public ExpressionEvaluator {
     std::unique_ptr<ExpressionEvaluator> lambdaRootEvaluator;
     std::vector<LambdaParamEvaluator*> lambdaParamEvaluators;
     std::vector<std::shared_ptr<common::ValueVector>> params;
+    std::vector<std::string> varNames;
```
andyfengHKU, line 92: I think paramNames is more accurate here? Consider applying this change.

andyfengHKU (changes requested):
Hi @excaliburwyj, thanks for the contribution! I was travelling in the past week but finally get the chance to review.

Overall LGTM, I think the only part that needs change is to resolve param index at compile time. I left a detailed comment in `list_reduce.cpp`. Feel free to discuss if there is any confusion.

src/function/list/list_reduce.cpp
excaliburwyj: Thanks for your reply.  I made the modifications and submitted a new commit for your suggestion.

src/include/expression_evaluator/lambda_evaluator.h
```diff
@@ -59,6 +74,27 @@ class ListLambdaEvaluator : public ExpressionEvaluator {
         return result;
     }
 
+    std::vector<common::idx_t> getIndexInVars() {
```
andyfengHKU, line 77: Let's put the implementation into cpp file. Some of the include will then be move to cpp as well.

Also since we use the terminology `param` everywhere. Maybe rename this function to `getParamIndices?`

src/include/expression_evaluator/lambda_evaluator.h
```diff
@@ -21,12 +30,15 @@ class LambdaParamEvaluator : public ExpressionEvaluator {
...
 
+    std::string getVarName() { return this->getExpression()->toString(); }
+
 protected:
     void resolveResultVector(const processor::ResultSet&, storage::MemoryManager*) override {}
 };
 
 struct ListLambdaBindData {
     std::vector<LambdaParamEvaluator*> lambdaParamEvaluators;
+    std::vector<common::idx_t> indexInVars;
```
andyfengHKU, line 41: ditto. Align this with function name so call it `paramIndices`.

andyfengHKU (approved):
Hi @excaliburwyj,

the PR look good to me. I'll just approve here. Let me know once u cleaned it up and I'll merge it into master.

mewim (approved):

mewim:
Hi @excaliburwyj,

Thanks for the contribution!

Could you please take a look at our [CLA](https://github.com/kuzudb/kuzu/blob/master/CLA.md) add following line in the PR description if you agree with it?

```
I have read and agree to the terms under [CLA.md](https://github.com/kuzudb/kuzu/blob/master/CLA.md).
```

mewim (changes requested):
Pending CLA agreement

excaliburwyj:
Hi @mewim  ,
well, I have added the description.

mewim (approved):

PR #4178, opened 2024-09-03
Remove local storage import from transaction.h

benjaminwinger:
I also had to fix some subsequent compilation errors due to things not being included where they are used and relying instead on this include (which admittedly is made more complicated by forward includes meaning that the return type of a function may be incomplete). In addition to a few places which now had to deal with storing unique_ptrs to a forward declared class and couldn't use the implicit default destructor.

This was pulling in a bunch of storage stuff into the single file header (increasing the size by ~50%) and, as a result, the C++ API docs.
We probably need some checks on what gets included. Maybe even just a list of files/classes which we know shouldn't get included as a sanity check.


src/transaction/transaction.cpp
```diff
@@ -24,6 +24,19 @@ Transaction::Transaction(main::ClientContext& clientContext, TransactionType tra
...
+    : type{transactionType}, ID{DUMMY_TRANSACTION_ID}, startTS{DUMMY_START_TIMESTAMP},
+      commitTS{common::INVALID_TRANSACTION}, clientContext{nullptr}, undoBuffer{nullptr},
+      forceCheckpoint{false} {
+    currentTS = common::Timestamp::getCurrentTimestamp().value;
+}
+Transaction::Transaction(TransactionType transactionType, common::transaction_t ID,
+    common::transaction_t startTS) noexcept
+    : type{transactionType}, ID{ID}, startTS{startTS}, commitTS{common::INVALID_TRANSACTION},
+      clientContext{nullptr}, undoBuffer{nullptr}, forceCheckpoint{false} {
+    currentTS = common::Timestamp::getCurrentTimestamp().value;
```
benjaminwinger, line 37: Not sure how these lines are missing coverage if the DUMMY_CHECKPOINT_TRANSACTION uses them. Maybe it doesn't check code running in the global scope?

ray6080 (approved):
Is it a possible to maintain a list of files that will be included into the single header file? And we error whenever we need to include some files not being part of the list. This needs to be part of the CI workflows.
This can add burden as we need to manually maintain the list, but it does force us to be careful about what's being included. Currently, I think just few of us are checking the generated header, and when I was adding the include of local_storage into transaction, I wasn't aware of the consequence on the single file header at all.

benjaminwinger:
> Is it a possible to maintain a list of files that will be included into the single header file? And we error whenever we need to include some files not being part of the list.

Sure. We could add that to the header generator script.

ray6080:
> > Is it a possible to maintain a list of files that will be included into the single header file? And we error whenever we need to include some files not being part of the list.
> 
> Sure. We could add that to the header generator script.

Can u open an issue on this? We can get to it later.

benjaminwinger:
I'd actually already done it earlier today. I've added it to this PR.
I think there's probably some further pruning that could be done though.

src/transaction/transaction.cpp
ray6080: `DUMMY_CHECKPOINT_TRANSACTION` should've used this interface. maybe it's because `DUMMY_CHECKPOINT_TRANSACTION` is static?

PR #4184, opened 2024-09-03
Royi/node batch insert populate error info

royi-luo:
# Description

Partially Fixes https://github.com/kuzudb/kuzu/issues/1872

# Contributor agreement


src/include/processor/warning_context.h
```diff
@@ -19,20 +40,41 @@ struct PopulatedCSVError {
...
-    WarningInfo(processor::PopulatedCSVError warning, uint64_t queryID)
+    WarningInfo(PopulatedCSVError warning, uint64_t queryID)
         : queryID(queryID), warning(std::move(warning)) {}
 };
 
-struct WarningContext {
-    std::vector<WarningInfo> warnings;
-    main::ClientConfig* clientConfig;
-    std::mutex mtx;
+using populate_func_t = std::function<PopulatedCSVError(const SharedWarning*)>;
```
royi-luo, line 49: Will remove/merge with `populate_error_func_t` once `WarningContext` is cleaned up

src/include/processor/warning_context.h
```diff
@@ -19,20 +40,41 @@ struct PopulatedCSVError {
...
+class WarningContext {
+public:
     explicit WarningContext(main::ClientConfig* clientConfig);
 
-    void appendWarningMessages(const std::vector<PopulatedCSVError>& messages, uint64_t queryID);
+    void appendWarningMessages(const std::vector<std::shared_ptr<SharedWarning>>& messages,
+        const WarningSource& warningSource);
+    void populateWarnings(const WarningSource& warningSource, uint64_t queryID,
+        populate_func_t populateFunc);
+    const std::vector<WarningInfo>& getPopulatedWarnings() const;
```
royi-luo, line 65: Warnings have the following lifecycle:
1. Reported warning cached in operator's local warning cache in raw form
2. When local warning cache is full, raw warnings are flushed to shared warning cache
3. In finalize stage, raw warnings are converted to populated warnings (which have line numbers, reconstructed lines, etc.)

Currently step 2 is partially done in the `WarningContext` (the local cache directly flushes raw warnings to `WarningContext` which stores both raw and populated warnings separately). This is unnecessary + more complicated so I will move step 2 to be done in operators (cached in shared state) with common code.

src/include/processor/operator/persistent/index_builder.h
```diff
@@ -25,14 +25,21 @@ namespace processor {
 
 const size_t SHOULD_FLUSH_QUEUE_SIZE = 32;
 
+constexpr size_t BUFFER_SIZE = 64;
+using OptionalExtraDataBuffer = std::optional<common::StaticVector<WarningSourceData, BUFFER_SIZE>>;
```
royi-luo, line 29: This structure is quite large (2.5kB) so the `IndexBuilder` takes up quite a bit of memory (~660kB, all of which is on the stack during construction) so want to double-check that this won't cause any stack-usage related issues (I ran into some during this PR)

royi-luo:
Some quick benchmarks (All datasets have 10M rows, single INT32 column):
branch | IGNORE_ERRORS | % of rows that are errors | copy time
---|---|---|---
master | false | 0 | 668ms
new | false | 0 | 668ms
new | true | 0 | 883ms
master | true | 1% | 1010ms
new | true | 1% | 1204ms
master | true | 10% | 1374ms
new | true | 10% | 1541ms
master | true | 100% | 5389ms
new | true | 100% | 5684ms

It looks like the overhead for copying the extra data is about 200ms for `10M` rows.

src/include/processor/operator/persistent/index_builder.h
royi-luo: Updated to allocate `IndexBuilder` on the heap, let me know if we want to resolve this some other way (e.g. making `IndexBuilder` smaller) and we can discuss

src/include/processor/warning_context.h
royi-luo: removed `populate_error_func_t`

src/processor/operator/persistent/reader/csv/parallel_csv_reader.cpp
```diff
@@ -215,24 +226,19 @@ static std::unique_ptr<TableFuncSharedState> initSharedState(TableFunctionInitIn
...
+    TableFuncSharedState* /*state*/, storage::MemoryManager* /*mm*/) {
     auto localState = std::make_unique<ParallelCSVLocalState>();
-    auto sharedState = state->ptrCast<ParallelCSVScanSharedState>();
-    localState->fileIdx = 0;
-    localState->errorHandler = std::make_unique<LocalCSVFileErrorHandler>(
-        sharedState->context->getClientConfig()->warningLimit, sharedState->csvOption.ignoreErrors,
-        &sharedState->errorHandlers[localState->fileIdx]);
-    localState->reader = std::make_unique<ParallelCSVReader>(sharedState->readerConfig.filePaths[0],
-        sharedState->csvOption.copy(), sharedState->columnInfo.copy(), sharedState->context,
-        localState->errorHandler.get());
```
royi-luo, line 235: This was removed as the error handler and reader are already constructed in `tableFunc` and this prevents `reportFinishedBlock` from being called twice for `blockIdx == 0`

src/processor/operator/persistent/reader/csv/csv_error_handler.cpp
```diff
@@ -228,27 +189,38 @@ void LocalCSVFileErrorHandler::handleError(BaseCSVReader* reader, const CSVError
...
+    }
 }
 
-void LocalCSVFileErrorHandler::reportFinishedBlock(BaseCSVReader* reader, uint64_t blockIdx,
-    uint64_t numRowsRead) {
-    sharedErrorHandler->reportFinishedBlock(reader, blockIdx, numRowsRead);
+void LocalCSVFileErrorHandler::reportFinishedBlock(uint64_t blockIdx, uint64_t numRowsRead) {
+    linesPerBlock[blockIdx].validLines += numRowsRead;
+    linesPerBlock[blockIdx].doneParsingBlock = true;
 }
```
royi-luo, line 201: Line number data is now also cached locally in the error handler (without locks) and occassionally flushed to the shared error handler

src/binder/bind/bind_file_scan.cpp
```diff
@@ -87,6 +88,38 @@ std::unique_ptr<BoundBaseScanSource> Binder::bindScanSource(BaseScanSource* sour
...
+          (TypeUtils::getPhysicalTypeIDForType<decltype(processor::WarningSourceData::fileIdx)>() ==
+              extraDataColumns[2]->getDataType().getPhysicalType());
+    ret =
+        ret &&
+        (TypeUtils::getPhysicalTypeIDForType<decltype(processor::WarningSourceData::blockIdx)>() ==
+            extraDataColumns[3]->getDataType().getPhysicalType());
+    ret = ret && (TypeUtils::getPhysicalTypeIDForType<
+                      decltype(processor::WarningSourceData::rowOffsetInBlock)>() ==
+                     extraDataColumns[4]->getDataType().getPhysicalType());
+    return ret;
```
royi-luo, line 119: This pattern may be better here
```
if (...)
  return false
if (...)
  return false
...
```

src/include/common/constants.h
```diff
@@ -150,6 +150,14 @@ struct CopyConstants {
     static constexpr const char* INT_CSV_PARSING_OPTIONS[] = {"SKIP", "SAMPLE_SIZE"};
     static constexpr uint64_t DEFAULT_CSV_SKIP_NUM = 0;
     static constexpr uint64_t DEFAULT_CSV_TYPE_DEDUCTION_SAMPLE_SIZE = 256;
+
+    // metadata columns used to populate CSV warnings
+    static constexpr std::array WARNING_METADATA_COLUMN_NAMES = {"startByteOffset", "endByteOffset",
+        "fileIdx", "blockIdx", "rowOffsetInBlock"};
```
royi-luo, line 156: I feel like there is no simple way to calculate `blockIdx` for `SerialCSVReader` so I left `blockIdx` in here (since we already minimize performance cost if `IGNORE_ERRORS=false`, this performance hit will only happen if users are expecting warnings). Let me know if there is a simple way and I can make this minor optimization.

src/include/common/copier_config/reader_config.h
```diff
@@ -46,6 +46,18 @@ struct ReaderConfig {
...
+    T getOption(std::string optionName, T defaultValue) {
+        T ret = defaultValue;
+
+        const auto optionIt = options.find(optionName);
+        if (optionIt != options.end()) {
+            ret = optionIt->second.getValue<T>();
+        }
+
+        return ret;
+    }
```
royi-luo, line 59: NIT:
```suggestion
    template<typename T>
    T getOption(std::string optionName, T defaultValue) {
        const auto optionIt = options.find(optionName);
        if (optionIt != options.end()) {
            return optionIt->second.getValue<T>();
        } else {
            return defaultValue;
        }
    }
```

src/include/processor/operator/persistent/index_builder.h
```diff
@@ -25,14 +25,21 @@ namespace processor {
...
+using OptionalWarningSourceData = std::optional<WarningSourceData>;
+
 class IndexBuilderGlobalQueues {
 public:
     explicit IndexBuilderGlobalQueues(transaction::Transaction* transaction,
         storage::NodeTable* nodeTable);
 
     template<typename T>
-    void insert(size_t index, storage::IndexBuffer<T> elem,
+    void insert(size_t index, storage::IndexBuffer<T> elem, OptionalExtraDataBuffer extraData,
```
royi-luo, line 39: I'm thinking of making all the parameter data rvalue references to show that `insert` should take ownership of the data, something like:
```suggestion
    void insert(size_t index, storage::IndexBuffer<T> &&elem, OptionalExtraDataBuffer &&extraData,
```
@ray6080 do you think this is a good idea or does that make things more confusing? Maybe I should just go with the normal pattern of using const references for both?

royi-luo: I've been using the copy-then-move pattern quite a bit and I feel like that one probably isn't the correct pattern to use.

src/planner/plan/append_table_function_call.cpp
```diff
@@ -15,22 +15,24 @@ void Planner::appendTableFunctionCall(const BoundTableScanSourceInfo& info, Logi
...
     const binder::BoundTableScanSourceInfo& info) {
     return std::make_shared<LogicalTableFunctionCall>(info.func, info.bindData->copy(),
-        info.columns, nullptr /* offset */);
+        info.columns, info.extraData, nullptr /* offset */);
 }
 
 std::shared_ptr<LogicalOperator> Planner::getTableFunctionCall(
     const BoundReadingClause& readingClause) {
     auto& call = readingClause.constCast<BoundTableFunctionCall>();
+    // bound table function calls do not make use of extra data
```
royi-luo, line 32:
```suggestion
    // bound table function calls do not make use of extra data so we pass an empty optional
```

src/storage/store/chunked_node_group.cpp
```diff
@@ -13,16 +13,25 @@ namespace kuzu {
...
         KU_ASSERT(this->chunks[columnID]->getResidencyState() == residencyState);
     }
+
+    if (hasExtraData) {
+        extraData = decltype(extraData)::value_type{};
+        for (auto columnType : CopyConstants::WARNING_METADATA_COLUMN_TYPES) {
+            extraData->emplace_back(std::make_unique<ColumnChunk>(LogicalType(columnType), capacity,
+                this->chunks[0]->isCompressionEnabled(), residencyState));
+        }
+    }
```
royi-luo, line 34: It looks like this constructor is only called during `flushAsNewChunkedNodeGroup` and `deserialize`. In this case would it make sense to leave `extraData` empty since I don't think we need it in these use cases?

src/processor/operator/persistent/index_builder.cpp
```diff
@@ -80,11 +89,23 @@ IndexBuilderLocalBuffers::IndexBuilderLocalBuffers(IndexBuilderGlobalQueues& glo
         [](auto) { KU_UNREACHABLE; });
 }
 
+void IndexBuilderLocalBuffers::optionalAppendExtraData(OptionalExtraDataBuffer& buffer,
+    OptionalWarningSourceData data) {
+    if (data.has_value()) {
+        if (!buffer.has_value()) {
+            // if the buffer isn't populated default-construct it
```
royi-luo, line 96:
```suggestion
            // if the buffer is unpopulated construct it as an empty buffer that we can append the data to
```

src/processor/operator/persistent/index_builder.cpp
```diff
@@ -99,25 +120,44 @@ void IndexBuilderSharedState::quitProducer() {
...
+OptionalWarningSourceData getWarningDataFromChunks(
+    const std::optional<std::vector<storage::ColumnChunkData*>>& chunks, common::idx_t posInChunk) {
+    OptionalWarningSourceData ret;
+    if (chunks.has_value()) {
+        KU_ASSERT(chunks->size() == CopyConstants::WARNING_METADATA_NUM_COLUMNS);
+        ret = WarningSourceData{chunks.value()[0]->getValue<uint64_t>(posInChunk),
+            chunks.value()[1]->getValue<uint64_t>(posInChunk),
+            chunks.value()[2]->getValue<uint32_t>(posInChunk),
+            chunks.value()[3]->getValue<uint64_t>(posInChunk),
+            chunks.value()[4]->getValue<uint32_t>(posInChunk)};
```
royi-luo, line 133: should probably add comment saying that these types should match those in `WarningSourceData`

royi-luo: could probably also use `decltype` here

src/processor/operator/persistent/reader/csv/csv_error_handler.cpp
```diff
@@ -21,70 +22,55 @@ void LineContext::setEndOfLine(uint64_t end) {
...
-        const auto lineNumber = getLineNumber(error.blockIdx, error.numRowsReadInBlock);
-        errorMessages.push_back(getPopulatedError(reader, error, lineNumber));
-    }
-    return errorMessages;
+populate_func_t SharedCSVFileErrorHandler::getPopulateCSVErrorFunc(SerialCSVReader* reader) {
+    return [reader, this](CSVError error) -> PopulatedCSVError {
+        const auto lineNumber =
+            getLineNumber(error.warningData.blockIdx, error.warningData.rowOffsetInBlock);
+        return getPopulatedError(reader, std::move(error), lineNumber);
+    };
```
royi-luo, line 37: Pass `reader` as a parameter to the functor instead of having it be captured

src/processor/operator/persistent/reader/csv/serial_csv_reader.cpp
```diff
@@ -156,8 +166,9 @@ static std::unique_ptr<TableFuncSharedState> initSharedState(TableFunctionInitIn
     auto columnInfo = CSVColumnInfo(bindData->getNumColumns(), bindData->getColumnSkips());
     auto sharedState = std::make_unique<SerialCSVScanSharedState>(bindData->config.copy(), numRows,
         bindData->context, csvOption.copy(), columnInfo.copy());
-    for (auto filePath : sharedState->readerConfig.filePaths) {
-        auto reader = std::make_unique<SerialCSVReader>(filePath, csvOption.copy(),
+    for (idx_t i = 0; i < sharedState->readerConfig.filePaths.size(); ++i) {
+        auto filePath = sharedState->readerConfig.filePaths[i];
```
royi-luo, line 170:
```suggestion
        const auto &filePath = sharedState->readerConfig.filePaths[i];
```

src/processor/operator/table_function_call.cpp
```diff
@@ -54,7 +66,7 @@ void TableFunctionCall::initLocalStateInternal(ResultSet* resultSet, ExecutionCo
         sharedState->funcState.get()};
 }
 
-void TableFunctionCall::initGlobalStateInternal(ExecutionContext*) {
+void TableFunctionCall::initGlobalStateInternal(ExecutionContext* /*context*/) {
```
royi-luo, line 69: revert

src/binder/bind/bind_file_scan.cpp
royi-luo: updated to
```
KU_ASSERT(...)
KU_ASSERT(...)
...
```

src/processor/operator/persistent/reader/csv/csv_error_handler.cpp
```diff
@@ -228,27 +189,38 @@ void LocalCSVFileErrorHandler::handleError(BaseCSVReader* reader, const CSVError
...
+
+void LocalCSVFileErrorHandler::finalize() {
     flushCachedErrors();
-}
+};
 
 void LocalCSVFileErrorHandler::flushCachedErrors() {
-    sharedErrorHandler->addCachedErrors(cachedErrors, linesPerBlock);
-    cachedErrors.clear();
-    linesPerBlock.clear();
```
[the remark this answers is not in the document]
royi-luo: updated to separate the clearing of line number data and cached errors

src/storage/store/chunked_node_group.cpp
royi-luo: Removing this for now since we don't actually have any code paths that make use of the `extraData` in this constructor, let me know if it is something we should leave in

src/include/storage/store/chunked_node_group.h
```diff
@@ -162,6 +173,7 @@ class ChunkedNodeGroup {
     uint64_t capacity;
     std::atomic<common::row_idx_t> numRows;
     std::vector<std::unique_ptr<ColumnChunk>> chunks;
+    std::optional<std::vector<std::unique_ptr<ColumnChunk>>> extraData;
```
ray6080, line 176: As we discussed, `ChunkedNodeGroup ` shouldn't be aware of `extraData`, it should just store column chunks based on `columnTypes`, and I think your PR should touch no lines in `ChunkedNodeGroup` ideally.

src/processor/operator/persistent/reader/csv/csv_error_handler.cpp
```diff
@@ -228,27 +189,39 @@ void LocalCSVFileErrorHandler::handleError(BaseCSVReader* reader, const CSVError
...
 }
 
 LocalCSVFileErrorHandler::~LocalCSVFileErrorHandler() {
+    finalize();
+};
+
+void LocalCSVFileErrorHandler::finalize() {
     flushCachedErrors();
-}
+};
```
ray6080, line 213: Do we need `;` at the end? Ditto on L209.

ray6080 (commented):
I've some comments left, please take a look. Can u also profile the memory usage a bit on top of your benchmark? just to make sure memory usage is as expected in different cases.

src/processor/operator/persistent/index_builder.cpp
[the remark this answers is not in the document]
royi-luo: It might be best to wait until I finish addressing the current review comments because if I combine the [output data chunks with the extra data chunk](https://github.com/kuzudb/kuzu/pull/4184#discussion_r1750662538) I'll probably make pretty large change to the index builder (hopefully the diff will be smaller). I'll ping when the new changes are pushed.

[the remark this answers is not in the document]
royi-luo: Never mind I couldn't think of a way to clean up the changes to the index builder, @benjaminwinger let me know if you can think of a way to integrate the extra data that's less coupled.

src/include/function/table_functions.h
```diff
@@ -57,6 +57,8 @@ struct TableFuncInput {
 // to use ResultSet as TableFuncOutput.
 struct TableFuncOutput {
     common::DataChunk dataChunk;
+    common::DataChunk extraDataChunk;
```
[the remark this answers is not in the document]
royi-luo: `dataChunk` and `extraDataChunk` have been merged

src/processor/operator/persistent/node_batch_insert.cpp
```diff
@@ -57,7 +57,7 @@ void NodeBatchInsert::initLocalStateInternal(ResultSet* resultSet, ExecutionCont
     const auto nodeLocalState = localState->ptrCast<NodeBatchInsertLocalState>();
     // NOLINTBEGIN(bugprone-unchecked-optional-access)
     if (nodeSharedState->globalIndexBuilder) {
```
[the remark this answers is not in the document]
royi-luo: Updated

src/processor/operator/persistent/reader/csv/csv_error_handler.cpp
royi-luo: updated

src/include/binder/copy/bound_table_scan_info.h
```diff
@@ -11,15 +11,19 @@ struct BoundTableScanSourceInfo {
     function::TableFunction func;
     std::unique_ptr<function::TableFuncBindData> bindData;
     binder::expression_vector columns;
+    std::optional<binder::expression_vector> extraData;
```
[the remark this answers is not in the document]
royi-luo: renamed

src/include/processor/operator/persistent/node_batch_insert.h
```diff
@@ -61,7 +64,7 @@ struct NodeBatchInsertSharedState final : BatchInsertSharedState {
     // Primary key info
     common::column_id_t pkColumnID;
     common::LogicalType pkType;
-    std::optional<IndexBuilder> globalIndexBuilder = std::nullopt;
+    std::unique_ptr<IndexBuilder> globalIndexBuilder = nullptr;
```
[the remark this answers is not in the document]
royi-luo: updated

src/processor/operator/persistent/reader/parquet/parquet_reader.cpp
```diff
@@ -721,11 +722,19 @@ static double progressFunc(TableFuncSharedState* sharedState) {
     return static_cast<double>(totalReadSize) / state->totalRowsGroups;
 }
 
+static void finalizeFunc(ExecutionContext* ctx, TableFuncSharedState* sharedState,
+    TableFuncLocalState*) {
+    auto state = ku_dynamic_cast<TableFuncSharedState*, ParquetScanSharedState*>(sharedState);
+    for (idx_t i = 0; i < state->readerConfig.getNumFiles(); ++i) {
+        ctx->clientContext->getWarningContextUnsafe().populateWarnings(i, ctx->queryID);
+    }
+}
```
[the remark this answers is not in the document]
royi-luo: updated

src/include/common/constants.h
royi-luo: Added todo to `WarningSourceData` struct

royi-luo:
Some quick memory usage benchmarks (1 thread, 1M nodes, single UINT32 column):
Branch|`IGNORE_ERRORS`|% PK Errors|Peak Memory Usage (bytes)
----|----|----|----
master|false|0|123,815,528
current|false|0|123,551,248
current|true|0|230,898,584
current|true|100|137,272,672

Some notes on the results:
- The run with 100% errors uses less memory because the table + index only contain one value (the rest of the rows error out)
- The extra memory usage with `IGNORE_ERRORS=true` comes from needing to store an extra 5 columns of warning metadata used to populate error messages. The index builder also needs to store this data, so it goes from storing 2 values per row (key, row offset) to 7.

src/include/processor/operator/persistent/index_builder.h
```diff
@@ -106,6 +136,8 @@ class IndexBuilderLocalBuffers {
         UniqueBuffers<uint32_t>, UniqueBuffers<uint16_t>, UniqueBuffers<uint8_t>,
         UniqueBuffers<common::int128_t>, UniqueBuffers<float>, UniqueBuffers<double>>
         buffers;
+
+    std::array<OptionalExtraDataBuffer, storage::NUM_HASH_INDEXES> extraDataBuffers;
```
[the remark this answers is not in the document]
royi-luo: updated

ray6080 (approved):
Looks fine to me. Thanks!

src/include/processor/operator/persistent/index_builder.h
```diff
@@ -25,13 +25,28 @@ namespace processor {
 
 const size_t SHOULD_FLUSH_QUEUE_SIZE = 32;
 
+constexpr size_t BUFFER_SIZE = 64;
```
benjaminwinger, line 28: I feel a little hypocritical saying this given that `in_mem_hash_index.h` has a very similar variable, but particularly since there is another variable called `BUFFER_SIZE` and the namespace alone isn't really enough to explain what they're for, maybe they should both be named, e.g. `INDEX_BUFFER_SIZE` and `WARNING_DATA_BUFFER_SIZE`. That or we should add a static constexpr `capacity` function to StaticVector so that the buffer size can be inlined and you can just call `OptionalWarningDataBuffer::capacity()` if you need to reference it.
That said, I don't actually see anywhere it's being referenced. The IndexBuffer was using a `BUFFER_SIZE` constant because there's a spot where a temporary array of the same length was being initialized, and even there it's something which should be removed since the `hashes` array in `InMemHashIndex::append` is no longer necessary.

Having two almost identically named values of the same type that are used in the same part of the code but which have very different values seems like it will be too easy for someone to accidentally use the wrong one.

src/binder/bind/bind_file_scan.cpp
```diff
@@ -103,7 +124,23 @@ std::unique_ptr<BoundBaseScanSource> Binder::bindFileScanSource(const BaseScanSo
     for (auto i = 0u; i < bindData->columnTypes.size(); i++) {
         inputColumns.push_back(createVariable(bindData->columnNames[i], bindData->columnTypes[i]));
     }
-    auto info = BoundTableScanSourceInfo(func, std::move(bindData), inputColumns);
+
+    const bool ignoreErrors = config->getOption(CopyConstants::IGNORE_ERRORS_OPTION_NAME,
+        CopyConstants::DEFAULT_IGNORE_ERRORS);
+    column_id_t numWarningDataColumns = 0;
+    if (ignoreErrors) {
```
andyfengHKU, line 131: shouldn't this be `if (!ignoreErrors)`?

src/include/processor/operator/persistent/reader/csv/parallel_csv_reader.h
```diff
@@ -15,7 +15,7 @@ class ParallelCSVReader final : public BaseCSVReader {
     friend class ParallelParsingDriver;
 
 public:
-    ParallelCSVReader(const std::string& filePath, common::CSVOption option,
+    ParallelCSVReader(const std::string& filePath, uint64_t fileIdx, common::CSVOption option,
```
andyfengHKU, line 18: use `common::idx` for `fileIdx`

src/binder/bind/bind_file_scan.cpp
royi-luo: The intention is to output warnings only if `ignoreErrors=true` which is why the columns are set if `ignoreErrors=true`

benjaminwinger (approved):
The index builder changes look generally good to me

src/include/binder/copy/bound_table_scan_info.h
```diff
@@ -12,14 +12,20 @@ struct BoundTableScanSourceInfo {
     std::unique_ptr<function::TableFuncBindData> bindData;
     binder::expression_vector columns;
 
+    // the last {numWarningDataColumns} columns are for temporary internal use
+    common::column_id_t numWarningDataColumns;
```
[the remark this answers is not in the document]
royi-luo: updated, I named it `numWarningDataColumns` since the columns store metadata used to populate warnings, not the warnings themselves so I wanted to differentiate between those two

src/include/function/table_functions.h
```diff
@@ -57,6 +57,7 @@ struct TableFuncInput {
 // to use ResultSet as TableFuncOutput.
 struct TableFuncOutput {
     common::DataChunk dataChunk;
+
```
royi-luo, line 60: remove

src/include/function/table_functions.h
```diff
@@ -65,8 +66,11 @@ struct TableFuncOutput {
 
 struct TableFunctionInitInput {
     TableFuncBindData* bindData;
+    common::column_id_t numWarningDataColumns;
```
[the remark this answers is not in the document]
royi-luo: updated

src/include/processor/operator/persistent/reader/csv/parallel_csv_reader.h
royi-luo: updated

src/include/processor/operator/table_function_call.h
```diff
@@ -31,6 +31,7 @@ struct TableFunctionCallInfo {
     function::TableFunction function;
     std::unique_ptr<function::TableFuncBindData> bindData;
     std::vector<DataPos> outPosV;
+    common::column_id_t numWarningDataColumns;
```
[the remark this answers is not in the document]
royi-luo: updated

src/include/processor/operator/table_function_call.h
```diff
@@ -115,6 +117,8 @@ class TableFunctionCall : public PhysicalOperator {
 
     double getProgress(ExecutionContext* context) const override;
 
+    common::column_id_t getNumWarningDataColumns() const;
```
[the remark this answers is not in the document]
royi-luo: updated

extension/duckdb/src/duckdb_scan.cpp
```diff
@@ -18,7 +18,7 @@ void getDuckDBVectorConversionFunc(PhysicalTypeID physicalTypeID,
 DuckDBScanBindData::DuckDBScanBindData(std::string query,
     std::vector<common::LogicalType> columnTypes, std::vector<std::string> columnNames,
     const DuckDBConnector& connector)
-    : TableFuncBindData{std::move(columnTypes), std::move(columnNames)}, query{std::move(query)},
+    : TableFuncBindData{std::move(columnTypes), std::move(columnNames)}, 0, query{std::move(query)},
```
andyfengHKU, line 21: As a convention, we always annotate constant `0 /* numWarningColumns */`. Apply this across this PR

andyfengHKU (approved):
LGTM

extension/duckdb/src/duckdb_scan.cpp
royi-luo: updatede

PR #4188, opened 2024-09-04
Spill to disk

benjaminwinger:
Based on #3743

This feature allows Column chunk data from the partitioner during rel table copies to be written to disk and freed (to save memory) if the ChunkedNodeGroup in question will no longer be needed until the end of the copy.
Every full ChunkedNodeGroup (in the InMemChunkedGroupCollection) gets passed to the Spiller, which stores them in a locked unordered_set (so they can be quickly removed once they are needed again; we could probably use an EvictionQueue to store them, but I'm not convinced that performance benefit would be noticeable).

Data starts to be spilled to disk when the buffer pool is full and at least 50% of the buffer pool is used by column chunks (though this calculation includes chunks which aren't in the fullPartitionerGroups set and can't be spilled).

When it comes time for the data spilled to be used, it gets loaded back into memory and removed from the fullPartitionerGroups set (which could cause other chunks to be spilled if there is not enough memory; ideally we could make sure to use all the chunks already in memory first, but that would be complicated). The memory is then freed afterwards (originally all the chunks would be held in memory until the end of RelBatchInsert execution, but that's no longer possible so we now need mutable access to that data and move it out of the partitioner shared state when processing it so that it can be freed).

### Future work
This is currently only enabled when the database is using the LocalFileSystem and it is not in in-memory mode. I feel like spilling to disk should not write to other types of filesystems, but it seems reasonable that this could spill to the local filesystem for databases stored on other filesystems. That will be more complicated though, and I don't know if there will be much demand for it.


benjaminwinger:
> We should remove the copy.ovf at the end of the query. (actually I think copy.tmp or just .tmp, considering we might spill other stuff in the future, is a better name, as we reserve ovf for overflow file already)

Done.

> We should provide a configuration option for users to specify the path of the temp file. (we should limit the path of temp file to local file system only for now)

I've set up the framework to pass this through the buffermanager, but how should it be configured?

> I think we should allow spilling under in-memory mode with a default .tmp file path eventually (though I think it's hard to how beneficial this can be, as if bm runs out of memory, spilling usually is not going to make the copy work in the end) , but maybe disable the spilling by default, so users have to turn it on manually and they are aware that we will use their local file system.

At the very least it would have to be manually with the current defaults since the would be no database directory to store it in. The only benefit I can see is that after compression the size may be smaller than during the copy. But it does seem to me like the easiest way to enable spilling to disk if you're using in-memory mode and running out of memory would be to switch to an on-disk database.

src/storage/buffer_manager/spiller.cpp
```diff
@@ -0,0 +1,93 @@
...
+
+#include <cstdint>
+#include <cstring>
+#include <memory>
+
+#include "common/assert.h"
+#include "common/exception/io.h"
+#include "common/file_system/virtual_file_system.h"
+#include "common/types/types.h"
+#include "storage/buffer_manager/spiller.h"
```
ray6080, line 11: Remove L11

src/include/storage/buffer_manager/memory_manager.h
```diff
@@ -69,12 +84,18 @@ class KUZU_API MemoryManager {
 
     BufferManager* getBufferManager() const { return bm; }
 
+    void addUnusedChunk(ChunkedNodeGroup* nodeGroup);
+    void clearUnusedChunk(ChunkedNodeGroup* nodeGroup);
+    void loadFromDisk(ChunkedNodeGroup* nodeGroup);
```
ray6080, line 89: Are these functions defined and used anywhere?

src/storage/store/chunked_node_group.cpp
```diff
@@ -461,5 +464,36 @@ std::unique_ptr<ChunkedNodeGroup> ChunkedNodeGroup::deserialize(MemoryManager& m
...
+        spiller.clearUnusedChunk(this);
+        for (auto& chunk : chunks) {
+            chunk->loadFromDisk();
+        }
+        dataInUse = true;
+    });
+}
+
+uint64_t ChunkedNodeGroup::spillToDisk() {
+    KU_ASSERT(!dataInUse);
```
ray6080, line 485: Should L485 be after the lock?

src/storage/store/chunked_node_group.cpp
```diff
@@ -461,5 +464,36 @@ std::unique_ptr<ChunkedNodeGroup> ChunkedNodeGroup::deserialize(MemoryManager& m
...
 }
 
+void ChunkedNodeGroup::setUnused(MemoryManager& mm) {
+    dataInUse = false;
+    mm.getBufferManager()->getSpillerOrSkip([&](auto& spiller) { spiller.addUnusedChunk(this); });
+}
+
+void ChunkedNodeGroup::loadFromDisk(MemoryManager& mm) {
+    mm.getBufferManager()->getSpillerOrSkip([&](auto& spiller) {
+        std::unique_lock lock{spillToDiskMutex};
```
ray6080, line 474: Would it be better to assert on `dataInUse` here?

src/storage/store/chunked_node_group.cpp
```diff
@@ -461,5 +464,36 @@ std::unique_ptr<ChunkedNodeGroup> ChunkedNodeGroup::deserialize(MemoryManager& m
     return chunkedGroup;
 }
 
+void ChunkedNodeGroup::setUnused(MemoryManager& mm) {
+    dataInUse = false;
```
ray6080, line 468: Should this be protected with lock too?

ray6080 (approved):
Can u add some benchmark numbers you've done for now?

We should also add free pages management in the spiller as a future TODO item. Currently, we always append to the end of the spilling file, while there should be cases we can potentially reuse pages.

src/processor/operator/persistent/rel_batch_insert.cpp
```diff
@@ -71,10 +72,13 @@ void RelBatchInsert::appendNodeGroup(transaction::Transaction* transaction, CSRN
     const RelBatchInsertInfo& relInfo, const RelBatchInsertLocalState& localState,
     BatchInsertSharedState& sharedState, const PartitionerSharedState& partitionerSharedState) {
     const auto nodeGroupIdx = localState.nodeGroupIdx;
-    auto& partitioningBuffer =
+    // TODO: I don't think we will need this buffer again. It should be moved out of its original
+    // location and destroyed at the end of this function to free memory (should help with spill to
+    // disk, otherwise we slowly move everything back into memory and run out)
```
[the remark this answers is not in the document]
benjaminwinger: Oh, it's not obvious with the `auto`, but it was already done and that comment can be removed (or moved into getPartitionBuffer to describe why we need to move the data). I don't particularly like that we're leaving holes in the `partitionerSharedState` by moving the buffers out, but it works.

src/include/storage/buffer_manager/memory_manager.h
benjaminwinger: One of them was defined, but no, those should have been removed. I'd initially implemented this stuff in the memory manager before realising that it needed to be better integrated into the buffer manager.

src/include/processor/operator/partitioner.h
```diff
@@ -61,11 +63,19 @@ struct PartitionerSharedState {
     void resetState();
     void merge(std::vector<std::unique_ptr<PartitioningBuffer>> localPartitioningStates);
 
-    storage::InMemChunkedNodeGroupCollection& getPartitionBuffer(common::idx_t partitioningIdx,
-        common::partition_idx_t partitionIdx) const {
+    // Must only be called once for any given parameters
+    std::unique_ptr<storage::InMemChunkedNodeGroupCollection> getPartitionBuffer(
```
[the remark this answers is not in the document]
benjaminwinger: For what part? It only can be called once for a given parameter pair since it moves the InMemChunkedNodeGroupCollection out of the shared state.
And loadFromDisk is called every time since it doesn't do anything if the data is already in memory.

src/storage/buffer_manager/buffer_manager.cpp
```diff
@@ -72,13 +79,25 @@ void EvictionQueue::clear(std::atomic<EvictionCandidate>& candidate) {
...
     vmRegions[1] = std::make_unique<VMRegion>(TEMP_PAGE, bufferPoolSize);
+
+    // TODO(bmwinger): It may be better to spill to disk in a different location for remote file
+    // systems, or even in general.
+    // Ideally we want to spill to disk in some temporary location such
+    // as /var/tmp (not /tmp since that may be backed by memory). However we also need to be able to
+    // support multiple databases spilling at once (can't be the same file), and handle different
+    // platforms.
+    if (!main::DBConfig::isDBPathInMemory(databasePath) &&
+        dynamic_cast<LocalFileSystem*>(vfs->findFileSystem(spillToDiskPath))) {
```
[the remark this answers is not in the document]
benjaminwinger: Maybe. But when would we want to use this feature for another type of filesystem?
I'm also not sure how we would describe it. `isSpillable` sounds more like a question of whether it's possible, but we're not skipping this for remote filesystems just because it's not possible, but because it's impractical since the performance is poor.

On the other hand maybe it's not impractical if you're using fast network storage on your local network, so maybe there's an argument to allowing any filesystem when not using the default path (though I don't know why you wouldn't use something like NFS to treat it as a local filesystem in that case).

src/storage/store/chunked_node_group.cpp
benjaminwinger: I don't think it needs to be. The point of the lock is that once it gets added to the set of unused chunks any thread may attempt to spill it to disk, but before that point it should only be accessed by a single thread.

src/storage/store/chunked_node_group.cpp
benjaminwinger: No, if the chunk hasn't already been spilled to disk it's possible for it to be spilled to disk as this is running, if another thread needs the memory.
`dataInUse` could actually be true here since the last chunked group in each collection won't ever get marked as unused.

src/storage/store/chunked_node_group.cpp
benjaminwinger: I think what's more important here is that it might be possible for a thread to try and spill this chunk to disk, but have it be loaded from disk and accessed between where its removed from the full group set and when its actually spilled. In which case, if we want to support concurrent spilling to disk (the other solution would be to call spillToDisk while the Spiller's lock is held), I think we'd want to wrap the inner part (after locking) in an `if (!dataInUse) { ... }` conditional, so that if its loaded by another thread before the lock is acquired it won't spill it.

benjaminwinger:
Benchmarks: 
Copying the first 194 million relationships in the twitter-2010 dataset (the hash index is 1.1GB)
- Unrestricted: ~12GB peak memory, ~23s copy time
- Buffer pool size of 8GB: ~9GB peak memory, ~25s copy time, ~2.1GB spilled
- Buffer pool size of 4GB: ~6GB peak memory, ~29s copy time, ~6.8GB spilled
- Buffer pool size of 3GB: ~4GB peak memory, ~30s copy time, ~8GB spilled
- Buffer pool size of 2GB runs out of memory

I'm also running into a segfault in the database initialization when trying to load a database which failed to copy. It doesn't seem to be related to this PR.

src/include/processor/operator/partitioner.h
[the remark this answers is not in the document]
ray6080: ah sorry, I got it after a second look, but forgot to remove the comment here. mark it as resolved now.

src/storage/buffer_manager/buffer_manager.cpp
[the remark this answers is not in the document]
ray6080: hmm i didn't make myself clear here. To clarify a bit more, I mainly meant if that is a better way to check if the vfs is local or not compared to the check using `dynamic_cast`. maybe the function should be `isLocalFS()`.

PR #4196, opened 2024-09-05
Optimize version vector array

ray6080:
# Description

This is on top of #4210.

Skip allocation of insertion/deletion versions array if all insertions and deletions are of the same version.
It's common that all values within the same vector are of the same version as they are inserted through `COPY FROM` statement, though all deletions of the same version happen only when all tuples within the vector are deleted.
The optimization is to keep a `sameInsertionVersion` and a `sameDeletionVersion` inside the `VectorVersionInfo`. Further, each `VersionInfo` can keep a global `sameInsertionVersion` and `sameDeletionVersion` too, though that won't contribute much to the reduction of memory usage.

I did a simple benchmark with lsqb-100 Comment table, which only contains only one property, `ID INT64`. The benchmarking is done under in-memory mode, the mem usage after COPY reduced from ~14GB to ~10G.

TODO:
- A common data structure for both insertion and deletion VectorVersionInfo.
- The `insertionStatus` and `deletionStatus` can perhaps also be simplified.


src/storage/store/version_info.cpp
```diff
@@ -9,37 +9,92 @@ using namespace kuzu::common;
...
     deletionStatus = DeletionStatus::CHECK_VERSION;
-    if (deletedVersions[rowIdx] == transactionID) {
+    if (sameDeletionVersion == transactionID) {
+        // All are deleted in the same transaction.
+        return false;
+    }
+    if (isSameDeletionVersion()) {
+        // All are deleted in a different transaction.
+        throw RuntimeException(
+            "Write-write conflict: deleting a row that is already deleted by another transaction.");
```
ray6080, line 44: Not sure why this line is not covered. We do have tests on this exception.

src/storage/store/node_table.cpp
```diff
@@ -293,7 +293,7 @@ void NodeTable::commit(Transaction* transaction, LocalTable* localTable) {
     for (auto localNodeGroupIdx = 0u; localNodeGroupIdx < localNodeTable.getNumNodeGroups();
          localNodeGroupIdx++) {
         const auto localNodeGroup = localNodeTable.getNodeGroup(localNodeGroupIdx);
-        if (localNodeGroup->getNumDeletedRows(transaction) > 0) {
+        if (localNodeGroup->hasDeletions(transaction) > 0) {
```
benjaminwinger, line 296: While this is a valid way of checking if hasDeletions returned `true`, I'm guessing you probably forgot to remove the `> 0`?

benjaminwinger (approved):

src/storage/store/node_table.cpp
ray6080: Yeah. good catch. removed that now.

PR #4200, opened 2024-09-05
Enable read only DB tests on windows

benjaminwinger:
Support for this was implemented a while ago (I don't recall exactly when), but evidently these tests were missed.

### DIFF
--- a/tools/python_api/test/test_database.py
+++ b/tools/python_api/test/test_database.py
@@ -35,13 +35,11 @@ def test_database_close(tmp_path: Path, build_dir: Path) -> None:
     assert db._database is not None
 
     # Open the database on a subprocess. It should raise an exception.
-    # TODO: Enable this test on Windows when the read-only mode is implemented.
-    if sys.platform != "win32":
-        with pytest.raises(
-            RuntimeError,
-            match=r"RuntimeError: IO exception: Could not set lock on file",
-        ):
-            open_database_on_subprocess(db_path, build_dir)
+    with pytest.raises(
+        Exception,
+        match=r"RuntimeError: IO exception: Could not set lock on file",
+    ):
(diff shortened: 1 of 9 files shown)


ray6080 (approved):

tools/python_api/test/test_database.py
```diff
@@ -37,13 +37,11 @@ def test_database_close(tmp_path: Path, build_dir: Path) -> None:
...
-        with pytest.raises(
-            RuntimeError,
-            match=r"RuntimeError: IO exception: Could not set lock on file",
-        ):
-            open_database_on_subprocess(db_path, build_dir)
+    with pytest.raises(
+        Exception,
+        match=r"RuntimeError: IO exception: Could not set lock on file",
+    ):
+        open_database_on_subprocess(db_path, build_dir)
```
benjaminwinger, line 44: I did some tests with the shell, and as far as I can tell on windows you can open the same database in read-write mode, operate on the files, and the files get clobbered by whichever writes last. The locks don't seem to be working properly. It works as expected if you open one in read-only mode though.
The C++ test that does something similar is being skipped, presumably because in addition to windows not supporting this initially, it also wasn't written to work with msvc (among other things it uses fork to create the process, which doesn't exist on windows).

ray6080: Yeah, the lock is not supported on windows yet. https://github.com/kuzudb/kuzu/issues/921

benjaminwinger: I guess not. I'd seen locking code from #2089, but it doesn't appear to work.
I have a suspicion that the issue is that it locks a region of the file, not the entire file, so locking write access to a zero-length region (as our code currently does) doesn't actually do anything. But maybe it would work if we lock even a single byte instead.

Issue #4129, opened 2024-08-22
Bug: Unexpected error message when copying from csv files with mismatched num of columns


ray6080:
### Kùzu version

master

### What operating system are you using?

MacOS Sonoma

### What happened?

When `COPY FROM` a csv file, if I give the wrong delimiter, which leads to whole line parsed as one column, I expect to see an error message thrown immediately along the line of `found mismatched num of columns, check your file and schema, or delimiter, etc.`. Instead, we throw a conversion exception, which makes sense but not good for usability, and it takes a while to throw the conversion exception on my local.

### Are there known steps to reproduce?

Try on ldbc-1 comment table.
```
kuzu> CREATE NODE TABLE Comment(ID INT64, creationDate TIMESTAMP, locationIP STRING, browserUsed STRING, content STRING, length INT64, PRIMARY KEY(ID));
┌─────────────────────────────────┐
│ result                          │
│ STRING                          │
├─────────────────────────────────┤
│ Table Comment has been created. │
└─────────────────────────────────┘
(1 tuple)
(1 column)
Time: 0.08ms (compiling), 1.24ms (executing)
kuzu> COPY Comment FROM '/Users/guodong/Downloads/ldbc/ldbc-1/csv/comment_0_0.csv' (DELIM=",", header=true);
Error: Copy exception: Error in file /Users/guodong/Downloads/ldbc/ldbc-1/csv/comment_0_0.csv on line 2: Conversion exception: Cast failed. Could not convert "1236950581249|2011-08-17T14:26:59.961+0000|92.39.58.88|Chrome|yes|3" to INT64. Line containing the error: '1236950581249|2011-08-17T14:26:59.961+0000|92.39.58.88|Chrome|yes|3'

```

SterlingT3485:
After investigation for a while, I found a possible solution of making use of `bool SniffCSVNameAndTypeDriver::addRow(uint64_t, common::column_id_t columnCount)` inside `driver.cpp`, it dumbly returns true before, we could add a check of `columnCount` here, similar to what we did in `bool ParsingDriver::addRow(uint64_t /*rowNum*/, common::column_id_t columnCount)`. 

And also in `serial_csv_reader.cpp` the `numColumn` of `columnInfo` inside the `Reader` need to be initialized properly as `auto columnInfo = CSVColumnInfo(bindInput->expectedColumnNames.size() /* numColumns */, {} /* columnSkips */);`,before it was simply zero. 

[...]

PR #4202, opened 2024-09-05
Fixing issue 4129: Unexpected error message when copying from csv files with mismatched num of columns

SterlingT3485:
# Description
This PR aims to solve issue [4129](https://github.com/kuzudb/kuzu/issues/4129#issuecomment-2331899406),

The solution is to make use of 
`bool SniffCSVNameAndTypeDriver::addRow(uint64_t, common::column_id_t columnCount)` inside `driver.cpp`, 

it dumbly returns true before, we could add a check of `columnCount` here, similar to what we did in 
`bool ParsingDriver::addRow(uint64_t /*rowNum*/, common::column_id_t columnCount)`. 

And also in `serial_csv_reader.cpp` the `numColumn` of `columnInfo` inside the `Reader` need to be initialized properly as `auto columnInfo = CSVColumnInfo(bindInput->expectedColumnNames.size() /* numColumns */, {} /* columnSkips */);`,before it was simply zero. 

A new field `"rowEmpty"` is introduced in `SniffCSVNameAndTypeDriver`, which is checked by `addRow` function to skip any empty row during the sniffing, similar to what we did in `ParsingDriver`.

After these changes, the mismatched num of columns problem can be detected when sniffing the header, as expected in the BUG description.

Three test cases are revised accordingly, which is `521/1219/1223`, they was expecting exception to be thrown in content before, now they should expect exception to be thrown in header.

Fixes # (4129)

# Contributor agreement

### DIFF
--- a/src/include/processor/operator/persistent/reader/csv/driver.h
+++ b/src/include/processor/operator/persistent/reader/csv/driver.h
@@ -49,9 +49,9 @@ private:
 class SerialCSVReader;
 
 class SerialParsingDriver : public ParsingDriver {
-    SerialCSVReader* reader;
-
+    
 public:
+    SerialCSVReader* reader;
     SerialParsingDriver(common::DataChunk& chunk, SerialCSVReader* reader);
 
     bool doneEarly() override;
@@ -61,19 +61,17 @@ private:
 };
 
 class SniffCSVNameAndTypeDriver : public SerialParsingDriver {
-    SerialCSVReader* reader;
-    // static common::DataChunk dummyChunk;
 
 public:
     SniffCSVNameAndTypeDriver(SerialCSVReader* reader,
         const function::ScanTableFuncBindInput* bindInput);
 
+    bool done(uint64_t rowNum) const;
+    bool addValue(uint64_t rowNum, common::column_id_t columnIdx, std::string_view value);
+
     std::vector<std::pair<std::string, common::LogicalType>> columns;
     std::vector<bool> sniffType;
     // if the type isn't declared in the header, sniff it
-
-    bool done(uint64_t rowNum) const;
-    bool addValue(uint64_t rowNum, common::column_id_t columnIdx, std::string_view value);
 };
 
 } // namespace processor
--- a/src/include/processor/operator/persistent/reader/csv/driver.h
+++ b/src/include/processor/operator/persistent/reader/csv/driver.h
@@ -62,7 +62,7 @@ private:
 
 class SniffCSVNameAndTypeDriver : public SerialParsingDriver {
     SerialCSVReader* reader;
-    static common::DataChunk dummyChunk;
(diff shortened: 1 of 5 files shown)


ray6080 (commented):
Thanks! This generally looks good to me. Though we should do a refactoring on `SniffCSVNameAndTypeDriver`. The logic of `addRow` inside `SniffCSVNameAndTypeDriver` now duplicates the one in `ParsingDriver`.

Can u make `SniffCSVNameAndTypeDriver` extend `SerialParsingDriver`?
Let me know when u are done with the refactoring, I'd like to take another look at this PR.

ray6080:
Btw, please make the PR title be more meaningful: 1) you don't need to include your username in the title; 2) issue ID is not the best reference as you don't know what it refers to unless you clicks into the issue.

SterlingT3485:
Thank Guodong for the reviewing, refactored the `SniffCSVNameAndTypeDriver` to be extended from `SerialParsingDriver`

src/include/processor/operator/persistent/reader/csv/driver.h
```diff
@@ -58,17 +60,20 @@ class SerialParsingDriver : public ParsingDriver {
     BaseCSVReader* getReader() override;
 };
 
-struct SniffCSVNameAndTypeDriver {
-    std::vector<std::pair<std::string, common::LogicalType>> columns;
-    std::vector<bool> sniffType;
-    // if the type isn't declared in the header, sniff it
+class SniffCSVNameAndTypeDriver : public SerialParsingDriver {
     SerialCSVReader* reader;
```
ray6080, line 64: Isn't this shadowing the `reader` of `SerialParsingDriver` at L50? I don't think this is properly set during construction.

SterlingT3485: let me try delete this` reader` thus the reader inherited from `SerialParsingDriver` is set properly

src/include/processor/operator/persistent/reader/csv/driver.h
```diff
@@ -16,6 +16,8 @@ namespace processor {
 class BaseCSVReader;
 class ParsingDriver {
     common::DataChunk& chunk;
+
+protected:
     bool rowEmpty;
```
ray6080, line 21: Can u move these two fields below functions? Our code convention is class member fields are below functions. For example https://github.com/kuzudb/kuzu/blob/master/src/include/storage/storage_manager.h.

Same for other driver class in this file.

ray6080 (approved):

src/include/processor/operator/persistent/reader/csv/driver.h
SterlingT3485: Thanks for pointing this out, let me fix it.

PR #4220, opened 2024-09-08
Fix compilation warnings and enable Werror on gcc and clang

ray6080:
# Description

Turn on `-Werror` by default and fix compilation warnings on linux and macOS platforms.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -15,10 +15,10 @@ set(CMAKE_VISIBILITY_INLINES_HIDDEN ON)
 # On Linux, symbols in executables are not accessible by loaded shared libraries (e.g. via dlopen(3)). However, we need to export public symbols in executables so that extensions can access public symbols. This enables that behaviour.
 set(CMAKE_ENABLE_EXPORTS TRUE)
 
-option(ENABLE_WERROR "Treat all warnings as errors" FALSE)
+option(ENABLE_WERROR "Treat all warnings as errors" TRUE)
 if(ENABLE_WERROR)
     if (CMAKE_VERSION VERSION_GREATER "3.24.0" OR CMAKE_VERSION VERSION_EQUAL "3.24.0")
-        set(COMPILE_WARNING_AS_ERROR TRUE)
+        set(CMAKE_COMPILE_WARNING_AS_ERROR TRUE)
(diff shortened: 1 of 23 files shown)


src/include/function/udf_function.h
```diff
@@ -67,7 +67,7 @@ struct UDF {
     template<typename RESULT_TYPE>
     static function::scalar_func_exec_t createEmptyParameterExecFunc(RESULT_TYPE (*udfFunc)(),
         const std::vector<common::LogicalTypeID>&) {
-        (void*)(udfFunc); // Disable compiler warnings.
+        KU_UNUSED(udfFunc); // Disable compiler warnings.
```
benjaminwinger, line 70: There's actually a built-in [[[maybe_unused]]](https://en.cppreference.com/w/cpp/language/attributes/maybe_unused) attribute that can be used instead. I'd added `KU_UNUSED` because I'd either forgotten about it or hadn't known about it. We should move to using that instead.

benjaminwinger (approved):

MSebanc (approved):
Shell changes are good 👍

CMakeLists.txt
```diff
@@ -15,10 +15,10 @@ set(CMAKE_VISIBILITY_INLINES_HIDDEN ON)
 # On Linux, symbols in executables are not accessible by loaded shared libraries (e.g. via dlopen(3)). However, we need to export public symbols in executables so that extensions can access public symbols. This enables that behaviour.
 set(CMAKE_ENABLE_EXPORTS TRUE)
 
-option(ENABLE_WERROR "Treat all warnings as errors" FALSE)
+option(ENABLE_WERROR "Treat all warnings as errors" TRUE)
```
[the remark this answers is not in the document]
ray6080: Make sense. I was more thinking from our perspective that easier to catch compilation warnings locally first. but it makes sense to revert this.

src/include/function/udf_function.h
ray6080: Does `maybe_unused` also work with clang-tidy checks? If so, we should probably get rid of `KU_UNUSED` for good.

benjaminwinger: I think so

PR #4232, opened 2024-09-10
Skip extension test from the main pipeline

mewim:
# Description

This PR skips the extension tests from the main test pipeline and remove extension tests from the API tests. Additionally, it adds a test case (as part of Python API) to verify the correctness of extension installation by comparing with the manually downloaded file (with Python's `urllib`). This should guarantee that extension installation works correctly without having to actually load an extension. The functionality tests of extensions should have already been covered with the extension test step.

Together with https://github.com/kuzudb/kuzu/pull/4228, these two PR should eliminate the need to build extensions and re-run CI due to code change.


acquamarin (approved):

ray6080 (approved):
There seems to be quite a lot code formatting changes, are they expected?

PR #4233, opened 2024-09-10
Fix return same column with different alias

### LINKED ISSUE #3913: Bug: `LOAD FROM` does not allow retrurning the same column twice
### Kùzu version

v0.4.2

_No response_

### What happened?

`LOAD FROM "demo-csv/city.csv" RETURN column0, column0;` is throwing the error `Error: Binder exception: Multiple result column with the same name column0 are not supported.` as expected.

However, it should be possible to return the column twice by renaming: `LOAD FROM "demo-csv/city.csv" RETURN column0 AS C0, column0 AS C1;`. This query fails with a similar error: `Error: Binder exception: Multiple result column with the same name C1 are not supported.`.

_No response_

andyfengHKU:
# Description

When we return the same column with different aliases, current master treat these columns with the same column name and triggers an exception. This PR fixes this by passing `columnNames` to `StatementResult`.

An alternative solution is to copy column expression so that each column expression is of its own object and thus can have its own alias/column name. Both solution make sense to me so I picked the one with less work.

Fixes #3913

# Contributor agreement

### DIFF
--- a/src/binder/bind/bind_projection_clause.cpp
+++ b/src/binder/bind/bind_projection_clause.cpp
@@ -85,7 +85,8 @@ void validateColumnNamesAreUnique(const std::vector<std::string>& columnNames) {
     }
 }
 
-std::vector<std::string> getColumnNames(const expression_vector& exprs, const std::vector<std::string>& aliases) {
+std::vector<std::string> getColumnNames(const expression_vector& exprs,
+    const std::vector<std::string>& aliases) {
     std::vector<std::string> columnNames;
     for (auto i = 0u; i < exprs.size(); ++i) {
         if (aliases[i].empty()) {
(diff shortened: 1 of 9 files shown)


src/binder/bind/bind_projection_clause.cpp
```diff
@@ -74,10 +74,55 @@ static std::pair<expression_vector, std::vector<std::string>> rewriteProjectionI
...
+        }
+    }
+    auto columnNames = getColumnNames(projectionExprs, aliases);
+    validateColumnNamesAreUnique(columnNames);
+    // Rewrite projection list
+    auto originalProjectionExprs = projectionExprs;
+    auto originalAliases = aliases;
+    auto [a, b] = rewriteProjectionInWithClause(projectionExprs, aliases);
+    projectionExprs = a;
+    aliases = b;
```
ray6080, line 117: Why not just use new variable names `rewrittenProjectionExprs`, and `rewriteenAliases`?

ray6080 (approved):

Issue #4211, opened 2024-09-07
Feature: show_tables() and table_ino() enhancements


ansarizafar:
### API

Node.js

### Description

show_tables() should return table ID. This will help in building schema migration tools.
table_info() should return column's default value and other constraints.

PR #4234, opened 2024-09-10
Feature: show_tables() and table_info()

WWW0030:
# Description

Added features described in https://github.com/kuzudb/kuzu/issues/4211#event-14176599306

Fixes # (issue)
https://github.com/kuzudb/kuzu/issues/4211#event-14176599306

# Contributor agreement


ray6080 (commented):
Can we add some more tests on `table_info()` for those with default values? Try a few different default values, for example:
```
CREATE NODE TABLE person(id INT64, prop STRING DEFAULT 'Alice', PRIMARY KEY(id));
CREATE NODE TABLE person(id INT64, prop STRING DEFAULT gen_random_uuid(), PRIMARY KEY(id));
CREATE NODE TABLE person(id INT64, prop INT64 default 10+20, PRIMARY KEY(id));
```

src/function/table/call/table_info.cpp
```diff
@@ -101,6 +102,8 @@ static std::unique_ptr<TableFuncBindData> bindFunc(main::ClientContext* context,
     columnTypes.push_back(LogicalType::INT32());
     columnNames.emplace_back("name");
     columnTypes.push_back(LogicalType::STRING());
+    columnNames.emplace_back("deault expression");
```
ray6080, line 105: I'd just call this `default`, which is shorter but still clear.

ray6080 (approved):
LGTM. Thanks!

PR #4249, opened 2024-09-12
Add function call to clear cached warnings

### LINKED ISSUE #4226: Add function call to clean up cached warnings
When copying from files there is an option `IGNORE_ERRORS` that caches any errors encountered as warnings and continues execution skipping whatever row caused the error. Warnings can then be retrieved using the function `show_warnings()`.

Warnings are cached globally across a connection, we should offer users the ability to clean up the global warning cache if they are no longer interested in warnings from previous queries. This will help especially in the case where the global warning limit is hit (and any new warnings will not be cached).

Users will be able to do this by executing the query `CALL clear_warnings();`

royi-luo:
# Description

Fixes #4226

Adds function call `CALL clear_warnings() RETURN status` which clears accumulated warnings in the global `WarningContext`

# Contributor agreement


ray6080 (approved):
Can u open an issue on supporting CALL func() without return clause?

royi-luo:
> Can u open an issue on supporting CALL func() without return clause?

Create issue https://github.com/kuzudb/kuzu/issues/4250

PR #4253, opened 2024-09-13
Changed cli default mode to multiline

MSebanc:
# Description

Made multiline mode the default mode of the shell. Tests for different editing features now run both multiline and singleline versions of the test.

# Contributor agreement


andyfengHKU (approved):
LGTM!

PR #4261, opened 2024-09-16
Skip scan of rel id when possible

ray6080:
# Description

Currently, we all scan `REL_ID` regardless if it's projected or not.
This PR removed the logic of always projecting `REL_ID` property when scan rel tables. This can improve performance of scan rel in some level.

LDBC30 dataset on my local Mac mini 2023 (M2 Pro).
```
// Q1: match (a)-[e:likeComment]->(b) return count(*);
|  master  | current |
| -------- | ------- |
| 696.65ms | 357.96ms |

// Q2: match (a)-[e:knows]->(b) return count(*);
|  master | current |
| ------- | ------- |
| 96.18ms | 51.72ms |
```

Future todo: we don't always need to scan NBR ID from rel tables, which should also be optimized away later.

### DIFF
--- a/src/optimizer/projection_push_down_optimizer.cpp
+++ b/src/optimizer/projection_push_down_optimizer.cpp
@@ -185,7 +185,8 @@ void ProjectionPushDownOptimizer::visitDelete(LogicalOperator* op) {
             auto& rel = info.pattern->constCast<RelExpression>();
             collectExpressionsInUse(rel.getSrcNode()->getInternalID());
             collectExpressionsInUse(rel.getDstNode()->getInternalID());
-            if (!rel.isEmpty() && rel.getRelType() == QueryRelType::NON_RECURSIVE) {
+            KU_ASSERT(rel.getRelType() == QueryRelType::NON_RECURSIVE);
+            if (!rel.isEmpty()) {
                 collectExpressionsInUse(rel.getInternalIDProperty());
             }
         }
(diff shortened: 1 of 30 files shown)


src/optimizer/projection_push_down_optimizer.cpp
```diff
@@ -185,7 +185,9 @@ void ProjectionPushDownOptimizer::visitDelete(LogicalOperator* op) {
             auto& rel = info.pattern->constCast<RelExpression>();
             collectExpressionsInUse(rel.getSrcNode()->getInternalID());
             collectExpressionsInUse(rel.getDstNode()->getInternalID());
-            collectExpressionsInUse(rel.getInternalIDProperty());
+            if (!rel.isEmpty() && rel.getRelType() == QueryRelType::NON_RECURSIVE) {
```
andyfengHKU, line 188: So you don't need relID to delete?

First I think `rel.getRelType() == QueryRelType::NON_RECURSIVE` this should be assertion.

Second, `!rel.isEmpty()` this looks like a bug we didn't capture before. Is it true?

andyfengHKU (approved):

andyfengHKU (commented):
Oh btw, normally we add some performance numbers for this type of PR. But up to u.

src/optimizer/projection_push_down_optimizer.cpp
ray6080: > So you don't need relID to delete?

I do need. but not need it when rel is empty (no matched tables).

Changed `rel.getRelType() == QueryRelType::NON_RECURSIVE` to assertion.

> Second, !rel.isEmpty() this looks like a bug we didn't capture before. Is it true?

Not sure if it was a bug, but it gets exposed now when we don't always add rel ID property.

ray6080:
> Oh btw, normally we add some performance numbers for this type of PR. But up to u.

I do have some numbers locally. was waiting for the benchmark report. Adding some numbers to PR description now.

PR #4264, opened 2024-09-16
Remove unncessary scan of nbrNodeID from rel tables

ray6080:
# Description

`nbrNodeID` is always scanned regardless needed or not currently, similar to `relID`.
This PR optimizes the unnecessary scan of `nbrNodeID` away when possible by replacing the `NBR_ID_COLUMN_ID` with `INVALID_COLUMN_ID`.

Also, add more strict checks to optimize away unnecessary rel table scans in multi-label or undirected rel table scans.

Some benchmark numbers here:
LDBC30 dataset on my local Mac mini 2023 (M2 Pro).
```
// Q1: match (a)-[e:likeComment]->(b) return count(*);
|  master  | current |
| -------- | ------- |
| 355.13ms | 18.17ms |

// Q2: match (a)-[e:knows]->(b) return count(*);
|  master | current |
| ------- | ------- |
| 51.12ms | 19.11ms |
```

@andyfengHKU I found that under many cases involving undirected and mutli-labled rel table scans, we apply NodeLabelFilter, which first is invisible in our query plan somehow, second I wonder if we can have more restricted apply of it.

### DIFF
--- a/src/processor/operator/scan/scan_multi_rel_tables.cpp
+++ b/src/processor/operator/scan/scan_multi_rel_tables.cpp
@@ -20,7 +20,7 @@ bool DirectionInfo::needFlip(RelDataDirection relDataDirection) const {
     return false;
 }
 
-bool RelTableCollectionScanner::scan(const SelectionVector& selVector, Transaction* transaction) {
+bool RelTableCollectionScanner::scan(Transaction* transaction) {
     while (true) {
         const auto& relInfo = relInfos[currentTableIdx];
         auto& scanState = *relInfo.scanState;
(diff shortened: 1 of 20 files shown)


src/processor/operator/scan/scan_multi_rel_tables.cpp
```diff
@@ -90,7 +97,7 @@ bool ScanMultiRelTable::getNextTuplesInternal(ExecutionContext* context) {
         }
         const auto currentIdx = boundNodeIDVector->state->getSelVector()[0];
         if (boundNodeIDVector->isNull(currentIdx)) {
-            outState->getSelVectorUnsafe().setSelSize(0);
+            currentScanner = nullptr;
```
andyfengHKU, line 100: We should cover this line. Can u open an issue.

andyfengHKU (approved):

src/processor/operator/scan/scan_multi_rel_tables.cpp
ray6080: See https://github.com/kuzudb/kuzu/issues/4267. I believe we might have more cases not covered yet. Will address the issue later.

Issue #4235, opened 2024-09-10
Missing progress report for second and third pipelines in COPY REL statements


ray6080:
### Kùzu version

master

### What happened?

Current the progress bar only reports the progress of the first pipeline during the execution of COPY REL statement.
We should integrate the progress bar into RelBatchInsert, too.

PR #4266, opened 2024-09-17
Adding Progress Report for REL_BATCH_INSERT

WWW0030:
# Description
Adding Progress report for REL_BATCH_INSERT
https://github.com/kuzudb/kuzu/issues/4235#event-14234443894

# Contributor agreement


ray6080 (changes requested):

src/processor/operator/persistent/rel_batch_insert.cpp
ray6080, line 1: I think you should use `initGlobalStateInternal` to init the progress bar shared state and set total num of partitions. With `initGlobalStateInternal`, you don't need to initialize the `progressSharedState` during mapping and pass through constructor.
```
void RelBatchInsert::initGlobalStateInternal(ExecutionContext*) {
    progressSharedState = std::make_unique<RelBatchInsertProgressSharedState>();
    progressSharedState->partitionsTotal =
        partitionerSharedState->numPartitions[info->ptrCast<RelBatchInsertInfo>()->partitioningIdx];
}
```

src/processor/map/map_copy_from.cpp
```diff
@@ -164,9 +164,11 @@ std::unique_ptr<PhysicalOperator> PlanMapper::createCopyRel(
         offsetVectorIdx, std::move(columnTypes));
     auto printInfo =
         std::make_unique<RelBatchInsertPrintInfo>(copyFrom.getInfo()->tableEntry->getName());
+    auto progressSharedState = std::make_shared<RelBatchInsertProgressSharedState>();
```
ray6080, line 167: I don't think you need this if you move the logic inside `RelBatchInsert::initGlobalStateInternal`.

src/processor/operator/persistent/rel_batch_insert.cpp
```diff
@@ -53,6 +53,7 @@ void RelBatchInsert::executeInternal(ExecutionContext* context) {
     while (true) {
         relLocalState->nodeGroupIdx =
             partitionerSharedState->getNextPartition(relInfo->partitioningIdx);
```
ray6080, line 55: Can u pass `progressSharedState` into `partitionerSharedState->getNextPartition`? so you don't need to take another lock inside the `progressSharedState` to sync on `partitionsDone`. 

```
partition_idx_t PartitionerSharedState::getNextPartition(idx_t partitioningIdx,
    RelBatchInsertProgressSharedState& progress) {
    std::unique_lock xLck{mtx};
    if (nextPartitionIdx >= numPartitions[partitioningIdx]) {
        return INVALID_PARTITION_IDX;
    }
    progress.partitionsDone++;
    return nextPartitionIdx++;
}
```

src/processor/operator/persistent/rel_batch_insert.cpp
```diff
@@ -225,5 +226,15 @@ void RelBatchInsert::finalizeInternal(ExecutionContext* context) {
...
 
+void RelBatchInsert::updateProgress(ExecutionContext* context) {
+    std::lock_guard<std::mutex> lock(progressSharedState->mtx);
+    progressSharedState->partitionsDone = partitionerSharedState->nextPartitionIdx;
+    progressSharedState->partitionsTotal =
+        partitionerSharedState->numPartitions[partitionerSharedState->numPartitions.size() - 1];
+    double progress =
+        double(progressSharedState->partitionsDone) / double(progressSharedState->partitionsTotal);
+    context->clientContext->getProgressBar()->updateProgress(context->queryID, progress);
+}
```
ray6080, line 237: This can be simplified.

```
void RelBatchInsert::updateProgress(ExecutionContext* context) {
    double progress =
        double(progressSharedState->partitionsDone) / double(progressSharedState->partitionsTotal);
    context->clientContext->getProgressBar()->updateProgress(context->queryID, progress);
}
```

src/processor/operator/persistent/rel_batch_insert.cpp
```diff
@@ -53,6 +53,7 @@ void RelBatchInsert::executeInternal(ExecutionContext* context) {
     while (true) {
         relLocalState->nodeGroupIdx =
             partitionerSharedState->getNextPartition(relInfo->partitioningIdx);
+        updateProgress(context);
```
ray6080, line 56: Can u move this after `appendNodeGroup`? I think that makes more sense.

src/include/processor/operator/persistent/rel_batch_insert.h
```diff
@@ -29,6 +29,14 @@ struct RelBatchInsertPrintInfo final : OPPrintInfo {
         : OPPrintInfo(other), tableName(other.tableName) {}
 };
 
+struct RelBatchInsertProgressSharedState {
+    std::mutex mtx;
```
ray6080, line 33: If you make the change suggested on `PartitionerSharedState::getNextPartition`, you don't need a mutex here.

src/processor/operator/persistent/rel_batch_insert.cpp
```diff
@@ -225,5 +226,15 @@ void RelBatchInsert::finalizeInternal(ExecutionContext* context) {
     partitionerSharedState->partitioningBuffers[relInfo->partitioningIdx].reset();
 }
 
+void RelBatchInsert::updateProgress(ExecutionContext* context) {
+    std::lock_guard<std::mutex> lock(progressSharedState->mtx);
+    progressSharedState->partitionsDone = partitionerSharedState->nextPartitionIdx;
+    progressSharedState->partitionsTotal =
+        partitionerSharedState->numPartitions[partitionerSharedState->numPartitions.size() - 1];
```
ray6080, line 233: This is not the correct way to fetch `numPartitions`, you should make use of `partitioningIdx` inside `RelBatchInsertInfo` to grab the correct num partitions for the current directed rel table data being ingested.

src/processor/operator/persistent/rel_batch_insert.cpp
```diff
@@ -50,6 +50,7 @@ void RelBatchInsert::executeInternal(ExecutionContext* context) {
     const auto relInfo = info->ptrCast<RelBatchInsertInfo>();
     const auto relTable = sharedState->table->ptrCast<RelTable>();
     const auto relLocalState = localState->ptrCast<RelBatchInsertLocalState>();
+
```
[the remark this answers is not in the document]
WWW0030: Fixed

src/processor/operator/persistent/rel_batch_insert.cpp
WWW0030: Sounds good, changed in the latest commit

src/processor/operator/persistent/rel_batch_insert.cpp
WWW0030: This ran into some issues because it would require `partitioner.h` to include `rel_batch_insert.h`, I changed it by having `RelBatchInsertProgressSharedState` in its separate file, is that alright?

src/include/processor/operator/partitioner.h
```diff
@@ -2,6 +2,7 @@
 
 #include "common/enums/column_evaluate_type.h"
 #include "expression_evaluator/expression_evaluator.h"
+#include "processor/operator/persistent/rel_batch_insert_progress_state.h"
```
ray6080, line 5: Make this as a forward declaration instead of include.

src/include/processor/operator/persistent/rel_batch_insert.h
```diff
@@ -63,23 +64,28 @@ class RelBatchInsert final : public BatchInsert {
...
         : BatchInsert{std::move(info), std::move(sharedState), std::move(resultSetDescriptor), id,
               std::move(printInfo)},
-          partitionerSharedState{std::move(partitionerSharedState)} {}
+          partitionerSharedState{std::move(partitionerSharedState)},
+          progressSharedState{std::move(progressSharedState)} {}
 
     bool isSource() const override { return true; }
 
     void initLocalStateInternal(ResultSet* resultSet_, ExecutionContext* context) override;
+    void initGlobalStateInternal(ExecutionContext* context) override;
```
ray6080, line 77: Can u put this above L76?

src/processor/operator/partitioner.cpp
```diff
@@ -46,11 +46,13 @@ void PartitionerSharedState::initialize(const PartitionerDataInfo& dataInfo) {
     Partitioner::initializePartitioningStates(dataInfo, partitioningBuffers, numPartitions);
 }
 
-partition_idx_t PartitionerSharedState::getNextPartition(idx_t partitioningIdx) {
+partition_idx_t PartitionerSharedState::getNextPartition(idx_t partitioningIdx,
+    std::shared_ptr<RelBatchInsertProgressSharedState> progressSharedState) {
```
ray6080, line 50:
```suggestion
    RelBatchInsertProgressSharedState& progressSharedState) {
```
pass in reference should be enough here, as you don't actually share the ownership.

src/processor/map/map_copy_from.cpp
ray6080: hmm I thought you want to revert the change here? Don't you initialize the progressSharedState inside `   initGlobalStateInternal` already?

You can keep a private constructor for `RelBatchInsert` that's only for `clone`.

src/include/processor/operator/persistent/rel_batch_insert_progress_state.h
ray6080, line 1: Can u try move this back into `rel_batch_insert.h` after you remove `nclude "processor/operator/persistent/rel_batch_insert_progress_state.h"` from `partitioner.h`?

src/processor/operator/persistent/rel_batch_insert.cpp
```diff
@@ -225,5 +233,11 @@ void RelBatchInsert::finalizeInternal(ExecutionContext* context) {
     partitionerSharedState->partitioningBuffers[relInfo->partitioningIdx].reset();
 }
 
+void RelBatchInsert::updateProgress(ExecutionContext* context) {
+    double progress =
```
ray6080, line 237: Though not likely to happen, but I think you should guard against the cast when `progressSharedState->partitionsTotal` is `0`.

src/processor/map/map_copy_from.cpp
WWW0030: There are no instance of progressSharedState to construct the RelBatchInsert class in this part of the code, so I think that I need to make a dummy copy here (From my understanding it's not getting used?) Would this pose a problem?

src/include/processor/operator/persistent/rel_batch_insert.h
```diff
@@ -59,6 +66,17 @@ struct RelBatchInsertLocalState final : BatchInsertLocalState {
 
 class RelBatchInsert final : public BatchInsert {
 public:
+    RelBatchInsert(std::unique_ptr<BatchInsertInfo> info,
```
ray6080, line 69: Can u comment on the top that this constructor is only used for clone.

```
Note: This constructor is used for clone() only.
```

ray6080 (approved):
Let's do some sanity check of rel COPY performance. then this is all good to me. Thanks!

WWW0030:
Local testing the new progress report runtime stayed around the same.
On new progress bar implementation 19M rows took 3767.77ms, and 65M took 11858.82ms.
On master 19M rows took 3289.76ms and 65M took 12437.50ms

PR #4277, opened 2024-09-19
Avoid unnecessary allocation of CSR header chunks when scanning from local rel tables

ray6080:
# Description

Improved scan from local rel tables by skipping unnecessary allocation of CSR header chunks inside `CSRNodeGroupScanState`.

Did a simple benchmark by looping over each line in ldbc1 knows csv file with 180623 rows, and for each row, perform `match (a:person {id: rows[0]}) match (b:person {id: rows[1]} merge (a)-[e:knows]->(b);` in a single transaction.
The end-to-end execution time improves from 622s to 384s.

### DIFF
--- a/src/include/storage/store/csr_node_group.h
+++ b/src/include/storage/store/csr_node_group.h
@@ -122,16 +122,23 @@ struct CSRNodeGroupScanState final : NodeGroupScanState {
     // States at the csr list level. Cached during scan over a single csr list.
     CSRNodeGroupScanSource source = CSRNodeGroupScanSource::COMMITTED_PERSISTENT;
 
-    CSRNodeGroupScanState(MemoryManager& memoryManager, common::idx_t numChunks)
-        : NodeGroupScanState{numChunks},
-          csrHeader{std::make_unique<ChunkedCSRHeader>(memoryManager, false,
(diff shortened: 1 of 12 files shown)


src/include/storage/store/csr_node_group.h
```diff
@@ -122,16 +122,23 @@ struct CSRNodeGroupScanState final : NodeGroupScanState {
...
 
-    CSRNodeGroupScanState(MemoryManager& memoryManager, common::idx_t numChunks)
-        : NodeGroupScanState{numChunks},
-          csrHeader{std::make_unique<ChunkedCSRHeader>(memoryManager, false,
-              common::StorageConstants::NODE_GROUP_SIZE, ResidencyState::IN_MEMORY)} {}
+    explicit CSRNodeGroupScanState(common::idx_t numChunks)
+        : NodeGroupScanState{numChunks}, csrHeader{nullptr} {}
+
+    void initCSRHeader(MemoryManager& mm) {
+        if (!csrHeader) {
```
andyfengHKU, line 129: Normally this is not good practice. If u can have the assumption that csrHeader is only initialized once, then this if check shouldn't be needed.

andyfengHKU (approved):

PR #4303, opened 2024-09-25
Refactor Validate Property Name

WWW0030:
# Description

Was looking through `bind_ddl.cpp`, noticed that the checks for unique and reserved can probably both go into the same function/for loop. Just a suggestion :)

# Contributor agreement

### DIFF
--- a/src/include/binder/binder.h
+++ b/src/include/binder/binder.h
@@ -121,6 +121,8 @@ public:
         const std::vector<parser::ParsedPropertyDefinition>& parsedDefinitions,
         const std::string& tableName);
 
+    static void validatePropertyName(const std::vector<PropertyDefinition>& definitions);
+    
     /*** bind copy ***/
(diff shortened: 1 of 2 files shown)


src/include/binder/binder.h
```diff
@@ -121,6 +121,8 @@ class Binder {
         const std::vector<parser::ParsedPropertyDefinition>& parsedDefinitions,
         const std::string& tableName);
 
+    static void validatePropertyName(const std::vector<PropertyDefinition>& definitions);
```
andyfengHKU, line 124: Remove this. Let's try not add static function to header. Hide them in cpp file.

andyfengHKU (approved):
LGTM. But let's avoid having these small tiny refactor PRs. Just hide them in your next PR. A PR should either be

- a feature
- a refactor that is critical, i.e. affects multiple files/components.

Having too many PR records will cause extra work when we glance through PR list at release time.

Issue #4285, opened 2024-09-22
GDS and Recursive Joins TODOs


semihsalihoglu-uw:
This issue is intended to keep track of the suite of optimizations, refactoring, enhancements, and experimentations, we plan to implement in both the new fast recursive join implementations and the upcoming Kuzu GDS functions.

## Recursive Joins TODOs
- [ ]  Refactor: Implement getParameterTypeIDs() in each rjFunction separately in rec_joins.h. (Xiyang)
- [ ] Enhancement: Currently paths return only the intermediate node IDs. Also return edgeIDs. 2 main things need to be done on the backend: (i) ParentIterAndNextPtr* need to change; (ii) My suggestion is to change SingleSPPathsEdgeCompute to also use ParentIterAndNextPtr. The SinglePaths data structure can be thrown away. See my comment in SinglePaths::setParentEdge. (Xiyang)
- [ ] Enhancement: Integrate predicates to recursive joins, e.g., `(a)-[r:* |  WHERE r.since < 2022 AND n.age > 45]->(b)`. (Xiyang)
- [ ] Enhancement: Make recursive joins work on undirected recursive relationship patterns (e.g., `(a)-[r:*]-(b)`) as well as when scanning the graph in backwards direction. Currently we only work for scanning the graph in the forward direction.
- [ ] Optimization: Add early stopping to shortest paths computation when there are few destinations and have reached all.
- [ ] Optimization: Fix the isMasked(offset, offset) to have a fast path. We currently use masks in Frontier implementations but we need to check a single location. So we need a fast function here. This should be very quick. (Xiyang).
- [ ] Optimization: Reimplement Fast scans in on_disk_graph. (Ben)
- [ ] Optimization: Graph::scan functions should return iterators instead of copying the data from the buffer manager to a value vector. This may requires keeping pages in the buffer manager pinned for longer time but it should be fine as we do very quick work on each edge in edgeCompute() functions. (Ben). 
- [ ] Optimization: Bidirectional and direction-optimizing joins. Bidirectional joins is the optimization for cases when there is a single source and a single destination and implements the heuristic to take a step from the smaller frontier at each step. This can be used both for shortest as well as variable length joins (and will likely have an annoying backtracking algorithm).
[...]

PR #4310, opened 2024-09-27
Track edge ID in gds

andyfengHKU:
# Description

This PR tracks edgeID in GDS functions. Changes are stated as follows

- Add edgeID information to `ParentList`
- Change `edgeCompute` interface to include edgeID
- Change graph interface to also scan edgeID.

@benjaminwinger Pls check changes in `Graph & OnDiskGraph`. Feel free to make changes. My implementation is just a POC to check validate correctness. It won't be performant.

Fixes # (issue)

# Contributor agreement

### DIFF
--- a/src/function/gds/all_shortest_paths.cpp
+++ b/src/function/gds/all_shortest_paths.cpp
@@ -20,9 +20,7 @@ namespace function {
 // TODO(Xiyang): optimize if edgeID is not needed.
 class ParentList {
 public:
-    ParentList(uint16_t iter_, nodeID_t nodeID, relID_t edgeID) {
-        store(iter_, nodeID, edgeID);
-    }
+    ParentList(uint16_t iter_, nodeID_t nodeID, relID_t edgeID) { store(iter_, nodeID, edgeID); }
 
     void store(uint16_t iter_, nodeID_t nodeID, relID_t edgeID) {
         iter.store(iter_, std::memory_order_relaxed);
@@ -328,8 +326,8 @@ public:
         auto sourceNodeID = output->sourceNodeID;
         PathVectorWriter writer(pathNodeIDsVector.get(), pathEdgeIDsVector.get());
         dstNodeIDVector->setValue<common::nodeID_t>(0, dstNodeID);
-        auto firstParent = bfsGraph.getCurFixedParentPtrs()[dstNodeID.offset].load(
-            std::memory_order_relaxed);
+        auto firstParent =
+            bfsGraph.getCurFixedParentPtrs()[dstNodeID.offset].load(std::memory_order_relaxed);
 
         if (firstParent == nullptr) {
             // This case should only run for variable length joins.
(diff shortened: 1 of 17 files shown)


src/function/gds/all_shortest_paths.cpp
```diff
@@ -18,15 +18,15 @@ namespace kuzu {
 namespace function {
 
 class ParentList {
```
semihsalihoglu-uw, line 20: Not in this PR but I think we should think about how we should implement the optimization when edgeIDs are not needed. In that case we should not keep track of them here or in the FactorizedTables. What would be a simple way to implement that optimization? That can decrease the memory footprint of these algorithms by more than 1/3 and should just speed things up. Would be good to think about this now and put that optimization and your general approach in this [issue](https://github.com/kuzudb/kuzu/issues/4285).

src/function/gds/all_shortest_paths.cpp
```diff
@@ -365,7 +379,9 @@ class VarLenAndAllSPOutputWriterPaths : public RJOutputWriter {
                     curPath.pop_back();
                 }
             } else {
-                auto parent = allSpOutputs->bfsGraph.getInitialParentAndNextPtr(topNodeID);
+                auto parent = bfsGraph.getInitialParentAndNextPtr(topNodeID);
+                // TODO(Semih): do we actually need the following while loop?
+                KU_ASSERT(parent->getIter() == top->getIter() - 1);
                 while (parent->getIter() != top->getIter() - 1) {
```
semihsalihoglu-uw, line 385: Yes we do and I think the ASSERT is wrong. Consider running var-len joins on the following data starting from s: 
s->u->v, s->z1->u, s->z2->u, ..., s->zk->u.

Suppose we are backtracking from v. It will have iter 2, so we will be looking for "iter 1" parent edges of u. However u will first have k iter 2 edges because of the s-z_i->u paths, so we need to iterate through all.

Since all tests pass, it means this case is not covered. Can you add a simple test case for this?

src/function/gds/single_shortest_paths.cpp
```diff
@@ -146,12 +147,12 @@ class SingleSPOutputWriterPaths : public SingleSPOutputWriterLengths {
...
         writer.beginWritingNewPath(length);
-        common::nodeID_t curIntNode = dstNodeID;
-        while (writer.nextPathPos > 0) {
-            curIntNode =
-                rjOutputs->ptrCast<SingleSPOutputsPaths>()->singlePaths->getParent(curIntNode);
-            writer.addNewNodeID(curIntNode);
+        common::nodeID_t nodeID = dstNodeID;
+        for (auto i = 0u; i < length - 1; ++i) {
+            nodeID = rjOutputs->ptrCast<SingleSPOutputsPaths>()->singlePaths->getParent(nodeID);
+            writer.addNode(nodeID, length - 2 - i);
```
semihsalihoglu-uw, line 153: Single SPs is still not returning edgeIDs. Just so we are on the same page: one design I would consider is to remove SingleSP function completely and give a bool parameter "isSinglePaths" to all_sp_paths (since it will be using the same algorithm). But then use the parameter to change this block to a single cas operation instead of a while loop (if the first cas fails, it's OK, because there is another winner. That is we can skip this loop: 
```
while (!curPtr[childNodeID.offset].compare_exchange_strong(expected, parentEdgePtr))
            ;
```
I would similarly optimize the output DSF writer code (e.g., the iter-1 search that I commented about inside the while loop is not needed). Or you might still want to keep a separate outputwriter but at least the computation I think can be completely shared.

semihsalihoglu-uw (approved):
Let's add that test I mentioned please.

src/function/gds/all_shortest_paths.cpp
andyfengHKU: Done

src/function/gds/all_shortest_paths.cpp
andyfengHKU: Left a TODO in the codebase.

src/function/gds/single_shortest_paths.cpp
andyfengHKU: Yeah I'm making this change in the coming PR.

Issue #2911, opened 2024-02-19
Allow specifying PRIMARY KEY after the column definition in CREATE NODE TABLE statements


ray6080:
Currently, we only allow specifying `PRIMARY KEY` constraint at the end of `CREATE NODE TABLE` statements.
An example:
```
CREATE NODE TABLE person(id INT64, age INT64, name STRING, PRIMARY KEY(id));
```

We should also allow specifying the constraint after column definition as SQL does:
```
CREATE NODE TABLE person(id INT64 PRIMARY KEY, age INT64, name STRING);
```

andyfengHKU:
Fixed in #4314

PR #4314, opened 2024-09-27
Allowing PRIMARY KEY specification after Column Definition

WWW0030:
# Description

Added feature to allow PRIMARY KEY Specification after Column Definition to support the following kinds of syntax:

`CREATE NODE TABLE Person(id STRING PRIMARY KEY, firstName STRING, lastName STRING, gender STRING, birthday STRING, creationDate STRING, ip STRING, browserUsed STRING);`

`CREATE NODE TABLE Person(id STRING, firstName STRING, lastName STRING, gender STRING, birthday STRING PRIMARY KEY, creationDate STRING, ip STRING, browserUsed STRING);`

# Issue: 
https://github.com/kuzudb/kuzu/issues/2911#event-14202915593

# Testing:
Tested with Kuzu CLI on the following cases:
Should support:
- Specifying PRIMARY KEY on 1st column
- Specifying PRIMARY KEY on other columns

Should return exception:
- Using both PRIMARY KEY after column definition and PRIMARY KEY after all definitions
- Not having any PRIMARY KEY statements
- Writing PRIMARY KEY in wrong locations
- Specifying PRIMARY KEY on a unsupported type

# Contributor agreement


src/antlr4/Cypher.g4
WWW0030, line 1: I defined `kU_PropertyDefinitionBeforeConstraint` and `kU_PropertyDefinitionAfterConstraint` to allow the parser to know exactly where in the table to put the column definition with constraint.

src/parser/transform/transform_ddl.cpp
```diff
@@ -33,11 +33,19 @@ std::unique_ptr<Statement> Transformer::transformCreateNodeTable(
...
     }
+    else if (ctx.kU_PropertyDefinitionsWithConstraint()) {
+        pkName = transformPrimaryKey(*ctx.kU_PropertyDefinitionsWithConstraint());
+    }
     auto createTableInfo = CreateTableInfo(TableType::NODE, tableName,
         ctx.kU_IfNotExists() ? common::ConflictAction::ON_CONFLICT_DO_NOTHING :
                                common::ConflictAction::ON_CONFLICT_THROW);
-    createTableInfo.propertyDefinitions =
-        transformPropertyDefinitions(*ctx.kU_PropertyDefinitions());
+    if (ctx.kU_PropertyDefinitionsWithConstraint()) {
```
WWW0030, line 41: took a while to figure out that this is where is causing segmentation fault 🫠

test/test_files/call/call.test
ray6080, line 1: I don't understand the logic of newly added tests.
```
-STATEMENT CREATE NODE TABLE personAlt(id INT64 PRIMARY KEY, prop STRING DEFAULT 'Alice');
---- ok
-STATEMENT CREATE NODE TABLE personAlt2(id INT64, prop STRING DEFAULT 'Alice' PRIMARY KEY);
---- ok
```
Don't you need to verify that the new tables are created with expected primary keys through the `call table_info` statement?

WWW0030: Good point, I tested this manually, but we should have it in the test case. I have added `call table_info` checks into the test. Thanks!

scripts/antlr4/Cypher.g4
```diff
@@ -280,7 +280,7 @@ kU_AttachDatabase
     : ATTACH SP StringLiteral (SP AS SP oC_SchemaName)? SP '(' SP? DBTYPE SP oC_SymbolicName (SP? ',' SP? kU_Options)? SP? ')' ;
 
 kU_Option
-    : oC_SymbolicName SP? '=' SP? oC_Literal ;
+    : oC_SymbolicName (SP? '=' SP? | SP*) oC_Literal | oC_SymbolicName;
```
andyfengHKU, line 283: Why is this grammar rule changing? It doesn't relate to what PR is trying to implement.

scripts/antlr4/Cypher.g4
```diff
@@ -954,10 +963,6 @@ kU_NonReservedKeywords
         | TYPE
         | USE
         | WRITE
-        | SINGLE
```
andyfengHKU, line 957: Why are these being removed?

scripts/antlr4/Cypher.g4
```diff
@@ -393,8 +394,16 @@ kU_PropertyDefinitions : kU_PropertyDefinition ( SP? ',' SP? kU_PropertyDefiniti
 
 kU_PropertyDefinition : kU_ColumnDefinition ( SP kU_Default )? ;
 
+kU_PropertyDefinitionBeforeConstraint : kU_ColumnDefinition ( SP kU_Default )? ;
```
andyfengHKU, line 397: Don't add these new rules. You should modify kU_PropertyDefinition to make it also accept primary key. Something like

```
kU_PropertyDefinition : kU_ColumnDefinition ( SP kU_Default )?  ( SP PRIMARY SP KEY)?;
```

src/parser/transform/transform_ddl.cpp
```diff
@@ -277,6 +284,56 @@ std::vector<ParsedPropertyDefinition> Transformer::transformPropertyDefinitions(
     return definitions;
 }
 
+std::vector<ParsedPropertyDefinition> Transformer::transformPropertyDefinitions(
```
andyfengHKU, line 287: All these should be reverted once u change the grammar.

andyfengHKU (changes requested):
We should keep parser, i.e. Cypher.g4 simple and let binder handles cases like duplicate primary key, etc. Because the error message thrown from the parser is very implicit.

scripts/antlr4/Cypher.g4
WWW0030: I think this is from Sterling's commit:
https://github.com/kuzudb/kuzu/commit/7c69437d92083f9ad6fd6f8c21cda740dbca9984
I believe it's because he didn't update `scripts/antlr4/Cypher.g4` in his PR, so it got updated in mine.

scripts/antlr4/Cypher.g4
WWW0030: I believe this will cause problems, where a faulty statement with multiple `PRIMARY KEY` specified will be accepted. In my understanding:
`kU_PropertyDefinition` doesn't know whether the other `kU_PropertyDefinition` within the same `kU_PropertyDefinitions` has `PRIMARY KEY` defined. If we use the original rules and only change `kU_PropertyDefinition`, cases where no `PRIMARY KEY` or multiple `PRIMARY KEY` will pass while they are suppose to be flagged.

scripts/antlr4/Cypher.g4
WWW0030: These should be related to Sterling's commit above

src/parser/transform/transform_ddl.cpp
```diff
@@ -26,13 +26,36 @@ std::unique_ptr<Statement> Transformer::transformAlterTable(
     }
 }
 
-std::unique_ptr<Statement> Transformer::transformCreateNodeTable(
-    CypherParser::KU_CreateNodeTableContext& ctx) {
-    auto tableName = transformSchemaName(*ctx.oC_SchemaName());
+std::string Transformer::getPKName(CypherParser::KU_CreateNodeTableContext& ctx) {
+    int pkCount = 0;
```
andyfengHKU, line 30: auto

src/parser/transform/transform_ddl.cpp
```diff
@@ -26,13 +26,36 @@ std::unique_ptr<Statement> Transformer::transformAlterTable(
...
+        pkCount++;
         pkName = transformPrimaryKey(*ctx.kU_CreateNodeConstraint());
     }
+    if (pkCount == 0) {
+        // Raise exception when no PRIMARY KEY is specified.
+        throw ParserException(
+            "Error while handling PRIMARY KEY Definition: please specify PRIMARY KEY.");
+    } else if (pkCount > 1) {
+        throw ParserException(
+            "Error while handling PRIMARY KEY Definition: multiple PRIMARY KEY found.");
```
andyfengHKU, line 49: change to `throw ParserException("Found multiple primary keys.")`

src/parser/transform/transform_ddl.cpp
```diff
@@ -26,13 +26,36 @@ std::unique_ptr<Statement> Transformer::transformAlterTable(
...
+        if (definition->PRIMARY() && definition->KEY()) {
+            pkCount++;
+            pkName = transformPrimaryKey(*definition->kU_ColumnDefinition());
+        }
+    }
     if (ctx.kU_CreateNodeConstraint()) {
+        pkCount++;
         pkName = transformPrimaryKey(*ctx.kU_CreateNodeConstraint());
     }
+    if (pkCount == 0) {
```
andyfengHKU, line 43: Double check if binder already handles this. I vaguely remember there was similar logic in binder.

If so, no need to check here.

Otherwise, change to `throw ParserException("Cannot find primary key.")`

src/parser/transform/transform_ddl.cpp
```diff
@@ -26,13 +26,36 @@ std::unique_ptr<Statement> Transformer::transformAlterTable(
...
     std::string pkName;
+    auto& propertyDefinitions = *ctx.kU_PropertyDefinitions();
+    for (auto& definition : propertyDefinitions.kU_PropertyDefinition()) {
+        if (definition->PRIMARY() && definition->KEY()) {
+            pkCount++;
+            pkName = transformPrimaryKey(*definition->kU_ColumnDefinition());
+        }
+    }
     if (ctx.kU_CreateNodeConstraint()) {
+        pkCount++;
```
andyfengHKU, line 40: We probably doesn't want to error in the following case

```
CREATE NODE TABLE (id INT64 PRIMARY KEY,  PRIMARY KEY(id));
```

Try handle this by also checking if primary key names are the name.

andyfengHKU (approved):
LGTM. Just left some minor comments.

src/parser/transform/transform_ddl.cpp
WWW0030: Yes, binder currently handle this, but I think this logic should still stay in here because it is more informative. Right now binder's exception returns:
`Error: Binder exception: Primary key  does not match any of the predefined node properties.`
which doesn't really fit with what the problem is, which is that there is no primary key.

Issue #4250, opened 2024-09-12
Add support for CALL function without RETURN clause


royi-luo:
We have some CALL functions such as the one for [clearing cached warnings](https://github.com/kuzudb/kuzu/issues/4226) that change some state in the connection/DB without needing to return anything.

It would be nice to not need to return any value for these types of functions (similar to having a void return type).

ray6080:
Done in https://github.com/kuzudb/kuzu/pull/4390.

prrao87:
Is there any documentation for this? How is this going to look when we actually call the `CALL` statement now?

ray6080:
Let me ping @acquamarin here to add the docs for generic standalone call function. Also @royi-luo can u update the dev docs on `CALL show_warnings() RETURN *;` (e.g., http://dev-docs.kuzudb.com/import/csv/#ignoring-erroneous-rows)? The syntax should have changed to `CALL show_warnings()` now.

PR #4315, opened 2024-09-27
Allow CALL functions without RETURN clause

royi-luo:
# Description

We have some "void" CALL functions such as `clear_warnings()`, this removes the need for these functions to return any values.

To do this, we modify the grammar to break `oC_ReadingClause` into `oC_MandatoryReturnReadingClause` (for clauses like `MATCH` and `LOAD_FROM`) and `oC_OptionalReturnReadingClause` (for CALL clauses). For optional return reading clauses, the check for `RETURN` is moved to the binding phase and only throws if the `CALL` clause returns non-empty output.

Fixes https://github.com/kuzudb/kuzu/issues/4250

# Contributor agreement


scripts/antlr4/Cypher.g4
```diff
@@ -478,17 +481,33 @@ oC_UpdatingClause
...
 kU_InQueryCall
+    : kU_InQueryTableFunctionCall
+        | kU_InQueryBuiltInCall
+        ;
+
+kU_InQueryTableFunctionCall
+    : ( kU_ProjectGraph SP? )? CALL SP oC_TableFunctionCall (SP? oC_Where)? ;
+
+kU_InQueryBuiltInCall
     : ( kU_ProjectGraph SP? )? CALL SP oC_FunctionInvocation (SP? oC_Where)? ;
```
royi-luo, line 511: Is `CALL COUNT(*)` or `CALL CAST` a thing or can I remove `kU_InQueryBuiltInCall` altogether?

andyfengHKU (changes requested):
hmmmm, this is wrong way to handle `CALL clear_warning()`.

Don't compile it as a `SingleQuery` or `TableFunction`.

Compile it as as a `StandaloneCall`

royi-luo:
> hmmmm, this is wrong way to handle `CALL clear_warning()`.
> 
> Don't compile it as a `SingleQuery` or `TableFunction`.
> 
> Compile it as a `StandaloneCall`.
> 
> I suggest let @acquamarin take over or mentor.

I see, will update and let @acquamarin know when I'm done

andyfengHKU:
> > hmmmm, this is wrong way to handle `CALL clear_warning()`.
> > Don't compile it as a `SingleQuery` or `TableFunction`.
> > Compile it as a `StandaloneCall`.
> > I suggest let @acquamarin take over or mentor.
> 
> I see, will update and let @acquamarin know when I'm done

Talk to @acquamarin before u start. I don't know if he plans to take over or not.

royi-luo:
Closing as @acquamarin is taking this over.

PR #4317, opened 2024-09-29
Deprecate single shortest path path tracking

andyfengHKU:
# Description

This PR removes the special code to track path for single shortest path and reuse the infrastructure for all shortest path.
It contains a lot of changes because I need to move classes around in order for them to be sharable across single/all shortest path.

I would suggest only review `src/function/gds/single_shortest_paths.cpp`.

Some refactors are listed as follows

We now have

```
RJOutputs
SPOutputs : RJOutputs

SingleSPOutputs : SPOutputs // track destination only, destination always has cardinality 1
AllSPMultiplicitiesOutputs : SPOutputs // track destination only, destination may have cardinality >= 1
PathsOutputs : SPOutputs // track path
```

Corresponding to `XXOutputs`, we have `XXOutputWriter`. Btw the naming is inconsistent, we should either use `Output` or `Outputs`

```
RJOutputWriter

DestinationsOutputWriter : RJOutputWriter // write destinations only
AllSPDestinationsOutputWriter : DestinationsOutputWriter // write destinations but with cardinality
AllSPLengthsOutputWriter : DestinationsOutputWriter // write destinations and length
SingleSPDestinationsLengthOutputWriter : DestinationsOutputWriter // write destinations but cardinality always =1

PathsOutputWriter : PathsOutputWriter // write paths
SPPathsOutputWriter : PathsOutputWriter
VarLenPathsOutputWriter : PathsOutputWriter // two different path writer have different skipping logic. See code for details.
```

Note that I have mad naming changes to OutputWriter, specifically
- use `Destinations` to replace `Dst`
- change `OutputWriterXXX` (e.g. `OutputWriterPaths`) to `XXXOutputWriter` so that we are more consistent with `XXXOutputs`

Fixes # (issue)

# Contributor agreement


src/function/gds/single_shortest_paths.cpp
```diff
@@ -127,74 +56,52 @@ class SingleSPOutputWriterLengths : public SPOutputWriterDsts {
...
+    }
+
+    bool edgeCompute(nodeID_t boundNodeID, nodeID_t nbrNodeID, relID_t edgeID) override {
+        auto shouldUpdate = frontierPair->pathLengths->getMaskValueFromNextFrontierFixedMask(
+                                nbrNodeID.offset) == PathLengths::UNVISITED;
+        if (shouldUpdate) {
+            if (!parentListBlock->hasSpace()) {
+                parentListBlock = bfsGraph->addNewBlock();
+            }
+            bfsGraph->addParent(frontierPair->curIter.load(std::memory_order_relaxed),
```
semihsalihoglu-uw, line 91: Hmm, you are still possibly keeping track of multiple parents per node if you dont have a special addParent function for single shortest paths. So the parent writing still happens inside a while loop and does not overwrite any previously written parents. This might be OK as long as the output writing takes care of outputting only one path (but it would still be slower). That said, I don't see that happening. So it looks like the current outputwriter used for single shortest paths, when outputting paths, can output multiple paths per node because it uses exactly the same `SPPathsOutputWriter` that AllSPAlgorithm uses. Does this make sense? I'm thinking of the case when 2 threads, T1 and T2 readh shouldUpdate as true for a neighbor node u here, then then they both call bfsGraph->addParent. That will end up inserting 2 parent edges for u. We need to fix this.  

So my suggestion was to have an if/else block inside addParent for single vs all shortest paths. All shortest paths uses a while loop while single sp does not.

src/include/function/gds/output_writer.h
```diff
@@ -0,0 +1,136 @@
...
+};
+
+class SPPathsOutputWriter : public PathsOutputWriter {
+public:
+    SPPathsOutputWriter(main::ClientContext* context, RJOutputs* rjOutputs, uint16_t upperBound)
+        : PathsOutputWriter(context, rjOutputs, 1 /* lower bound */, upperBound) {}
+
+    bool skipWriting(common::nodeID_t dstNodeID) const override {
+        auto pathsOutputs = rjOutputs->ptrCast<PathsOutputs>();
+        // For all shortest path computations, we do not output any results from source to source.
```
semihsalihoglu-uw, line 123: Let's update this comment. This comment sounds like we don't output any results from s to s only for "all shortest paths" but we should not output s to s for "single shortest paths" as well. So let's be clear and say "For single or all shortest paths computations ..."

semihsalihoglu-uw (approved):
Approving but I think there is a bug in the code.

src/include/function/gds/output_writer.h
andyfengHKU: Done

src/function/gds/single_shortest_paths.cpp
andyfengHKU: Done. Though it might be good if u take another look at `tryAddSingleParent` function.

PR #4327, opened 2024-10-01
Use an iterator instead of collecting Graph scan results into an std::vector

benjaminwinger:
This has a smaller benefit than it did initially (now roughly a 5-6% improvement in the comparison I'd done; I'll try some more benchmarks and add them in a later comment). I think this is because adding the edge IDs (in #4310) increased the time spent scanning but didn't increase the cost of the overhead for the results as much.

The iterator is actually producing chunks of values as I found that eliminating that had performance tradeoffs that seemed to outweigh the benefits of avoiding the extra copy (I don't think there's an efficient way of turning a nested loop into an iterator; it might be worth looking at [std::ranges::join_view](https://en.cppreference.com/w/cpp/ranges/join_view), though we've had issues with std::ranges and apple clang in the past).
The iterator is not polymorphic, but passes off the task of fetching the next chunk of results to the scan state. `Graph::Iterator` is just providing an easy to use interface to work with the scan state.

It's important to note that you cannot begin another scan using the same scan state and then continue the original scan. All of the iterator's state is stored in the scan state, not in the iterator. For page rank I added a second scan state to work around this, and in weakly connected components I made it collect the results into a vector so that the scan always can finish before the function gets called recursively (the alternative would be one scan state per function call, which would probably be slower), so it doesn't benefit from this at all.

### DIFF
--- a/src/function/gds/gds_task.cpp
+++ b/src/function/gds/gds_task.cpp
@@ -10,19 +10,19 @@ void FrontierTask::run() {
     auto numApproxActiveNodesForNextIter = 0u;
     std::unique_ptr<graph::Graph> graph = sharedState->graph->copy();
     auto state = graph->prepareScan(sharedState->relTableIDToScan);
-    auto scanResult = graph::GraphScanResult();
     auto localEc = sharedState->ec.copy();
     while (sharedState->frontierPair.getNextRangeMorsel(frontierMorsel)) {
         while (frontierMorsel.hasNextOffset()) {
             common::nodeID_t nodeID = frontierMorsel.getNextNodeID();
             if (sharedState->frontierPair.curFrontier->isActive(nodeID)) {
-                graph->scanFwd(nodeID, *state, scanResult);
-                for (auto i = 0u; i < scanResult.size(); ++i) {
-                    auto nbrNodeID = scanResult.nbrNodeIDs[i];
-                    auto edgeID = scanResult.edgeIDs[i];
-                    if (localEc->edgeCompute(nodeID, nbrNodeID, edgeID)) {
-                        sharedState->frontierPair.nextFrontier->setActive(nbrNodeID);
-                        numApproxActiveNodesForNextIter++;
+                for (const auto [nodes, edges] : graph->scanFwd(nodeID, *state)) {
+                    for (size_t i = 0; i < nodes.size(); i++) {
+                        const auto& nbrNodeID = nodes[i];
+                        const auto& edgeID = edges[i];
+                        if (localEc->edgeCompute(nodeID, nbrNodeID, edgeID)) {
+                            sharedState->frontierPair.nextFrontier->setActive(nbrNodeID);
+                            numApproxActiveNodesForNextIter++;
+                        }
                     }
                 }
             }
--- a/src/include/graph/graph.h
+++ b/src/include/graph/graph.h
@@ -1,7 +1,10 @@
 #pragma once
 
+#include <cstdint>
+#include <iterator>
 #include <memory>
 
+#include "common/copy_constructors.h"
 #include "common/types/types.h"
 #include <span>
 
@@ -17,22 +20,10 @@ struct RelTableIDInfo {
 class GraphScanState {
 public:
     virtual ~GraphScanState() = default;
-};
-
-// TODO(Ben): feel free to remove this class.
-struct GraphScanResult {
-    std::vector<common::nodeID_t> nbrNodeIDs;
-    std::vector<common::relID_t> edgeIDs;
-
-    common::idx_t size() const {
-        KU_ASSERT(nbrNodeIDs.size() == edgeIDs.size());
-        return nbrNodeIDs.size();
-    }
-
-    void clear() {
-        nbrNodeIDs.clear();
-        edgeIDs.clear();
-    }
+    virtual std::span<const common::nodeID_t> getNbrNodes() const = 0;
+    virtual std::span<const common::relID_t> getEdges() const = 0;
+    // Returns true if there are more values after the current batch
+    virtual bool next() = 0;
 };
 
 /**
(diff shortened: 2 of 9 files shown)


src/function/gds/gds_task.cpp
```diff
@@ -10,19 +10,19 @@ void FrontierTask::run() {
...
             common::nodeID_t nodeID = frontierMorsel.getNextNodeID();
             if (sharedState->frontierPair.curFrontier->isActive(nodeID)) {
-                graph->scanFwd(nodeID, *state, scanResult);
-                for (auto i = 0u; i < scanResult.size(); ++i) {
-                    auto nbrNodeID = scanResult.nbrNodeIDs[i];
-                    auto edgeID = scanResult.edgeIDs[i];
-                    if (localEc->edgeCompute(nodeID, nbrNodeID, edgeID)) {
-                        sharedState->frontierPair.nextFrontier->setActive(nbrNodeID);
-                        numApproxActiveNodesForNextIter++;
+                for (const auto [nodes, edges] : graph->scanFwd(nodeID, *state)) {
```
andyfengHKU, line 18: I just a minor question. In my future PRs, depends on configuration, I may need to both scanFwd and scanBwd.
Will the following works if I'm reusing the same state? Or do I need fwdState and bwdState?

```
for (const auto [nodes, edges] : graph->scanFwd(nodeID, *state)) {
}

for (const auto [nodes, edges] : graph->scanBwd(nodeID, *state)) {
}
```

benjaminwinger: Yes, that should be fine. You also should always be able to re-use the state for a new scan if you've finished the original, you just can't nest them (even with a backward scan, though that would actually be fairly easy to allow as long as it's not recursive).

src/include/graph/graph.h
```diff
@@ -45,6 +36,68 @@ struct GraphScanResult {
...
+        Iterator(const Iterator& other) = default;
+
+        using iterator_category = std::input_iterator_tag;
+        using difference_type = std::ptrdiff_t;
+        using value_type =
+            std::pair<std::span<const common::nodeID_t>, std::span<const common::relID_t>>;
+        using pointer = value_type;
+        using reference = value_type;
+
+        std::pair<std::span<const common::nodeID_t>, std::span<const common::nodeID_t>>
```
ray6080, line 60:
```suggestion
        value_type
```
I think you can directly use `value_type` here.

src/include/graph/graph.h
```diff
@@ -45,6 +36,68 @@ struct GraphScanResult {
...
+        uint64_t count() {
+            // TODO(bmwinger): avoid scanning if all that's necessary is to count the results
+            uint64_t result = 0;
+            do {
+                result += scanState->getNodes().size();
+            } while (scanState->next());
+            return result;
+        }
+
+        std::vector<common::nodeID_t> collectNodes() {
```
ray6080, line 85: Would `collectNbrNodes` be more accurate here?

src/include/graph/graph.h
```diff
@@ -45,6 +36,68 @@ struct GraphScanResult {
...
+            std::vector<common::nodeID_t> adjacentNodes;
```
ray6080, line 86: Would it better to rename `adjacentNodes` to `nbrNodes`, though they mean the same? I think we should be more constrained to use multiple different terms referring to the same thing.

src/include/graph/graph.h
```diff
@@ -17,22 +20,10 @@ struct RelTableIDInfo {
...
-    common::idx_t size() const {
-        KU_ASSERT(nbrNodeIDs.size() == edgeIDs.size());
-        return nbrNodeIDs.size();
-    }
-
-    void clear() {
-        nbrNodeIDs.clear();
-        edgeIDs.clear();
-    }
+    virtual std::span<const common::nodeID_t> getNodes() const = 0;
```
ray6080, line 23: `getNodes` -> `getNbrNodes`

src/include/graph/on_disk_graph.h
```diff
@@ -13,10 +22,54 @@ class MemoryManager;
...
-
-    explicit OnDiskGraphScanState(const storage::RelTable& table,
+    class InnerIterator {
+    public:
+        InnerIterator(const main::ClientContext* context, storage::RelTable* relTable,
+            std::unique_ptr<storage::RelTableScanState> tableScanState);
+
+        DELETE_COPY_DEFAULT_MOVE(InnerIterator);
+
+        std::span<const common::nodeID_t> getNodes() const {
```
ray6080, line 32: `getNodes` -> `getNbrNodes`

src/include/graph/on_disk_graph.h
```diff
@@ -13,10 +22,54 @@ class MemoryManager;
...
+            return std::span<const common::nodeID_t>(
+                &relIDVector().getValue<const common::nodeID_t>(firstElement),
+                dstSelVector().getSelSize());
+        }
+
+        bool next();
+        void initScan();
+
+    private:
+        const common::SelectionVector& dstSelVector() const {
```
ray6080, line 58: Can we also unify `dst` to `nbr`?

src/include/graph/graph.h
```diff
@@ -45,6 +36,68 @@ struct GraphScanResult {
...
+        explicit Iterator(GraphScanState* scanState) : scanState{scanState} {}
+        DEFAULT_BOTH_MOVE(Iterator);
+        Iterator(const Iterator& other) = default;
+
+        using iterator_category = std::input_iterator_tag;
+        using difference_type = std::ptrdiff_t;
+        using value_type =
+            std::pair<std::span<const common::nodeID_t>, std::span<const common::relID_t>>;
+        using pointer = value_type;
+        using reference = value_type;
```
ray6080, line 58: What are `pointer` and `reference` for here? Are they required by the input_iterator?

src/include/graph/graph.h
```diff
@@ -45,6 +36,68 @@ struct GraphScanResult {
...
+        std::vector<common::nodeID_t> collectNodes() {
+            std::vector<common::nodeID_t> adjacentNodes;
+            for (const auto [nbrs, edges] : *this) {
+                adjacentNodes.insert(adjacentNodes.end(), nbrs.begin(), nbrs.end());
+            }
+            return adjacentNodes;
+        }
+
+        Iterator& begin() noexcept { return *this; }
+        Iterator end() const noexcept { return Iterator(nullptr); }
```
ray6080, line 94: I think you can replace `end()` by a constexpr Iterator object.

ray6080 (approved):

src/include/graph/graph.h
benjaminwinger: Oh, so that's an abbreviation of neighbour? I thought it was number and was a little confused when I'd seen it elsewhere.

ray6080: 😄 yeah we use that for abbr of neighbour.

src/include/graph/on_disk_graph.h
benjaminwinger: I'd put dstSelVector there since it's the selection vector for the dstNodeIDVector. Or do you mean make the dstVectors be nbrVectors too? I guess that there is no longer a single output/destination storage with the edges being stored too, so that would probably be more appropriate.

PR #4336, opened 2024-10-03
Add edge direction to gds

andyfengHKU:
# Description

Track edge direction when we extend both fwd and bwd adjacency list. This is not the most performant implementation since we will always store a boolean field in `ParentList` which could be optimized. We can try to figure this out eventually. This is not yet my priority.

Fixes # (issue)

# Contributor agreement

### DIFF
--- a/src/function/gds/output_writer.cpp
+++ b/src/function/gds/output_writer.cpp
@@ -19,32 +19,33 @@ void PathsOutputs::beginWritingOutputsForDstNodesInTable(common::table_id_t tabl
     bfsGraph.pinNodeTable(tableID);
 }
 
+static std::unique_ptr<ValueVector> createVector(const LogicalType& type, storage::MemoryManager* mm, std::vector<ValueVector*>& vectors) {
+    auto vector = std::make_unique<ValueVector>(type.copy(), mm);
+    vector->state = DataChunkState::getSingleValueDataChunkState();
+    vectors.push_back(vector.get());
+    return vector;
+}
+
 RJOutputWriter::RJOutputWriter(main::ClientContext* context, RJOutputs* rjOutputs)
     : context{context}, rjOutputs{rjOutputs} {
     auto mm = context->getMemoryManager();
-    srcNodeIDVector = std::make_unique<ValueVector>(LogicalType::INTERNAL_ID(), mm);
-    dstNodeIDVector = std::make_unique<ValueVector>(LogicalType::INTERNAL_ID(), mm);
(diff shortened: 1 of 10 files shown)


src/function/gds/output_writer.cpp
```diff
@@ -19,32 +19,35 @@ void PathsOutputs::beginWritingOutputsForDstNodesInTable(common::table_id_t tabl
     bfsGraph.pinNodeTable(tableID);
 }
 
+static std::unique_ptr<ValueVector> createVector(const LogicalType& type,
```
semihsalihoglu-uw, line 22: Can you put this into another common utility function? This doesn't look like a gds-related utility function. Looks like something many other parts of the system can use as well. Do we have a VectorUtils-like or CommonUtils-like class?

src/function/gds/output_writer.cpp
```diff
@@ -19,32 +19,35 @@ void PathsOutputs::beginWritingOutputsForDstNodesInTable(common::table_id_t tabl
...
-    srcNodeIDVector->state = DataChunkState::getSingleValueDataChunkState();
+    srcNodeIDVector = createVector(LogicalType::INTERNAL_ID(), mm, vectors);
+    dstNodeIDVector = createVector(LogicalType::INTERNAL_ID(), mm, vectors);
     srcNodeIDVector->setValue<nodeID_t>(0, rjOutputs->sourceNodeID);
-    dstNodeIDVector->state = DataChunkState::getSingleValueDataChunkState();
-    vectors.push_back(srcNodeIDVector.get());
-    vectors.push_back(dstNodeIDVector.get());
 }
 
 PathsOutputWriter::PathsOutputWriter(main::ClientContext* context, RJOutputs* rjOutputs,
```
semihsalihoglu-uw, line 38: I would consider simplifying this by always writing the direction now. So always create a direction value vector and always write it and the outputter can ignore it.

semihsalihoglu-uw (approved):
Can you add a TODO in that gds optimizations issue to optimize this later?

src/function/gds/output_writer.cpp
andyfengHKU: I don't recall other parts using this. The part of assigning `SingleDataChunkState` is very specific to GDS.

src/function/gds/output_writer.cpp
andyfengHKU: I'll think about it. Right now it's not too much work yet. Later if I observe we need to do if check in multiple places I might apply this change.

Issue #4148, opened 2024-08-25
Add support for MAP data type for C API


mewim:
### API

C

### Description

The MAP data type support is missing for C API. As such, it also cannot be binded to Go API.

PR #4339, opened 2024-10-03
Add support for MAP data type for C API

WWW0030:
# Description

Added MAP data type to c_api
Added tests to MAP data type

Fixes # (issue)
https://github.com/kuzudb/kuzu/issues/4148

# Contributor agreement

### DIFF
--- a/src/include/common/types/types.h
+++ b/src/include/common/types/types.h
@@ -231,7 +231,6 @@ enum class PhysicalTypeID : uint8_t {
     ARRAY = 23,
     STRUCT = 24,
     POINTER = 25,
-    MAP = 26,
 };
 
 class ExtraTypeInfo;
--- a/src/include/common/types/types.h
+++ b/src/include/common/types/types.h
@@ -330,7 +330,6 @@ public:
 
     static KUZU_API LogicalType LIST(LogicalType childType);
 
-    static KUZU_API LogicalType MAP(std::unique_ptr<StructTypeInfo> typeInfo);
     template<class T>
     static inline LogicalType LIST(T&& childType) {
         return LogicalType::LIST(LogicalType(std::forward<T>(childType)));
--- a/src/include/common/types/types.h
+++ b/src/include/common/types/types.h
@@ -329,7 +329,6 @@ public:
     static KUZU_API LogicalType UNION(std::vector<StructField>&& fields);
 
     static KUZU_API LogicalType LIST(LogicalType childType);
-
     template<class T>
     static inline LogicalType LIST(T&& childType) {
         return LogicalType::LIST(LogicalType(std::forward<T>(childType)));
--- a/src/include/common/types/types.h
+++ b/src/include/common/types/types.h
@@ -231,6 +231,7 @@ enum class PhysicalTypeID : uint8_t {
     ARRAY = 23,
     STRUCT = 24,
     POINTER = 25,
+    MAP = 26,
 };
 
 class ExtraTypeInfo;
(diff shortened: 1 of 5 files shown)


src/include/common/types/types.h
```diff
@@ -231,6 +231,7 @@ enum class PhysicalTypeID : uint8_t {
     ARRAY = 23,
     STRUCT = 24,
     POINTER = 25,
+    MAP = 26,
```
mewim, line 234: `MAP` is not a physical type. You should NOT add a physical type for it.

mewim:
The test cases are not enough. You should also add value tests to make sure that a dict returned as query result can actually be used in C API (i.e., you can read from it).

mewim (changes requested):
See my comments

mewim:
Also, I do not think the implmentation is correct. Reading from `MAP` should be done with `NestedVal`. Please refer to https://github.com/kuzudb/kuzu/blob/fae65c679902ea73721784378e64b13421ff237a/tools/python_api/src_cpp/py_query_result.cpp#L280-L289

WWW0030:
> Also, I do not think the implmentation is correct. Reading from `MAP` should be done with `NestedVal`. Please refer to
> 
> https://github.com/kuzudb/kuzu/blob/fae65c679902ea73721784378e64b13421ff237a/tools/python_api/src_cpp/py_query_result.cpp#L280-L289

I think that it is already done with NestedVal inside of the called function `kuzu_value_get_list_element`. The added test seems to be working fine aswell. Please let me know if its OK.

mewim:
> > Also, I do not think the implmentation is correct. Reading from `MAP` should be done with `NestedVal`. Please refer to
> > https://github.com/kuzudb/kuzu/blob/fae65c679902ea73721784378e64b13421ff237a/tools/python_api/src_cpp/py_query_result.cpp#L280-L289
> 
> I think that it is already done with NestedVal inside of the called function `kuzu_value_get_list_element`. The added test seems to be working fine aswell. Please let me know if its OK.

I don't think you called `kuzu_value_get_map_field_name` in the test case. My expected behavior is that `kuzu_value_get_map_field_name` returns `audience1` and `kuzu_value_get_map_field_value` returns a value of `33`. The user SHOULD NOT need to call another `kuzu_value_get_struct_field_name` or `kuzu_value_get_struct_field_value` to read the key or value from the map.

mewim (changes requested):
The ergonomics of the API is not desirable. Further changes are needed.

test/c_api/value_test.cpp
```diff
@@ -556,6 +555,93 @@ TEST_F(CApiValueTest, GetStructFieldValue) {
...
+    ASSERT_EQ(state, KuzuSuccess);
+    ASSERT_TRUE(kuzu_query_result_is_success(&result));
+    ASSERT_TRUE(kuzu_query_result_has_next(&result));
+    state = kuzu_query_result_get_next(&result, &flatTuple);
+    ASSERT_EQ(state, KuzuSuccess);
+    ASSERT_FALSE(kuzu_query_result_has_next(&result));
+    kuzu_value value;
+    ASSERT_EQ(kuzu_flat_tuple_get_value(&flatTuple, 0, &value), KuzuSuccess);
+
+    char* mapName;
```
mewim, line 599: There is probably a memory leak here. The `mapName` is not destroyed.

WWW0030: Should be destroyed below, line 602

mewim (approved):
Can be merged if all CI passes

PR #4340, opened 2024-10-04
Add semi mask point lookup

andyfengHKU:
# Description

Add an optimization mentioned in #4285 

Fixes # (issue)

# Contributor agreement

### DIFF
--- a/src/include/common/mask.h
+++ b/src/include/common/mask.h
@@ -52,8 +52,11 @@ public:
         maskData = std::make_unique<MaskData>(maxOffset + 1);
     }
 
+    bool isMasked(common::offset_t offset) const {
+        return maskData->isMasked(offset, numMasks);
+    }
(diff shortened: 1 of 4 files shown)


src/include/common/mask.h
```diff
@@ -52,6 +52,7 @@ class MaskCollection {
         maskData = std::make_unique<MaskData>(maxOffset + 1);
     }
 
+    bool isMasked(common::offset_t offset) const { return maskData->isMasked(offset, numMasks); }
     // Return true if any offset between [startOffset, endOffset] is masked. Otherwise return false.
     bool isMasked(common::offset_t startOffset, common::offset_t endOffset) const {
```
ray6080, line 57: I feel this should be renamed to `isAnyMasked`.

ray6080 (approved):

PR #4341, opened 2024-10-04
CLI auto-complete highlighting and initial improvements

MSebanc:
# Description

Adds grey highlighting for completion and some other initial improvements. 

image

More work will be done to make the suggestions better and add suggestions for various other things like column names.

# Contributor agreement


andyfengHKU (approved):
Very cool! As always I just glanced through the changes and played with the shell a bit. Looks very interesting.

PR #4346, opened 2024-10-07
SelectionVector cleanup

benjaminwinger:
Three minor changes to the SelectionVector

1. getMu**l**tableBuffer -> getMutableBuffer
2. I made `INCREMENTAL_SELECTED_POS` be initialized with a constexpr function
3. I added a `forEach` function, which makes it easier to use the `isUnfiltered` optimization (the lambda function passed as an argument gets inlined by the compiler so there is no overhead). I also made it possible to use it when the first selected position is nonzero (as long as the `selectedPositions` points to somewhere in `INCREMENTAL_SELECTED_POS`). To demonstrate its use I replaced a bunch of places where this was done manually in binary_function_executor.h, as I saw `isUnfiltered` was being used a lot there.


src/common/data_chunk/data_chunk_state.cpp
```diff
@@ -15,256 +21,18 @@ void DataChunkState::slice(offset_t offset) {
...
-#elif (VECTOR_CAPACITY_LOG_2 == 2)
-const sel_t SelectionVector::INCREMENTAL_SELECTED_POS[] = {0, 1, 2, 3};
-#elif (VECTOR_CAPACITY_LOG_2 == 1)
-const sel_t SelectionVector::INCREMENTAL_SELECTED_POS[] = {0, 1};
-#elif (VECTOR_CAPACITY_LOG_2 == 0)
-const sel_t SelectionVector::INCREMENTAL_SELECTED_POS[] = {0};
-#endif
-
+// NOLINTNEXTLINE(cert-err58-cpp): always evaluated at compile time, and even not it would not throw
+const std::array<sel_t, DEFAULT_VECTOR_CAPACITY> SelectionVector::INCREMENTAL_SELECTED_POS =
```
ray6080, line 31: Should we move `SelectionVector` related implementation to a new `sel_vector.cpp` file?

src/include/common/data_chunk/sel_vector.h
```diff
@@ -38,11 +43,25 @@ class SelectionVector {
         selectedSize = size;
     }
 
-    std::span<sel_t> getMultableBuffer() {
+    std::span<sel_t> getMutableBuffer() {
```
ray6080, line 46: This function should be marked as `const`.

src/include/common/data_chunk/sel_vector.h
```diff
@@ -21,7 +22,11 @@ class SelectionVector {
...
     SelectionVector() : SelectionVector{DEFAULT_VECTOR_CAPACITY} {}
 
-    bool isUnfiltered() const { return selectedPositions == (sel_t*)&INCREMENTAL_SELECTED_POS; }
+    bool isUnfiltered() const { return selectedPositions == INCREMENTAL_SELECTED_POS.data(); }
+    bool isContiguous() const {
+        return selectedPositions >= &*INCREMENTAL_SELECTED_POS.begin() &&
+               selectedPositions <= &*INCREMENTAL_SELECTED_POS.end();
+    }
 
     void setToUnfiltered() { selectedPositions = (sel_t*)&INCREMENTAL_SELECTED_POS; }
```
ray6080, line 31: Since you're changing this class, can u also refactor away the c-style castings?

src/include/common/data_chunk/sel_vector.h
```diff
@@ -21,7 +22,11 @@ class SelectionVector {
 
     SelectionVector() : SelectionVector{DEFAULT_VECTOR_CAPACITY} {}
 
-    bool isUnfiltered() const { return selectedPositions == (sel_t*)&INCREMENTAL_SELECTED_POS; }
+    bool isUnfiltered() const { return selectedPositions == INCREMENTAL_SELECTED_POS.data(); }
+    bool isContiguous() const {
```
ray6080, line 26: Should this function be private for now?

ray6080 (approved):

src/common/data_chunk/data_chunk_state.cpp
benjaminwinger: I guess we could, though it's only this part that will get moved.

src/include/common/data_chunk/sel_vector.h
benjaminwinger: I guess so. It would make the const_cast more explicit, but I also don't really like that we're const-casting here. Admittedly it will segfault if you attempt to use `operator[]` to modify it when selectedPositions is pointing at `INCREMENTAL_SELECTED_POS`.

src/include/common/data_chunk/sel_vector.h
benjaminwinger: I don't know. I guess it depends if there are places where isUnfiltered cannot be replaced with `forEach`, but where we want to take advantage of the additional cases isContiguous covers.

PR #4348, opened 2024-10-07
Bug fixes + optimizations on Fintech benchmark

royi-luo:
# Description

Fixed some bugs in ALP code that were causing crashes in the Fintech benchmark.

Also made some optimizations to reduce the number of `malloc` calls in some of the most frequently-called spots (where it wasn't necessary).

# Contributor agreement

### DIFF
--- a/src/storage/compression/float_compression.cpp
+++ b/src/storage/compression/float_compression.cpp
@@ -56,6 +56,10 @@ void EncodeExceptionView<T>::setValue(EncodeException<T> exception,
 }
 
 template<std::floating_point T>
+FloatCompression<T>::FloatCompression()
+    : constantEncodedFloatBitpacker(getBitpackingLogicalType<T>()), encodedFloatBitpacker() {}
+
+template<std::floating_point T>
 uint64_t FloatCompression<T>::compressNextPage(const uint8_t*&, uint64_t, uint8_t*, uint64_t,
     const struct CompressionMetadata&) const {
     KU_UNREACHABLE;
@@ -100,7 +104,7 @@ uint64_t FloatCompression<T>::compressNextPageWithExceptions(const uint8_t*& src
     const auto* castedIntegerEncodedBuffer =
         reinterpret_cast<const uint8_t*>(integerEncodedValues.data());
     const auto compressedIntegerSize =
-        getEncodedFloatBitpacker(metadata)->compressNextPage(castedIntegerEncodedBuffer,
+        getEncodedFloatBitpacker(metadata).compressNextPage(castedIntegerEncodedBuffer,
             numValuesToCompress, dstBuffer, dstBufferSize, metadata.getChild(BITPACKING_CHILD_IDX));
 
     // zero out unused parts of the page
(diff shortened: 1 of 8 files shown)


src/storage/store/column.cpp
```diff
@@ -35,12 +35,21 @@ struct ReadInternalIDValuesToVector {
...
-        auto buffer = std::make_unique<offset_t[]>(numValuesToRead);
-        compressedReader(frame, pageCursor, reinterpret_cast<uint8_t*>(buffer.get()), 0,
+        std::unique_ptr<offset_t[]> bufferMgr;
+        offset_t fixedSizeBuffer[DEFAULT_VECTOR_CAPACITY];
+        offset_t* bufferPtr = nullptr;
+        if (numValuesToRead <= DEFAULT_VECTOR_CAPACITY) {
+            bufferPtr = fixedSizeBuffer;
+        } else {
+            bufferMgr = std::make_unique<offset_t[]>(numValuesToRead);
+            bufferPtr = bufferMgr.get();
```
royi-luo, line 45: Should this case be handled or is it fine to replace it with `KU_ASSERT(numValuesToRead <= DEFAULT_VECTOR_CAPACITY)`?

src/storage/store/column_reader_writer.cpp
```diff
@@ -186,10 +186,18 @@ class DefaultColumnReadWriter : public ColumnReadWriter {
...
+                // we pack all the arguments to readFunc into a tuple and capture the tuple by
+                // this makes it more likely that we avoid any heap allocations when the lambda is
+                // copied into an std::function
+                const auto startReadOffset = numValuesScanned + startOffsetInResult;
+                auto readFuncArgs = std::tie(pageCursor, result, startReadOffset,
+                    numValuesToScanInPage, chunkMeta.compMeta);
+                readFromPage(transaction, pageCursor.pageIdx,
+                    [&readFunc, &readFuncArgs](uint8_t* frame) -> void {
+                        std::apply(readFunc, std::tuple_cat(std::make_tuple(frame), readFuncArgs));
+                    });
```
royi-luo, line 200:
```suggestion
                // to keep the size of the lambda small
                // we pack all the arguments to readFunc into a tuple and capture the tuple by
                // this makes it more likely for the std::function to be stack-allocated
                const auto startReadOffset = numValuesScanned + startOffsetInResult;
                auto readFuncArgs = std::tie(pageCursor, result, startReadOffset,
                    numValuesToScanInPage, chunkMeta.compMeta);
                readFromPage(transaction, pageCursor.pageIdx,
                    [&readFunc, &readFuncArgs](uint8_t* frame) -> void {
                        std::apply(readFunc, std::tuple_cat(std::make_tuple(frame), readFuncArgs));
                    });
```

src/storage/compression/float_compression.cpp
```diff
@@ -158,18 +164,18 @@ void FloatCompression<T>::setValuesFromUncompressed(const uint8_t* srcBuffer,
...
         numValues, metadata.getChild(BITPACKING_CHILD_IDX), nullMask);
 }
 
 template<std::floating_point T>
-std::unique_ptr<CompressionAlg> FloatCompression<T>::getEncodedFloatBitpacker(
+const CompressionAlg& FloatCompression<T>::getEncodedFloatBitpacker(
     const CompressionMetadata& metadata) const {
     if (metadata.getChild(BITPACKING_CHILD_IDX).isConstant()) {
-        return std::make_unique<ConstantCompression>(getBitpackingLogicalType<T>());
+        return constantEncodedFloatBitpacker;
```
benjaminwinger, line 176: Is this still necessary with #4166? Constant compression should be handled higher up now. Or is there some special meaning to the float compression having a constant child compression type?

src/storage/store/column.cpp
benjaminwinger: I think you can just assert it. I don't think value vectors can have a size other than `DEFAULT_VECTOR_CAPACITY` as the constructor always initializes the value buffer to that (I feel like it should have a better name).
There's a doc comment that seems to suggest that the capacity may also be 1, but I think that must be out of date.
>  The capacity of a ValueVector is either 1 (sequence) or DEFAULT_VECTOR_CAPACITY.

src/storage/store/column_reader_writer.cpp
benjaminwinger: I'm just looking into this now you mention this, but one option I ran across is that I think you should be able to get a very similar result by creating the lambda first as a variable (on the stack) and then passing it to readFromPage via `readFromPage(..., std::cref(lambdaFunc))`. That way you can still have it use automatic capture by reference without any heap allocations.

benjaminwinger (approved):

src/storage/store/column.cpp
royi-luo: I think the auxiliary buffer can resize it to something else but I don't think that's applicable here

src/storage/compression/float_compression.cpp
royi-luo: Let me check on this

src/storage/store/column_reader_writer.cpp
royi-luo: Cool, let me try this out

src/storage/compression/float_compression.cpp
royi-luo: I feel like the changes in #4166 only affect how the compression metadata is generated, I don't think we have a similar mechanism for handling constant compression higher up when picking a compression algorithm to use since we still have code like [this](https://github.com/kuzudb/kuzu/blob/master/src/storage/compression/compression.cpp#L739).

ray6080 (approved):

PR #4349, opened 2024-10-07
Add support for MAP data type for JAVA API

WWW0030:
# Description

Adding Support for MAP data type within JAVA API
(Currently draft)

Fixes # (issue)

# Contributor agreement

### DIFF
--- a/tools/java_api/src/jni/kuzu_java.cpp
+++ b/tools/java_api/src/jni/kuzu_java.cpp
@@ -1152,7 +1152,8 @@ JNIEXPORT jlong JNICALL Java_com_kuzudb_Native_kuzu_1value_1get_1map_1index(JNIE
     return -1;
 }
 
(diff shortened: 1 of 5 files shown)


tools/java_api/src/jni/kuzu_java.cpp
```diff
@@ -1119,6 +1132,26 @@ JNIEXPORT jlong JNICALL Java_com_kuzudb_Native_kuzu_1value_1get_1struct_1index(J
     }
 }
 
+JNIEXPORT jlong JNICALL Java_com_kuzudb_Native_kuzu_1value_1get_1map_1index(JNIEnv* env, jclass,
```
mewim, line 1135: I suggest to just implement a function to get value by index (similar to C API) and pass a `jlong index` instead of `jstring field_name`. Then, if iterate through the object to get by field name is required, the logic of iterating can be pushed to Java side.

mewim (changes requested):
See my comment

mewim (approved):

PR #4351, opened 2024-10-08
Add missing numOutputTuples in physical operators

ray6080:
# Description

Fill missing stats info (`numOutputTuples`) in some physical operators.

### DIFF
--- a/src/processor/operator/projection.cpp
+++ b/src/processor/operator/projection.cpp
@@ -36,7 +36,7 @@ bool Projection::getNextTuplesInternal(ExecutionContext* context) {
         resultSet->multiplicity *=
             resultSet->getNumTuplesWithoutMultiplicity(discardedDataChunksPos);
     }
-    metrics->numOutputTuple.increase(1);
+    metrics->numOutputTuple.increase(resultSet->multiplicity);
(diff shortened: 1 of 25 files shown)


andyfengHKU (approved):

src/processor/operator/projection.cpp
```diff
@@ -36,7 +36,10 @@ bool Projection::getNextTuplesInternal(ExecutionContext* context) {
         resultSet->multiplicity *=
             resultSet->getNumTuplesWithoutMultiplicity(discardedDataChunksPos);
     }
-    metrics->numOutputTuple.increase(1);
+    // The if statement is added to avoid the cost of calculating numTuples when metric is disabled.
+    if (metrics->numOutputTuple.enabled) [[unlikely]] {
+        metrics->numOutputTuple.increase(resultSet->getNumTuples(expressionsOutputDataChunksPos));
```
andyfengHKU, line 41: I think it's better to track number of factorized tuple rather than flattened tuple but up to u.

ray6080: I'm less sure about this. We have to iterate this later I guess.

PR #4352, opened 2024-10-08
Vectorize EdgeCompute

benjaminwinger:
The `edgeCompute` functions now take a range of values, and a mask indicating which values are actual inputs, and modify the mask to indicate the active values (i.e. the values where previously the function would have returned true). This improves performance by roughly 10% in the benchmarks I'd done.

### DIFF
--- a/src/function/gds/single_shortest_paths.cpp
+++ b/src/function/gds/single_shortest_paths.cpp
@@ -1,3 +1,4 @@
+#include "common/data_chunk/sel_vector.h"
 #include "function/gds/bfs_graph.h"
 #include "function/gds/gds_frontier.h"
 #include "function/gds/gds_function_collection.h"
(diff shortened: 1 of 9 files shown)


src/function/gds/single_shortest_paths.cpp
```diff
@@ -81,18 +89,23 @@ class SingleSPPathsEdgeCompute : public EdgeCompute {
...
+            auto shouldUpdate = frontierPair->pathLengths->getMaskValueFromNextFrontierFixedMask(
+                                    nbrNodeIDs[i].offset) == PathLengths::UNVISITED;
+            if (shouldUpdate) {
+                if (!parentListBlock->hasSpace()) {
+                    parentListBlock = bfsGraph->addNewBlock();
+                }
+                bfsGraph->tryAddSingleParent(frontierPair->curIter.load(std::memory_order_relaxed),
+                    parentListBlock, nbrNodeIDs[i] /* child */, boundNodeID /* parent */,
+                    edgeIDs[i], isFwd);
+                mask.getMutableBuffer()[activeCount++] = i;
```
semihsalihoglu-uw, line 105: I think we can directly get the edgeCompute() functions to update the next frontiers, since at this point we lost the the benefit of just returning a true/false to delegate the work of setting the frontier to the GDSTask. Can you test how much difference that makes?

semihsalihoglu-uw (approved):

src/function/gds/single_shortest_paths.cpp
benjaminwinger: Copying what I'd said on discord here for the record.

I benchmarked that change on the ldbc graph500-30 dataset using the same query I had been using on the smaller datasets:
- master: average of 43.4 seconds
- the vectorized version: average of 38.2 seconds
- the vectorized version which updates the frontier in edgeCompute: average of 37.4 seconds

The difference is relatively small, and I had to expose stuff from the internals GDSFrontier to the edgeCompute functions to get it to run quickly, so I've left it as-is to keep things simple.

PR #4355, opened 2024-10-08
Added cli autocompletion for table columns

MSebanc:
# Description

Added column auto completion based on variables specified in query
image

# Contributor agreement

### DIFF
--- a/tools/shell/embedded_shell.cpp
+++ b/tools/shell/embedded_shell.cpp
@@ -163,7 +163,8 @@ bool isInsideCommentOrQuote(const std::string& buffer) {
            insideSingleQuote;
 }
 
-void findTableVariableNames(const std::string buf, std::regex tableRegex, std::vector<std::string> &tempTableNames, std::vector<std::string> &foundTableNames) {
+void findTableVariableNames(const std::string buf, std::regex tableRegex,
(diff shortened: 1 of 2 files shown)


tools/shell/embedded_shell.cpp
```diff
@@ -193,6 +206,33 @@ void completion(const char* buffer, linenoiseCompletions* lc) {
         return;
     }
 
+    std::vector<std::string> tempNodeNames;
+    std::vector<std::string> foundNodeTableNames;
+    for (auto& node : nodeTableNames) {
```
andyfengHKU, line 211: Similarly try refactor l211-221 and l224-234.

tools/shell/embedded_shell.cpp
```diff
@@ -203,6 +243,40 @@ void completion(const char* buffer, linenoiseCompletions* lc) {
     if (buf.empty()) {
         return;
     }
+
+    for (size_t i = 0; i < tempNodeNames.size(); ++i) {
```
andyfengHKU, line 247: Ditto try refactor into one function.

andyfengHKU (approved):
LGTM. I have some implementation detail comments. I feel some common node/rel table logic can be refactored into a common function. But u can make the final call.

PR #4357, opened 2024-10-10
Optimize rel label prune under both direction

ted-wq-x:
# Description

As the title suggests.

# Contributor agreement


andyfengHKU (approved):
LGTM. Thanks a lot!

PR #4358, opened 2024-10-10
Optimize short query and avoid scheduling overhead

ted-wq-x:
# Description

Optimizing query like ‘match (a:person {id:24189255912498})-[f]-(b) return count(1);’ ,  Query time dropped from 50ms to 15ms

# Contributor agreement


andyfengHKU (commented):
I think is fine (or better) in general. But @ray6080 should take another look.

The general assumption is that primary key scan always lead to 1 output tuple so executing the entire pipeline with in single thread mode makes more sense.

ray6080 (approved):
LGTM. Thanks!

PR #4359, opened 2024-10-10
Adding DECIMAL Support for C_API

WWW0030:
# Description

Adding decimal support for C_API

Fixes # (issue)

# Contributor agreement

### DIFF
--- a/src/c_api/value.cpp
+++ b/src/c_api/value.cpp
@@ -616,6 +616,37 @@ kuzu_state kuzu_value_get_timestamp_tz(kuzu_value* value, kuzu_timestamp_tz_t* o
     return KuzuSuccess;
 }
 
+kuzu_state kuzu_value_get_decimal(kuzu_value* value, std::string* out_result) {
+    auto logical_type_id = static_cast<Value*>(value->_value)->getDataType().getLogicalTypeID();
+    auto physical_type_id = static_cast<Value*>(value->_value)->getDataType().getPhysicalType();
+    if (logical_type_id != LogicalTypeID::DECIMAL) {
+        return KuzuError;
+    }
+    switch (physical_type_id) {
+        case PhysicalTypeID::INT16:
+            *out_result = std::to_string(static_cast<Value*>(value->_value)->getValue<int16_t>());
+            return KuzuSuccess;
+
+        case PhysicalTypeID::INT32:
+            *out_result = std::to_string(static_cast<Value*>(value->_value)->getValue<int32_t>());
+            return KuzuSuccess;
+
+        case PhysicalTypeID::INT64:
+            *out_result = std::to_string(static_cast<Value*>(value->_value)->getValue<int64_t>());
+            return KuzuSuccess;
+
+        case PhysicalTypeID::INT128: {
+            auto int_value_representation = static_cast<Value*>(value->_value)->getValue<int128_t>();
+            *out_result = std::to_string(int_value_representation.low) + std::to_string(int_value_representation.high);
+            return KuzuSuccess;
+        }
+
+        default:
+            return KuzuError;
+    }
+    return KuzuSuccess;
+}
+
 kuzu_state kuzu_value_get_interval(kuzu_value* value, kuzu_interval_t* out_result) {
     auto logical_type_id = static_cast<Value*>(value->_value)->getDataType().getLogicalTypeID();
     if (logical_type_id != LogicalTypeID::INTERVAL) {
--- a/src/c_api/value.cpp
+++ b/src/c_api/value.cpp
@@ -9,6 +9,7 @@
 #include "common/types/value/recursive_rel.h"
 #include "common/types/value/rel.h"
 #include "function/cast/functions/cast_from_string_functions.h"
+#include <cstring>
 
 using namespace kuzu::common;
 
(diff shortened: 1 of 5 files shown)


src/include/c_api/kuzu.h
```diff
@@ -2,6 +2,9 @@
 #include <stdbool.h>
 #include <stdint.h>
 #include <time.h>
+
+#include <string>
```
mewim, line 6: `kuzu.h` is a C header. not a C++ header. It is a binding to C++ with [opaque pointer pattern](https://en.wikipedia.org/wiki/Opaque_pointer). You should NOT import any C++ header into this file at all. `<string>` is a C++ header, so it is not allowed in this file.

src/include/common/types/types.h
```diff
@@ -224,6 +224,7 @@ enum class PhysicalTypeID : uint8_t {
     INTERNAL_ID = 14,
     ALP_EXCEPTION_FLOAT = 15,
     ALP_EXCEPTION_DOUBLE = 16,
+    DECIMAL = 17,
```
mewim, line 227: As always, to implement an API binding, not a single line from C++ side should be changed!

src/include/c_api/kuzu.h
```diff
@@ -1225,6 +1228,14 @@ KUZU_C_API kuzu_state kuzu_value_get_timestamp_tz(kuzu_value* value,
...
  */
 KUZU_C_API kuzu_state kuzu_value_get_interval(kuzu_value* value, kuzu_interval_t* out_result);
+/**
+ * @brief Returns the decimal value of the given value as a string. The value must be of type
+ * DECIMAL.
+ * @param value The value to return.
+ * @param[out] out_result The output parameter that will hold the decimal value.
+ * @return The state indicating the success or failure of the operation.
+ */
+KUZU_C_API kuzu_state kuzu_value_get_decimal(kuzu_value* value, std::string* out_result);
```
mewim, line 1238: Instead using a C++ string, this should be a C string (`char**`)

mewim: The API signature should be `kuzu_value_get_decimal_as_string`

src/c_api/value.cpp
```diff
@@ -610,6 +610,38 @@ kuzu_state kuzu_value_get_timestamp_tz(kuzu_value* value, kuzu_timestamp_tz_t* o
...
+    case PhysicalTypeID::INT128: {
+        auto int_value_representation = static_cast<Value*>(value->_value)->getValue<int128_t>();
+        *out_result = std::to_string(int_value_representation.low) +
+                      std::to_string(int_value_representation.high);
+        return KuzuSuccess;
+    }
+
+    default:
+        return KuzuError;
+    }
```
mewim, line 641: You should NOT reinvent the logic of converting decimal to string. There is already a function `toString` you can use.

See: https://github.com/kuzudb/kuzu/blob/f15d8783345ee7b18cd23c09e07334b2540912c8/tools/java_api/src/jni/kuzu_java.cpp#L875-L879

and 
https://github.com/kuzudb/kuzu/blob/f15d8783345ee7b18cd23c09e07334b2540912c8/tools/python_api/src_cpp/py_query_result.cpp#L210-L213

test/c_api/value_test.cpp
```diff
@@ -642,6 +642,93 @@ TEST_F(CApiValueTest, getMapValue) {
     kuzu_flat_tuple_destroy(&flatTuple);
 }
 
+TEST_F(CApiValueTest, getDecimal) {
```
mewim, line 645: After changing to C string as required, please re-visit this test to make sure adding all necessary calls to destructors.

mewim (changes requested):
See my comments

src/c_api/value.cpp
```diff
@@ -610,6 +610,36 @@ kuzu_state kuzu_value_get_timestamp_tz(kuzu_value* value, kuzu_timestamp_tz_t* o
...
+    case PhysicalTypeID::INT128: {
+        *out_result = kuzu_value_to_string(value);
+        return KuzuSuccess;
+    }
+
+    default:
+        return KuzuError;
+    }
+    return KuzuSuccess;
+}
```
mewim, line 641: You did not change this part as I suggested. 
> You should NOT reinvent the logic of converting decimal to string. There is already a function `toString` you can use.

test/c_api/value_test.cpp
mewim: You did not add destructors as I requested. 
> After changing to C string as required, please re-visit this test to make sure adding all necessary calls to destructors.

src/include/c_api/kuzu.h
```diff
@@ -2,6 +2,7 @@
 #include <stdbool.h>
 #include <stdint.h>
 #include <time.h>
+
```
mewim, line 5: Remove this new line

mewim (changes requested):
You need to make sure all my comments are addressed before requesting a re-review.

test/c_api/value_test.cpp
mewim: Also, the test's name should be changed to `getDecimalAsString` to be consistent with the function you are testing

WWW0030: sorry, missed a destructor while double-checking

src/c_api/value.cpp
WWW0030: Is there a reason we want to use the `toString` function instead of calling another c_api? the `toString` method will be returning a `const char*` variable which I can't seem to find an easy way of translating into a `char*`, and `kuzu_value_to_string` returns exactly the datatype we want.

I reimplemented this with `toString`, currently just testing locally to see if everything still works.

mewim: You are working on an API binding, which should not care about the underlying physical type. If a new physical type is added or the underlying core implementation is changed, as long as the same interface function is provided, no changes should be required at the API-level. Your implementation is depending on the underlying physical type, which:

1. May create additional maintenance burden in the future.
2. You are duplicating some logic. I am sure that the `toString` function is already able to handle the type switching you have done.

If you have question about a change I request, you should ask me first. Not doing it and requesting a re-review is unproductive as I have to read the same code twice.

src/c_api/value.cpp
```diff
@@ -610,6 +611,23 @@ kuzu_state kuzu_value_get_timestamp_tz(kuzu_value* value, kuzu_timestamp_tz_t* o
...
+kuzu_state kuzu_value_get_decimal_as_string(kuzu_value* value, char** out_result) {
+    auto v = static_cast<Value*>(value->_value);
+    auto logical_type_id = v->getDataType().getLogicalTypeID();
+    auto physical_type_id = v->getDataType().getPhysicalType();
+    if (logical_type_id != LogicalTypeID::DECIMAL) {
+        return KuzuError;
+    }
+    if (physical_type_id != PhysicalTypeID::INT16 && physical_type_id != PhysicalTypeID::INT32 && physical_type_id != PhysicalTypeID::INT64 && physical_type_id != PhysicalTypeID::INT128) {
+        return KuzuError;
+    }
```
mewim: DO NOT check the physical id type. As I said, this has nothing to do with the API level.

src/c_api/value.cpp
```diff
@@ -610,6 +612,24 @@ kuzu_state kuzu_value_get_timestamp_tz(kuzu_value* value, kuzu_timestamp_tz_t* o
...
+    if (physical_type_id != PhysicalTypeID::INT16 && physical_type_id != PhysicalTypeID::INT32 &&
+        physical_type_id != PhysicalTypeID::INT64 && physical_type_id != PhysicalTypeID::INT128) {
+        return KuzuError;
+    }
+
+    std::string v_string = v->toString();
+    *out_result = new char[v_string.length() + 1];
+    strcpy(*out_result, v_string.c_str());
```
mewim, line 629: There is a `convertToOwnedCString` function you can use. You should refer to other functions in this file regarding how to use it.

src/c_api/value.cpp
```diff
@@ -1,5 +1,7 @@
 #include "common/types/value/value.h"
 
+#include <cstring>
```
mewim, line 3: Remove this

src/c_api/value.cpp
```diff
@@ -610,6 +610,17 @@ kuzu_state kuzu_value_get_timestamp_tz(kuzu_value* value, kuzu_timestamp_tz_t* o
     return KuzuSuccess;
 }
 
+kuzu_state kuzu_value_get_decimal_as_string(kuzu_value* value, char** out_result) {
+    auto v = static_cast<Value*>(value->_value);
```
mewim, line 614: Rename this to decimal_val.

mewim (changes requested):
Minor change required

mewim (approved):

mewim:
Can be merged after CI passes

PR #4361, opened 2024-10-10
Add gds edge predicate

andyfengHKU:
# Description

This PR allows executing recursive pattern with edge predicate in GDS, e.g `[e* (r, n |  WHERE r.score = 1)]`. 

Changes includes,

- Add rel property expressions and rel predicate to `GraphEntry`.
- Instead of compile only internal ID vectors, we also compile property vectors and save them into a resultSet. This allows us to create expression evaluator for rel predicate.
- Add predicate evaluator to `OnDiskGraphScanStates`
- Apply predicate after table scan.

@benjaminwinger
TODO: make iterator respect non-consecutive selected positions. 
 

Fixes # (issue)

# Contributor agreement


src/graph/on_disk_graph.cpp
```diff
@@ -34,38 +45,52 @@ static std::unique_ptr<RelTableScanState> getRelScanState(MemoryManager& mm, con
...
-
-OnDiskGraphScanStates::OnDiskGraphScanStates(ClientContext* context, std::span<RelTable*> tables)
-    : iteratorIndex{0}, direction{RelDataDirection::FWD} {
-    scanStates.reserve(tables.size());
-    srcNodeIDVectorState = DataChunkState::getSingleValueDataChunkState();
-    dstNodeIDVectorState = std::make_shared<DataChunkState>();
+OnDiskGraphScanStates::OnDiskGraphScanStates(ClientContext* context, std::span<RelTable*> tables,
+    const GraphEntry& graphEntry)
+    : iteratorIndex{0}, direction{RelDataDirection::INVALID} {
+    auto schema = Schema();
```
semihsalihoglu-uw, line 60: I don't have a very concrete suggestion here but I'm a bit worried that some of this logic of how we map an expression to scan operators must be also implemented in other parts of the mapper (e.g., when we compile scan + filter operators) and is being duplicated here. I wonder if there is some way we can refactor this logic. For example some function that takes in predicate expression and figures out the properties that needs to be scanned from the expression and constructs a predicateEvaluator and update a ResultSet. And maybe it's used here and in the mapper?

semihsalihoglu-uw (approved):

src/include/graph/on_disk_graph.h
```diff
@@ -42,6 +43,7 @@ struct OnDiskGraphScanState {
         std::span<const common::nodeID_t> getEdges() const {
             // When scanning a single node, all returned results should be consecutive
             auto firstElement = dstSelVector().getSelectedPositions()[0];
+            // TODO(bmwinder): make this function adapt to selected positions.
```
benjaminwinger, line 46: This is already done as of #4352

src/graph/on_disk_graph.cpp
```diff
@@ -225,26 +240,18 @@ Graph::Iterator OnDiskGraph::scanBwd(nodeID_t nodeID, GraphScanState& state) {
...
+    // I don't think it's necessary because scan doesn't read SelVector.
+    if (tableScanState->source == TableScanSource::NONE ||
+        !relTable->scan(context->getTx(), *tableScanState)) {
+        return false;
+    }
+    if (predicate != nullptr) {
+        predicate->select(tableScanState->outState->getSelVectorUnsafe());
+        tableScanState->outState->getSelVectorUnsafe().setToFiltered();
+    }
+    return dstSelVector().getSelSize() > 0;
```
benjaminwinger, line 254: This function should only return false once the scan is finished, as it indicates to `Graph::Iterator` that the iterator is exhausted. You should probably put this in a loop and keep scanning until the result after the predicate is nonempty. Always returning true on this line should also work, but would be a bit slower as it will just process the empty data and call next again.

dataset/gds-rec-joins-small/vKnows.csv
```diff
@@ -1,6 +1,6 @@
-0	1
-1	2
-2	3
-3	4
-2	4
-2	4
+0	1	1
```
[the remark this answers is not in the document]
andyfengHKU: Done

src/graph/on_disk_graph.cpp
andyfengHKU: I only make a minimal change. I understand your comment. Though the problem needs a relatively bigger change which is beyond the scope of this PR. I'll make a following up PR soon to resolve this.

src/graph/on_disk_graph.cpp
andyfengHKU: Good catch. This is done.

PR #4362, opened 2024-10-11
Optimize column&ChunkedCSRHeader memory initialize

ted-wq-x:
# Description

Optimize short query QPS.

1. ChunkedCSRHeader no need initialize offset/length field memory to 0.
2. Inside Column::scan method , columnChunk->resize() no need reserve data.

# Contributor agreement

### DIFF
--- a/src/storage/store/column_chunk_data.cpp
+++ b/src/storage/store/column_chunk_data.cpp
@@ -390,17 +390,31 @@ void ColumnChunkData::setToInMemory() {
     }
 }
 
-void ColumnChunkData::resize(uint64_t newCapacity, bool reserveDataAndInitializeToZero) {
+void ColumnChunkData::resize(uint64_t newCapacity) {
     if (newCapacity > capacity) {
         capacity = newCapacity;
     }
     const auto numBytesAfterResize = getBufferSize(newCapacity);
     if (numBytesAfterResize > getBufferSize()) {
-        auto resizedBuffer = buffer->getMemoryManager()->mallocBuffer(
-            reserveDataAndInitializeToZero, numBytesAfterResize);
-        if (reserveDataAndInitializeToZero) {
-            memcpy(resizedBuffer->getBuffer().data(), buffer->getBuffer().data(), getBufferSize());
-        }
+        auto resizedBuffer = buffer->getMemoryManager()->mallocBuffer(false, numBytesAfterResize);
+        auto bufferSize = getBufferSize();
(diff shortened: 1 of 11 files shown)


src/storage/store/column_chunk_data.cpp
```diff
@@ -388,14 +390,17 @@ void ColumnChunkData::setToInMemory() {
...
+void ColumnChunkData::resize(uint64_t newCapacity, bool reserveDataAndInitializeToZero) {
     if (newCapacity > capacity) {
         capacity = newCapacity;
     }
     const auto numBytesAfterResize = getBufferSize(newCapacity);
     if (numBytesAfterResize > getBufferSize()) {
-        auto resizedBuffer = buffer->getMemoryManager()->mallocBuffer(true, numBytesAfterResize);
-        memcpy(resizedBuffer->getBuffer().data(), buffer->getBuffer().data(), getBufferSize());
+        auto resizedBuffer = buffer->getMemoryManager()->mallocBuffer(
+            reserveDataAndInitializeToZero, numBytesAfterResize);
```
benjaminwinger, line 400: If we're going to overwrite the data afterwards, it might make more sense to not initialize the buffer to zero, overwrite that data and then zero just the part exceeding the original buffer's size. Otherwise we're initializing the part that already exists twice.

src/storage/store/column_chunk_data.cpp
```diff
@@ -388,14 +390,17 @@ void ColumnChunkData::setToInMemory() {
     }
 }
 
-void ColumnChunkData::resize(uint64_t newCapacity) {
+void ColumnChunkData::resize(uint64_t newCapacity, bool reserveDataAndInitializeToZero) {
```
benjaminwinger, line 393: I don't think this name very clearly indicates what's occurring here.
Maybe this should be two different functions (particularly since, aside from adding the argument name in a comment, boolean arguments are not generally very easy to read). 

While admittedly we haven't been doing this, maybe we should follow std containers like `std::vector` and have a `reserve` function which resizes the buffer without changing the numElements and copies the original data into the buffer (i.e. `reserveDataAndInitializeToZero=true`), and then a separate function that just reallocates a larger buffer without preserving the original data (`reserveDataAndInitializeToZero=false`).

ted-wq-x: Good advice!

ray6080 (approved):
LGTM. Thanks!

benjaminwinger (approved):

PR #4365, opened 2024-10-12
fix path semantic no work

excaliburwyj:
# Description

path semantic do not work in this case

call recursive_pattern_semantic='TRAIL';
match p=(a )-[f*2]-(b) where a.id=1 return p;

In the graph, there is only a relationship from 1->2 and not 2->1. This means that when the 'TRAIL' option is set, the end node b should not include the start node, which id=1. However, the result set includes: {_NODES:[1,2,1],_RELS:[1->2,1->2]}.

The cause of this issue is that in the trailSemanticCheck Function, some edgeIDs have been processed with markFlip. When comparing these edges with those that have not undergone markFlip, they cannot be correctly filtered.

Additionally, the path semantics do not work in the case:
match p=(a)-[f*2]-(b) where a.id=1 return b;

This is because the optimizer has set TRACK_NONE, and this option  only support WALK. My approach is to set TRACK_NONE only when WALK is specified.

# Contributor agreement

### DIFF
--- a/src/include/optimizer/projection_push_down_optimizer.h
+++ b/src/include/optimizer/projection_push_down_optimizer.h
@@ -1,5 +1,6 @@
 #pragma once
 
+#include "common/enums/path_semantic.h"
 #include "logical_operator_visitor.h"
 #include "planner/operator/logical_plan.h"
 
@@ -24,6 +25,7 @@ namespace optimizer {
 class ProjectionPushDownOptimizer : public LogicalOperatorVisitor {
 public:
     void rewrite(planner::LogicalPlan* plan);
+    inline void setPathSemantic(common::PathSemantic _semantic) { this->semantic = _semantic; }
 
 private:
     void visitOperator(planner::LogicalOperator* op);
@@ -58,6 +60,7 @@ private:
     binder::expression_set propertiesInUse;
     binder::expression_set variablesInUse;
     binder::expression_set nodeOrRelInUse;
+    common::PathSemantic semantic;
 };
 
 } // namespace optimizer
--- a/src/include/optimizer/projection_push_down_optimizer.h
+++ b/src/include/optimizer/projection_push_down_optimizer.h
@@ -25,7 +25,7 @@ namespace optimizer {
 class ProjectionPushDownOptimizer : public LogicalOperatorVisitor {
 public:
     void rewrite(planner::LogicalPlan* plan);
-    inline void setPathSemantic(common::PathSemantic _semantic) { this->semantic = _semantic; }
+    explicit ProjectionPushDownOptimizer(common::PathSemantic semantic) : semantic(semantic) {};
 
 private:
     void visitOperator(planner::LogicalOperator* op);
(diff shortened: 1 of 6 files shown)


src/include/optimizer/projection_push_down_optimizer.h
```diff
@@ -24,6 +25,7 @@ namespace optimizer {
 class ProjectionPushDownOptimizer : public LogicalOperatorVisitor {
 public:
     void rewrite(planner::LogicalPlan* plan);
+    inline void setPathSemantic(common::PathSemantic _semantic) { this->semantic = _semantic; }
```
andyfengHKU, line 28: I would suggest making this part of the constructor because we always expect a semantic. In this case, we will need to assign a default `PathSemantic`.

So better make it
```
ProjectionPushDownOptimizer(common::PathSemantic)
```

src/include/processor/operator/recursive_extend/frontier.h
```diff
@@ -48,6 +48,9 @@ struct RelIDMasker {
     static void markFlip(common::internalID_t& relID) { relID.offset |= FLIP_SRC_DST_MASK; }
     static bool needFlip(common::internalID_t& relID) { return relID.offset & FLIP_SRC_DST_MASK; }
     static void clearMark(common::internalID_t& relID) { relID.offset &= CLEAR_FLIP_SRC_DST_MASK; }
+    static common::internalID_t getWithoutMark(common::internalID_t& relID) {
```
andyfengHKU, line 51: the input should be `const common::internalID_t&`

src/processor/operator/recursive_extend/frontier_scanner.cpp
```diff
@@ -51,10 +51,15 @@ bool PathScanner::trailSemanticCheck(const std::vector<nodeID_t>&,
     const std::vector<relID_t>& edgeIDs) {
     common::rel_id_set_t set;
     for (auto i = 0u; i < edgeIDs.size() - 1; ++i) {
-        if (set.contains(edgeIDs[i])) {
+        internalID_t edgeID = edgeIDs[i];
+        if (RelIDMasker::needFlip(const_cast<internalID_t&>(edgeIDs[i]))) {
```
andyfengHKU, line 55: Good catch. This is indeed a bug.

I would suggest some minor coding convention changes.

```
if (RelIDMasker::needFlip(const_cast<internalID_t&>(edgeIDs[i])))
```
The `const_cast` doesn't seem to be needed. You can probably just do
```
if (RelIDMasker::needFlip(edgeID))
```

src/processor/operator/recursive_extend/frontier_scanner.cpp
```diff
@@ -51,10 +51,15 @@ bool PathScanner::trailSemanticCheck(const std::vector<nodeID_t>&,
...
+            edgeID = RelIDMasker::getWithoutMark(const_cast<internalID_t&>(edgeIDs[i]));
+        } else {
+        }
```
andyfengHKU, line 58: remove empty else statement so that the coding style is consistent in the codebase.

andyfengHKU (approved):
LGTM. This is indeed a bug. Thanks for the fix!

PR #4367, opened 2024-10-14
Howe/java api query result

WWW0030:
# Description

Adding `hasNextQueryResult` and `getNextQueryResult` to JAVA_API

Fixes # (issue)

# Contributor agreement

### DIFF
--- a/tools/java_api/src/main/java/com/kuzudb/QueryResult.java
+++ b/tools/java_api/src/main/java/com/kuzudb/QueryResult.java
@@ -128,6 +128,26 @@ public class QueryResult {
     }
 
     /**
+     * Return if the query result has next query result or not.
+     * @return Whether there are more query results to read.
+     * @throws ObjectRefDestroyedException If the query result has been destroyed.
+     */
+    public boolean hasNextQueryResult() throws ObjectRefDestroyedException {
+        checkNotDestroyed();
+        return Native.kuzu_query_result_has_next_query_result(this);
+    }
+
+    /**
+     * Get the next query result.
+     * @return The next query result.
+     * @throws ObjectRefDestroyedException If the query result has been destroyed.
+     */
+    public QueryResult getNextQueryResult() throws ObjectRefDestroyedException {
+        checkNotDestroyed();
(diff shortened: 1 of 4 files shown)


mewim (changes requested):
This implementation is not complete. Note that the `QueryResult` returned by `getNextQueryResult` can not be safely destructed. Calling `destroy` on it can lead to crash. You need to handle such case, because `destroy` function can also be called automatically by the finalizer. If this is not handled, Java garbage collection can cause crash. You should refer to 

https://github.com/kuzudb/kuzu/blob/74be70279222a03742e36114e0c528635d5f6aff/src/c_api/query_result.cpp#L10-L19

and 

https://github.com/kuzudb/kuzu/blob/74be70279222a03742e36114e0c528635d5f6aff/tools/java_api/src/main/java/com/kuzudb/Value.java#L71-L77

Essentially, a boolean flag needs to be added and checked to properly handle this.

tools/java_api/src/test/java/com/kuzudb/QueryResultTest.java
```diff
@@ -140,15 +140,18 @@ void getMultipleQueryResult() throws ObjectRefDestroyedException {
         String str = result.toString();
         assertEquals(str, "1\n1\n");
 
+        QueryResult next_result;
```
mewim, line 143: In Java `cammelCase` should be used

tools/java_api/src/main/java/com/kuzudb/QueryResult.java
```diff
@@ -144,7 +151,9 @@ public boolean hasNextQueryResult() throws ObjectRefDestroyedException {
      */
     public QueryResult getNextQueryResult() throws ObjectRefDestroyedException {
         checkNotDestroyed();
-        return Native.kuzu_query_result_get_next_query_result(this);
+        QueryResult ret = Native.kuzu_query_result_get_next_query_result(this);
+        ret.isOwnedByCPP = true;
```
mewim, line 155: Similar to `Value`, I think we should set this flag on the C++ side, not here.

mewim (approved):
Can be merged if CI passes

PR #4370, opened 2024-10-16
fix the execution plan containing NodeLabelFiler always results in empty

ted-wq-x:
# Description

As the title suggests.

# Contributor agreement

### DIFF
--- a/src/include/optimizer/logical_operator_visitor.h
+++ b/src/include/optimizer/logical_operator_visitor.h
@@ -75,6 +75,12 @@ protected:
         return op;
     }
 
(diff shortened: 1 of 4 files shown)


andyfengHKU (approved):
LGTM. But was this causing trouble? I thought `Extend` will always put the filtered nodeID into scope. If so, let's also add a test case.

ted-wq-x:
> LGTM. But was this causing trouble? I thought `Extend` will always put the filtered nodeID into scope. If so, let's also add a test case.

I found this problem when optimizing the code. Normally there will be no problem.

PR #4373, opened 2024-10-16
Enable zone map for scan node/rel tables

royi-luo:
# Description
Skip scanning a chunked node group if for any column, the metadata (min/max value) indicates that no value in the chunk can satisfy one of the query predicates. This should have zone maps working for both node and rel tables.

Also maintain statistics (min/max value) for in-memory versions of column chunks to support the above (before we only update the statistics during checkpoint).

# Contributor agreement

### DIFF
--- a/src/include/main/client_config.h
+++ b/src/include/main/client_config.h
@@ -13,7 +13,7 @@ struct ClientConfigDefault {
     static constexpr uint64_t TIMEOUT_IN_MS = 0;
     static constexpr uint32_t VAR_LENGTH_MAX_DEPTH = 30;
     static constexpr bool ENABLE_SEMI_MASK = true;
-    static constexpr bool ENABLE_ZONE_MAP = false;
+    static constexpr bool ENABLE_ZONE_MAP = true;
     static constexpr bool ENABLE_GDS = false;
     static constexpr bool ENABLE_PROGRESS_BAR = false;
     static constexpr uint64_t SHOW_PROGRESS_AFTER = 1000;
--- a/src/include/storage/predicate/column_predicate.h
+++ b/src/include/storage/predicate/column_predicate.h
@@ -7,7 +7,7 @@
 namespace kuzu {
 namespace storage {
 
-struct CompressionMetadata;
+struct InMemoryColumnChunkStats;
 
 class ColumnPredicate;
 class KUZU_API ColumnPredicateSet {
@@ -20,7 +20,7 @@ public:
     }
     bool isEmpty() const { return predicates.empty(); }
 
-    common::ZoneMapCheckResult checkZoneMap(const CompressionMetadata& metadata);
+    common::ZoneMapCheckResult checkZoneMap(const InMemoryColumnChunkStats& metadata) const;
 
     std::string toString() const;
 
@@ -38,7 +38,8 @@ public:
 
     virtual ~ColumnPredicate() = default;
 
-    virtual common::ZoneMapCheckResult checkZoneMap(const CompressionMetadata& metadata) const = 0;
+    virtual common::ZoneMapCheckResult checkZoneMap(
+        const InMemoryColumnChunkStats& metadata) const = 0;
 
     virtual std::string toString() = 0;
 
--- a/src/include/storage/predicate/column_predicate.h
+++ b/src/include/storage/predicate/column_predicate.h
@@ -7,7 +7,7 @@
 namespace kuzu {
 namespace storage {
 
-struct InMemoryColumnChunkStats;
+struct ColumnChunkStats;
 
 class ColumnPredicate;
 class KUZU_API ColumnPredicateSet {
@@ -20,7 +20,7 @@ public:
     }
     bool isEmpty() const { return predicates.empty(); }
 
-    common::ZoneMapCheckResult checkZoneMap(const InMemoryColumnChunkStats& metadata) const;
+    common::ZoneMapCheckResult checkZoneMap(const ColumnChunkStats& metadata) const;
 
     std::string toString() const;
 
@@ -38,8 +38,7 @@ public:
 
     virtual ~ColumnPredicate() = default;
 
-    virtual common::ZoneMapCheckResult checkZoneMap(
-        const InMemoryColumnChunkStats& metadata) const = 0;
+    virtual common::ZoneMapCheckResult checkZoneMap(const ColumnChunkStats& metadata) const = 0;
 
     virtual std::string toString() = 0;
 
--- a/src/include/storage/store/column_chunk.h
+++ b/src/include/storage/store/column_chunk.h
@@ -35,16 +35,6 @@ struct ColumnCheckpointState {
     }
 };
 
-struct ColumnChunkStats {
-    std::optional<StorageValue> max;
-    std::optional<StorageValue> min;
-
-    void update(std::optional<StorageValue> min, std::optional<StorageValue> max,
-        common::PhysicalTypeID dataType);
-    void update(StorageValue val, common::PhysicalTypeID dataType);
-    void reset();
-};
-
 class ColumnChunk {
 public:
     // TODO(bmwinger): the dataType reference is copied when passing it to the ColumnChunkData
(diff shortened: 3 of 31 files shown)


src/include/main/client_config.h
```diff
@@ -13,7 +13,7 @@ struct ClientConfigDefault {
     static constexpr uint64_t TIMEOUT_IN_MS = 0;
     static constexpr uint32_t VAR_LENGTH_MAX_DEPTH = 30;
     static constexpr bool ENABLE_SEMI_MASK = true;
-    static constexpr bool ENABLE_ZONE_MAP = false;
+    static constexpr bool ENABLE_ZONE_MAP = true;
```
royi-luo, line 16: Should probably keep this disabled by default until this also works for rel tables

src/include/storage/store/column_chunk.h
```diff
@@ -113,6 +136,7 @@ class ColumnChunk {
     std::unique_ptr<ColumnChunkData> data;
     // Update versions.
     std::unique_ptr<UpdateInfo> updateInfo;
+    InMemoryColumnChunkStats inMemoryUpdatedStats;
```
royi-luo, line 139: maybe rename

src/include/storage/store/column_chunk.h
```diff
@@ -89,18 +99,31 @@ class ColumnChunk {
...
         if (updateInfo) {
             updateInfo.reset();
         }
+        inMemoryUpdatedStats.reset();
     }
 
     void loadFromDisk() { data->loadFromDisk(); }
     uint64_t spillToDisk() { return data->spillToDisk(); }
 
+    InMemoryColumnChunkStats getMergedUpdateStats(const CompressionMetadata& o) const;
```
royi-luo, line 121: This could probably have a better name

src/storage/store/chunked_node_group.cpp
```diff
@@ -315,6 +345,17 @@ void ChunkedNodeGroup::addColumn(Transaction* transaction,
...
+    if (numRows > 0) {
+        TypeUtils::visit(
+            chunkData.getDataType().getPhysicalType(),
+            [this, &chunkData]<StorageValueType T>(T) {
+                auto defaultVal = chunkData.getValue<T>(0);
+                if (!chunkData.hasNullData() || !chunkData.isNull(0)) {
+                    chunks.back()->updateStats(StorageValue{defaultVal}, StorageValue{defaultVal});
+                }
+            },
+            [](auto) {});
```
royi-luo, line 357: move to static function

src/storage/store/column_chunk_data.cpp
```diff
@@ -601,6 +657,8 @@ void BoolChunkData::write(const ValueVector* vector, offset_t offsetInVector,
         nullData->write(vector, offsetInVector, offsetInChunk);
     }
     numValues = offsetInChunk >= numValues ? offsetInChunk + 1 : numValues;
+    static constexpr uint64_t numValuesToWrite = 1;
+    updateInMemoryStats(inMemoryUpdateStats, *vector, offsetInChunk, numValuesToWrite);
```
royi-luo, line 661:
```suggestion
    updateInMemoryStats(inMemoryUpdateStats, *vector, offsetInVector, numValuesToWrite);
```

src/storage/store/column_chunk_stats.cpp
```diff
@@ -0,0 +1,36 @@
+#include "storage/store/column_chunk_stats.h"
+
+#include "common/type_utils.h"
+
+namespace kuzu {
+namespace storage {
+
+void ColumnChunkStats::update(uint8_t* data, uint64_t offset, uint64_t numValues,
+    common::PhysicalTypeID physicalType) {
+    if (common::TypeUtils::visit(physicalType, []<typename T>(T) { return StorageValueType<T>; })) {
```
royi-luo, line 10:
```suggestion
    bool isStorageValueType = common::TypeUtils::visit(physicalType, []<typename T>(T) { return StorageValueType<T>; });
    if (isStorageValueType) {
```

test/test_files/dml_node/set/set_tinysnb.test
```diff
@@ -40,6 +40,20 @@
...
+---- ok
+-STATEMENT MATCH (a:person) WHERE a.age > 85 RETURN a.ID, a.age
+---- 1
+0|90
+-STATEMENT CREATE (a:person {id: 100, age: 100})
+---- ok
+-STATEMENT MATCH (a:person) WHERE a.age > 95 RETURN a.ID, a.age
+---- 1
+100|100
+
```
royi-luo, line 56: move to `filter/node.test`

andyfengHKU (commented):
LGTM @benjaminwinger or @ray6080 should take a look at nodeGroup and stat update.

src/storage/store/column_chunk_data.cpp
```diff
@@ -368,6 +420,7 @@ void ColumnChunkData::copy(ColumnChunkData* srcChunk, offset_t srcOffsetInChunk,
         }
     }
     append(srcChunk, srcOffsetInChunk, numValuesToCopy);
+    updateInMemoryStats(inMemoryUpdateStats, srcChunk, srcOffsetInChunk, numValuesToCopy);
```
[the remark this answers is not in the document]
royi-luo: Looks like the `copy()` interface isn't used, should I remove it? Maybe @benjaminwinger can comment on this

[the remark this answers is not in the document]
benjaminwinger: There's a comment in the header saying it's used in applyDeletionsToChunk, but that appears to have been removed when MVCC was implemented, so I guess it's no longer needed.

src/include/storage/store/column_chunk_data.h
```diff
@@ -272,6 +278,7 @@ class ColumnChunkData {
 
     // On-disk metadata for column chunk.
     ColumnChunkMetadata metadata;
+    ColumnChunkStats inMemoryUpdateStats;
```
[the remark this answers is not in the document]
benjaminwinger: These stats are supposed to be universally valid right? So I think stats is probably a sufficiently descriptive name (or even inMemoryStats, if you want to distinguish updates to them from updates to the on-disk version).

That said, I notice that the stats aren't being updated in `setValue`? Was that an oversight, or are these only being updated in certain contexts?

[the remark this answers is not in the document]
royi-luo: Good catch, I think the current zone map use cases happen to not need the stats updated in `setValue` but I should definitely do so

src/storage/store/column_chunk_data.cpp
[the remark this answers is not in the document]
royi-luo: ok removed

royi-luo:
Some quick benchmarks on ldbc-sf100:

# Scan Node Table
**Table: `Comment`**
Query: `MATCH (c:Comment) WHERE c.length > {val} RETURN *;`
Query|Time(master)|Time(new)
----|----|----
length>2000|3013ms|46ms
length>1900|3038ms|745ms
length>1000|4558ms|3975ms

# Scan Rel Table
**Table: `likesComment` (first 100M tuples)**
Query: `MATCH (a:person)-[l:likes_comment]->(b:comment) WHERE l.creationDate > {val} RETURN MIN(a.id + b.id);`
Query|Time(master)|Time(new)
----|----|----
`creationDate > timeStamp("2013-09-13 09:35:59.391")`|674ms|74ms
`creationDate > timeStamp("2012-09-10 09:35:59.391")`|3393ms|3143ms
`creationDate > timeStamp("2012-03-10 09:35:59.391")`|16213ms|15577ms

benjaminwinger (approved):

src/include/storage/predicate/column_predicate.h
```diff
@@ -20,7 +20,7 @@ class KUZU_API ColumnPredicateSet {
     }
     bool isEmpty() const { return predicates.empty(); }
 
-    common::ZoneMapCheckResult checkZoneMap(const CompressionMetadata& metadata);
+    common::ZoneMapCheckResult checkZoneMap(const ColumnChunkStats& metadata) const;
```
ray6080, line 23: Can u make the class name and variable name be consistent? I think you'd want to rename `metadata` as `stats`.

ray6080 (commented):
I have a very minor comment as I was browsing the changes. Also, can u add the queries you used to demonstrate the benefits of scan on node/rel tables into our benchmark query set?

src/include/storage/predicate/column_predicate.h
royi-luo: Sure I'll add this in a future PR.

PR #4374, opened 2024-10-16
GDS node filter

andyfengHKU:
# Description

This PR adds support to run node predicate in GDS framework. The general workflow is

- Construct a pipeline populating a semi-mask for nodes passing predicate
- When writing path, apply this semi-mask against all intermediate nodes (excluding src and dst)

Fixes # (issue)

# Contributor agreement


src/function/gds/output_writer.cpp
```diff
@@ -144,6 +148,18 @@ void PathsOutputWriter::write(processor::FactorizedTable& fTable, nodeID_t dstNo
     }
 }
 
+bool PathsOutputWriter::checkPathNode(const std::vector<ParentList*>& path) const {
+    if (pathNodeMask == nullptr) {
+        return true;
+    }
+    for (auto i = 0u; i < path.size() - 1; ++i) {
```
semihsalihoglu-uw, line 155: The logic of applying the node predicate is not very simple. Adding a bit of comment here would help. What does path represent, the sequence of edges? Why are you looping  from 0 to path.size()-2 (inclusive), etc.

src/function/gds/output_writer.cpp
```diff
@@ -82,30 +84,32 @@ void PathsOutputWriter::write(processor::FactorizedTable& fTable, nodeID_t dstNo
...
         if (topNodeID == sourceNodeID) {
-            beginWritingNewPath(curPath.size());
-            if (extendFromSource) {
-                // Write path in reverse direction because we append ParentList from dst to src.
-                for (auto i = 1u; i < curPath.size(); i++) {
-                    auto p = curPath[curPath.size() - 1 - i];
-                    addNode(p->getNodeID(), i - 1);
-                    addEdge(p->getEdgeID(), p->isFwdEdge(), i);
+            // check path
+            if (checkPathNode(curPath)) {
```
semihsalihoglu-uw, line 88: I think this is not the efficient way to apply the node predicate. Your algorithm is: enumerate every path and then check if all the nodes in it pass the predicate. Instead, you should do it while running the DFS logic in the path construction, so in this PR I would apply it inside 140-147 branch:
```
...
else {
            auto parent = bfsGraph.getInitialParentAndNextPtr(topNodeID);
            while (parent->getIter() != top->getIter() - 1 && checkNodePredicateIfIterNotOne(parent)) {
                parent = parent->getNextPtr();
            }
            if (parent != null) { 
               curPath.push_back(parent);
               backtracking = false;
           }
        }
```
First I am assuming that there is a`checkNodePredicateIfIterNotOne` function that checks that the parent is not at iteration 1, in which case we don't need to check if a node predicate applies to it, because it is guaranteed to be the source, and then applies the node predicate (if any) if the iteration is not 1. Second, now you need the `if (parent != null)` condition because now there is a chance that the parent is null. before you could guarantee that some parent must always satisfy this condition: parent->getIter() != top->getIter() - 1.

Further lines 129-137 also need to change (the if condition there needs to be a while loop now):
```
            while (top->getNextPtr() != nullptr  &&
                ((curPath.size() == 1 && top->getNextPtr()->getIter() >= lowerBound) ||
                    top->getNextPtr()->getIter() == top->getIter())) {
                if (checkNodePredicateIfIterNotOne(top->getNextPtr()) {
                  curPath[curPath.size() - 1] = top->getNextPtr();
                  backtracking = false;
                  if (curPath.size() == 1) {
                     lengthVector->setValue<int64_t>(0, curPath[0]->getIter());
                  }
                  break;
                }
            }
            if (backtracking) {
                curPath.pop_back();
            }
```
Something like this. The if condition needs to be replaced with a while loop because we need to check the next parent, which may be at the right level but not pass the node predicate. If it does not pass the node predicate (the inner `if (checkNodePredicateIfIterNotOne(top->getNextPtr())` branch, we should continue to see if there are other parents who are the right level. The previous else branch (currently on line 137) now needs to be an if condition that determines whether or not the while loop was able to find an appropriate parent.

The backtracking logic is hard to understand. So you should have test cases that tests all of these edge cases. This is very important because it will be a nightmare to debug these unless we make sure that we test every single line of this path outputting logic. For example, there should be a test case that tests that each of the while loops in this complex logic iterates multiple times.  Let's talk about your test cases.

Also: if you don't want to do this optimization now, then you should open an issue about it. Though I strongly, strongly suggest we do it now, it's the right time to do all these optimizations and be done with them. The next person will have harder time doing them I think.

src/include/processor/operator/sink.h
```diff
@@ -44,5 +44,24 @@ class Sink : public PhysicalOperator {
     std::unique_ptr<ResultSetDescriptor> resultSetDescriptor;
 };
 
+class DummySink : public Sink {
```
semihsalihoglu-uw, line 47: Hmm, I wasn't expecting that we need a DummySink operator. We have built prior subplans that build semi-masking pipelines for other parts of the query plan, did we not? If so, what was the Sink operator in those subplans?

test/test_files/function/gds/rec_joins_small.test
```diff
@@ -13,6 +13,27 @@
 
 -STATEMENT CALL enable_gds = true;
 ---- ok
+-STATEMENT MATCH p = (a:person1)-[e:knows11*1..4 (r, n | WHERE n.age < 15)]->(b:person1)
```
semihsalihoglu-uw, line 16: I didn't look closely at your tests but these tests look cursory. We need more thorough test cases:
1) Covering all queries: so cover single and shortest path queries.
2) For variable length: covering lower bounds that are not 1.
3) Covering when the node predicate applies to some but not all of the edges in an adjacency list. We need to ensure there are cases when there are multiple parents p1, .., pk, of a node u and that sometimes all of them pass pass the filter, sometimes none, and sometimes some.
4) Further we should test cyclic cases when the source does not pass the filter but we still output the paths when the source node is actually the source and not an intermediate.

Let's talk about your test cases. We need to be very rigorous here.

semihsalihoglu-uw (changes requested):

ray6080 (approved):

src/binder/bind/bind_graph_pattern.cpp
```diff
@@ -449,8 +445,13 @@ std::shared_ptr<RelExpression> Binder::createRecursiveQueryRel(const parser::Rel
     recursiveInfo->nodeCopy = std::move(nodeCopy);
     recursiveInfo->rel = std::move(rel);
     recursiveInfo->lengthExpression = std::move(lengthExpression);
-    recursiveInfo->nodePredicateExecFlag = std::move(nodePredicateExecutionFlag);
-    recursiveInfo->nodePredicate = std::move(nodePredicate);
+    recursiveInfo->nodePredicateExecFlag = nodePredicateExecutionFlag;
```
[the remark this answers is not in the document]
andyfengHKU: As we discussed, this is a piece of code that we are going to remove. So I'm not changing this part.

src/include/binder/expression/rel_expression.h
```diff
@@ -39,6 +39,10 @@ struct RecursiveInfo {
     // During recursive computation, we set 'Flag' to True to avoid executing 'P'.
     std::shared_ptr<Expression> nodePredicateExecFlag;
     std::shared_ptr<Expression> nodePredicate;
+    // The node predicate we use for recursive join & gds is different. Until we migrate recursive
+    // extend to GDS, we need to keep this field.
```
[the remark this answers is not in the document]
andyfengHKU: Ditto

src/include/processor/operator/sink.h
andyfengHKU: As we discussed, this is needed.

src/function/gds/output_writer.cpp
andyfengHKU: As we discussed, I'll do this in a separate PR. And I have updated the issue to track this.

semihsalihoglu-uw (approved):

PR #4379, opened 2024-10-16
Rework logical op print info interface

ray6080:
# Description

Follow up of #3987. Most changes are boilerplate.
- Removed `OPPrintInfo` being passed as parameter of LogicalOperator's constructor, instead generated only when needed and fed to the PlanPrinter.
- Added cardinality to be printed. Always initialized as 0 now. Will feed the actual values in next PR.


andyfengHKU (approved):
LGTM

PR #4386, opened 2024-10-17
Allow scanning edge properties in GDS edgeCompute

benjaminwinger:
This just adds support for a single edge property. It should be relatively simple to extend it so that it provides a list of vectors instead of a single vector.
Note that this is assuming it's not being used in conjunction with the predicates added in #4361. If they were used together, it would scan the same property twice (though it could be changed to have the duplicates share the same output vector).

### DIFF
--- a/src/function/gds/all_shortest_paths.cpp
+++ b/src/function/gds/all_shortest_paths.cpp
@@ -1,5 +1,11 @@
+#include "common/types/types.h"
+#include "function/gds/bfs_graph.h"
+#include "function/gds/gds_frontier.h"
 #include "function/gds/gds_function_collection.h"
 #include "function/gds/rec_joins.h"
+#include "function/gds_function.h"
+#include "graph/graph.h"
+#include "main/client_context.h"
 #include "processor/execution_context.h"
 
 using namespace kuzu::processor;
@@ -140,12 +146,12 @@ public:
         PathMultiplicities* multiplicities)
         : frontierPair{frontierPair}, multiplicities{multiplicities} {};
 
-    void edgeCompute(nodeID_t boundNodeID, std::span<const nodeID_t> nbrIDs,
-        std::span<const relID_t>, SelectionVector& mask, bool) override {
-        size_t activeCount = 0;
-        mask.forEach([&](auto i) {
+    std::vector<nodeID_t> edgeCompute(nodeID_t boundNodeID, GraphScanState::Chunk& resultChunk,
+        bool) override {
+        std::vector<nodeID_t> activeNodes;
+        resultChunk.forEach([&](auto nbrNodeID, auto /*edgeID*/) {
             auto nbrVal =
-                frontierPair->pathLengths->getMaskValueFromNextFrontierFixedMask(nbrIDs[i].offset);
+                frontierPair->pathLengths->getMaskValueFromNextFrontierFixedMask(nbrNodeID.offset);
             // We should update the nbrID's multiplicity in 2 cases: 1) if nbrID is being visited
             // for the first time, i.e., when its value in the pathLengths frontier is
             // PathLengths::UNVISITED. Or 2) if nbrID has already been visited but in this
@@ -155,14 +161,14 @@ public:
             if (shouldUpdate) {
                 // Note: This is safe because curNodeID is in the current frontier, so its
                 // shortest paths multiplicity is guaranteed to not change in the current iteration.
-                multiplicities->incrementTargetMultiplicity(nbrIDs[i].offset,
+                multiplicities->incrementTargetMultiplicity(nbrNodeID.offset,
                     multiplicities->getBoundMultiplicity(boundNodeID.offset));
             }
             if (nbrVal == PathLengths::UNVISITED) {
-                mask.getMutableBuffer()[activeCount++] = i;
+                activeNodes.push_back(nbrNodeID);
             }
         });
-        mask.setToFiltered(activeCount);
+        return activeNodes;
     }
 
     std::unique_ptr<EdgeCompute> copy() override {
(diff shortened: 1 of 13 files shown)


src/graph/on_disk_graph.cpp
```diff
@@ -30,11 +31,15 @@ namespace graph {
 static std::unique_ptr<RelTableScanState> getRelScanState(MemoryManager& mm,
     const TableCatalogEntry& relEntry, const RelTable& table, RelDataDirection direction,
     ValueVector* srcVector, ValueVector* dstVector, ValueVector* relIDVector,
-    expression_vector properties, const Schema& schema, const ResultSet& resultSet) {
+    expression_vector predicateEdgeProperties, std::optional<column_id_t> edgePropertyID,
```
andyfengHKU, line 34: This is fine for this PR.

@acquamarin ultimately u should just pass `edgeProperties` from front-end to replace `predicateEdgeProperties & edgePropertyID`

src/graph/on_disk_graph.cpp
```diff
@@ -30,11 +31,15 @@ namespace graph {
...
+    ValueVector* propertyVector, const Schema& schema, const ResultSet& resultSet) {
     auto columnIDs = std::vector<column_id_t>{NBR_ID_COLUMN_ID, REL_ID_COLUMN_ID};
-    for (auto property : properties) {
+    for (auto property : predicateEdgeProperties) {
         columnIDs.push_back(property->constCast<PropertyExpression>().getColumnID(relEntry));
     }
+    if (edgePropertyID) {
```
andyfengHKU, line 40: @benjaminwinger I would call them `edgePropertyColumnID`. `PropertyID` is terminology in the binding phase. In the processor we usually use `ColumnID`. 

If u decide to change, apply this change across this PR.

src/graph/on_disk_graph.cpp
```diff
@@ -70,6 +78,21 @@ OnDiskGraphScanStates::OnDiskGraphScanStates(ClientContext* context, std::span<R
     relIDVector =
         std::make_unique<ValueVector>(LogicalType::INTERNAL_ID(), context->getMemoryManager());
     relIDVector->state = state;
+    std::optional<column_id_t> edgePropertyID;
+    if (edgePropertyIndex) {
+        // Edge property scans are only supported for single table scans at the moment
+        KU_ASSERT(tables.size() == 1);
+        // TODO(bmwinger): If there are both a predicate and a custom edgePropertyIndex, they will
```
andyfengHKU, line 85: You can probably leave this TODO to me or ziyi. I think the ultimate solution is to pass columnIDs or property expressions directly from the front-end. And we don't need to resolve the case of duplicate scanning.

src/include/catalog/catalog_entry/table_catalog_entry.h
```diff
@@ -58,6 +59,7 @@ class KUZU_API TableCatalogEntry : public CatalogEntry {
     const binder::PropertyDefinition& getProperty(const std::string& propertyName) const;
     const binder::PropertyDefinition& getProperty(common::idx_t idx) const;
     virtual common::column_id_t getColumnID(const std::string& propertyName) const;
+    common::column_id_t getColumnID(common::idx_t idx) const;
```
andyfengHKU, line 62: Remove this. Always get through propertyName instead of idx because there will be internal properties causing index to shift.

andyfengHKU (approved):
LGTM

src/graph/on_disk_graph.cpp
benjaminwinger: Okay

src/include/catalog/catalog_entry/table_catalog_entry.h
benjaminwinger: @acquamarin, is passing a property name to prepareScan acceptable? You'd originally said to pass a column id, I'd changed it to a property index (since you can look up properties to get their type by index and name, but not by column id, and can't look up the index or name using the column ID) as I'd assumed that would be preferable than passing names around.

Unfortunately it seems I then mixed up the types in the signatures, so it's currently using property indices internally, but the type and argument names in prepareScan indicate that it expects a column ID (unfortunately `column_id_t` and `idx_t` are aliases of the same type, so the compiler won't catch that).

src/function/gds/all_shortest_paths.cpp
```diff
@@ -140,12 +145,10 @@ class AllSPDestinationsEdgeCompute : public EdgeCompute {
         PathMultiplicities* multiplicities)
         : frontierPair{frontierPair}, multiplicities{multiplicities} {};
 
-    void edgeCompute(nodeID_t boundNodeID, std::span<const nodeID_t> nbrIDs,
-        std::span<const relID_t>, SelectionVector& mask, bool) override {
-        size_t activeCount = 0;
-        mask.forEach([&](auto i) {
+    void edgeCompute(nodeID_t boundNodeID, GraphScanState::Chunk& resultChunk, bool) override {
+        resultChunk.filter([&](auto nbrNodeID, auto /*edgeID*/) {
```
semihsalihoglu-uw, line 149: I am trying to understand what this filter does. Looking at graph.h, it seems to update the selection vector of the internal scanstate but why? Shouldn't this just be a blind forEach? When do we need to filter the selection vector?

semihsalihoglu-uw (approved):

src/function/gds/all_shortest_paths.cpp
benjaminwinger: It's because we're re-using the SelectionVector in the chunk to set the next frontier. The filter function acts like forEach, but also updates the SelectionVector to include only the neighbours for which the function passed to filter returns true.
Maybe filter isn't the ideal name, but I thought that this was better than manually modifying the selection vector.

https://github.com/kuzudb/kuzu/blob/63840d0593bbd894e7c27258a289bff8c0844252/src/function/gds/gds_task.cpp#L15-L17
(the above is slightly different in this PR, but github doesn't show the inline code chunk for PR links)

benjaminwinger: I've updated it to have edgeCompute return an `std::vector<nodeID_t>` instead for all the active nodes. The cost of this seems to be quite small compared to the rest of the work being done (benchmarking some of the algorithms currently implemented I couldn't find a measurable difference).

semihsalihoglu-uw (approved):

PR #4387, opened 2024-10-17
Support different path semantic for gds

andyfengHKU:
# Description

This PR implements `TRAIL & ACYCLIC` semantic for GDS framework.

Fixes # (issue)

# Contributor agreement


src/common/enums/query_rel_type.cpp
```diff
@@ -0,0 +1,23 @@
+#include "common/enums/query_rel_type.h"
+#include "common/assert.h"
+
+namespace kuzu {
+namespace common {
+
+PathSemantic QueryRelTypeUtils::getPathSemantic(QueryRelType queryRelType) {
+    switch (queryRelType) {
+    case QueryRelType::VARIABLE_LENGTH_WALK:
+    case QueryRelType::SHORTEST:
```
semihsalihoglu-uw, line 10: Shortest and All Shortest (unweighted) are by default both acyclic and trail though I'd mark them as trail. Why did you mark them as walk?

src/function/gds/output_writer.cpp
```diff
@@ -83,31 +81,10 @@ void PathsOutputWriter::write(processor::FactorizedTable& fTable, nodeID_t dstNo
...
-                    for (auto i = 0u; i < curPath.size() - 1; i++) {
-                        auto p = curPath[i];
-                        addNode(p->getNodeID(), i);
-                        addEdge(p->getEdgeID(), p->isFwdEdge(), i);
-                    }
-                    auto lastPathElement = curPath[curPath.size() - 1];
-                    addEdge(lastPathElement->getEdgeID(), lastPathElement->isFwdEdge(),
-                        curPath.size() - 1);
-                }
+            if (checkPathNodeMask(curPath) && checkSemantic(curPath)) {
```
semihsalihoglu-uw, line 86: My comment is the same. Doing these check after each path enumeration can be arbitrarily slower than doing it as you do the DFS-based path enumeration. So similar to my comment in the other PR, I would do this as part of the DFS-based enumeration. The most obvious way is to have some auxiliary c++ map that you keep each time addEdge is called if the semantics is trail or acyclic and call checkSemantic before adding a new edge. The logic around where you make the addEdge calls might need to change somewhat as well.

test/test_files/function/gds/basic.test
```diff
@@ -114,3 +114,13 @@ Hubert Blaine Wolfeschlegelsteinhausenbergerdorff|0.018750
 [Alice,Dan,Bob]|[2021-06-30,1950-05-14]|[0:0,0:3]|[0:3,0:1]|Alice|Bob
 [Alice,Dan,Carol]|[2021-06-30,2000-01-01]|[0:0,0:3]|[0:3,0:2]|Alice|Carol
 [Alice,Dan]|[2021-06-30]|[0:0]|[0:3]|Alice|Dan
+-STATEMENT MATCH p = (a:person)-[e:knows* 3..3]->(b:person) WHERE a.ID = 0 RETURN COUNT(*)
```
semihsalihoglu-uw, line 117: One edge case I'm curious about is whether the source and destination nodes are the same for acyclic. I believe such paths are also not allowed, right? To be more specific, I'm curious about what to do in the following query:
```
MATCH (a:Person)-[r:Knows* ACYCLIC]->(b:Person)
WHERE a.name = Alice and b.name=Alice
RETURN *
```
So suppose there is an Alice node, node 5, and node 5 has a self loop edge to itself. Does that qualify as a valid output? Or say an edge like node 5 -[Knows]-> node 7 -[Know]-> node 5. The question is the [r:Knows*] part of this matched pattern only involves -[Knows]-> node 7 -[Know]-, which is acyclic (though is not really a path), but if you consider the source and end nodes, the full path, i.e., p, is acyclic.  I would learn what should be done here (maybe look at Neo4j) and then add a test case for it.

semihsalihoglu-uw (changes requested):

ray6080 (approved):

semihsalihoglu-uw (approved):

test/test_files/function/gds/basic.test
andyfengHKU: Neo4j doesn't have this ACYCLIC keyword in the recursive pattern.

```
MATCH p = (a:Person)-[r:Knows* ACYCLIC]->(b:Person)
WHERE a.name = Alice and b.name=Alice
RETURN *
```

In your example, `r` is acyclic as it only involves `-[Knows]-> node 7 -[Know]-`. But `p` is cyclic.

src/function/gds/output_writer.cpp
andyfengHKU: As we discussed, we will do this optimization together with node filter optimization.

src/common/enums/query_rel_type.cpp
andyfengHKU: They don't make much difference when we do the BFS backtrack because the graph itself doesn't contain duplicate nodes anyways. But I have changed semantic to acyclic as you requested.

Issue #2974, opened 2024-02-29
[Java API] Change `finalize` to `AutoCloseable`


mewim:
The current API uses finalizers to auto release the underlying C++ object, but it is deprecated. We should impanelment  `AutoCloseable` instead.

PR #4388, opened 2024-10-17
Replacing Deprecated Finalize() with autoCloseable from JAVA_API

WWW0030:
# Description

Replacing Deprecated `Finalize()` with `autoCloseable` from JAVA_API

Fixes # (issue) 
https://github.com/kuzudb/kuzu/issues/2974#event-14662134856

# Contributor agreement

### DIFF
--- a/tools/java_api/src/main/java/com/kuzudb/Connection.java
+++ b/tools/java_api/src/main/java/com/kuzudb/Connection.java
@@ -6,7 +6,7 @@ import java.util.Map;
  * Connection is used to interact with a Database instance. Each Connection is thread-safe. Multiple
  * connections can connect to the same Database instance in a multi-threaded environment.
  */
-public class Connection {
+public class Connection implements AutoCloseable {
 
     long conn_ref;
     boolean destroyed = false;
@@ -31,11 +31,11 @@ public class Connection {
     }
 
     /**
-    * Finalize.
+    * Close.
     * @throws ObjectRefDestroyedException If the connection has been destroyed.
     */
     @Override
-    protected void finalize() throws ObjectRefDestroyedException {
+    public void close() throws ObjectRefDestroyedException {
         destroy();
     }
 
@@ -43,7 +43,7 @@ public class Connection {
     * Destroy the connection.
     * @throws ObjectRefDestroyedException If the connection has been destroyed.
     */
-    public void destroy() throws ObjectRefDestroyedException {
+    private void destroy() throws ObjectRefDestroyedException {
         checkNotDestroyed();
         Native.kuzu_connection_destroy(this);
         destroyed = true;
--- a/tools/java_api/src/main/java/com/kuzudb/Connection.java
+++ b/tools/java_api/src/main/java/com/kuzudb/Connection.java
@@ -31,7 +31,7 @@ public class Connection implements AutoCloseable {
     }
 
     /**
-    * Close.
+    * Used to destroy all allocated memory of the data. Automatically called by AutoCloseable within try-with-resouce clauses.
     * @throws ObjectRefDestroyedException If the connection has been destroyed.
     */
     @Override
--- a/tools/java_api/src/main/java/com/kuzudb/Connection.java
+++ b/tools/java_api/src/main/java/com/kuzudb/Connection.java
@@ -31,7 +31,7 @@ public class Connection implements AutoCloseable {
     }
 
     /**
-    * Used to destroy all allocated memory of the data. Automatically called by AutoCloseable within try-with-resouce clauses.
+    * Close the connection and release the underlying resources. This method is invoked automatically on objects managed by the try-with-resources statement.    
     * @throws ObjectRefDestroyedException If the connection has been destroyed.
     */
     @Override
--- a/tools/java_api/src/main/java/com/kuzudb/Connection.java
+++ b/tools/java_api/src/main/java/com/kuzudb/Connection.java
@@ -31,7 +31,7 @@ public class Connection implements AutoCloseable {
     }
 
     /**
-    * Close the Connection and release the underlying resources. This method is invoked automatically on objects managed by the try-with-resources statement.    
+    * Close the connection and release the underlying resources. This method is invoked automatically on objects managed by the try-with-resources statement.    
     * @throws ObjectRefDestroyedException If the connection has been destroyed.
     */
     @Override
--- a/tools/java_api/src/main/java/com/kuzudb/Connection.java
+++ b/tools/java_api/src/main/java/com/kuzudb/Connection.java
@@ -31,7 +31,7 @@ public class Connection implements AutoCloseable {
     }
 
     /**
-    * Close the connection and release the underlying resources. This method is invoked automatically on objects managed by the try-with-resources statement.    
+    * Close the Connection and release the underlying resources. This method is invoked automatically on objects managed by the try-with-resources statement.    
     * @throws ObjectRefDestroyedException If the connection has been destroyed.
     */
     @Override
--- a/tools/java_api/src/main/java/com/kuzudb/DataType.java
+++ b/tools/java_api/src/main/java/com/kuzudb/DataType.java
@@ -30,7 +30,7 @@ public class DataType implements AutoCloseable {
     }
 
     /**
-    * Close the connection and release the underlying resources. This method is invoked automatically on objects managed by the try-with-resources statement.
+    * Close the dataType and release the underlying resources. This method is invoked automatically on objects managed by the try-with-resources statement.
     * @throws ObjectRefDestroyedException If the data type instance has been destroyed.
     */
     @Override
--- a/tools/java_api/src/main/java/com/kuzudb/DataType.java
+++ b/tools/java_api/src/main/java/com/kuzudb/DataType.java
@@ -30,7 +30,7 @@ public class DataType implements AutoCloseable {
     }
 
     /**
-    * Used to destroy all allocated memory of the data. Automatically called by AutoCloseable within try-with-resouce clauses.
+    * Close the connection and release the underlying resources. This method is invoked automatically on objects managed by the try-with-resources statement.
     * @throws ObjectRefDestroyedException If the data type instance has been destroyed.
     */
     @Override
--- a/tools/java_api/src/main/java/com/kuzudb/DataType.java
+++ b/tools/java_api/src/main/java/com/kuzudb/DataType.java
@@ -30,7 +30,7 @@ public class DataType implements AutoCloseable {
     }
 
     /**
-    * Close the dataType and release the underlying resources. This method is invoked automatically on objects managed by the try-with-resources statement.
+    * Close the DataType and release the underlying resources. This method is invoked automatically on objects managed by the try-with-resources statement.
     * @throws ObjectRefDestroyedException If the data type instance has been destroyed.
     */
     @Override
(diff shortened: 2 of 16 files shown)


mewim (changes requested):
There is a major issue in this PR. The `AutoCloseable` will only call `close()` automatically "when exiting a try-with-resources block for which the object has been declared" (as mentioned in Java docs). There is NO implication with garbage collection activities. But you have decided to remove all the manual calls to destroy the objects, which causes a lot of memory leak. I agree that we should not have both `destroy` and `close` public methods (which can be confusing). But instead of removing all the calls to `destroy`, you should change it to `close`.

tools/java_api/src/main/java/com/kuzudb/Connection.java
```diff
@@ -31,19 +31,19 @@ private void checkNotDestroyed() throws ObjectRefDestroyedException {
     }
 
     /**
-    * Finalize.
+    * Close.
```
mewim, line 34: The description here should be more descriptive.

tools/java_api/src/main/java/com/kuzudb/DataType.java
```diff
@@ -30,19 +30,19 @@ private void checkNotDestroyed() throws ObjectRefDestroyedException {
     }
 
     /**
-    * Finalize.
+    * Close.
```
mewim, line 33: Ditto

tools/java_api/src/main/java/com/kuzudb/Database.java
```diff
@@ -60,19 +60,19 @@ private void checkNotDestroyed() throws ObjectRefDestroyedException {
     }
 
     /**
-    * Finalize.
+    * Close.
```
mewim, line 63: Ditto

tools/java_api/src/main/java/com/kuzudb/FlatTuple.java
```diff
@@ -17,19 +17,19 @@ private void checkNotDestroyed() throws ObjectRefDestroyedException {
     }
 
     /**
-     * Finalize.
+     * Close.
```
mewim, line 20: Ditto

tools/java_api/src/main/java/com/kuzudb/PreparedStatement.java
```diff
@@ -17,19 +17,19 @@ private void checkNotDestroyed() throws ObjectRefDestroyedException {
     }
 
     /**
-    * Finalize.
+    * Close.
```
mewim, line 20: Ditto

tools/java_api/src/main/java/com/kuzudb/QueryResult.java
```diff
@@ -18,11 +18,11 @@ private void checkNotDestroyed() throws ObjectRefDestroyedException {
     }
 
     /**
-     * Finalize.
+     * Close.
```
mewim, line 21: Ditto

tools/java_api/src/main/java/com/kuzudb/Value.java
```diff
@@ -52,11 +52,11 @@ public void checkNotDestroyed() throws ObjectRefDestroyedException {
     }
 
     /**
-     * Finalize.
+     * Close.
```
mewim, line 55: Dittto

mewim (changes requested):
Also, see my comments. All the `close()` function needs to have a descriptive comment. Since they are public methods, the comment will be used in generating the javadoc.

tools/java_api/src/main/java/com/kuzudb/Connection.java
WWW0030: Changed, please let me know if the new description is more fitting 👍

mewim (commented):
I think the comments should be more specific. For example:
`Close the connection and release the underlying resources. This method is invoked automatically on objects managed by the try-with-resources statement.`

tools/java_api/src/test/java/com/kuzudb/ConnectionTest.java
```diff
@@ -18,9 +18,7 @@ public class ConnectionTest extends TestBase {
 
     @Test
     void ConnCreationAndDestroy() {
-        try {
-            Connection conn = new Connection(db);
-            conn.destroy();
+        try ( Connection conn = new Connection(db);) {
```
mewim, line 21: Fix the format here. It should not requre `;`.

tools/java_api/src/test/java/com/kuzudb/DatabaseTest.java
```diff
@@ -15,35 +15,35 @@ public class DatabaseTest extends TestBase {
     void DBCreationAndDestroyWithArgs() {
         try {
             String dbPath = tempDir.toFile().getAbsolutePath();
-            Database database = new Database(
+            try (Database database = new Database(
```
mewim, line 18: I think you should add seperate `catch` for database creation. If your database creation failed, your test case should be aware that it is caused by database creation failure. 
So something like:
```java
String dbPath;
try {
  String dbPath = tempDir.toFile().getAbsolutePath();
} catch(Exception e) {
  fail("Cannot get database path: " + e.getMessage();
}

try (Database database = new Database(...){
}
catch(e){
  fail("DBCreationAndDestroyWithArgs failed: " + e.getMessage();
}
```

mewim (changes requested):
See my comments.

tools/java_api/src/test/java/com/kuzudb/DatabaseTest.java
mewim: Note that this applies to all functions in the same file

tools/java_api/src/main/java/com/kuzudb/DataType.java
```diff
@@ -30,19 +30,19 @@ private void checkNotDestroyed() throws ObjectRefDestroyedException {
     }
 
     /**
-    * Finalize.
+    * Close the connection and release the underlying resources. This method is invoked automatically on objects managed by the try-with-resources statement.
```
mewim, line 33: My example is for `Connection` class. You should change it for each class and not blindly copy it everywhere.

tools/java_api/src/main/java/com/kuzudb/Database.java
```diff
@@ -60,19 +60,19 @@ private void checkNotDestroyed() throws ObjectRefDestroyedException {
     }
 
     /**
-    * Finalize.
+    * Close the connection and release the underlying resources. This method is invoked automatically on objects managed by the try-with-resources statement.
```
mewim, line 63: Ditto

tools/java_api/src/main/java/com/kuzudb/FlatTuple.java
```diff
@@ -17,19 +17,19 @@ private void checkNotDestroyed() throws ObjectRefDestroyedException {
     }
 
     /**
-     * Finalize.
+    * Close the connection and release the underlying resources. This method is invoked automatically on objects managed by the try-with-resources statement.
```
mewim, line 20: Ditto

tools/java_api/src/main/java/com/kuzudb/PreparedStatement.java
```diff
@@ -17,19 +17,19 @@ private void checkNotDestroyed() throws ObjectRefDestroyedException {
     }
 
     /**
-    * Finalize.
+    * Close the connection and release the underlying resources. This method is invoked automatically on objects managed by the try-with-resources statement.
```
mewim, line 20: Ditto

tools/java_api/src/main/java/com/kuzudb/QueryResult.java
```diff
@@ -18,11 +18,11 @@ private void checkNotDestroyed() throws ObjectRefDestroyedException {
     }
 
     /**
-     * Finalize.
+    * Close the connection and release the underlying resources. This method is invoked automatically on objects managed by the try-with-resources statement.
```
mewim, line 21: Ditto

tools/java_api/src/main/java/com/kuzudb/Value.java
```diff
@@ -52,11 +52,11 @@ public void checkNotDestroyed() throws ObjectRefDestroyedException {
     }
 
     /**
-     * Finalize.
+    * Close the connection and release the underlying resources. This method is invoked automatically on objects managed by the try-with-resources statement.
```
mewim, line 55: Ditto

mewim (changes requested):
See my comments

tools/java_api/src/main/java/com/kuzudb/DataType.java
WWW0030: Yes, sorry didn't check the comment closely enough, will fix

tools/java_api/src/main/java/com/kuzudb/QueryResult.java
```diff
@@ -18,11 +18,11 @@ private void checkNotDestroyed() throws ObjectRefDestroyedException {
     }
 
     /**
-     * Finalize.
+    * Close the queryResult and release the underlying resources. This method is invoked automatically on objects managed by the try-with-resources statement.
```
mewim, line 21: Using `camelCase` does not make sense here. It should either be `query result` (plain English) or `QueryResult` (consistent with the class name)

tools/java_api/src/main/java/com/kuzudb/DataType.java
```diff
@@ -30,19 +30,19 @@ private void checkNotDestroyed() throws ObjectRefDestroyedException {
     }
 
     /**
-    * Finalize.
+    * Close the dataType and release the underlying resources. This method is invoked automatically on objects managed by the try-with-resources statement.
```
mewim, line 33: Ditto

tools/java_api/src/main/java/com/kuzudb/FlatTuple.java
```diff
@@ -17,19 +17,19 @@ private void checkNotDestroyed() throws ObjectRefDestroyedException {
     }
 
     /**
-     * Finalize.
+    * Close the flatTuple and release the underlying resources. This method is invoked automatically on objects managed by the try-with-resources statement.
```
mewim, line 20: Ditto

tools/java_api/src/main/java/com/kuzudb/PreparedStatement.java
```diff
@@ -17,19 +17,19 @@ private void checkNotDestroyed() throws ObjectRefDestroyedException {
     }
 
     /**
-    * Finalize.
+    * Close the perparedStatement and release the underlying resources. This method is invoked automatically on objects managed by the try-with-resources statement.
```
mewim, line 20: Ditto

mewim (changes requested):
See my comments

tools/java_api/src/main/java/com/kuzudb/QueryResult.java
mewim: If you decide to change it to `QueryResult`, you need to change `database` -> `Database` etc. to be consistent in other comments

mewim: So I'd suggest to just go with `query result`

mewim (approved):
Can be merged if CI passes

PR #4393, opened 2024-10-18
Zone map enable follow-up

### LINKED ISSUE #4389: Bug: Zone map doesn't work on integer fields smaller than `INT64`
### Kùzu version

master

Ubuntu 22.04.5 LTS

### What happened?

In the transformer we treat integer literals as either INT64 or INT128.
If we have a table `create node table someTable(..., value INT32, ...)`, the expression `someTable.value < {some literal}` will have the LHS be of type `CAST_INT64(value)` which is of type `ExpressionType::FUNCTION`. The operator `<` can be replaced with any comparison operator.

This becomes a problem in `ColumnPredicateSet::getPredicateSet` because when we try to convert the predicate to a `ColumnConstantPredicate` in `ColumnPredicateUtil::tryConvert()` this will fail as we don't consider predicates with an LHS of type `FUNCTION` to be convertible to `ColumnConstantPredicate`.
As a result, zone mapping will never succeed as there is no `ColumnConstantPredicate` to perform zone mapping on. This will happen for any type that requires a cast to `INT64` to compare against a literal.

_No response_

royi-luo:
# Description

- Zone map now works if LHS of predicate is a casted column reference
- Add zone map benchmarks

Fixes #4389

# Contributor agreement


src/storage/store/chunked_node_group.cpp
```diff
@@ -203,11 +203,17 @@ static ZoneMapCheckResult getZoneMapResult(const TableScanState& scanState,
...
             }
-            KU_ASSERT(i < scanState.columnPredicateSets.size());
-            const auto columnZoneMapResult = scanState.columnPredicateSets[i].checkZoneMap(
-                chunks[columnID]->getData().getMergedColumnChunkStats());
-            if (columnZoneMapResult == common::ZoneMapCheckResult::SKIP_SCAN) {
-                return common::ZoneMapCheckResult::SKIP_SCAN;
+
+            const bool columnHasStorageValueType =
+                TypeUtils::visit(chunks[columnID]->getDataType().getPhysicalType(),
+                    []<typename T>(T) { return StorageValueType<T>; });
```
royi-luo, line 209: Found some weirdness going on in the RDF tests where we can cast an RDF value (physical type struct) to an integer. I'm not sure if we can do that with normal structs but here it only checks the zone map if the column has a storage value type just in case.

src/storage/predicate/column_predicate.cpp
```diff
@@ -33,21 +34,38 @@ static bool isColumnRef(ExpressionType type) {
     return type == ExpressionType::PROPERTY || type == ExpressionType::VARIABLE;
 }
 
+static bool isCastedColumnRef(const Expression& expr) {
+    if (expr.expressionType == ExpressionType::FUNCTION) {
+        const auto& funcExpr = expr.constCast<ScalarFunctionExpression>();
+        if (funcExpr.getFunction().name.starts_with("CAST") && funcExpr.getNumChildren() > 0) {
```
andyfengHKU, line 40: Isn't checking function name starts_with "CAST" guarantees numChildren > 0? I think the alter part should be an KU_ASSERT.

src/storage/store/chunked_node_group.cpp
andyfengHKU: I would remove those RDF test since we are removing RDF anyways and make this an assertion.

andyfengHKU (approved):

src/storage/predicate/column_predicate.cpp
royi-luo: changed to assert

src/storage/store/chunked_node_group.cpp
royi-luo: added assertion that we only skip scan if `columnHasStorageValueType` is true

PR #4395, opened 2024-10-21
Added function names to cli autocomplete

MSebanc:
# Description

Added function names to cli autocomplete. Currently, keywords are prioritized over function names.

# Contributor agreement

### DIFF
--- a/tools/shell/embedded_shell.cpp
+++ b/tools/shell/embedded_shell.cpp
@@ -77,6 +77,7 @@ std::vector<std::string> relTableNames;
 std::unordered_map<std::string, std::vector<std::string>> tableColumnNames;
 
 std::vector<std::string> functionNames;
+std::vector<std::string> tableFunctionNames;
 
 bool continueLine = false;
 std::string currLine;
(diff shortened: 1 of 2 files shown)


tools/shell/embedded_shell.cpp
```diff
@@ -176,6 +186,27 @@ void findTableVariableNames(const std::string buf, std::regex tableRegex,
...
 }
 
+void keywordCompletion(std::string buf, std::string prefix, std::string keyword,
+    linenoiseCompletions* lc) {
+    std::string bufEscaped = regex_replace(buf, specialChars, R"(\$&)");
+    if (regex_search(keyword, std::regex("^" + bufEscaped, std::regex_constants::icase))) {
+        std::string transformedKeyword = keyword;
+        for (size_t i = 0; i < buf.size() && i < keyword.size(); ++i) {
+            if (islower(buf[i])) {
+                transformedKeyword[i] = tolower(keyword[i]);
```
andyfengHKU, line 196: I don't know if it makes sense to keep a lower version and upper version of keyword list make sense or not?

andyfengHKU (approved):
LGTM!

tools/shell/embedded_shell.cpp
MSebanc: This part of the code makes sure that the keyword we are completing matches the case typed by the user when we do the autocomplete. Without this, if "Ma" was typed, since our keyword is "MATCH", later on in the linenoise code this autocompletion would be rejected

andyfengHKU: I see. Sounds good.

PR #4396, opened 2024-10-21
Recursive relationships support rel label pruning

ted-wq-x:
# Description

As the title suggests.

Fixes # (issue)

# Contributor agreement

### DIFF
--- a/src/binder/query/query_graph_label_analyzer.cpp
+++ b/src/binder/query/query_graph_label_analyzer.cpp
@@ -22,7 +22,6 @@ void QueryGraphLabelAnalyzer::pruneLabel(QueryGraph& graph) {
 }
 
 void QueryGraphLabelAnalyzer::pruneNode(const QueryGraph& graph, NodeExpression& node) {
-    auto catalog = clientContext.getCatalog();
     for (auto i = 0u; i < graph.getNumQueryRels(); ++i) {
         auto queryRel = graph.getQueryRel(i);
         if (queryRel->isRecursive()) {
@@ -32,7 +31,6 @@ void QueryGraphLabelAnalyzer::pruneNode(const QueryGraph& graph, NodeExpression&
         std::unordered_set<std::string> candidateNamesSet;
         auto isSrcConnect = *queryRel->getSrcNode() == node;
         auto isDstConnect = *queryRel->getDstNode() == node;
-        auto tx = clientContext.getTx();
         if (queryRel->getDirectionType() == RelDirectionType::BOTH) {
             if (isSrcConnect || isDstConnect) {
                 for (auto entry : queryRel->getEntries()) {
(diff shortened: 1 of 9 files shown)


andyfengHKU (commented):
Hi @ted-wq-x,

Thanks for the contribution. I quickly glanced through this PR and believe the overall direction is good. 

Though I will need to put this PR on hold for a week because we are migrating recursive_join implementation to GDS. I don't think there will be too much conflicts. But for simplicity I'll prioritize the migration and then come back to this PR.

Hope this is fine for you.

ted-wq-x:
Yes,that's fine for me.

ted-wq-x:
Is there any progress about this pr?

andyfengHKU:
> Is there any progress about this pr?

Whoops. Forget to review this. Let me do it today.

src/binder/query/query_graph_label_analyzer.cpp
```diff
@@ -93,49 +91,182 @@ void QueryGraphLabelAnalyzer::pruneNode(const QueryGraph& graph, NodeExpression&
...
+                for (const auto& entry : item) {
+                    activationTabelIDs.push_back(entry->getTableID());
+                }
+                stepActivationRelInfos.push_back(activationTabelIDs);
+            }
+
+            rel.getRecursiveInfo()->stepActivationRelInfos = stepActivationRelInfos;
+            std::vector<catalog::TableCatalogEntry*> newRelEntries{temp.begin(), temp.end()};
+            if (!same(newRelEntries, rel.getEntries())) {
+                rel.setEntries(newRelEntries);
```
andyfengHKU, line 231: As I understand you are trying prune src & dst node labels here. This is not the best place though.

Recall at the very top level of this rewriter we have

```
void QueryGraphLabelAnalyzer::pruneLabel(QueryGraph& graph) {
    for (auto i = 0u; i < graph.getNumQueryNodes(); ++i) {
        pruneNode(graph, *graph.getQueryNode(i));
    }
    for (auto i = 0u; i < graph.getNumQueryRels(); ++i) {
        pruneRel(*graph.getQueryRel(i));
    }
}

```

Ideally this should be an iterative process until coverage. So something like
```
while (!converged) {
      pruneNodes()
      pruneRels()
}
```

We don't do the iterative way because we don't wanna spend too much time in this rewrite stage. 

In your case, if this rewrite of node labels is important. I would suggest the following

```
pruneNodes()
pruneRels()
pruneNodes()
```

src/include/binder/expression/node_rel_expression.h
```diff
@@ -32,6 +32,7 @@ class NodeOrRelExpression : public Expression {
     common::table_id_vector_t getTableIDs() const;
     common::table_id_set_t getTableIDsSet() const;
     const std::vector<catalog::TableCatalogEntry*>& getEntries() const { return entries; }
+    std::vector<catalog::TableCatalogEntry*>& getEntries() { return entries; }
```
andyfengHKU, line 35: I would remove this. All modification should be done be setEntries()

andyfengHKU (changes requested):
Overall LGTM. This optimization make sense. I left some comments regarding implementation details.

Though one major thing has changed. The recursive join operator is deprecated and we are running recursive computation through GDS.

I suggest we keep the QueryGraphLabelAnalyzer part of this PR and merge to branch `recursive_pattern_label_pruning` branch. And I can plugin the part for GDS.

Let me know what's your thought on this.

ted-wq-x:
I agree with your idea.

src/binder/query/query_graph_label_analyzer.cpp
ted-wq-x: Because only the src of the first hop and the dst of the last hop need to be adjusted, the logic is quite different from prunNode, so I have handled it here.

andyfengHKU:
> I agree with your idea, and i fixed all of your comments except `https://github.com/kuzudb/kuzu/pull/4396#discussion_r1862493053`

Sounds good. Let me change the base and take over.

ted-wq-x:
Is there any progress on this pr?

PR #4398, opened 2024-10-21
remove  REL TABLE GROUP and make REL TABLE subsume it

SterlingT3485:
The high level idea is to remove the notion of REL TABLE GROUP and make REL TABLE subsume its functionality which is defining relationship between different pairs of node tables . 

## DDL

Allowing multiple `FROM TO` in `CREATE REL TABLE`
```
CREATE REL TABLE R (FROM N1 TO N1, FROM N1 TO N2，date STRING)
```

- Internally, we create `rel table group` if found multiple `from to pairs`
- `R` rel table group
- `R_N1_N1, R_N1_N2` rel table 

All DDLs should work on both `rel table` & `rel table group`.  Check for
- Drop table
- Rename table
- Alter table drop/rename/add column
`Rel table group` should just 

Drop an internal table under `rel table group` should be protected.

## DML

Bind `rel table group` as multi-labeled relationship.

SET & DELETE should work directly.

CREATE is allowed only if both src and destination node labels are given, e.g.
```
CREATE REL TABLE R (FROM N1 TO N1, FROM N1 TO N2)
Create (a) // Error
Create (a:N1)-[R]->(b:N1)
```

## Query

Bind `rel table group` as multi-labeled relationship.  All existing cases should pass.

```
MATCH (a)-[:R]->(b)

MATCH (a)-[:R_N1_N1|:R_N1_N2]
```

## Copy

First we should provide backward compatibility meaning we still expose internal rel group table for copy, e.g.
```
Copy R FROM '' (from=N1, to=N1);
```

we should ask for two extra columns along side `from, to` to indicate node labels, e.g.
```
from, from_label, to, to_label
1, N1, 2, N1
1, N1, 2, N2
...
```
so that we can directly copy with
```
COPY R FROM
```

## Workflows

- [x] - Change grammar and binds to `rel table group entry` if multiple  `FROM TO` pairs. 
- [x] - Merge the `create rel table group` grammar into `create rel table`. And we can remove the `create rel table group` grammar.
- [x] - Check all DDLs works with `rel table group`.  Finish the #DDL section
- [x] - Finish #DML section. The only change should be checking the special case if src & dst node labels are given for a rel group relationship.
- [x] - Check all queries work as expected.
- [x] - Provide backward compatibility for `COPY`
- [ ] - Support directly `COPY` into `rel table group`with `from_label` & `to_label` columns
- [x] - Test export/import rel table group between different version. Backward compatibility.
- [ ] - show_tables() should not print rel 

```

### CODE REVIEW

File: scripts/antlr4/generate_grammar.cmake [hunk 1]
```diff
@@ -10,10 +10,10 @@ file(READ hash.md5 OLDHASH)
 execute_process(
     COMMAND ${Python3_EXECUTABLE} hash.py ${ROOT_DIR}/src/antlr4/keywords.txt ${ROOT_DIR}/src/antlr4/Cypher.g4 OUTPUT_VARIABLE NEWHASH)
 
-if("${OLDHASH}" STREQUAL "${NEWHASH}")
-    message(DEBUG " Not regenerating grammar files as Cypher.g4 and keywords.txt is unchanged.")
-    return() # Exit.
-endif()
+# if("${OLDHASH}" STREQUAL "${NEWHASH}")
```
andyfengHKU, line 13: Double check if we should comment this out?

andyfengHKU (commented):
Cool I think we are on the same page. Let's discuss one more time when we wanna support the second type of `Copy` for rel group.

andyfengHKU:
The is done in #4755

PR #4399, opened 2024-10-21
Implement single thread mode for task scheduler

mewim:
# Description

This PR adds single thread mode to scheduler. It is controlled by a compile-time flag `SINGLE_THREADED`. In single thread mode, no new thread will be spawned and everything will execute on main thread, so that including and linking against thread library is not needed. This is useful to create a single-threaded version of binary for WebAssembly later that does not require cross-origin isolation. Currently the single-threaded mode test has been enabled as part of WebAssembly test.


src/common/task_system/task_scheduler.cpp
```diff
@@ -78,6 +78,83 @@ void TaskScheduler::scheduleTaskAndWaitOrError(const std::shared_ptr<Task>& task
...
+// Single-threaded version of TaskScheduler
+TaskScheduler::TaskScheduler(uint64_t) : stopWorkerThreads{false}, nextScheduledTaskID{0} {}
+
+TaskScheduler::~TaskScheduler() {
+    stopWorkerThreads = true;
+}
+
+void TaskScheduler::scheduleTaskAndWaitOrError(const std::shared_ptr<Task>& task,
+    processor::ExecutionContext* context, bool) {
+
```
mewim, line 129: Remove this

semihsalihoglu-uw (approved):

PR #4401, opened 2024-10-21
Add rel multiplicity check for insertions

### LINKED ISSUE #4334: Bug: Relationship multiplicity not working
### Kùzu version

v0.6.0

Arch Linux x64

### What happened?

I feel the relationship constraints are not working as documented in https://docs.kuzudb.com/cypher/data-definition/create-table/#relationship-multiplicities.

This Cypher executes successfully on an empty in-memory DB:
```cypher
CREATE NODE TABLE User(id STRING, PRIMARY KEY (id));
CREATE NODE TABLE City(id STRING, PRIMARY KEY (id));
CREATE REL TABLE LivesIn(FROM User TO City, ONE_ONE);
CREATE (u:User { id: 'user1' });
CREATE (u:User { id: 'user2' });
CREATE (c:City { id: 'city1' });
CREATE (c:City { id: 'city2' });
MATCH (u:User), (c:City) WHERE u.id = 'user1' AND c.id = 'city1' CREATE (u)-[e:LivesIn]->(c);
MATCH (u:User), (c:City) WHERE u.id = 'user2' AND c.id = 'city1' CREATE (u)-[e:LivesIn]->(c);
MATCH (u:User), (c:City) WHERE u.id = 'user1' AND c.id = 'city2' CREATE (u)-[e:LivesIn]->(c);
MATCH (u)-[e]->(c) RETURN *;
```
This is the result of the last `MATCH` statement in the Explorer:
image

1 User Lives in 2 Cities while 1 City has 2 LivesIn Users. The point of stating `ONE_ONE` means that the relation must be 1 in both directions.
Is this a bug or am I doing something wrong?

Start Explorer with an empty DB
```
docker run -p 8000:8000 --rm -e KUZU_IN_MEMORY=true kuzudb/explorer:latest
```
Run the queries
```cypher
```cypher
CREATE NODE TABLE User(id STRING, PRIMARY KEY (id));
CREATE NODE TABLE City(id STRING, PRIMARY KEY (id));
CREATE REL TABLE LivesIn(FROM User TO City, ONE_ONE);
CREATE (u:User { id: 'user1' });
CREATE (u:User { id: 'user2' });
CREATE (c:City { id: 'city1' });
CREATE (c:City { id: 'city2' });
MATCH (u:User), (c:City) WHERE u.id = 'user1' AND c.id = 'city1' CREATE (u)-[e:LivesIn]->(c);
MATCH (u:User), (c:City) WHERE u.id = 'user2' AND c.id = 'city1' CREATE (u)-[e:LivesIn]->(c);
MATCH (u:User), (c:City) WHERE u.id = 'user1' AND c.id = 'city2' CREATE (u)-[e:LivesIn]->(c);
MATCH (u)-[e]->(c) RETURN *;
```
Kuzu should complain about the relation `user2 to city1` and `user1 to city2` as it violates the constraint.

royi-luo:
# Description
Add a check that throws an exception if an insertion is performed that would violate a rel multiplicity constraint. This code will need to be slightly reworked if in the future we wish to support constraints in the form `degree< N` .

Fixes #4334

# Contributor agreement

### DIFF
--- a/src/include/storage/store/rel_table.h
+++ b/src/include/storage/store/rel_table.h
@@ -121,7 +121,7 @@ struct RelTableDeleteState final : TableDeleteState {
 
 class RelTable final : public Table {
 public:
-    using rel_table_entry_exception_t =
+    using rel_multiplicity_constraint_throw_func_t =
         std::function<void(const std::string&, common::offset_t, common::RelDataDirection)>;
 
     RelTable(catalog::RelTableCatalogEntry* relTableEntry, const StorageManager* storageManager,
@@ -148,7 +148,7 @@ public:
         common::RelDataDirection direction, common::ValueVector* srcNodeIDVector) const;
     void throwIfNodeHasRels(transaction::Transaction* transaction,
         common::RelDataDirection direction, common::ValueVector* srcNodeIDVector,
-        const rel_table_entry_exception_t& throwFunc) const;
+        const rel_multiplicity_constraint_throw_func_t& throwFunc) const;
 
     void addColumn(transaction::Transaction* transaction,
         TableAddColumnState& addColumnState) override;
--- a/src/include/storage/store/rel_table.h
+++ b/src/include/storage/store/rel_table.h
@@ -121,6 +121,9 @@ struct RelTableDeleteState final : TableDeleteState {
 
 class RelTable final : public Table {
 public:
+    using rel_table_entry_exception_t =
+        std::function<void(const std::string&, common::offset_t, common::RelDataDirection)>;
+
     RelTable(catalog::RelTableCatalogEntry* relTableEntry, const StorageManager* storageManager,
         MemoryManager* memoryManager, common::Deserializer* deSer = nullptr);
 
@@ -141,8 +144,11 @@ public:
 
     void detachDelete(transaction::Transaction* transaction, common::RelDataDirection direction,
         RelTableDeleteState* deleteState);
-    void checkIfNodeHasRels(transaction::Transaction* transaction,
+    bool checkIfNodeHasRels(transaction::Transaction* transaction,
         common::RelDataDirection direction, common::ValueVector* srcNodeIDVector) const;
+    void throwIfNodeHasRels(transaction::Transaction* transaction,
+        common::RelDataDirection direction, common::ValueVector* srcNodeIDVector,
+        const rel_table_entry_exception_t& throwFunc) const;
 
     void addColumn(transaction::Transaction* transaction,
         TableAddColumnState& addColumnState) override;
@@ -198,6 +204,9 @@ private:
     void detachDeleteForCSRRels(transaction::Transaction* transaction,
         const RelTableData* tableData, const RelTableData* reverseTableData,
         RelTableScanState* relDataReadState, RelTableDeleteState* deleteState);
+
+    void checkRelMultiplicityConstraint(transaction::Transaction* transaction,
+        const TableInsertState& state) const;
 
 private:
     common::table_id_t fromNodeTableID;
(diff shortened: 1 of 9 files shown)


src/processor/operator/persistent/delete_executor.cpp
```diff
@@ -23,11 +24,17 @@ void NodeTableDeleteInfo::init(const ResultSet& resultSet) {
 
 void NodeTableDeleteInfo::deleteFromRelTable(Transaction* transaction,
     ValueVector* nodeIDVector) const {
+    const auto deleteFunc = [](const std::string& tableName, offset_t nodeOffset,
```
royi-luo, line 27:
```suggestion
    const auto throwFunc = [](const std::string& tableName, offset_t nodeOffset,
```

src/storage/store/rel_table.cpp
```diff
@@ -166,7 +168,30 @@ bool RelTable::scanInternal(Transaction* transaction, TableScanState& scanState)
...
+            throwFunc);
+    }
+    if (bwdRelTableData->getMultiplicity() == common::RelMultiplicity::ONE) {
+        throwIfNodeHasRels(transaction, common::RelDataDirection::BWD, &insertState.dstNodeIDVector,
+            throwFunc);
+    }
+}
+
 void RelTable::insert(Transaction* transaction, TableInsertState& insertState) {
+    checkRelMultiplicityConstraint(transaction, insertState);
```
royi-luo, line 193: No need to worry about this for updates right? This should be the case as long as updates cannot modify the source/dest nodes of an edge.

src/include/storage/store/rel_table.h
```diff
@@ -121,6 +121,9 @@ struct RelTableDeleteState final : TableDeleteState {
 
 class RelTable final : public Table {
 public:
+    using rel_table_entry_exception_t =
```
ray6080, line 124: `rel_table_entry_exception_t` -> `rel_multiplicity_constraint_throw_func_t`.

src/storage/store/rel_table.cpp
ray6080: yep.

src/storage/store/rel_table.cpp
```diff
@@ -166,7 +168,30 @@ bool RelTable::scanInternal(Transaction* transaction, TableScanState& scanState)
...
+    const TableInsertState& state) const {
+    const auto& insertState = state.constCast<RelTableInsertState>();
+    KU_ASSERT(insertState.srcNodeIDVector.state->getSelVector().getSelSize() == 1 &&
+              insertState.dstNodeIDVector.state->getSelVector().getSelSize() == 1);
+
+    const auto throwFunc = [](const std::string& tableName, offset_t nodeOffset,
+                               RelDataDirection direction) {
+        throw RuntimeException(ExceptionMessage::violateRelMultiplicityConstraint(tableName,
+            std::to_string(nodeOffset), RelDataDirectionUtils::relDirectionToString(direction)));
+    };
```
ray6080, line 181: I wonder how L177-L181 differs from defining a static function and passed as lambda into `throwIfNodeHasRels`. Will that trigger copying?

ray6080 (approved):

src/storage/store/rel_table.cpp
royi-luo: In this case I think there's a good chance that copying will trigger. It's possible to avoid the copy while keeping the lambda but it's probably best to keep it simple and make it a static function.

src/include/storage/store/rel_table.h
royi-luo: updated

src/storage/store/rel_table.cpp
royi-luo: updated

Issue #3906, opened 2024-07-23
Bug: None field may truncate normal comparison expression


Aurora-yzj:
### Kùzu version

v0.4.2

### What operating system are you using?

Ubuntu 22.04

### What happened?

Hi! I discovered a bug that the None field in the result could truncate normal comparison expressions. Reproduce steps are as follows.
1. Initialize the database.
```cypher
CREATE NODE TABLE L1 (k5 BOOLEAN, k6 STRING, k7 BOOLEAN, k8 INT64, k9 INT64, id INT64, PRIMARY KEY (id));
CREATE NODE TABLE L2 (k10 STRING, k11 BOOLEAN, k12 STRING, k13 STRING, k14 BOOLEAN, k15 INT64, id INT64, PRIMARY KEY (id));
CREATE NODE TABLE L3 (k16 INT64, k17 BOOLEAN, k18 BOOLEAN, k19 BOOLEAN, k20 INT64, id INT64, PRIMARY KEY (id));
CREATE REL TABLE T3 (FROM L3 TO L1, k68 INT64, k69 INT64, k70 STRING, k71 BOOLEAN, k72 STRING, k73 INT64, k74 BOOLEAN, id INT64);
CREATE REL TABLE T4 (FROM L1 TO L2, k75 BOOLEAN, k76 INT64, k77 STRING, k78 INT64, k79 INT64, k80 INT64, k81 INT64, id INT64);
CREATE (n0 :L3{k17 : false, k16 : 8993382861735600205, k19 : true, k18 : true, k20 : 627801758695764003, id : 1});
CREATE (n0 :L1{k5 : true, k6 : "0sE0v", id : 3, k7 : true, k8 : 4583869781704786832, k9 : -2450552654359393402});
CREATE (n0 :L3{k17 : true, k16 : -6767407109008028552, k19 : true, k18 : true, k20 : 3222522150000917167, id : 7});
CREATE (n0 :L2{k11 : true, k10 : "hxS0uD", id : 8, k13 : "GW6VMHfM", k12 : "hSWWBpac0", k15 : -8129440654474488646, k14 : true});
MATCH (n0 {id : 3}), (n1 {id : 8}) CREATE(n0)-[r :T4{k80 : 1623934273299180333, k81 : 2691253309671402116, k75 : false, k77 : "OdcmX8O", k76 : 8922449714764930060, id : 12, k79 : 2358311490140127332, k78 : -8031617625991434006}]->(n1);
[...]

PR #4407, opened 2024-10-22
Always update null mask in vector comparison functions

royi-luo:
# Description

We do not reset the null mask of the output vector in function evaluators. This means that in some cases the null result of a previously-evaluated chunk will affect the result of a currently-evaluate chunk. This is fixed by always updating the null mask in binary function executors.

Fixes https://github.com/kuzudb/kuzu/issues/3906

# Contributor agreement

### DIFF
--- a/src/include/function/binary_function_executor.h
+++ b/src/include/function/binary_function_executor.h
@@ -110,6 +110,7 @@ struct BinaryFunctionExecutor {
         if (left.isNull(lPos)) {
             result.setAllNull();
         } else if (right.hasNoNullsGuarantee()) {
+            result.setAllNonNull();
             rightSelVector.forEach([&](auto i) {
                 executeOnValue<LEFT_TYPE, RIGHT_TYPE, RESULT_TYPE, FUNC, OP_WRAPPER>(left, right,
                     result, lPos, i, i, dataPtr);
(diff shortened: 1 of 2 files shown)


src/include/function/binary_function_executor.h
```diff
@@ -110,6 +110,7 @@ struct BinaryFunctionExecutor {
         if (left.isNull(lPos)) {
             result.setAllNull();
         } else if (right.hasNoNullsGuarantee()) {
+            result.setAllNonNull();
```
royi-luo, line 113: Should the null mask be updated here or should be reset it higher up (e.g. in `FunctionExpressionEvaluator::evaluate()` or in the operator's `getNextTuplesInternal()` function)?

andyfengHKU: Was this causing bug? The `executeOnValue` never read from `result nullMask` but instead always read from `left & right nullMask` and write to `result nullMask`. So the previous implementation should be safe.

royi-luo: The problem was that this branch doesn't modify the result nullmask so something like below can happen:
1. Perform vector comparison, result at pos[0] is null, result nullmask correctly updated so that pos[0] is null
2. Perform vector comparison, result should have no nulls (children are both non-null). We reuse the result nullmask from step 1 but never update it so pos[0] is still reported as null

andyfengHKU: sounds good. Then I would expect test cases to reproduce this bug.

andyfengHKU (approved):

src/include/function/binary_function_executor.h
royi-luo: ah right thanks for the reminder I'll add them to this PR

royi-luo: added test

PR #4415, opened 2024-10-24
Properly enable progress bar

royi-luo:
# Description

Undo changes in https://github.com/kuzudb/kuzu/pull/4115 instead of setting `ENABLE_PROGRESS_BAR` in `ClientConfig` to true so that progress bar is only enabled by default in the shell (and not in tests, other APIs, etc.`

# Contributor agreement


andyfengHKU (approved):
Good to go！

PR #4417, opened 2024-10-24
Fix bugs in GDS vertex property scanning

benjaminwinger:
Based on #4416

I fixed some edge cases in the NodeGroup::scan variant used by GDS vertex property scanning, and also fixed handling scans that run across node group boundaries.

I renamed `NodeGroupScanState::numScannedRows` to `nextRowToScan` since they are functionally equivalent and nothing was ever using `numScannedRows` as a size, just as an indication of the next row to scan. There was also a field called `nextRowToScan` already in `CSRNodeGroupScanState`, so that was just removed and the other field is now used in its place.

### DIFF
--- a/src/graph/on_disk_graph.cpp
+++ b/src/graph/on_disk_graph.cpp
@@ -18,6 +18,7 @@
 #include "processor/expression_mapper.h"
 #include "storage/buffer_manager/memory_manager.h"
 #include "storage/storage_manager.h"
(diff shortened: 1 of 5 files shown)


benjaminwinger:
Do we not have CI for PRs which don't target the master branch? This was passing all tests locally (notably including the two tests which were failing for me initially on #4416).

PR #4418, opened 2024-10-25
Removing PathSemiMasker unnecessary mask node

ted-wq-x:
# Description

As the title suggest.

# Contributor agreement

### DIFF
--- a/src/include/planner/operator/sip/logical_semi_masker.h
+++ b/src/include/planner/operator/sip/logical_semi_masker.h
@@ -1,5 +1,6 @@
 #pragma once
 
+#include "common/enums/extend_direction.h"
 #include "common/exception/runtime.h"
 #include "planner/operator/logical_operator.h"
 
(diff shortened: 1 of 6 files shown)


src/include/planner/operator/sip/logical_semi_masker.h
```diff
@@ -79,6 +85,8 @@ class LogicalSemiMasker : public LogicalOperator {
     std::vector<common::table_id_t> nodeTableIDs;
     // Operators accepting semi masker
     std::vector<LogicalOperator*> targetOps;
+    // for path semi masker
```
andyfengHKU, line 88: `For path semi masker marking src or dst nodes`

src/include/planner/operator/sip/logical_semi_masker.h
```diff
@@ -62,14 +64,18 @@ class LogicalSemiMasker : public LogicalOperator {
     void addTarget(LogicalOperator* op) { targetOps.push_back(op); }
     std::vector<LogicalOperator*> getTargetOperators() const { return targetOps; }
 
+    void setDirection(common::ExtendDirection extendDirection) { direction = extendDirection; }
```
andyfengHKU, line 67: Better to also assert keyType is `PATH` when setting direction

andyfengHKU (approved):
Make sense to me. Thanks!

ted-wq-x:
Done

PR #4419, opened 2024-10-25
Gds no path track optimization

andyfengHKU:
# Description

This PR allows GDS framework to skip writing path to factorized table if not needed in later query processing. The optimization logic is written in projection_push_down_optimizer.

This PR contains a major refactor that removes some `SINGLE_SP_LENGTH` & `ALL_SP_LENGTH` functions. The previous organization of these functions is based on output columns. This is not ideal because we may have different combinations, e.g. write dst only, dst & length, dst & path, etc.

The new design keeps 5 functions
- SINGLE_SP_DESTINATIONS
- SINGLE_SP_PATHS
- ALL_SP_DESTINATIONS
- ALL_SP_PATHS
- VAR_LENGTH_PATHS

3 XX_PATHS functions are trivial to understand because they use `BFSGraph` to keep track path. `SINGLE_SP_DESTINATIONS` & `ALL_SP_DESTINATIONS` are able to track intermediate result with a more performant data structure so we create a separate function for them. 

Technically this is also doable for `VAR_LENGTH` but since it won't work with `trail` or `acyclic` semantic, we fall back to the easy solution to always track with `BFSGraph`

Fixes # (issue)

# Contributor agreement

### DIFF
--- a/src/include/function/gds/output_writer.h
+++ b/src/include/function/gds/output_writer.h
@@ -33,18 +33,18 @@ public:
 
 struct SPOutputs : public RJOutputs {
 public:
-    SPOutputs(std::unordered_map<common::table_id_t, uint64_t> nodeTableIDAndNumNodes,
-        common::nodeID_t sourceNodeID, storage::MemoryManager* mm);
+    SPOutputs(common::table_id_map_t<common::offset_t> numNodesMap, common::nodeID_t sourceNodeID,
+        storage::MemoryManager* mm);
 
 public:
+    // TODO: I don't think
     std::shared_ptr<PathLengths> pathLengths;
 };
 
 struct PathsOutputs : public SPOutputs {
-    PathsOutputs(std::unordered_map<common::table_id_t, uint64_t> nodeTableIDAndNumNodes,
+    PathsOutputs(common::table_id_map_t<common::offset_t> numNodesMap,
         common::nodeID_t sourceNodeID, storage::MemoryManager* mm = nullptr)
-        : SPOutputs(nodeTableIDAndNumNodes, sourceNodeID, mm),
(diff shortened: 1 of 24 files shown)


src/include/function/gds/output_writer.h
```diff
@@ -108,6 +114,8 @@ struct PathsOutputWriterInfo {
     // Direction
     bool extendFromSource = false;
     bool writeEdgeDirection = false;
+
```
semihsalihoglu-uw, line 117: Remove empty line maybe.

src/optimizer/projection_push_down_optimizer.cpp
```diff
@@ -44,18 +48,34 @@ void ProjectionPushDownOptimizer::visitOperator(LogicalOperator* op) {
...
+            }
         }
         return;
     }
-
-    if (pathPropertyProbe.getChild(0)->getOperatorType() !=
-        planner::LogicalOperatorType::RECURSIVE_EXTEND) {
-        return;
-    }
+    KU_ASSERT(child->getOperatorType() == LogicalOperatorType::RECURSIVE_EXTEND);
```
semihsalihoglu-uw, line 78: Will we keep the RECURSIVE_EXTEND operator after we deprecate?

semihsalihoglu-uw (approved):

src/include/function/gds/output_writer.h
andyfengHKU: Done

src/optimizer/projection_push_down_optimizer.cpp
andyfengHKU: Yes

PR #4420, opened 2024-10-25
Change uncommitted local node offsets to be consecutive with committed ones

ray6080:
# Description

Change uncommitted transaction local node offsets to start with the max of committed node offset, instead of `MAX_NUM_ROWS_IN_TABLE`. The aim of this change is to simplify the mask on node tables for both committed and uncommitted data. #4421 relies on this change.

The key idea is to keep a map of min uncommitted node offsets (i.e., num of total rows) when the transaction starts. This may not be scalable for very small transactions when the database contains a lot of tables.

## TODO
- [ ] Change uncommitted local rel offsets to start with the max of committed rel offset.
- [ ] Apply semi mask to uncommitted transaction local nodes.

### DIFF
--- a/src/processor/operator/scan/offset_scan_node_table.cpp
+++ b/src/processor/operator/scan/offset_scan_node_table.cpp
@@ -34,10 +34,10 @@ bool OffsetScanNodeTable::getNextTuplesInternal(ExecutionContext* context) {
     auto nodeID = nodeIDVector->getValue<nodeID_t>(0);
     KU_ASSERT(tableIDToNodeInfo.contains(nodeID.tableID));
     auto& nodeInfo = tableIDToNodeInfo.at(nodeID.tableID);
-    if (nodeID.offset >= StorageConstants::MAX_NUM_ROWS_IN_TABLE) {
+    if (transaction->isUnCommitted(nodeID.tableID, nodeID.offset)) {
         nodeInfo.localScanState->source = TableScanSource::UNCOMMITTED;
-        nodeInfo.localScanState->nodeGroupIdx =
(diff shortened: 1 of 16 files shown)


src/processor/operator/scan/offset_scan_node_table.cpp
```diff
@@ -34,10 +34,10 @@ bool OffsetScanNodeTable::getNextTuplesInternal(ExecutionContext* context) {
     auto nodeID = nodeIDVector->getValue<nodeID_t>(0);
     KU_ASSERT(tableIDToNodeInfo.contains(nodeID.tableID));
     auto& nodeInfo = tableIDToNodeInfo.at(nodeID.tableID);
-    if (nodeID.offset >= StorageConstants::MAX_NUM_ROWS_IN_TABLE) {
+    if (transaction->isUnCommitted(nodeID.tableID, nodeID.offset)) {
         nodeInfo.localScanState->source = TableScanSource::UNCOMMITTED;
-        nodeInfo.localScanState->nodeGroupIdx =
-            StorageUtils::getNodeGroupIdx(nodeID.offset - StorageConstants::MAX_NUM_ROWS_IN_TABLE);
+        nodeInfo.localScanState->nodeGroupIdx = StorageUtils::getNodeGroupIdx(
```
andyfengHKU, line 39: Is this a false positive alert?

andyfengHKU (commented):
LGTM

src/processor/operator/scan/offset_scan_node_table.cpp
ray6080: Can u say a bit more here? What do you mean "alert"?

andyfengHKU (approved):

PR #4424, opened 2024-10-25
Store just one OnDiskGraph scan state for both fwd and backward

benjaminwinger:
Only one is used at a time, and we should always know which at the point that the state is created.

This halves the setup cost of preparing OnDiskGraph scans, even when scanning in both directions (since in gds_utils.cpp where the tasks are created we're creating a task for each direction anyway).

### DIFF
--- a/src/function/gds/gds_utils.cpp
+++ b/src/function/gds/gds_utils.cpp
@@ -1,6 +1,8 @@
 #include "function/gds/gds_utils.h"
 
+#include "common/enums/extend_direction.h"
 #include "common/task_system/task_scheduler.h"
+#include "common/types/types.h"
 #include "function/gds/gds_frontier.h"
 #include "function/gds/gds_task.h"
 #include "function/gds/rec_joins.h"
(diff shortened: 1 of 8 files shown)


src/include/common/task_system/task.h
```diff
@@ -72,6 +72,12 @@ class Task {
         return exceptionsPtr;
     }
 
+    // Resets a task so that it can be re-used and run again
+    virtual void reset() {
```
andyfengHKU, line 76: Normally I would avoid using `reset` as a function name because there is also `smart_pointer->reset()` which might cause confusion. Consider using `resetState` or some other naming.

src/function/gds/gds_utils.cpp
```diff
@@ -30,39 +24,43 @@ static void scheduleFrontierTask(table_id_t relTableID, graph::Graph* graph,
...
     clientContext->getTaskScheduler()->scheduleTaskAndWaitOrError(task, context,
         true /* launchNewWorkerThread */);
 }
 
 void GDSUtils::runFrontiersUntilConvergence(processor::ExecutionContext* context,
     RJCompState& rjCompState, graph::Graph* graph, ExtendDirection extendDirection,
     uint64_t maxIters) {
+    auto maxThreads =
+        context->clientContext->getCurrentSetting(main::ThreadsSetting::name).getValue<uint64_t>();
+    auto info = FrontierTaskInfo(0, graph, extendDirection, *rjCompState.edgeCompute);
```
andyfengHKU, line 37: I would use INVALID_TABLE_ID instead of 0 if it means un-initialized field.

andyfengHKU (approved):

PR #4428, opened 2024-10-28
Add limit push down optimization

ted-wq-x:
# Description

As the title suggests.

# Contributor agreement

### DIFF
--- a/src/include/optimizer/limit_push_down_optimizer.h
+++ b/src/include/optimizer/limit_push_down_optimizer.h
@@ -1,0 +1,44 @@
+#pragma once
+
+#include "planner/operator/logical_limit.h"
+#include "planner/operator/logical_plan.h"
+#include "planner/operator/sip/logical_semi_masker.h"
+
+namespace kuzu {
+namespace main {
+class ClientContext;
+}
+namespace optimizer {
+
+class LimitPushDownOptimizer {
+public:
+    explicit LimitPushDownOptimizer(main::ClientContext* context)
+        : limitOperator{nullptr}, context{context} {}
+
+    void rewrite(planner::LogicalPlan* plan);
+
+private:
+    std::shared_ptr<planner::LogicalOperator> visitOperator(
+        std::shared_ptr<planner::LogicalOperator> op);
+
+    std::shared_ptr<planner::LogicalOperator> finishPushDown(
+        std::shared_ptr<planner::LogicalOperator> op);
+
+    static planner::LogicalSemiMasker* findLogicalSemiMasker(
+        const std::shared_ptr<planner::LogicalOperator>& op) {
+        if (op->getOperatorType() == planner::LogicalOperatorType::SEMI_MASKER) {
+            return &op->cast<planner::LogicalSemiMasker>();
+        }
+        if (op->getNumChildren() != 1) {
+            return nullptr;
+        }
+        return findLogicalSemiMasker(op->getChild(0));
+    }
+
+private:
+    planner::LogicalLimit* limitOperator;
+    main::ClientContext* context;
+};
+
+} // namespace optimizer
+} // namespace kuzu
--- a/src/include/processor/operator/aggregate/hash_aggregate.h
+++ b/src/include/processor/operator/aggregate/hash_aggregate.h
@@ -12,7 +12,8 @@ class HashAggregateSharedState final : public BaseAggregateSharedState {
 public:
     explicit HashAggregateSharedState(
         const std::vector<function::AggregateFunction>& aggregateFunctions)
-        : BaseAggregateSharedState{aggregateFunctions} {}
+        : BaseAggregateSharedState{aggregateFunctions}, limitCounter{0},
+          limitNumber{common::INVALID_LIMIT} {}
 
     void appendAggregateHashTable(std::unique_ptr<AggregateHashTable> aggregateHashTable);
 
@@ -28,9 +29,22 @@ public:
 
     uint64_t getCurrentOffset() const { return currentOffset; }
 
+    // return whether limitNumber is exceeded
+    bool increaseAndCheckLimitCount(uint64_t num) {
+        if (limitNumber == common::INVALID_LIMIT) {
+            return false;
+        } else {
+            return limitCounter.fetch_add(num) >= limitNumber;
+        }
+    }
+
+    void setLimitNumber(uint64_t num) { limitNumber = num; }
+
 private:
     std::vector<std::unique_ptr<AggregateHashTable>> localAggregateHashTables;
     std::unique_ptr<AggregateHashTable> globalAggregateHashTable;
+    std::atomic_uint64_t limitCounter;
+    uint64_t limitNumber;
 };
 
 struct HashAggregateInfo {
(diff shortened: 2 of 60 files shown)


ray6080:
Can we add some benchmark queries related to this optimization into our benchmark query set, which is on top of ldbc100?

src/optimizer/limit_push_down_optimizer.cpp
```diff
@@ -0,0 +1,136 @@
...
+            op->setChild(i, optimizer.visitOperator(op->getChild(i)));
+        }
+        return op;
+    }
+    case LogicalOperatorType::HASH_JOIN: {
+        if (limitOperator == nullptr) {
+            return op;
+        }
+        auto& hashJoin = op->cast<LogicalHashJoin>();
+        // Push down limit to left side
```
andyfengHKU, line 92: Instead of saying left side, better say probe side.

src/optimizer/limit_push_down_optimizer.cpp
```diff
@@ -0,0 +1,136 @@
...
+        if (hashJoin.getJoinType() == JoinType::LEFT) {
+            // Add a limit to the probe side,and keep the original limit as well.
+            auto optimizer = LimitPushDownOptimizer(context);
+            optimizer.limitOperator = limitOperator;
+            auto newChild = optimizer.visitOperator(op->getChild(0));
+            op->setChild(0, newChild);
+        } else if (hashJoin.getJoinType() == JoinType::INNER) {
+            LogicalScanNodeTable* scanNode = isSimpleScanNode(hashJoin.getChild(0));
+            // If probe side is ScanNode and build side has semiMasker, we can push down limit to
```
andyfengHKU, line 101: I don't fully understand the reasoning here. If probe side is a simple ScanNode, how can we guarantee the limit we push to the build side will limit to a result that can be matched by probe side?

andyfengHKU (commented):
I just had a question regarding how can we push limit to an InnerHashJoin. Since there no guarantee that a tuple on probe/build side will find at least one match. Pushing down limit doesn't sound safe to me.

src/optimizer/limit_push_down_optimizer.cpp
```diff
@@ -0,0 +1,136 @@
...
+    }
+    case LogicalOperatorType::HASH_JOIN: {
+        if (limitOperator == nullptr) {
+            return op;
+        }
+        auto& hashJoin = op->cast<LogicalHashJoin>();
+        // Push down limit to left side
+        if (hashJoin.getJoinType() == JoinType::LEFT) {
+            // Add a limit to the probe side,and keep the original limit as well.
+            auto optimizer = LimitPushDownOptimizer(context);
```
andyfengHKU, line 95: Is there a specific benefit by pushing LIMIT to probe side? Hash join probe is a pipelined operator instead of a  sink operator, so the improvement is probably not significant in my opinion?

ted-wq-x:
I understand that in the current design, if a node is associated through an inner join and there is no filtering on this node, then it can be considered that this node must exist. In this case, the limit can be pushed down.

src/optimizer/limit_push_down_optimizer.cpp
ted-wq-x: The execution plan on the probe side may also be split into multiple stages, so after pushing down, the limit can be pushed down to the downstream stage.

src/include/processor/operator/aggregate/hash_aggregate.h
```diff
@@ -28,9 +29,22 @@ class HashAggregateSharedState final : public BaseAggregateSharedState {
 
     uint64_t getCurrentOffset() const { return currentOffset; }
 
+    // return whether limitNumber is exceeded
+    bool increaseAndCheckLimitCount(uint64_t num) {
```
ray6080, line 33: Can we move this to cpp file? I don't feel this can be inlined.

src/include/optimizer/limit_push_down_optimizer.h
```diff
@@ -0,0 +1,44 @@
...
+    void rewrite(planner::LogicalPlan* plan);
+
+private:
+    std::shared_ptr<planner::LogicalOperator> visitOperator(
+        std::shared_ptr<planner::LogicalOperator> op);
+
+    std::shared_ptr<planner::LogicalOperator> finishPushDown(
+        std::shared_ptr<planner::LogicalOperator> op);
+
+    static planner::LogicalSemiMasker* findLogicalSemiMasker(
```
ray6080, line 27: Move this to cpp file.

src/processor/operator/aggregate/hash_aggregate.cpp
```diff
@@ -117,6 +122,9 @@ void HashAggregate::executeInternal(ExecutionContext* context) {
     while (children[0]->getNextTuple(context)) {
         const auto numAppendedFlatTuples = localState.append(aggInputs, resultSet->multiplicity);
         metrics->numOutputTuple.increase(numAppendedFlatTuples);
+        if (sharedState->increaseAndCheckLimitCount(numAppendedFlatTuples)) {
```
ray6080, line 125: Better add a comment here to explain that the limit will only be applied if this is distinct hash aggregate. It's not to figure that out inside the operator without checking the limit push down optimizer.

src/optimizer/limit_push_down_optimizer.cpp
```diff
@@ -0,0 +1,136 @@
...
+                    auto optimizer = LimitPushDownOptimizer(context);
+                    optimizer.limitOperator = limitOperator;
+                    auto newChild = optimizer.visitOperator(op->getChild(1));
+                    op->setChild(1, newChild);
+                    limitOperator = nullptr;
+                    return op;
+                }
+            }
+            scanNode = isSimpleScanNode(hashJoin.getChild(1));
+            // If build side is ScanNode ,we can push down limit to probe side
```
ray6080, line 115: I'm confused on the assumption here. Will this work if the query is like `match (a:A), (b:B) where a.prop1 = b.prop2 return * limit 10`?

src/optimizer/limit_push_down_optimizer.cpp
```diff
@@ -0,0 +1,136 @@
...
+            // Add a limit to the probe side,and keep the original limit as well.
+            auto optimizer = LimitPushDownOptimizer(context);
+            optimizer.limitOperator = limitOperator;
+            auto newChild = optimizer.visitOperator(op->getChild(0));
+            op->setChild(0, newChild);
+        } else if (hashJoin.getJoinType() == JoinType::INNER) {
+            LogicalScanNodeTable* scanNode = isSimpleScanNode(hashJoin.getChild(0));
+            // If probe side is ScanNode and build side has semiMasker, we can push down limit to
+            // build side
+            if (scanNode) {
```
ray6080, line 103: This seems like having strong implicit assumptions on how plan should be generated. Also, should you check if the semiMasker is associated with the current probe side? I'm not confident if this can work under all cases actually. I'd suggest to revert the push down into hash join part in this PR. We can do it in a separate one with more test cases and more careful designs if necessary.

ray6080 (approved):
Hi @ted-wq-x , I'm going to merge this soon as we need the optimization on GDS side, and I will clean up comments in a follow-up PR.

PR #4429, opened 2024-10-28
Fix cost model for Extend

ray6080:
# Description

Fix the bug that `Extend` is not keeping its child's cost.
Also, update the cost of `Extend` itself to be aligned with num of rels scanned, instead of num of bound nodes, which is not reflecting where the main cost comes from. 

The change can fix the plan for the following query to start scan from `c1`, instead of `c2`.
```
match (c1:person)-[e1:knows]-(c2:person)-[e2:knows]-(c3:person) where c1.ID=30786325638641 return min(c1.ID), min(c2.ID), min(c3.ID);
```

@andyfengHKU let me know if there are other tests I should add for this change.


andyfengHKU (approved):
LGTM

PR #4430, opened 2024-10-28
GDS dst node pruning optimization

andyfengHKU:
# Description

This PR pass a semi masker to GDS if the number of destination nodes is small. So we end up writing a much smaller intermediate result.

A quick sanity check benchmark. On LDBC10 person knows person
```
MATCH (a:Person)-[e:Knows* 1..4]->(b:Person) WHERE a.ID = 933 AND b.ID = 15393162795677 RETURN e;
```
Query time improves from 400ms -> 18ms

Fixes # (issue)

# Contributor agreement

### DIFF
--- a/src/function/gds/output_writer.cpp
+++ b/src/function/gds/output_writer.cpp
@@ -20,8 +20,9 @@ void PathsOutputs::beginWritingOutputsForDstNodesInTable(common::table_id_t tabl
     bfsGraph.pinNodeTable(tableID);
 }
 
-RJOutputWriter::RJOutputWriter(main::ClientContext* context, RJOutputs* rjOutputs)
-    : context{context}, rjOutputs{rjOutputs} {
+RJOutputWriter::RJOutputWriter(main::ClientContext* context, RJOutputs* rjOutputs,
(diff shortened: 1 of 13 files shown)


src/function/gds/output_writer.cpp
```diff
@@ -36,9 +37,24 @@ std::unique_ptr<common::ValueVector> RJOutputWriter::createVector(const LogicalT
     return vector;
 }
 
+void RJOutputWriter::beginWritingForDstNodesInTable(common::table_id_t tableID) {
+    rjOutputs->beginWritingOutputsForDstNodesInTable(tableID);
+    if (outputNodeMask != nullptr) {
```
semihsalihoglu-uw, line 42: Is it easy/possible to get rid of outputNodeMask != nullptr and always ensure its not null (but a default one that always returns true?)? That would make the code cleaner and safer I think.

semihsalihoglu-uw (approved):

src/function/gds/output_writer.cpp
andyfengHKU: I think this would be solved once #4356 is merged.

PR #4443, opened 2024-10-29
set canParallel for clear_warnings() to false

royi-luo:
# Description

Parallel calls to `WarningContext::clearPopulatedWarnings()` was causing spurious test crashes. Also modified test runner so that tests could be repeatedly run with `--gtest_repeat` (previously since we moved the test statements all test runs after the 1st would run on an empty vector of statements).

# Contributor agreement


ray6080 (approved):
Can u rebase the latest master? It should fix your CI failures.

PR #4446, opened 2024-10-30
Fixed ctrl c not interrupting shell bug

### LINKED ISSUE #4106: BUG: Using `CTRL C` to interrupt a query exits the shell
When running a query and printing query results in the CLI, if `CTRL C` is pressed it exits the shell instead of terminating the query and starting a new line to edit. 

This is due to the way we handle `SIGINT` for the `embedded_shell.cpp` code. 

https://github.com/kuzudb/kuzu/blob/87de7f9128ff9946b95e62bad69bd8d8053a176c/tools/shell/embedded_shell.cpp#L304 https://github.com/kuzudb/kuzu/blob/87de7f9128ff9946b95e62bad69bd8d8053a176c/tools/shell/embedded_shell.cpp#L429-L438

It seems like SIGINT is not being handled by this function or it still terminates afterwards. 

One possibility is switching to using [sigaction](https://www.ibm.com/docs/en/zos/2.4.0?topic=functions-sigaction-examine-change-signal-action) instead of `signal` for more flexibility but a different solution would be needed for Windows. 

Our `CTRL C` handling works in the `linenoise.cpp` file where we change the terminal modes to process the raw input. A similar solution may need to be applied to adjust the terminal mode to correctly send `SIGINT` to our signal handler.
https://github.com/kuzudb/kuzu/blob/87de7f9128ff9946b95e62bad69bd8d8053a176c/tools/shell/linenoise.cpp#L448-L519

MSebanc:
# Description

When ctrl c is pressed during query processing/printing, the shell would exit instead of just terminating the query. Turns out, this was an issue with surrounding `signal` with `KU_ASSERT`. Fixed this and made sure that ctrl c interrupts printing as well

Fixes #4106 

# Contributor agreement


tools/shell/embedded_shell.cpp
```diff
@@ -774,6 +774,9 @@ std::string EmbeddedShell::printJsonExecutionResult(QueryResult& queryResult) co
         printString = jsonDrawingCharacters->ArrayOpen;
     }
     while (queryResult.hasNext()) {
+        if (printInterrupted) {
```
andyfengHKU, line 777: I wonder if this printInterrupted check should be outside the while loop. Also we seem to add this check in multiple print function (e.g. latex, html, ...). I wonder if there should be a refactor so that we can check once and then trigger different type of printing.

andyfengHKU (approved):
LGTM. I think u need to rebase and rerun the CI. All CI failures do not seem to relate to this PR.

tools/shell/embedded_shell.cpp
MSebanc: Yeah, I was thinking about that too. The reason why it is inside the while loop and in so many functions is because if the user sets their printing to print a higher number of tuples (for example they want more rows printing for csv output), if we don't check inside the while loop, they can't terminate the printing by pressing ctrl c. If you have a better idea for how to combat this let me know 👍

andyfengHKU: I see. I don't have a good idea here either. In such case you can go ahead and merge.

PR #4448, opened 2024-10-30
Add shortest path early stop

andyfengHKU:
# Description

See title.

As a sanity check, on live journal
```
MATCH (a:N)-[e*SHORTEST 1..5]->(b:N) WHERE a.id=56174 AND b.id=952528 RETURN b.id, length(e) AS x ORDER BY x;
```
This query drops from 180ms to 80ms.

Fixes # (issue)

# Contributor agreement

### DIFF
--- a/src/function/gds/gds_utils.cpp
+++ b/src/function/gds/gds_utils.cpp
@@ -38,30 +38,35 @@ void GDSUtils::runFrontiersUntilConvergence(processor::ExecutionContext* context
     RJCompState& rjCompState, graph::Graph* graph, ExtendDirection extendDirection,
     uint64_t maxIters) {
     auto frontierPair = rjCompState.frontierPair.get();
+    auto outputNodeMask = rjCompState.outputWriter->getOutputNodeMask();
+    rjCompState.edgeCompute->resetSingleThreadState();
     while (frontierPair->hasActiveNodesForNextLevel() && frontierPair->getNextIter() <= maxIters) {
         frontierPair->beginNewIteration();
+        if (outputNodeMask->enabled() && rjCompState.edgeCompute->terminate(*outputNodeMask)) {
+            break;
+        }
         for (auto& relTableIDInfo : graph->getRelTableIDInfos()) {
             switch (extendDirection) {
             case ExtendDirection::FWD: {
                 rjCompState.beginFrontierComputeBetweenTables(relTableIDInfo.fromNodeTableID,
                     relTableIDInfo.toNodeTableID);
-                scheduleFrontierTask(relTableIDInfo.relTableID, graph, common::ExtendDirection::FWD,
+                scheduleFrontierTask(relTableIDInfo.relTableID, graph, ExtendDirection::FWD,
                     rjCompState, context);
             } break;
             case ExtendDirection::BWD: {
                 rjCompState.beginFrontierComputeBetweenTables(relTableIDInfo.toNodeTableID,
                     relTableIDInfo.fromNodeTableID);
-                scheduleFrontierTask(relTableIDInfo.relTableID, graph, common::ExtendDirection::BWD,
+                scheduleFrontierTask(relTableIDInfo.relTableID, graph, ExtendDirection::BWD,
                     rjCompState, context);
             } break;
             case ExtendDirection::BOTH: {
                 rjCompState.beginFrontierComputeBetweenTables(relTableIDInfo.fromNodeTableID,
                     relTableIDInfo.toNodeTableID);
-                scheduleFrontierTask(relTableIDInfo.relTableID, graph, common::ExtendDirection::FWD,
+                scheduleFrontierTask(relTableIDInfo.relTableID, graph, ExtendDirection::FWD,
                     rjCompState, context);
                 rjCompState.beginFrontierComputeBetweenTables(relTableIDInfo.toNodeTableID,
                     relTableIDInfo.fromNodeTableID);
-                scheduleFrontierTask(relTableIDInfo.relTableID, graph, common::ExtendDirection::BWD,
+                scheduleFrontierTask(relTableIDInfo.relTableID, graph, ExtendDirection::BWD,
                     rjCompState, context);
             } break;
             default:
--- a/src/include/function/gds/gds_frontier.h
+++ b/src/include/function/gds/gds_frontier.h
@@ -6,6 +6,7 @@
 #include "common/types/types.h"
 #include "graph/graph.h"
 #include "storage/buffer_manager/memory_manager.h"
+#include "processor/operator/gds_call_shared_state.h"
 
 namespace kuzu {
 namespace function {
@@ -26,6 +27,12 @@ public:
     // **do not** call setActive. Helper functions in GDSUtils will do that work.
     virtual std::vector<common::nodeID_t> edgeCompute(common::nodeID_t boundNodeID,
         graph::GraphScanState::Chunk& results, bool fwdEdge) = 0;
+
+    virtual void resetSingleThreadState() {}
+
+    virtual bool terminate(processor::NodeOffsetMaskMap&) {
+        return false;
+    }
 
     virtual std::unique_ptr<EdgeCompute> copy() = 0;
 };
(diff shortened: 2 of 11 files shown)


src/processor/operator/gds_call_shared_state.cpp
```diff
@@ -5,6 +5,17 @@ using namespace kuzu::common;
 namespace kuzu {
 namespace processor {
 
+offset_t NodeOffsetMaskMap::getNumMaskedNode() const {
+    if (!enabled_) {
```
semihsalihoglu-uw, line 9: This looks a bit prone to bugs to me because you return 0 for two different reasons. One is if the NodeMask is not enabled second is if it's enabled but no node is marked in the mask. As a general principle outputs of functions should have one interpretation and lack of indicates a problem with the design. I would try to give clear meaning to each possible output of this function, e.g., return MAX_OFFSET or something or just throw an exception if this function should not be called if the mask was disabled.

Also, if for some reason you don't need to differentiate between the above 2 cases, will the code from 12-16 just return 0 if the mask was disabled? If so, you might consider removing the first if clause.

src/function/gds/rec_joins.cpp
```diff
@@ -202,5 +202,22 @@ void RJAlgorithm::exec(processor::ExecutionContext* executionContext) {
     sharedState->mergeLocalTables();
 }
 
+bool SPAlgorithm::terminateFunc(GDSFrontier& frontier,
+    const processor::NodeOffsetMaskMap& maskMap) {
```
semihsalihoglu-uw, line 206: is the maskMap here outputMaskMap? If so, let's rename maskMap -> outputMaskMap.

I think we can remove the first if block here.

src/function/gds/rec_joins.cpp
```diff
@@ -202,5 +202,22 @@ void RJAlgorithm::exec(processor::ExecutionContext* executionContext) {
...
+    if (!maskMap.enabled()) {
+        return false;
+    }
+    for (auto& [tableID, mask] : maskMap.getMaskMap()) {
+        if (frontier.pinTableID(tableID)) {
+            for (auto i = 0u; i < mask->getMaxOffset(); ++i) {
+                if (mask->isMasked(i) && !frontier.isActive(i)) {
```
semihsalihoglu-uw, line 213: This termination condition looks wrong to me. What if there are 2 output nodes vi and vj, so maskMap contains 2 nodes. But vi of them is reached in iteration 1, so will be active in frontier1 and vj in iteration 5, so in frontier 5. So there will never be a frontier in which both vi and vj are in the same frontier, although both of them have been reached. If this is a bug, we should add a test case for this.

src/include/function/gds/gds_frontier.h
```diff
@@ -77,6 +77,7 @@ class FrontierMorsel {
     }
 
 private:
+    // TODO(Xiyang): we probably not
```
semihsalihoglu-uw, line 80: Remove this.

src/function/gds/gds_utils.cpp
```diff
@@ -34,34 +34,42 @@ static void scheduleFrontierTask(table_id_t relTableID, graph::Graph* graph,
         true /* launchNewWorkerThread */);
 }
 
+static constexpr uint64_t SMALL_DST_NODES_NUMBER = 100;
```
semihsalihoglu-uw, line 37: Something like EARLY_TERM_NUM_NODES_THRESHOLD?

src/function/gds/rec_joins.cpp
```diff
@@ -202,5 +202,22 @@ void RJAlgorithm::exec(processor::ExecutionContext* executionContext) {
...
+                    return false;
+                }
+            }
+        }
+    }
+    return true;
```
semihsalihoglu-uw, line 219: return numDstReached == mask.getNumNodes();

src/function/gds/rec_joins.cpp
```diff
@@ -202,5 +202,22 @@ void RJAlgorithm::exec(processor::ExecutionContext* executionContext) {
...
+bool SPAlgorithm::terminateFunc(GDSFrontier& frontier,
+    const processor::NodeOffsetMaskMap& maskMap) {
+    if (!maskMap.enabled()) {
+        return false;
+    }
+    for (auto& [tableID, mask] : maskMap.getMaskMap()) {
+        if (frontier.pinTableID(tableID)) {
+            for (auto i = 0u; i < mask->getMaxOffset(); ++i) {
+                if (mask->isMasked(i) && !frontier.isActive(i)) {
+                    return false;
```
semihsalihoglu-uw, line 214: During initializing SPAlgorithm, you set numDstReached=0, or it's constructed that way.
numDstReached++;

semihsalihoglu-uw (changes requested):

src/include/processor/operator/gds_call_shared_state.h
```diff
@@ -11,6 +11,13 @@ namespace processor {
 
 class NodeOffsetMaskMap {
 public:
+    NodeOffsetMaskMap() : enabled_{false} {}
+
+    void enable() { enabled_ = true; }
```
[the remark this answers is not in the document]
andyfengHKU: Done

src/processor/operator/gds_call_shared_state.cpp
andyfengHKU: Done

src/function/gds/rec_joins.cpp
andyfengHKU: Done

src/include/function/gds/gds_frontier.h
andyfengHKU: Done

src/function/gds/rec_joins.cpp
andyfengHKU: Done

src/function/gds/rec_joins.cpp
andyfengHKU: Done

src/function/gds/rec_joins.cpp
andyfengHKU: I don't think this is a bug. Or at least it's not a bug that can be captured by e2e test. It's just we won't trigger early termination.

src/include/function/gds/rec_joins.h
```diff
@@ -57,19 +58,24 @@ struct RJBindData final : public GDSBindData {
...
+    std::function<bool(GDSFrontier& frontier, const processor::NodeOffsetMaskMap& maskMap)>;
+
 // Wrapper around the data that needs to be stored during the computation of a recursive joins
 // computation from one source. Also contains several initialization functions.
 struct RJCompState {
     std::unique_ptr<function::FrontierPair> frontierPair;
     std::unique_ptr<function::EdgeCompute> edgeCompute;
     std::unique_ptr<RJOutputs> outputs;
     std::unique_ptr<RJOutputWriter> outputWriter;
+    gds_terminate_func_t terminateFunc;
```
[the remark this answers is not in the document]
andyfengHKU: Either is fine but I don't think creating a class is necessary because there is no state. Even if we use a class, it will not be maintained with unique_ptr because there is no inheritance.

src/processor/operator/gds_call_shared_state.cpp
```diff
@@ -5,6 +5,18 @@ using namespace kuzu::common;
 namespace kuzu {
 namespace processor {
 
+offset_t NodeOffsetMaskMap::getNumMaskedNode() const {
+    KU_ASSERT(enabled_);
+    if (!enabled_) {
+        throw InternalException("asd");
```
semihsalihoglu-uw, line 11: What's asd? Let's have a better error message.

src/include/function/gds/gds_frontier.h
semihsalihoglu-uw, line 80: I still see this TODO with an incomplete sentence.

src/processor/operator/gds_call_shared_state.cpp
andyfengHKU: Remove this check

src/function/gds/gds_utils.cpp
```diff
@@ -38,30 +38,37 @@ void GDSUtils::runFrontiersUntilConvergence(processor::ExecutionContext* context
     RJCompState& rjCompState, graph::Graph* graph, ExtendDirection extendDirection,
     uint64_t maxIters) {
     auto frontierPair = rjCompState.frontierPair.get();
+    auto outputNodeMask = rjCompState.outputWriter->getOutputNodeMask();
+    common::offset_t numReachedNodes = 0;
```
semihsalihoglu-uw, line 42: The numReachedNodes is a shortest-paths-specific auxiliary for terminating early. It should be defined as part of the SPAlgorithms' shared state. And the SharedState or something similar should be passed to terminateFunction.

semihsalihoglu-uw (approved):
I'll leave when to do this fix, but I have a comment that I think is important.

PR #4450, opened 2024-10-30
Modify BM so behaviour can be overriden during tests + some robustness improvements

royi-luo:
# Description

Robustness improvements:

- Throw exception before deleting erroneous row after node batch insert failure (delete should only happen with `IGNORE_ERRORS=true`)
- `DiskArrayInternal::WriteIterator::unpin()` resets shadow pages to `INVALID_PAGE_IDX`
- Avoid ownership transfer of `Column` from `NodeTable` to `NodeGroupCheckpointState` during checkpoint so that checkpoint can be interrupted without `NodeTable` having null columns
- `TransactionManager::checkpointNoLock` releases lock for starting transaction if exception is thrown in the middle

There are more fixes to come in a future PR with more tests. For now most of the cases are difficult to test without triggering other bugs.

# Contributor agreement

### DIFF
--- a/src/include/main/database.h
+++ b/src/include/main/database.h
@@ -97,8 +97,6 @@ public:
      */
     KUZU_API static std::unique_ptr<Database> construct(std::string_view databasePath,
         SystemConfig systemConfig = SystemConfig());
-    KUZU_API static std::shared_ptr<Database> constructShared(std::string_view databasePath,
-        SystemConfig systemConfig = SystemConfig());
 
     /**
      * @brief Destructs the database object.
--- a/src/include/main/database.h
+++ b/src/include/main/database.h
@@ -95,13 +95,12 @@ public:
      *        an in-memory database.
      * @param systemConfig System configurations (buffer pool size and max num threads).
      */
-    KUZU_API static std::unique_ptr<Database> construct(std::string_view databasePath,
+    KUZU_API explicit Database(std::string_view databasePath,
         SystemConfig systemConfig = SystemConfig());
-
     /**
      * @brief Destructs the database object.
      */
-    KUZU_API virtual ~Database();
+    KUZU_API ~Database();
 
     // TODO(Ziyi): Instead of exposing a dedicated API for adding a new function, we should consider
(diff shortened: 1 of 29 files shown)


src/include/main/database.h
```diff
@@ -95,12 +95,15 @@ class Database {
      *        an in-memory database.
      * @param systemConfig System configurations (buffer pool size and max num threads).
      */
-    KUZU_API explicit Database(std::string_view databasePath,
+    KUZU_API static std::unique_ptr<Database> construct(std::string_view databasePath,
         SystemConfig systemConfig = SystemConfig());
+    KUZU_API static std::shared_ptr<Database> constructShared(std::string_view databasePath,
```
benjaminwinger, line 100: `std::shared_ptr` has a constructor that takes a `std::unique_ptr`, so I'm not sure if we really need a second constructor just to produce a shared_ptr.

I also don't understand the purpose of separating out `initMembers` or making the constructor private in favour of a static `construct` function. They don't seem like they do anything that couldn't be achieved with a regular constructor.

benjaminwinger (approved):
The fixes generally look good to me.

src/include/main/database.h
royi-luo: Thanks pointing that out, I'll remove the second constructor.

I separated `initMembers` out of the constructor because the vtable for a class isn't initialized until after the constructor for the base class completes execution. I wanted to have `initBufferManager` be overrideable for tests but this is only possible outside of the constructor for `Database`

ray6080 (approved):

src/include/main/database.h
```diff
@@ -95,12 +95,13 @@ class Database {
      *        an in-memory database.
      * @param systemConfig System configurations (buffer pool size and max num threads).
      */
-    KUZU_API explicit Database(std::string_view databasePath,
+    KUZU_API static std::unique_ptr<Database> construct(std::string_view databasePath,
```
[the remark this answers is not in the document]
royi-luo: updated, the public API should be the same now

PR #4451, opened 2024-10-31
Fix PathSingleTableSemiMasker not working as expected

ted-wq-x:
# Description

See title.

# Contributor agreement


andyfengHKU (approved):
LGTM. Can u also add a test case to capture this bug?

ted-wq-x:
Done

PR #4454, opened 2024-11-01
Fix primary key scan not working

ted-wq-x:
# Description

This enables index scan on Scan of node table as child of Extend.

An example query:

`MATCH (p1:person)-[e:knows]->(b:person) WHERE p1.ID=1 RETURN *;`

The optimizer will turn SCAN(p1) into INDEX_SCAN(p1).

# Contributor agreement


ray6080 (approved):
LGTM! Thanks! @andyfengHKU do you want to take a second look?

PR #4455, opened 2024-11-01
Implementing Parallel Computing for WCC Algorithm

WWW0030:
# Description

TODO:
 - [ ] Currently the program runs, but it is returning the wrong result. I suspect that this is due to the initization not initializing the frontier correctly. Taking a look into it
 - [ ] fix convergence checker. As opposed to RJ Algorithms which converge when frontier becomes zero, WCC's frontier will always be the entire nodes table. It needs a different way of checking for convergence, i.e. checking if any componentIDs change in the last iteration (could be done by setting a boolean flag which resets per iteration?) Currently the algorithm runs until max iteration is achieved.

# Contributor agreement


src/function/gds/gds_frontier.cpp
```diff
@@ -41,6 +41,48 @@ bool FrontierMorselDispatcher::getNextRangeMorsel(FrontierMorsel& frontierMorsel
...
+ComponentIDs::ComponentIDs(const common::table_id_map_t<common::offset_t>& numNodesMap_, storage::MemoryManager* mm) {
+    curIter.store(0);
+    for (const auto& [tableID, numNodes] : numNodesMap_) {
+        numNodesMap[tableID] = numNodes;
+        auto memBuffer = mm->allocateBuffer(false, numNodes * sizeof(std::atomic<uint64_t>));
+        std::atomic<uint16_t>* memBufferPtr = 
+            reinterpret_cast<std::atomic<uint16_t>*>(memBuffer.get()->getData());
+
+        // Cast a unique number to each node
+        for (uint64_t i = 0; i < numNodes; ++i) {
```
WWW0030, line 61: I think that with WCC, this should become uint64_t, since each node will start off with an unique ComponentID

src/include/function/gds/par_wcc.h
WWW0030, line 1: Currently not in use, used during PR draft to cook up implementation

src/function/gds/par_wcc.cpp
WWW0030, line 1: Also currently not in use, will delete after everything is working.

it might also be a good idea to refactor/organize the class definitions into different files, currently a bit all over the place

PR #4460, opened 2024-11-04
Add filter pushdown with extend

ray6080:
# Description

This enables index scan on Scan of node table as chid of Extend.

An example query:
```
MATCH (p1:person)-[e:knows]->(b:person) WHERE p1.ID=1 RETURN *;
```
The optimizer will turn `SCAN(p1)` into `INDEX_SCAN(p1)`.


ray6080:
Didn't realize that this is already covered in #4454.

PR #4461, opened 2024-11-04
Gds node predicate push down

andyfengHKU:
# Description

Push down node predicate and semantic checking during DFS backtracking. Now we can handle the following case efficiently (the query timeout on master)

```
CREATE (a)-[]->(b)-[]->(c), (c)-[]->(a)
MATCH (a)-[e*1..30]->(c)
```
Fixes # (issue)

# Contributor agreement


src/function/gds/output_writer.cpp
```diff
@@ -104,99 +107,139 @@ void PathsOutputWriter::write(processor::FactorizedTable& fTable, nodeID_t dstNo
...
+    const std::vector<ParentList*>& path) const {
+    if (next == nullptr) {
+        return false;
+    }
+    auto nextIter = next->getIter();
+    // (1) if this is the first element in the stack (curPath.size() == 1), i.e., we
+    // are enumerating the parents of the destination, then we should switch to
+    // forward-tracking if the next parent has visited the destination at a length
+    // that's greater than or equal to the lower bound of the recursive join. Otherwise,
+    // we'll enumerate paths that are smaller than the lower; OR
```
semihsalihoglu-uw, line 185: Let's be more explicit on this comment (which I know I wrote before): "... Otherwise, we would enumerate paths that are smaller than the lower bound from the start element, so we can stop here.

I also think there is a bug here. The code should be:
```
    if (path.size() == 1) {
        return nextIter >= info.lowerBound;
    }
```
If the path.size() == 1, then you should return true if nextIter >= info.lowerBound and return false otherwise. You are not returning false. This may not strictly be a bug but let's be explicit in the logic of the enumeration.

src/function/gds/output_writer.cpp
```diff
@@ -104,99 +107,139 @@ void PathsOutputWriter::write(processor::FactorizedTable& fTable, nodeID_t dstNo
...
     }
 }
 
-bool PathsOutputWriter::isTrail(const std::vector<ParentList*>& path) const {
-    rel_id_set_t set;
-    for (auto i = 0u; i < path.size(); ++i) {
-        auto edgeID = path[i]->getEdgeID();
-        if (set.contains(edgeID)) {
+bool PathsOutputWriter::isTrail(const std::vector<ParentList*>& path, ParentList* candidate) const {
+    for (auto& element : path) {
```
semihsalihoglu-uw, line 226: I don't think we should check it this way. You are implementing a quadratic algorithm here, looping through the entire path each time you add a new element. Instead, I would maintain an unordered map of nodes (for acyclic) or edges (for trail) along with the currentPath as you construct path and add/remove elements from it each time I add/remove a new node/edge from the path. Then you can do these checks with 1 map call (using a node or edge). We also need a fast path that does not do any map operations for the walk semantics.

You need to benchmark and compare both these implementations to see that using a map is faster than your quadratic/for-loop-based implementation, and then that using map is also faster than the post-enumeration checks (which is what's done prior to this PR). 

This requires some serious work. This won't be quick unfortunately, but needs to be done.

src/function/gds/output_writer.cpp
```diff
@@ -104,99 +107,139 @@ void PathsOutputWriter::write(processor::FactorizedTable& fTable, nodeID_t dstNo
...
     }
     return true;
 }
 
-bool PathsOutputWriter::isAcyclic(const std::vector<ParentList*>& path) const {
-    node_id_set_t set;
-    for (auto i = 0u; i < path.size() - 1; ++i) {
-        auto nodeID = path[i]->getNodeID();
-        if (set.contains(nodeID)) {
+bool PathsOutputWriter::isAcyclic(const std::vector<ParentList*>& path,
```
semihsalihoglu-uw, line 234: Ditto. I would try and benchmark this against an implementation that uses unordered or some other fast map implementations.

src/include/function/gds/output_writer.h
```diff
@@ -113,6 +113,7 @@ struct PathsOutputWriterInfo {
     // Semantic
     common::PathSemantic semantic = common::PathSemantic::WALK;
     // Range
+    //    uint16_t upperBound = 0;
```
semihsalihoglu-uw, line 116: What's this commented line?

src/function/gds/output_writer.cpp
```diff
@@ -104,99 +107,139 @@ void PathsOutputWriter::write(processor::FactorizedTable& fTable, nodeID_t dstNo
...
-            // same iter value as the current parent. That's because the levels/iter need to
-            // decrease by 1 each time we add a new node in the stack.
-            if (top->getNextPtr() != nullptr &&
-                ((curPath.size() == 1 && top->getNextPtr()->getIter() >= info.lowerBound) ||
-                    top->getNextPtr()->getIter() == top->getIter())) {
-                curPath[curPath.size() - 1] = top->getNextPtr();
-                backtracking = false;
+            auto next = getTop(curPath)->getNextPtr();
+            while (true) {
+                if (!canNextReplacePathTop(next, curPath)) {
```
semihsalihoglu-uw, line 118: I believe this code is fine algorithmically but still looks like slow code because of all the functions calls that are being made and the structure of the while loop, continue, and break, seems could be improved. Can you still open an issue to continue optimizing this and assign to me?

semihsalihoglu-uw, line 118: Let's rename this function to isNextParentViable(next, curPath)? The current name sounds grammatically a bit off to me.

src/function/gds/output_writer.cpp
```diff
@@ -104,99 +107,139 @@ void PathsOutputWriter::write(processor::FactorizedTable& fTable, nodeID_t dstNo
...
+                    // No more forward tracking candidates. Switch to backward tracking.
+                    backtracking = true;
+                    break;
+                }
+                if (parent->getIter() == top->getIter() - 1 &&
+                    checkNewPathElement(curPath, parent)) {
+                    // A forward tracking candidate should decrease the iteration by one and also
+                    // pass node predicate checking.
+                    curPath.push_back(parent);
+                    backtracking = false;
```
semihsalihoglu-uw, line 149: It looks like this backtracking = false line is redundant because are you not guaranteed that backtracking is false here?

src/function/gds/output_writer.cpp
```diff
@@ -104,99 +107,139 @@ void PathsOutputWriter::write(processor::FactorizedTable& fTable, nodeID_t dstNo
...
-                curPath.pop_back();
+                backtracking = false;
+                break;
             }
         } else {
-            auto parent = bfsGraph.getInitialParentAndNextPtr(topNodeID);
-            while (parent->getIter() != top->getIter() - 1) {
+            auto top = getTop(curPath);
+            auto parent = bfsGraph.getInitialParentAndNextPtr(top->getNodeID());
+            while (true) {
```
semihsalihoglu-uw, line 138: Isn't this while loop redundant too? This while (true) loop is inside the following broader code structure:
```
while (!curPath.empty()) {
   ...
   if (backtracking) {
       ...
   } else {
      .... // your while (true) loop is here
   }
}
```
So if you don't add the while (true) loop in that else branch, if you remove it, the above while loop will still loop. So let's remove it to simplify the code.

semihsalihoglu-uw (changes requested):
For test, the immediate test that comes to my mind is to have a graph with a 2-cycle (0<->1), 3-cycle (2->3->4->2)  and then test:
(a)-[:k..t]->(b) with all three semantics, for k=[1,..3], t=[k,3]. so 3+2+1=6 possible combinations of k and t, and also 3 semantics, so a total of 18 tests. I would start with that.

src/include/function/gds/output_writer.h
andyfengHKU: Done

src/function/gds/output_writer.cpp
andyfengHKU: Done

src/function/gds/output_writer.cpp
andyfengHKU: Consider the size of path is always small (less than 30). I doubt using a map will be noticeable faster than for loop. But I have opened an issue to track this. Will take a look in the future.

src/function/gds/output_writer.cpp
andyfengHKU: Ditto

src/function/gds/output_writer.cpp
andyfengHKU: I have written a fast path, i.e. keep the original one when there is no semantic or predicate.

andyfengHKU: Done

src/function/gds/output_writer.cpp
andyfengHKU: I don't think this is the case. This for loop is trying to get the parent of parent. If we remove the while loop directly, we will always access the parent of top. So it's not as simple as removing a while loop. Also the previous version is doing something very similar

```
while (!curPath.empty) {
      if (backtracking) {
      } else {
                auto parent = bfsGraph.getParentListHead(topNodeID);
                while (parent->getIter() != top->getIter() - 1) {
                    parent = parent->getNextPtr();
                }
      }
}
```

So I would not change.

mewim (approved):

mewim (approved):

PR #4470, opened 2024-11-05
Rework limit push down optimizer

andyfengHKU:
# Description

Rework limit push down optimizer so that we don't remove LIMIT operator. Instead we just document skip and limit number and push them into other operators (DISTINCT for now).

Fixes # (issue)

# Contributor agreement


ray6080 (approved):

ray6080:
Can you also add a test case for distinct together with skip and limit?

PR #4471, opened 2024-11-05
Gds limit push down

andyfengHKU:
# Description

This PR adds limit push down to GDS which can be quite expensive if we are generating a large number of path.

Some sanity check on LiveJournal
```
MATCH p=(a)-[e* 2..2]-(b) WHERE a.id=0 RETURN p LIMIT 10;
```
Goes from 680ms -> 280ms

And
```
MATCH p=(a)-[e* 3..3]-(b) WHERE a.id=0 RETURN p LIMIT 10;
```
Goes from more than 2 mins-> 380ms

Fixes # (issue)

# Contributor agreement

### DIFF
--- a/src/function/gds/all_shortest_paths.cpp
+++ b/src/function/gds/all_shortest_paths.cpp
@@ -124,10 +124,10 @@ public:
         processor::NodeOffsetMaskMap* outputNodeMask)
         : DestinationsOutputWriter{context, rjOutputs, outputNodeMask} {}
 
-    void write(FactorizedTable &fTable, nodeID_t dstNodeID, processor::GDSOutputCounter* counter) override {
+    void write(FactorizedTable& fTable, nodeID_t dstNodeID,
+        processor::GDSOutputCounter* counter) override {
         auto outputs = rjOutputs->ptrCast<AllSPDestinationsOutputs>();
-        auto length = outputs->pathLengths->getMaskValueFromCurFrontierFixedMask(
-                dstNodeID.offset);
+        auto length = outputs->pathLengths->getMaskValueFromCurFrontierFixedMask(dstNodeID.offset);
         dstNodeIDVector->setValue<nodeID_t>(0, dstNodeID);
         lengthVector->setValue<uint16_t>(0, length);
         auto multiplicity = outputs->multiplicities.getTargetMultiplicity(dstNodeID.offset);
--- a/src/function/gds/all_shortest_paths.cpp
+++ b/src/function/gds/all_shortest_paths.cpp
@@ -124,19 +124,23 @@ public:
         processor::NodeOffsetMaskMap* outputNodeMask)
         : DestinationsOutputWriter{context, rjOutputs, outputNodeMask} {}
 
-    std::unique_ptr<RJOutputWriter> copy() override {
-        return std::make_unique<AllSPDestinationsOutputWriter>(context, rjOutputs, outputNodeMask);
-    }
-
-protected:
-    void writeMoreAndAppend(processor::FactorizedTable& fTable, common::nodeID_t dstNodeID,
-        uint16_t) const override {
-        auto multiplicity =
-            rjOutputs->ptrCast<AllSPDestinationsOutputs>()->multiplicities.getTargetMultiplicity(
+    void write(FactorizedTable &fTable, nodeID_t dstNodeID, processor::GDSOutputCounter* counter) override {
+        auto outputs = rjOutputs->ptrCast<AllSPDestinationsOutputs>();
+        auto length = outputs->pathLengths->getMaskValueFromCurFrontierFixedMask(
                 dstNodeID.offset);
-        for (uint64_t i = 0; i < multiplicity; ++i) {
+        dstNodeIDVector->setValue<nodeID_t>(0, dstNodeID);
+        lengthVector->setValue<uint16_t>(0, length);
+        auto multiplicity = outputs->multiplicities.getTargetMultiplicity(dstNodeID.offset);
+        for (auto i = 0u; i < multiplicity; ++i) {
             fTable.append(vectors);
         }
+        if (counter != nullptr) {
+            counter->increase(multiplicity);
+        }
+    }
+
+    std::unique_ptr<RJOutputWriter> copy() override {
+        return std::make_unique<AllSPDestinationsOutputWriter>(context, rjOutputs, outputNodeMask);
     }
 };
 
--- a/src/function/gds/gds_task.cpp
+++ b/src/function/gds/gds_task.cpp
@@ -56,7 +56,6 @@ void VertexComputeTask::run() {
             localVc->vertexCompute(nodeID);
         }
     }
-    localVc->finalizeWorkerThread();
 }
 } // namespace function
 } // namespace kuzu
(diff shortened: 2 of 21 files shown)


src/function/gds/gds_task.cpp
```diff
@@ -53,10 +53,12 @@ void VertexComputeTask::run() {
     while (sharedState->morselDispatcher.getNextRangeMorsel(frontierMorsel)) {
         while (frontierMorsel.hasNextOffset()) {
```
semihsalihoglu-uw, line 54: I see two options. 
Option 1 (General early stopping logic):  Either have a general breaking early optimization similar to edgeCompute() but not exactly the same based on implementing some abstract functions. 
Option 2 (Only algorithm-specific early stopping logic): Do not have any early stopping logic here, so this code does not change and then push all early stopping logics to the algorithms, which can implement early stopping inside their vertexCompute functions. This means, we always call vertexCompute() on each node, but those vertexCompute() functions would just return without doing any work.

I'm fine with either approach frankly. You can choose the simpler one. But I would probably not do something in between, like this one, which has some logic of general early stopping implemented in algorithm-agnostic parts of the code like gds_task.cpp, plus some algorithm-specific code (but not abstracted away cleanly).

src/function/gds/output_writer.cpp
```diff
@@ -80,7 +80,7 @@ static void setLength(ValueVector* vector, uint16_t length) {
     vector->setValue<uint16_t>(0, length);
 }
 
-void PathsOutputWriter::write(processor::FactorizedTable& fTable, nodeID_t dstNodeID) const {
+offset_t PathsOutputWriter::write(processor::FactorizedTable& fTable, nodeID_t dstNodeID) {
```
semihsalihoglu-uw, line 83: I think you can push limit down all the way here too. That would be the most efficient. That is, you can push the limit to both here but you still have to implement some logic to stop the frontier morsel grabbing and calling vertex computes. That would be the most efficient but it's your call.

src/function/gds/output_writer.cpp
```diff
@@ -111,6 +113,10 @@ void PathsOutputWriter::write(processor::FactorizedTable& fTable, nodeID_t dstNo
             if (checkPathNodeMask(curPath) && checkSemantic(curPath)) {
                 writePath(curPath);
                 fTable.append(vectors);
+                numOutputTuples++;
+                if (numOutputTuples >= info.limitNumber) {
```
semihsalihoglu-uw, line 117: As I said, i would probably use a global numOutputTuples atomic value in a shared state and have the write() function and the vertexCompute function related to output writers stop early, instead of doing a thread-level limit counting and then further aggregating. I think for limit, that does not seem necessary.

semihsalihoglu-uw (changes requested):
I have some suggestions that would be good to integrate.

src/include/processor/operator/gds_call_shared_state.h
```diff
@@ -63,13 +63,38 @@ class NodeOffsetMaskMap {
     bool enabled_;
 };
 
+class RJOutputCounter {
```
andyfengHKU, line 66: Rename to GDSOutputCounter.

src/function/gds/all_shortest_paths.cpp
```diff
@@ -124,19 +124,20 @@ class AllSPDestinationsOutputWriter : public DestinationsOutputWriter {
...
-    }
-
-protected:
-    void writeMoreAndAppend(processor::FactorizedTable& fTable, common::nodeID_t dstNodeID,
-        uint16_t) const override {
-        auto multiplicity =
-            rjOutputs->ptrCast<AllSPDestinationsOutputs>()->multiplicities.getTargetMultiplicity(
-                dstNodeID.offset);
-        for (uint64_t i = 0; i < multiplicity; ++i) {
+    offset_t write(FactorizedTable& fTable, nodeID_t dstNodeID) override {
```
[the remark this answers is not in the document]
andyfengHKU: It represents number of output which falls into the definition of offset I assume.

src/function/gds/gds_task.cpp
andyfengHKU: Took the second approach.

src/function/gds/output_writer.cpp
andyfengHKU: I already pushed into output writer. Which push down are u referring here?

src/function/gds/output_writer.cpp
andyfengHKU: Done

src/function/gds/all_shortest_paths.cpp
```diff
@@ -124,19 +124,23 @@ class AllSPDestinationsOutputWriter : public DestinationsOutputWriter {
...
-        uint16_t) const override {
-        auto multiplicity =
-            rjOutputs->ptrCast<AllSPDestinationsOutputs>()->multiplicities.getTargetMultiplicity(
-                dstNodeID.offset);
-        for (uint64_t i = 0; i < multiplicity; ++i) {
+    void write(FactorizedTable& fTable, nodeID_t dstNodeID,
+        processor::RJOutputCounter* counter) override {
+        auto outputs = rjOutputs->ptrCast<AllSPDestinationsOutputs>();
+        auto length = outputs->pathLengths->getMaskValueFromCurFrontierFixedMask(dstNodeID.offset);
+        dstNodeIDVector->setValue<nodeID_t>(0, dstNodeID);
```
semihsalihoglu-uw, line 131: Was this a bug? Why were these code not there before?

src/function/gds/output_writer.cpp
```diff
@@ -96,6 +98,10 @@ void PathsOutputWriter::write(processor::FactorizedTable& fTable, nodeID_t dstNo
                 beginWritePath(0);
             }
             fTable.append(vectors);
+            // No need to check against limit number since this is the first output.
+            if (counter != nullptr) {
+                counter->increase(1);
```
semihsalihoglu-uw, line 103: I don't understand this comment. What if the limit is 1?

src/include/processor/operator/gds_call_shared_state.h
semihsalihoglu-uw: Reminding this TODO.

src/optimizer/limit_push_down_optimizer.cpp
```diff
@@ -18,15 +19,18 @@ void LimitPushDownOptimizer::visitOperator(planner::LogicalOperator* op) {
     switch (op->getOperatorType()) {
     case LogicalOperatorType::LIMIT: {
         auto& limit = op->constCast<LogicalLimit>();
-        skipNumber = limit.getSkipNum();
-        limitNumber = limit.getLimitNum();
+        if (limit.hasSkipNum()) {
```
semihsalihoglu-uw, line 22: Was this another bug?

src/function/gds/all_shortest_paths.cpp
andyfengHKU: Nah I simply removed one virtual function and choose to duplicate several lines of code.

src/function/gds/output_writer.cpp
andyfengHKU: Then we will continue and terminate when we find the second path. There is a LIMIT operator at the very end which guarantees we will not exceed 1.

src/optimizer/limit_push_down_optimizer.cpp
andyfengHKU: Not really. It doesn't have an effect on DISTINCT. But does have an effect on the way I implement GDS. So I wouldn't call this as a bug.

PR #4474, opened 2024-11-06
Push down dst table id filter to GDS

andyfengHKU:
# Description

Improve GDS performance on multi-labeled query.

```
MATCH p=(a:Person)-[*4..4]->(b:TagClass) WHERE a.id=96 RETURN COUNT(*);
```
The above query goes from 27s to 0.8s after this PR.

Fixes # (issue)

# Contributor agreement

### DIFF
--- a/src/planner/plan/append_extend.cpp
+++ b/src/planner/plan/append_extend.cpp
@@ -152,6 +152,21 @@ void Planner::appendRecursiveExtendAsGDS(const std::shared_ptr<NodeExpression>& 
         createPathNodeFilterPlan(recursiveInfo->node, recursiveInfo->originalNodePredicate, p);
         gdsCall->ptrCast<LogicalGDSCall>()->setNodePredicateRoot(p.getLastOperator());
     }
+    // E.g. Given schema person-knows->person & person-knows->animal
+    // And query MATCH (a:person:animal)-[e*]->(b:person)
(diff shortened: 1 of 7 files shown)


src/planner/plan/append_extend.cpp
```diff
@@ -152,6 +152,13 @@ void Planner::appendRecursiveExtendAsGDS(const std::shared_ptr<NodeExpression>&
         createPathNodeFilterPlan(recursiveInfo->node, recursiveInfo->originalNodePredicate, p);
         gdsCall->ptrCast<LogicalGDSCall>()->setNodePredicateRoot(p.getLastOperator());
     }
+    // E.g. Given schema person-knows->person & person-knows->animal
+    // And query MATCH (a:person:animal)-[e*]->(b:person)
+    // The destination node b after GDS will contain both person & animal label. We need to prune
+    // the animal out.
+    if (nbrNode->getTableIDsSet().size() < recursiveInfo->node->getTableIDsSet().size()) {
```
ray6080, line 159: Should we take intersect of the two tableIDsSet here? I'm not sure how the labels are pruned here. but is it possible that there are cases, for example, `recursiveInfo->node` has the tableIDsSet of `person, animal, student`, while `nbrNode`'s tableIDsSet is `person, animal, teacher`?

ray6080 (approved):

PR #4479, opened 2024-11-06
CSV parallel parsing fixes

royi-luo:
# Description
Fixes:
- After skipping an ignored (errored) row, check if the end of the block is reached and stop parsing if it has
- CSV errors are now never thrown the moment they are found (before they are thrown immediately if we can get the line number). Instead they are cached and thrown periodically when threads report that they have finished parsing blocks. This has a few benefits:
  - Since errors are cached in the shared state, they are visible from all threads. Before throwing an exception could terminate one thread but remaining threads would continue execution, delaying the end of the query (potentially forcing the entire file to be parsed before the error is reported)
  - Line number information (lines per block) for exceptions are always accurate as we don't throw in the middle of parsing blocks

# Contributor agreement


acquamarin (approved):

ray6080 (approved):
Thanks! A minor comment, irrelevant to changes in this PR, is that I found the include `#include "processor/operator/persistent/reader/csv/base_csv_reader.h"` is useless in `file_error_handler.cpp`. Can you remove it?

PR #4497, opened 2024-11-11
Apply semi mask to path property

andyfengHKU:
# Description

Fixes # (issue)

# Contributor agreement

### DIFF
--- a/src/planner/plan/append_extend.cpp
+++ b/src/planner/plan/append_extend.cpp
@@ -197,12 +197,11 @@ void Planner::appendRecursiveExtendAsGDS(const std::shared_ptr<NodeExpression>& 
     pathPropertyProbe->extendFromLeft = *boundNode == *rel->getLeftNode();
     pathPropertyProbe->pathNodeIDs = recursiveInfo->pathNodeIDsExpr;
     pathPropertyProbe->pathEdgeIDs = recursiveInfo->pathEdgeIDsExpr;
-    pathPropertyProbe->getSIPInfoUnsafe().position = SemiMaskPosition::PROHIBIT;
     pathPropertyProbe->computeFactorizedSchema();
     probePlan.setLastOperator(pathPropertyProbe);
     probePlan.setCost(plan.getCardinality());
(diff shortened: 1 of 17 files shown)


src/processor/map/map_gds.cpp
```diff
@@ -80,6 +80,7 @@ std::unique_ptr<PhysicalOperator> PlanMapper::mapGDSCall(LogicalOperator* logica
         logicalOpToPhysicalOpMap.erase(logicalOperator);
     }
 
+    logicalOpToPhysicalOpMap.insert({logicalOperator, gdsCall.get()});
```
ray6080, line 83: Let's open an issue on refactoring the mapping to be consistent that we always keep child sink op, instead of parent scan op of the logical operator.

src/planner/plan/append_extend.cpp
```diff
@@ -197,12 +197,12 @@ void Planner::appendRecursiveExtendAsGDS(const std::shared_ptr<NodeExpression>&
     pathPropertyProbe->extendFromLeft = *boundNode == *rel->getLeftNode();
     pathPropertyProbe->pathNodeIDs = recursiveInfo->pathNodeIDsExpr;
     pathPropertyProbe->pathEdgeIDs = recursiveInfo->pathEdgeIDsExpr;
-    pathPropertyProbe->getSIPInfoUnsafe().position = SemiMaskPosition::PROHIBIT;
     pathPropertyProbe->computeFactorizedSchema();
     probePlan.setLastOperator(pathPropertyProbe);
     probePlan.setCost(plan.getCardinality());
-    auto cardinality = UINT32_MAX;
-    probePlan.setCardinality(cardinality);
+    auto extensionRate =
```
ray6080, line 203: Can you add a comment here that we're adding a penalty factor UINT32_MAX here to ensure that in almost all cases, the GDS pipeline is on the probe side? This is a heuristic rule we keep before we have a better way to figure out a good estimation on output cardinality of recursive joins, such as based on the max degree of the rel table in the direction.

PR #4505, opened 2024-11-12
More ALP bug fixes

royi-luo:
# Description
- ALP chunks contain two types of pages: one for storing bitpacked data and one for storing exceptions. When determining if we can update in place, we should only use the number of pages to used bitpacked data to determine if we can perform updates (insertions) in place. Before we also included the number of pages for storing exceptions, giving a larger capacity than expected which caused some overflow errors.
- Modify the ALP vector size to match column chunk (rowgroup) size. For ALP we find a single set of parameters per column chunk and the disagreement between the ALP vector size and our column chunk size could potentially cause the parameters to be selected incorrectly.
- Have `ALPEncode::encode_value()` return int32_t if encoding `float` types (instead of `double`). This ensures that doing an encode then a decode of the return value will properly reflect the success of the operation. Before `encode_value()` always returned int64_t which caused us to miss some failed encode + decode for floats (which should be bitpacked as int32_t).
  - Technically always encoding as `int64_t` can be correct. However, we encode `float` as `int32_t` so that we can decompress the bitpacked values using the same memory that we return the decoded `float` values in, allowing us to avoid heap allocations.

# Contributor agreement


third_party/alp/include/alp/config.hpp
```diff
@@ -8,16 +8,16 @@
...
+inline constexpr size_t ROWGROUP_SIZE = VECTOR_SIZE;
 /// Vectors from the rowgroup from which to take samples; this will be used to then calculate the jumps
-inline constexpr size_t ROWGROUP_VECTOR_SAMPLES = 8;
+inline constexpr size_t ROWGROUP_VECTOR_SAMPLES = 1;
 /// We calculate how many equidistant vector we must jump within a rowgroup
 inline constexpr size_t ROWGROUP_SAMPLES_JUMP = (ROWGROUP_SIZE / ROWGROUP_VECTOR_SAMPLES) / VECTOR_SIZE;
 /// Values to sample per vector
-inline constexpr size_t SAMPLES_PER_VECTOR   = 32;
-inline constexpr size_t SAMPLES_PER_ROWGROUP = 128 * SAMPLES_PER_VECTOR;
+inline constexpr size_t SAMPLES_PER_VECTOR   = 128 * 32;
```
royi-luo, line 19:
```suggestion
inline constexpr size_t SAMPLES_PER_VECTOR   = 4 * 1024;
```

src/storage/store/column_chunk_metadata.cpp
```diff
@@ -128,7 +143,7 @@ ColumnChunkMetadata getConstantFloatMetadata(PhysicalTypeID physicalType, uint64
 template<std::floating_point T>
 alp::state getAlpMetadata(const T* buffer, uint64_t numValues) {
     alp::state alpMetadata;
-    std::vector<uint8_t> sampleBuffer(alp::config::SAMPLES_PER_ROWGROUP);
+    std::vector<uint8_t> sampleBuffer(alp::config::SAMPLES_PER_ROWGROUP * sizeof(T));
```
benjaminwinger, line 146: Since this is just being used by casting it to `T*` on the next line, why not have it be a `std::vector<T>` instead?

benjaminwinger (approved):

src/storage/store/column_chunk_metadata.cpp
royi-luo: updated

PR #4523, opened 2024-11-14
Wasm buffer manager support

benjaminwinger:
This adds a simple alternative implementation to the BufferManager that allocates pages using malloc (via unique_ptr) and stores the page pointer in the PageState object. 

### Performance
There is a small performance penalty, pinning pages appears to be ~2-5x slower with this method, however even when doing a lot of new pins (i.e. where it's also allocating; there should be little difference for pins of pages which are already in memory) the overall cost is minimal.

E.g. summing the float edge property in the LDBC datagen-9_0-fb dataset took ~700ms using the normal buffer manager and ~900ms using this variant (obviously not a webassembly build; I overrode the macro used to enable/disable this feature).

Interestingly, it was actually faster when comparing the following query on the much larger graph500-30 dataset: 
`kuzu kugraph500-30 -d512 --noprogressbar --read_only <<< 'MATCH (v:N)-[e:E]->(:N) RETURN SUM(v.ID);'`, where the original runtime was ~4s, and improved to ~2.8s with this change (also on a machine with 128 threads; the difference may be that this approach scales better or has less of a setup cost, even though the allocation cost is higher; I'll look into it and try and see what the difference is).

### Outstanding Webassembly bugs
Unfortunately it appears that this is not the issue with the web assembly tests which were running out of memory. They are still running out of memory with this change, so I have left them disabled; as far as I can tell the huge memory allocations in those tests are unrelated to the BufferManager/MemoryManager.

### Other changes of note
I also lowered the default buffer pool size for testing back to 64MB. This should have been lowered before, but I guess was missed after it was raised when I was adding ColumnChunk memory to the tracked limit. With the higher limit most tests never do buffer manager evictions (including some of the tests which are disabled on webassembly due to this memory issue).


src/storage/buffer_manager/buffer_manager.cpp
```diff
@@ -193,25 +199,35 @@ void handleAccessViolation(unsigned int exceptionCode, PEXCEPTION_POINTERS excep
...
-#if defined(_WIN32)
+    const std::array<std::unique_ptr<VMRegion>, 2>& vmRegions [[maybe_unused]],
+    PageSizeClass pageSizeClass [[maybe_unused]]) {
+#if BM_MALLOC
+    if (frame == nullptr) {
+        return false;
+    }
+#endif
+
+#if defined(_WIN32) && !BM_MALLOC
```
ray6080, line 210: The macro here is becoming more a bit hard to follow now. I wonder if we should choose to duplicate the code a bit to separate them more clearly.

src/storage/buffer_manager/buffer_manager.cpp
```diff
@@ -193,25 +199,35 @@ void handleAccessViolation(unsigned int exceptionCode, PEXCEPTION_POINTERS excep
...
     try {
 #endif
         func(frame);
-#if defined(_WIN32)
+#if defined(_WIN32) && !BM_MALLOC
     } catch (AccessViolation& exc) {
         // If we encounter an acess violation within the VM region,
         // the page was decomitted by another thread
         // and is no longer valid memory
+#if BM_MALLOC
```
ray6080, line 219: How can this be true when `#if defined(_WIN32) && !BM_MALLOC` is true?

benjaminwinger: I think I added those at two different times.

I'm not sure if accessing freed memory may cause an access violation on Windows, in which case optimisticRead won't be safe even when using malloc unless we handle those access violations (at least, those that occur within the frame).
The CI is passing with it disabled at the moment, which makes me wonder if it's not covered by tests, or if it in practice just isn't happening (maybe it doesn't actually happen at all, or it's just that freed memory that's been re-used wouldn't cause the access violation).
I think I'll just remove this block for now and leave the access violation handling disabled with `BM_MALLOC`.

src/include/storage/buffer_manager/page_state.h
```diff
@@ -74,12 +81,28 @@ class PageState {
...
+        stateAndVersion.store(EVICTED << NUM_BITS_TO_SHIFT_FOR_STATE);
+#if BM_MALLOC
+        page.reset();
+#endif
+    }
+
+#if BM_MALLOC
+    uint8_t* getPage() const { return page.get(); }
+    uint8_t* allocatePage(uint64_t pageSize) {
+        page = std::make_unique<uint8_t[]>(pageSize);
```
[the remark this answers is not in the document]
benjaminwinger: Interestingly in webassembly the tests are compiled with the default [ABORTING_MALLOC=1](https://emscripten.org/docs/tools_reference/settings_reference.html?highlight=environment#aborting-malloc), which aborts the webassembly process if it runs out of memory instead of returning 0, but outside of the tests we're compiling with `ALLOW_MEMORY_GROWTH`, which has the default of having malloc return 0 on failure.
So I think we should set `ABORTING_MALLOC=0` for the tests just so it's handled the same (not that we have any tests where we expect this yet).

PR #4533, opened 2024-11-15
Using shared_mutex instead of mutex in CatalogSet

ted-wq-x:
# Description

With 50 concurrent short queries, mutex becomes a performance bottleneck.

DataStet: Sf100
Concurrency: 50
Query: match (a:person {id:24189255912498})-[f]-(b) return count(1) ;
Throughput(QPS): 11645.5-->19364.1

# Contributor agreement

### DIFF
--- a/src/catalog/catalog_set.cpp
+++ b/src/catalog/catalog_set.cpp
@@ -210,7 +210,7 @@ CatalogEntrySet CatalogSet::getEntries(const Transaction* transaction) {
 }
 
 void CatalogSet::iterateEntriesOfType(const Transaction* transaction, CatalogEntryType type,
-    const std::function<void(CatalogEntry*)>& func) {
+    const std::function<void(const CatalogEntry*)>& func) {
     std::shared_lock lck{mtx};
     for (auto& [_, entry] : entries) {
         if (entry->getType() != CatalogEntryType::DUMMY_ENTRY && entry->getType() != type) {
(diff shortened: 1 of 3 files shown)


src/catalog/catalog_set.cpp
```diff
@@ -211,7 +211,7 @@ CatalogEntrySet CatalogSet::getEntries(const Transaction* transaction) {
 
 void CatalogSet::iterateEntriesOfType(const Transaction* transaction, CatalogEntryType type,
     const std::function<void(CatalogEntry*)>& func) {
-    std::lock_guard lck{mtx};
+    std::shared_lock lck{mtx};
```
ray6080, line 214: I don't think shared_lock is safe here, as the `func` can modify entires.
One option is to change func signature to take in `const CatalogEntry&` instead of `CatalogEntry*`, or we should fall back to `unique_lock` here. Can you see if you can apply the first solution?

ted-wq-x: 👍 ,Done

ray6080 (approved):
Thanks! LGTM.

Can you provide a bit more details on the benchmarking you did? I want to keep this in record and build some understanding here.
1. what's the detailed query you run?
2. what's the concurrency number?
3. how is the throughput changing after the changes? detailed numbers would be great.

ray6080:
@ted-wq-x can you also fix the compilation warnings and clang-tidy checks?

PR #4556, opened 2024-11-20
Fix VersionInfo SelectionVector creation

### LINKED ISSUE #4271: Bug: ldbc sf01 deleteComment test fail randomly
### Kùzu version

master

_No response_

### What happened?

See log here https://github.com/kuzudb/kuzu/actions/runs/10858946549/job/30138088819

_No response_

benjaminwinger:
Fixes #4271.

The issue was that `VersionInfo::getSelVectorForScan` (and `VectorVersionInfo::getSelVectorForScan`) append repeatedly to a `SelectionVector` and in certain circumstances it will turn a non-empty unfiltered SelectionVector into a filtered one without preserving the former values and then just appending to the end. My solution was to add a new makeDynamic function that copies the old selectedPositions into the mutable buffer before switching them.

E.g. this can occur when the first few VectorVersionInfos have no deletions, such that the first branch is taken in `VectorVersionInfo::getSelVectorToScan` and it just increases the length of the unfiltered SelectionVector, but then later VectorVersionInfos have deletions, taking the second branch where there is no check for unfiltered data (since it has to be converted to filtered at that point).

### DIFF
--- a/src/include/common/data_chunk/sel_vector.h
+++ b/src/include/common/data_chunk/sel_vector.h
@@ -1,5 +1,7 @@
 #pragma once
 
+#include <string.h>
+
 #include <memory>
 
 #include "common/constants.h"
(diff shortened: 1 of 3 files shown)


ray6080 (approved):
Nice catch. LGTM!

PR #4557, opened 2024-11-21
Add sparse frontier implementation

andyfengHKU:
# Description

This optimization should improve the performance of `recursive-join-sparse` by order-of-magnitude. I'm not seeing it in PR benchmark. Though it performs pretty well when I test both locally and on ac5. So I'm merging to see the performance difference on our daily benchmark.

Fixes # (issue)

# Contributor agreement

### DIFF
--- a/src/function/gds/gds_utils.cpp
+++ b/src/function/gds/gds_utils.cpp
@@ -25,13 +25,13 @@ static uint64_t getNumThreads(processor::ExecutionContext& context) {
         .getValue<uint64_t>();
 }
 
-void GDSUtils::scheduleFrontierTask(table_id_t nbrTableID, table_id_t relTableID, graph::Graph* graph,
-    ExtendDirection extendDirection, GDSComputeState& gdsComputeState,
+void GDSUtils::scheduleFrontierTask(table_id_t nbrTableID, table_id_t relTableID,
+    graph::Graph* graph, ExtendDirection extendDirection, GDSComputeState& gdsComputeState,
     processor::ExecutionContext* context, std::optional<uint64_t> numThreads,
     std::optional<common::idx_t> edgePropertyIdx) {
     auto clientContext = context->clientContext;
-    auto info = FrontierTaskInfo(nbrTableID, relTableID, graph, extendDirection, *gdsComputeState.edgeCompute,
-        edgePropertyIdx);
+    auto info = FrontierTaskInfo(nbrTableID, relTableID, graph, extendDirection,
+        *gdsComputeState.edgeCompute, edgePropertyIdx);
     auto sharedState = std::make_shared<FrontierTaskSharedState>(*gdsComputeState.frontierPair);
     uint64_t maxThreads = numThreads ? numThreads.value() : getNumThreads(*context);
     auto task = std::make_shared<FrontierTask>(maxThreads, info, sharedState);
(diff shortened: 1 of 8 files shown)


src/include/function/gds/gds_frontier.h
```diff
@@ -278,8 +339,14 @@ class KUZU_API FrontierPair {
...
-    std::shared_ptr<GDSFrontier> curFrontier;
-    std::shared_ptr<GDSFrontier> nextFrontier;
+
+    // Dense frontiers are always updated.
+    std::shared_ptr<GDSFrontier> curDenseFrontier;
+    std::shared_ptr<GDSFrontier> nextDenseFrontier;
+    // Sparse frontiers
+    std::shared_ptr<SparseFrontier> curSparseFrontier;
+    std::shared_ptr<SparseFrontier> nextSparseFrontier;
+    std::shared_ptr<SparseFrontier> vertexComputeCandidates;
```
andyfengHKU, line 349: -> allPossibleDestinations

src/include/function/gds/gds_frontier.h
```diff
@@ -147,6 +140,67 @@ class GDSFrontier {
     common::table_id_map_t<common::offset_t> numNodesMap;
 };
 
+/*
+ * LocalFrontierSample keeps a small frontier per thread.
+ * If enabled, this frontier represents a complete local frontier.
+ * Otherwise, complete local frontier is larger than this sample.
+ * */
+class LocalFrontierSample {
```
semihsalihoglu-uw, line 148: Reading the code, this looks redundant to me. It looks like a specialized SparseFrontier implementation that keeps a set of nodeIDs instead of a set of set of nodeOffsets. It looks like we should be able to have a single SparseFrontier implementation that can be used both as a LocalFrontier and a global one, even if it is a slightly slower implementation.

src/include/function/gds/gds_frontier.h
```diff
@@ -147,6 +140,67 @@ class GDSFrontier {
...
+    const std::unordered_set<common::offset_t>& getOffsetSet() const { return offsetSet; }
+
+    bool empty() const { return offsetSet.empty(); }
+
+    void addNodes(const std::vector<common::nodeID_t> nodeIDs);
+
+    void checkSampleSize();
+
+private:
+    static constexpr uint16_t SAMPLE_SIZE = 20;
```
semihsalihoglu-uw, line 161: This looks quite small to me. It also should probably depend on the number of nodes in the graph and threads. Can you implement a rule somewhere that is something like: SAMPLE_SIZE = max(20, |V|/(8xnumThreads)). To explain |V|/(8xnumThreads): The intuition is that sparse execution has the advantage that it does not loop over |V| many nodes, so that goes in the numerator. The 8 x numThreads models the disadvantage of sparse execution, which is due to two components: (i) that it runs single threaded instead of running in numThreads; and (ii) it does some extra operations, such as updating both the sparse and the dense frontier. So I'm putting a constant 8 to accommodate for the latter. That is until V/(8xnumThreads)-size we're better of taking both costs and running in a sparse manner.

This is another case that requires some benchmarking. To understand what SAMPLE_SIZE should be, we need to see the behavior of different heuristics to pick a SAMPLE_SIZE on some benchmark.

src/function/gds/gds_utils.cpp
```diff
@@ -113,5 +119,23 @@ void GDSUtils::runVertexComputeIteration(processor::ExecutionContext* executionC
     }
 }
 
+void GDSUtils::runVertexComputeSparse(SparseFrontier& sparseFrontier, graph::Graph* graph,
```
semihsalihoglu-uw, line 122: FYI, ideally even if the frontier is sparse, we should run it in parallel. You need a different "frontiermorselizer" implementation that can give morsels of frontiers to threads. This would require some rethinking of the frontiermorsel and frontiermorselizer interfaces, so it can work against both a sparse frontier and a dense frontier. But it would also avoid us having to implement a separate vertex compute or edge compute functions.

semihsalihoglu-uw (approved):
I have several medium-level comments. Would be good to discuss more once you have some more extensive benchmarks.

src/include/function/gds/gds_frontier.h
andyfengHKU: I tried to make sample size proportional to maxOffset but this wasn't a trivial change. Also as we discussed, we might wanna experiment more on this. So instead I opened an issue.

src/include/function/gds/gds_frontier.h
andyfengHKU: Done

src/function/gds/gds_utils.cpp
andyfengHKU: Ditto opened an issue

PR #4558, opened 2024-11-21
Rel scan selection optimizations

benjaminwinger:
The biggest optimization here is avoiding using the bitset in `CSRNodeGroup::tryScanCachedTuples` if it's known that all values are selected. 

In some unfiltered rel table scans this improved performance of `CSRNodeGroup::tryScanCachedTuples` by about 4x, for a total of about a 15% on some queries I'd benchmarked which were mostly scans.

E.g. 'MATCH (v:N)-[e:E]->(v2:N) RETURN SUM(v.ID);' on the ldbc graph500-30 dataset (with 128 threads) went from ~2.65s to ~2.25s

I've also:
- removed a couple of unnecessary `std::shared_ptr` copies (which are immediately destroyed).
- Added `ValueVector::forEachNonNull` to consolidate some code in the aggregate functions

### DIFF
--- a/src/include/common/vector/value_vector.h
+++ b/src/include/common/vector/value_vector.h
@@ -4,6 +4,7 @@
 
 #include "common/assert.h"
 #include "common/cast.h"
+#include "common/constants.h"
 #include "common/copy_constructors.h"
 #include "common/data_chunk/data_chunk_state.h"
 #include "common/null_mask.h"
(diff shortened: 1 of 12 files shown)


ray6080 (approved):
LGTM! @andyfengHKU can you also take a quick look into the changes regarding to null mask and aggregation functions and see if there are other places we can apply the changes in similar pattern?

src/processor/result/result_set.cpp
```diff
@@ -31,7 +31,7 @@ uint64_t ResultSet::getNumTuplesWithoutMultiplicity(
     KU_ASSERT(!dataChunksPosInScope.empty());
     uint64_t numTuples = 1;
     for (auto& dataChunkPos : dataChunksPosInScope) {
-        auto state = dataChunks[dataChunkPos]->state;
+        auto& state = dataChunks[dataChunkPos]->state;
```
andyfengHKU, line 34: Let's simply inline this line

src/include/common/vector/value_vector.h
```diff
@@ -35,6 +36,32 @@ class KUZU_API ValueVector {
     DELETE_COPY_AND_MOVE(ValueVector);
     ~ValueVector() = default;
 
+    template<class Func>
+    void forEachNonNull(Func&& func) const {
```
andyfengHKU, line 40: Can u open an issue to propagate this to unary, binary & ternary evaluator and assign the issue to @acquamarin

andyfengHKU (approved):

PR #4559, opened 2024-11-22
Rollback hash index checkpoint

### LINKED ISSUE #4338: Buffer manager allocation failures causing corruption

royi-luo:
# Description

Fixes #4338 

- Separate execute/commit and checkpoint in `ClientContext::executeNoLock()` into separate catch statements which avoids us triggering rollback of a committed statement if an error occurs during checkpoint.
- Rollback changes made to PK index during checkpoint if an exception occurs so that the checkpoint can be retried successfully

# Contributor agreement


src/main/client_context.cpp
```diff
@@ -504,16 +506,28 @@ std::unique_ptr<QueryResult> ClientContext::executeNoLock(PreparedStatement* pre
...
+    if (autoCommitTriggered) {
+        try {
+            this->transactionContext->autoCheckpointIfNeeded();
+        } catch (common::CheckpointException& e) {
+            auto locks = e.getLocks();
+            this->transactionContext->rollbackCheckpoint(locks);
+            return handleFailedExecute(executionContext.get(), e);
+        } catch (std::exception& e) {
+            // exception was before checkpoint started, no need to rollback checkpoint
+            return handleFailedExecute(executionContext.get(), e);
```
royi-luo, line 527: add test to cover this

test/copy/copy_test.cpp
```diff
@@ -247,6 +258,70 @@ TEST_F(CopyTest, RelInsertBMExceptionDuringCommitRecovery) {
...
+            },
+        .executeFunc =
+            [](main::Connection* conn, int) {
+                return conn->query(common::stringFormat(
+                    "UNWIND RANGE(1,{}) AS i CREATE (a:account {ID:i})", numValues));
+            },
+        .earlyExitOnFailureFunc =
+            [this](main::QueryResult*) {
+                // make sure the checkpoint when closing the DB doesn't fail
+                failureFrequency = UINT64_MAX;
```
royi-luo, line 316: can remove this

src/transaction/transaction_manager.cpp
```diff
@@ -152,28 +183,37 @@ void TransactionManager::checkpointNoLock(main::ClientContext& clientContext) {
...
+        // Clear the wal, and also shadowing files.
+        wal.clearWAL();
+        clientContext.getStorageManager()->getShadowFile().clearAll(clientContext);
+        StorageUtils::removeWALVersionFiles(clientContext.getDatabasePath(),
+            clientContext.getVFSUnsafe());
+    } catch (std::exception& e) {
+        // since we want to rollback the checkpoint atomically
+        // we maintain ownership of the locks (which we will release when we are done handling the
+        // exception)
+        common::CheckpointException checkpointException(e);
```
ray6080, line 213: Can we not handle rollback here? instead of back to ClientContext? so lots of things can be simplified.

royi-luo: that's probably a good idea

royi-luo: updated

src/include/common/exception/checkpoint_exception.h
```diff
@@ -0,0 +1,27 @@
...
+    std::vector<const common::UniqLock*> getLocks() const {
+        std::vector<const common::UniqLock*> ret;
+        for (const auto& lock : locks) {
+            ret.push_back(lock.get());
+        }
+        return ret;
+    }
+
+private:
+    std::vector<std::shared_ptr<common::UniqLock>> locks;
```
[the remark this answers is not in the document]
royi-luo: removed since we handle the rollback in the transaction manager

src/include/transaction/transaction_context.h
```diff
@@ -5,6 +5,10 @@
 #include "transaction.h"
 
 namespace kuzu {
+namespace common {
+struct UniqLock;
```
ray6080, line 9: Do we need this forward declaration?

src/storage/index/hash_index.cpp
```diff
@@ -485,6 +491,12 @@ PrimaryKeyIndex::PrimaryKeyIndex(const DBFileIDAndName& dbFileIDAndName, bool re
             }
         },
         [&](auto) { KU_UNREACHABLE; });
+
+    if (newIndex && !inMemMode) {
+        // checkpoint the creation of the index so that if we need to rollback it will be to a
+        // state we can retry from (an empty index with the disk arrays initialized)
+        checkpoint(true);
```
ray6080, line 498: Add inline comment for `true` here.

src/include/main/client_context.h
```diff
@@ -177,6 +177,9 @@ class KUZU_API ClientContext {
 
     void runFuncInTransaction(const std::function<void(void)>& fun);
 
+    std::unique_ptr<QueryResult> handleFailedExecute(processor::ExecutionContext* executionContext,
```
ray6080, line 180: `handleFailedExecute` -> `handleFailedExecution` / `handleExecutionError`

src/main/client_context.cpp
```diff
@@ -504,16 +505,23 @@ std::unique_ptr<QueryResult> ClientContext::executeNoLock(PreparedStatement* pre
...
     } catch (std::exception& e) {
         transactionContext->rollback();
-        getMemoryManager()->getBufferManager()->getSpillerOrSkip(
-            [](auto& spiller) { spiller.clearFile(); });
-        progressBar->endProgress(executionContext->queryID);
-        return queryResultWithError(e.what());
+        return handleFailedExecute(executionContext.get(), e);
+    }
+
+    if (autoCommitTriggered) {
```
ray6080, line 517: I wonder if you can throw a CheckpointException and catch here instead of explicitly separating autoCheckpoint and commit in different places.

ray6080 (commented):
@benjaminwinger can you also take a look at changes to hash index and overflow file?

src/main/client_context.cpp
royi-luo: probably a good idea, that should clean up the changes quite a bit

src/include/transaction/transaction_context.h
royi-luo: removed

src/main/client_context.cpp
royi-luo: updated

src/storage/index/hash_index.cpp
```diff
@@ -425,7 +431,7 @@ PrimaryKeyIndex::PrimaryKeyIndex(const DBFileIDAndName& dbFileIDAndName, bool re
                                       vfs, context)},
       dbFileIDAndName{dbFileIDAndName}, shadowFile{*shadowFile} {
     KU_ASSERT(!(inMemMode && readOnly));
-    bool newIndex = fileHandle->getNumPages() == 0;
+    bool newIndex = (fileHandle->getNumPages() == 0);
```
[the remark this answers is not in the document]
royi-luo: updated

ray6080 (approved):

PR #4574, opened 2024-11-29
Fix config parameters binding for C API

mewim:
This PR fixes the config parameter binding for C API (the parameter `max_db_size` was taken but not actually passed into C++). It also added checkpoint parameters. Additionally, it changes the way of getting the default parameter for C API.

### DIFF
--- a/src/include/c_api/kuzu.h
+++ b/src/include/c_api/kuzu.h
@@ -131,7 +131,7 @@ typedef struct {
     bool auto_checkpoint;
     // The threshold of the WAL file size in bytes. When the size of the
     // WAL file exceeds this threshold, the database will checkpoint if autoCheckpoint is true.
-    uint64_t check_point_threshold;
+    uint64_t checkpoint_threshold;
 } kuzu_system_config;
 
 /**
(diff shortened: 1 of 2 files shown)


src/include/c_api/kuzu.h
```diff
@@ -126,6 +126,12 @@ typedef struct {
     //  will be removed once we implemente a better solution later. The value is default to 1 << 43
     //  (8TB) under 64-bit environment and 1GB under 32-bit one (see `DEFAULT_VM_REGION_MAX_SIZE`).
     uint64_t max_db_size;
+    // If true, the database will automatically checkpoint when the size of
+    // the WAL file exceeds the checkpoint threshold.bool auto_checkpoint;
```
ray6080, line 130: Remove `bool auto_checkpoint;` at the end?

src/include/c_api/kuzu.h
```diff
@@ -126,6 +126,12 @@ typedef struct {
...
+    bool auto_checkpoint;
+    // The threshold of the WAL file size in bytes. When the size of the
+    // WAL file exceeds this threshold, the database will checkpoint if autoCheckpoint is true.
```
ray6080, line 133: `autoCheckpoint` -> `auto_checkpoint`

ray6080 (approved):

PR #4606, opened 2024-12-06
Add optimizer pass to repopulate cardinalities + combine cardinalities in Logical Plan/Operator

royi-luo:
# Description

- The optimizer can add/remove new operators, we add a new pass at the end of the optimizer the recalculates all the cardinalities to make it correct for the new plan
- The cardinality variables in Logical plan/operator are redundant, this PR removes the one in plan (`LogicalPlan::getCardinality()` now gets the cardinality from the last operator)
- Populate cardinalities for a few more operators
  - Aggregate, accumulate
  - Load from (for sources parquet, numpy, pandas, pyarrow only)

# Contributor agreement


src/include/planner/join_order/cardinality_estimator.h
```diff
@@ -27,28 +27,31 @@ class CardinalityEstimator {
     void addNodeIDDomAndStats(const transaction::Transaction* transaction,
```
royi-luo, line 27: The main change in the cardinality estimator is to make all the functions take in a `LogicalOperator` as input instead of a plan, this allows it to work during the operator pass (when we want to estimate on cardinalities for operators in a plan instead of plans that are still in construction)

src/planner/plan/append_accumulate.cpp
```diff
@@ -36,6 +36,7 @@ void Planner::appendAccumulate(AccumulateType accumulateType, const expression_v
     appendFlattens(op->getGroupPositionsToFlatten(), plan);
     op->setChild(0, plan.getLastOperator());
     op->computeFactorizedSchema();
+    // TODO(Royi) set the cardinality here
     plan.setLastOperator(std::move(op));
```
royi-luo, line 40: How is the cardinality for accumulate calculated?

src/planner/plan/append_scan_node_table.cpp
```diff
@@ -26,8 +26,7 @@ void Planner::appendScanNodeTable(std::shared_ptr<Expression> nodeID,
         propertiesToScan_);
     scan->computeFactorizedSchema();
     // TODO(Guodong): Should remove cardinality from Plan.
```
royi-luo, line 28: remove

src/include/planner/operator/logical_aggregate.h
```diff
@@ -27,14 +27,16 @@ class LogicalAggregate final : public LogicalOperator {
     static constexpr LogicalOperatorType operatorType_ = LogicalOperatorType::AGGREGATE;
 
 public:
+    // TODO(Royi) make sure cardinality is being set correctly
     LogicalAggregate(binder::expression_vector keys, binder::expression_vector aggregates,
         std::shared_ptr<LogicalOperator> child)
-        : LogicalOperator{operatorType_, std::move(child)}, keys{std::move(keys)},
-          aggregates{std::move(aggregates)} {}
+        : LogicalOperator{operatorType_, std::move(child), aggregates.size()},
```
royi-luo, line 33: I'm assuming here that the cardinality of the aggregate is `aggregates.size()`, is this correct?

src/optimizer/cardinality_updater.cpp
```diff
@@ -0,0 +1,137 @@
...
+        *extend.getBoundNode(), transaction);
+    extend.setCardinality(cardinalityEstimator.estimateExtend(extensionRate, *op->getChild(0)));
+}
+
+void CardinalityUpdater::visitHashJoin(planner::LogicalOperator* op) {
+    auto& hashJoin = op->cast<planner::LogicalHashJoin&>();
+    KU_ASSERT(hashJoin.getNumChildren() >= 2);
+    hashJoin.setCardinality(cardinalityEstimator.estimateHashJoin(
+        hashJoin.isNodeIDOnlyJoin() ? hashJoin.getJoinNodeIDs() : binder::expression_vector{},
+        *hashJoin.getChild(0), *hashJoin.getChild(1)));
```
royi-luo, line 96: In the hash joins produced by the filter push down optimizer (I think those are the only ones with non-nodeID-only joins) how do we estimate the cardinalities?

src/include/planner/operator/logical_plan.h
```diff
@@ -12,22 +12,25 @@ class LogicalPlan {
     friend class CostModel;
 
 public:
-    LogicalPlan() : estCardinality{1}, cost{0} {}
```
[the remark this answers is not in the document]
royi-luo: changed default to 1 to match old behaviour

src/planner/plan/plan_join_order.cpp
```diff
@@ -265,7 +265,8 @@ void Planner::planNodeIDScan(uint32_t nodePos, const QueryGraphPlanningInfo& inf
     // later in the optimization stage. So we can assume the cardinality will not exceed outer
     // plan cardinality.
     if (plan->getCardinality() > info.corrExprsCard) {
-        plan->setCardinality(info.corrExprsCard);
+        plan->getLastOperator()->setCardinality(info.corrExprsCard);
+        cardinalityEstimator.addNodeIDDomOverride(*node->getInternalID(), info.corrExprsCard);
```
royi-luo, line 269: I was seeing a problem where this updated cardinality (which I guess is from us assuming we'll add a semi-mask later) isn't in sync with the denominator in the cardinality estimator so I added a second map with overrides for the denominators (that is reset for each query graph). Is this the right way to deal with this?

ray6080: If it is re-initialized for each query graph, can we directly update existing nodeID dom?

ray6080 (approved):

src/planner/plan/append_accumulate.cpp
royi-luo: cool, removed TODO

src/include/planner/operator/logical_aggregate.h
royi-luo: updated, cardinality will be 1 if `keys.size() == 0` and we'll take the child's card otherwise. Left a TODO to improve this estimation.

src/optimizer/cardinality_updater.cpp
royi-luo: update, for non-ID based joins for now we will go with the simplest approach of using the cross product cardinality multiplied by a magic constant

src/optimizer/optimizer.cpp
```diff
@@ -52,6 +55,9 @@ void Optimizer::optimize(planner::LogicalPlan* plan, main::ClientContext* contex
     // FactorizationRewriter.
     auto aggKeyDependencyOptimizer = AggKeyDependencyOptimizer();
     aggKeyDependencyOptimizer.rewrite(plan);
+
+    auto cardinalityUpdater = CardinalityUpdater(cardinalityEstimator, context->getTx());
```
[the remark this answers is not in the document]
royi-luo: updated

src/planner/plan/plan_join_order.cpp
royi-luo: Discussed offline, renamed to `addPerQueryGraphNodeIDDom` to be more clear

PR #4619, opened 2024-12-10
Add setting `enable_plan_optimizer`

royi-luo:
# Description

Add setting `enable_plan_optimizer` that allows us to control whether we want the plan optimizer to be enabled/disabled. Disabling the optimizer may be useful in cases such as when we wish to view logical plans (e.g. with `EXPLAIN LOGICAL`) before optimizations occur (e.g. when debugging join order).

# Contributor agreement

### DIFF
--- a/src/optimizer/optimizer.cpp
+++ b/src/optimizer/optimizer.cpp
@@ -15,43 +15,60 @@
 namespace kuzu {
 namespace optimizer {
 
-void Optimizer::optimize(planner::LogicalPlan* plan, main::ClientContext* context) {
-    // Factorization structure should be removed before further optimization can be applied.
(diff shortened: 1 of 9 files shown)


src/optimizer/optimizer.cpp
```diff
@@ -15,43 +15,60 @@
 namespace kuzu {
 namespace optimizer {
 
+static void populateSchemaRecursive(planner::LogicalOperator* op) {
```
andyfengHKU, line 18: I would use a child class of LogicalOperatorVisitor to do this.

andyfengHKU (approved):

ray6080 (approved):

PR #4624, opened 2024-12-11
Parallel K Core Decomposition Algorithm

WWW0030:
# Description
Parallel Implementation of K Core Decomposition Algorithm, which computes the maximum K Core value of all nodes within a given graph.

A node is said to be K Core if:
- It has at least K edges
- It is connected to at least K other K Core nodes

The following is a graph showcasing a graph's K-Core values:
image

# Algorithm Description
The algorithm utilizes the runUntilConvergence Algorithm twice, using different edge Computes to achieve different goals.
The first run uses the `KCoreInitEdgeCompute` to correctly initalize all nodes to their respective vertex degree
The second run uses the `KCoreEdgeCompute` to remove active nodes starting from smallest vertex degree and peeling until no more vertices are active.

# Possible race condition:
In one of the previous implementations, inwhich the smallest degree is updated by 1 everytime, a race condition can occur:
image
In essence, some vertices which forms a chain reaction of removing vertex degrees need the algorithm to stay on the same vertex degree for more than 1 iterations to compute. This is also what is happening within ligra (where the degree doesn't increase until all nodes of current degree are removed).

# Possible optimization
`updateSmallestDegree` can be optimized to run with vertexCompute (in parallel), though that might require a lot of changes, since the code will need to support vertexCompute between iterations of `runUntilConvergence`. 

```
class KCoreUpdateVertexCompute : public VertexCompute {
    // Runs between iterations
    // Finds the smallest degree to compute in the next vertex
    // Sets the nodes which need computation active inside the frontier
public:
    KCoreFrontierPair* kCoreFrontierPair;
public:
    KCoreUpdateVertexCompute(storage::MemoryManager* mm, processor::GDSCallSharedState* sharedState, KCoreFrontierPair* kCoreFrontierPair)
    : mm{mm}, sharedState{sharedState}, kCoreFrontierPair{kCoreFrontierPair} {}

    ~KCoreUpdateVertexCompute() override {}

    bool beginOnTable(common::table_id_t tableID) override {
        kCoreFrontierPair->pinCurrFrontier(tableID);
        kCoreFrontierPair->pinNextFrontier(tableID);
        return true;
    }

    void vertexCompute(common::offset_t startOffset, common::offset_t endOffset, common::table_id_t tableID) override {
        kCoreFrontierPair->updateSmallestDegree(startOffset, endOffset, tableID);
    }

    std::unique_ptr<VertexCompute> copy() override {
        return std::make_unique<KCoreUpdateVertexCompute>(mm, sharedState, kCoreFrontierPair);
    }

private:
    storage::MemoryManager* mm;
    processor::GDSCallSharedState* sharedState;
};
```
# Contributor agreement


src/function/gds/gds_utils.cpp
```diff
@@ -19,6 +21,19 @@ GDSComputeState::GDSComputeState(std::unique_ptr<function::FrontierPair> frontie
 
 GDSComputeState::~GDSComputeState() = default;
 
+void GDSUtils::initKCore(processor::ExecutionContext* context, RJCompState& rjCompState, graph::Graph* graph, std::unique_ptr<EdgeCompute> initEdgeCompute) {
```
andyfengHKU, line 24: Remove. GDSUtils should not hold any algorithm specific logic.

src/function/gds/K_Core_Decomposition.cpp
```diff
@@ -0,0 +1,192 @@
...
+struct KCoreInitEdgeCompute : public EdgeCompute {
+    KCoreFrontierPair* frontierPair;
+
+    explicit KCoreInitEdgeCompute(KCoreFrontierPair* frontierPair) : frontierPair{frontierPair} {}
+
+    std::vector<common::nodeID_t> edgeCompute(common::nodeID_t boundNodeID, graph::NbrScanState::Chunk& chunk, bool) override {
+        std::vector<common::nodeID_t> result;
+        uint64_t nbrAmount = 0;
+        chunk.forEach([&](auto nbrNodeID, auto) {nbrAmount++;
+        result.push_back(nbrNodeID);});
```
andyfengHKU, line 29: why do we need to put them in result? As I understand, this EdgeCompute is to init degree and should run for only 1 iteration.

src/include/function/gds/gds_frontier.h
```diff
@@ -319,6 +393,74 @@ class KUZU_API DoublePathLengthsFrontierPair : public FrontierPair {
...
+    }
+
+    uint64_t getSmallestDegree() {
+        return curSmallestDegree.load(std::memory_order_relaxed);
+    }
+
+private:
+    std::atomic<uint64_t>& getVertexDegreeAtomic(common::nodeID_t nodeID) const {
+        auto& memBuffer = curVertexValues.find(nodeID.tableID)->second;
+        std::atomic<uint64_t>* memBufferPtr = reinterpret_cast<std::atomic<uint64_t>*>(memBuffer->getData());
```
andyfengHKU, line 454: this should be saved as a member field and set in beginFrontierComputeBetweenTables. See how WCC handles this.

src/include/function/gds/gds_frontier.h
```diff
@@ -319,6 +393,74 @@ class KUZU_API DoublePathLengthsFrontierPair : public FrontierPair {
     void beginNewIterationInternalNoLock() override;
 };
 
+class KUZU_API KCoreFrontierPair : public FrontierPair {
```
andyfengHKU, line 396: This should be moved to k_core_decomposition.cpp.

src/include/function/gds/gds_frontier.h
```diff
@@ -319,6 +393,74 @@ class KUZU_API DoublePathLengthsFrontierPair : public FrontierPair {
...
+            return getVertexDegreeAtomic(nodeID).fetch_sub(1, std::memory_order_relaxed);
+        }
+        return curVertexDegree;
+    }
+
+    uint64_t getVertexDegree(common::nodeID_t nodeID) const {
+        return getVertexDegreeAtomic(nodeID).load(std::memory_order_relaxed);
+    }
+
+    bool isNodeActive(common::nodeID_t nodeID) const {
```
andyfengHKU, line 423: Why not use/overwrite isActive/setActive but instead create isNodeActive/setNodeActive?

andyfengHKU (commented):
Btw I would take a look at the implementation here

https://github.com/jshun/ligra/blob/master/apps/KCore.C

src/function/gds/gds_utils.cpp
WWW0030: Gotcha, found a way to use runUntilConvergence to do the calculation instead

src/function/gds/K_Core_Decomposition.cpp
```diff
@@ -0,0 +1,324 @@
...
+        for (const auto& [tableID, curNumNodes] : numNodesMap) {
+            curVertexValues.allocate(tableID, curNumNodes, mm);
+            auto data = curVertexValues.getData(tableID);
+            for (auto i = 0u; i < curNumNodes; ++i) {
+                data[i].store(0, std::memory_order_relaxed);
+            }
+        }
+    }
+
+    void initRJFromSource(common::nodeID_t /* source */) override { setActiveNodesForNextIter(); };
```
andyfengHKU, line 37: is this used? If not, why implement this function with `setActiveNodesForNextIter();`

src/function/gds/K_Core_Decomposition.cpp
```diff
@@ -0,0 +1,324 @@
...
+struct KCoreInitEdgeCompute : public EdgeCompute {
+    KCoreFrontierPair* frontierPair;
+
+    explicit KCoreInitEdgeCompute(KCoreFrontierPair* frontierPair) : frontierPair{frontierPair} {}
+
+    std::vector<common::nodeID_t> edgeCompute(common::nodeID_t boundNodeID,
+        graph::NbrScanState::Chunk& chunk, bool) override {
+        std::vector<common::nodeID_t> result;
+        uint64_t nbrAmount = 0;
+        chunk.forEach([&](auto /* nbrNodeID */, auto) { nbrAmount++; });
```
andyfengHKU, line 141: Just use the size of chunk instead looping through it.

src/function/gds/K_Core_Decomposition.cpp
```diff
@@ -0,0 +1,324 @@
...
+
+    void beginNewIterationInternalNoLock() override {
+        std::swap(curDenseFrontier, nextDenseFrontier);
+        updateSmallestDegree();
+        curDenseFrontier->ptrCast<PathLengths>()->incrementCurIter();
+        nextDenseFrontier->ptrCast<PathLengths>()->incrementCurIter();
+    }
+
+    uint64_t addToVertexDegree(common::nodeID_t nodeID, uint64_t degreeToAdd) {
+        return curVertexValues.getData(nodeID.tableID)[nodeID.offset].fetch_add(degreeToAdd,
```
andyfengHKU, line 66: curVertexValues should have been pinned when u called this function. So u should only use `vertexValues` and `nodeID.offset`

src/function/gds/K_Core_Decomposition.cpp
```diff
@@ -0,0 +1,324 @@
...
+        int curVertexDegree =
+            curVertexValues.getData(nodeID.tableID)[nodeID.offset].load(std::memory_order_relaxed);
+        if (curVertexDegree > curSmallest) {
+            return curVertexValues.getData(nodeID.tableID)[nodeID.offset].fetch_sub(1,
+                std::memory_order_relaxed);
+        }
+        return curVertexDegree;
+    }
+
+    void updateSmallestDegree(bool dontCheckOffset = false) {
```
andyfengHKU, line 81: Seems that this `dontCheckOffset` is always false from where u call this function. If so, clean this up.

src/function/gds/K_Core_Decomposition.cpp
```diff
@@ -0,0 +1,324 @@
...
+            sharedState->getOutputNodeMaskMap(), frontierPair.get());
+        auto computeState = GDSComputeState(std::move(frontierPair), std::move(initEdgeCompute),
+            sharedState->getOutputNodeMaskMap());
+
+        GDSUtils::runFrontiersUntilConvergence(context, computeState, graph, ExtendDirection::BOTH,
+            1);
+        auto edgeCompute = std::make_unique<KCoreEdgeCompute>(
+            computeState.frontierPair->ptrCast<KCoreFrontierPair>());
+        computeState.edgeCompute = std::move(edgeCompute);
+        computeState.frontierPair->ptrCast<KCoreFrontierPair>()->updateSmallestDegree();
```
andyfengHKU, line 299: Note this is not parallelized. It should be done through vertex compute.

src/function/gds/K_Core_Decomposition.cpp
```diff
@@ -0,0 +1,324 @@
...
+    bool isNodeActive(common::nodeID_t nodeID) const {
+        return curDenseFrontier->isActive(nodeID.offset);
+    }
+
+    uint64_t getVertexValue(common::offset_t offset) {
+        return vertexValues[offset].load(std::memory_order_relaxed);
+    }
+
+private:
+    bool updateDegreeFlag = false;
```
andyfengHKU, line 125: Double check if this is actually needed. My feeling after reading through this PR is that you can optimize your functions call chain and remove this flag.

src/function/gds/K_Core_Decomposition.cpp
```diff
@@ -0,0 +1,324 @@
...
+            }
+        }
+    }
+
+    uint64_t getSmallestDegree() { return curSmallestDegree.load(std::memory_order_relaxed); }
+
+    void setUpdateFlag() { updateDegreeFlag = true; }
+
+    bool isNodeActive(common::nodeID_t nodeID) const {
+        return curDenseFrontier->isActive(nodeID.offset);
```
andyfengHKU, line 117: again why not use isActive interface in GDSFrontier and yet expose an identical one in this class?

src/function/gds/K_Core_Decomposition.cpp
```diff
@@ -0,0 +1,324 @@
...
+    KCoreFrontierPair* frontierPair;
+
+    explicit KCoreEdgeCompute(KCoreFrontierPair* frontierPair) : frontierPair{frontierPair} {}
+
+    std::vector<common::nodeID_t> edgeCompute(common::nodeID_t boundNodeID,
+        graph::NbrScanState::Chunk& chunk, bool) override {
+        std::vector<common::nodeID_t> result;
+        uint64_t vertexDegree = frontierPair->getVertexValue(boundNodeID.offset);
+        uint64_t smallestDegree = frontierPair->getSmallestDegree();
+        if (frontierPair->isNodeActive(boundNodeID) && (vertexDegree == smallestDegree)) {
```
andyfengHKU, line 162: If you are able to enter edgeCompute then boundNodeID is guaranteed to be active right?

src/function/gds/K_Core_Decomposition.cpp
```diff
@@ -0,0 +1,324 @@
...
+            chunk.forEach([&](auto nbrNodeID, auto) {
+                if (frontierPair->isNodeActive(nbrNodeID)) {
+                    frontierPair->removeFromVertex(nbrNodeID);
+                    result.push_back(nbrNodeID);
```
andyfengHKU, line 166: You are making every node whose degree has been updated (fetch_sub(1)) as active nodes in next frontier. Though I think you only need those whose degree equals to new smallest degree as active nodes right? 

In K-core algorithm, as I understand, frontier size should keep decreasing because as K increases, more nodes are removed. Your implementation doesn't seem to have such property.

Also, I believe there is racing condition here. You are reading vertexDegree in l160 and updating vertexDegree in l165. There is no guarantee that the value in l160 has not been updated by the time you use it.

andyfengHKU (changes requested):
I think there is still more work & test needs to be done in this PR.

If my comment of racing condition is true, add test case to cover it and then fix it. Otherwise comment how you provide the concurrency guarantee.

Also, the idea of smallest degree is fine. But to compute smallest degree is still potential work, especially when the number of nodes are huge. Why not go through K 1 by 1 so you can omit that compute. In practice, your smallest degrees will most likely increase by 1 anyways. In the most optimal case, your smallest degree increase by more than 1, then doing an empty frontier compute also won't cost much.

Again I suggest read the logic in this implementation https://github.com/jshun/ligra/blob/master/apps/KCore.C

WWW0030:
> I think there is still more work & test needs to be done in this PR.
> 
> If my comment of racing condition is true, add test case to cover it and then fix it. Otherwise comment how you provide the concurrency guarantee.
> 
> Also, the idea of smallest degree is fine. But to compute smallest degree is still potential work, especially when the number of nodes are huge. Why not go through K 1 by 1 so you can omit that compute. In practice, your smallest degrees will most likely increase by 1 anyways. In the most optimal case, your smallest degree increase by more than 1, then doing an empty frontier compute also won't cost much.
> 
> Again I suggest read the logic in this implementation https://github.com/jshun/ligra/blob/master/apps/KCore.C

Thanks for the feedback, I explained concurrency guarantee in my previous comment. I do agree with your idea of smallest degree and have changed implementation to follow that. 
Thanks for pointing out ligra's implementation, I think that it is slightly different than neo4j's implementation in that they return different results. Could you also take a look at neo4j's k-core and let me know which one we perfer or maybe both?

https://neo4j.com/docs/graph-data-science/current/algorithms/k-core/

src/function/gds/K_Core_Decomposition.cpp
```diff
@@ -0,0 +1,294 @@
...
+
+struct KCoreInitEdgeCompute : public EdgeCompute {
+    KCoreFrontierPair* frontierPair;
+
+    explicit KCoreInitEdgeCompute(KCoreFrontierPair* frontierPair) : frontierPair{frontierPair} {}
+
+    std::vector<common::nodeID_t> edgeCompute(common::nodeID_t boundNodeID,
+        graph::NbrScanState::Chunk& chunk, bool) override {
+        std::vector<common::nodeID_t> result;
+        result.push_back(boundNodeID);
```
andyfengHKU, line 112: Why do we need to push `BoundNodeID` into result?

src/function/gds/K_Core_Decomposition.cpp
```diff
@@ -0,0 +1,294 @@
...
+        if (curVertexDegree > curSmallest) {
+            curVertexValues.getData(nodeID.tableID)[nodeID.offset].fetch_sub(1,
+                std::memory_order_relaxed);
+            return true;
+        } else {
+            return false;
+        }
+    }
+
+    void updateSmallestDegree() { curSmallestDegree.fetch_add(1, std::memory_order_relaxed); }
```
andyfengHKU, line 88: this is no longer accurate. Rename to increaseSmallestDegree()

src/function/gds/K_Core_Decomposition.cpp
```diff
@@ -0,0 +1,294 @@
...
+
+    uint64_t getSmallestDegree() { return curSmallestDegree.load(std::memory_order_relaxed); }
+
+    uint64_t getVertexValue(common::offset_t offset) {
+        return vertexValues[offset].load(std::memory_order_relaxed);
+    }
+
+private:
+    bool updated = false;
```
andyfengHKU, line 97: I have made a comment about this. Let's make sure all comments are resolved before you hold them

src/function/gds/K_Core_Decomposition.cpp
andyfengHKU: Let's discuss in person. I don't understand the difference between the output of Ligra & Neo4j.

Let me give a more concrete example of racing condition.

Suppose node A has degree 5 and node B has degree 4 in current iteration. There is one edge between A&B. And we are looking for smallestDegree == 4. 

Logically, you want to only consider B in the current iteration (core). However, in a racing condition, you might first visit B, compute it, removeFromVertex for its neighbours, i.e. A and reduce A's degree to 4. Then u see A, which now has degree == 4 and you will consider A in current iteration as well.

Is above a legit racing condition? If so, add a test case to reveal it and then fix it.

src/function/gds/K_Core_Decomposition.cpp
WWW0030: I reverted updateSmallestDegree back to manually finding the next smallest degree insteead of fetch_adding 1 everytime. The fetch_add has a chance to cause a race condition, I  put explanation in the description of the PR

PR #4633, opened 2024-12-14
Skip query result from rewritten queries

ray6080:
# Description

Query results from rewritten queries should only be kept as internal one, but not exposed as query result to users.

### DIFF
--- a/src/include/parser/statement.h
+++ b/src/include/parser/statement.h
@@ -8,13 +8,16 @@ namespace parser {
 
 class Statement {
 public:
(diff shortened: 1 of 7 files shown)


src/include/parser/statement.h
```diff
@@ -38,6 +41,7 @@ class Statement {
 
 private:
     common::StatementType statementType;
+    bool skipResult_;
```
andyfengHKU, line 44: I would call this `internal_`. And also comment, internal statements are those rewritten from input and whose result should not be rendered on client side.

andyfengHKU (approved):
Add tests if possible.

PR #4646, opened 2024-12-17
Support IGNORE_ERRORS when scanning from pyarrow/pandas

### LINKED ISSUE #4534: Feature: Support `IGNORE_ERRORS` when scanning from in-memory sources
### API

Python

### Description

`IGNORE_ERRORS` currently doesn't work when scanning from in-memory sources such as Dataframes/Polars tables and subqueries, we should make this work.

royi-luo:
# Description

Adds support for the default `IGNORE_ERRORS` behaviour (skipping rows that trigger PK-related exceptions) when scanning from pandas/pyarrow.

Fixes #4534 

# Contributor agreement

### DIFF
--- a/src/binder/bound_scan_source.cpp
+++ b/src/binder/bound_scan_source.cpp
@@ -13,14 +13,9 @@ expression_vector BoundTableScanSource::getWarningColumns() const {
     }
     return warningDataExprs;
 }
+
 bool BoundTableScanSource::getIgnoreErrorsOption() const {
-    bool ignoreErrors = common::CopyConstants::DEFAULT_IGNORE_ERRORS;
-    if (type == common::ScanSourceType::FILE) {
-        auto bindData = info.bindData->constPtrCast<function::ScanBindData>();
-        ignoreErrors = bindData->config.getOption(common::CopyConstants::IGNORE_ERRORS_OPTION_NAME,
-            ignoreErrors);
-    }
-    return ignoreErrors;
+    return info.bindData->getIgnoreErrorsOption();
 }
 
 } // namespace binder
--- a/tools/python_api/src_cpp/pyarrow/pyarrow_scan.cpp
+++ b/tools/python_api/src_cpp/pyarrow/pyarrow_scan.cpp
@@ -5,6 +5,7 @@
 #include "common/arrow/arrow_converter.h"
 #include "function/cast/functions/numeric_limits.h"
 #include "function/table/bind_input.h"
+#include "processor/execution_context.h"
 #include "py_connection.h"
 #include "pyarrow/pyarrow_bind.h"
 #include "pybind11/pytypes.h"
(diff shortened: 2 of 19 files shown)


CMakeLists.txt
```diff
@@ -317,7 +317,7 @@ add_subdirectory(third_party)
 if(${BUILD_KUZU})
 add_definitions(-DKUZU_ROOT_DIRECTORY="${PROJECT_SOURCE_DIR}")
 add_definitions(-DKUZU_CMAKE_VERSION="${CMAKE_PROJECT_VERSION}")
-add_definitions(-DKUZU_EXTENSION_VERSION="0.7.0")
```
ray6080, line 320: hmm Do we need to upgrade the extension version?

src/binder/bound_scan_source.cpp
```diff
@@ -14,13 +14,7 @@ expression_vector BoundTableScanSource::getWarningColumns() const {
     return warningDataExprs;
 }
```
ray6080, line 15: Add an empty line between these two functions.

tools/python_api/src_cpp/pyarrow/pyarrow_scan.cpp
```diff
@@ -17,6 +18,7 @@ namespace kuzu {
 PyArrowScanConfig::PyArrowScanConfig(const common::case_insensitive_map_t<Value>& options) {
```
ray6080, line 18: I wonder why PyArrowScan and PandasScan have different ways to convert scanConfig. I need to take another look to see if we can unify them. but can you also take a look?

ray6080 (commented):
There are several KUZU_API changes which I don't follow why they're now needed in this PR?

CMakeLists.txt
royi-luo: I added a virtual function to `TableFuncBindData` which is inherited by some classes in the extensions (e.g. JSONScanBindData) so I probably need to bump the extension version.

That single change (me needing to export `ScanBindData/TableFuncBindData) caused a bunch of dynamic casts in the macos extension test to fail due to the casted targets not being exported which is why I added all those `KUZU_API` changes. I'm not completely sure why they showed up now and not before though.

tools/python_api/src_cpp/pyarrow/pyarrow_scan.cpp
royi-luo: I think the reason is that we don't support the pyarrow scan options (`skipNum` and `limitNum`) in pandas scan so before this PR we don't actually support any configuration for pandas scan. I could look into if it's doable to support those options in pandas scan and if it is I'll add the support and unify the config in a separate PR.

CMakeLists.txt
ray6080: okay. We don't need to bump extension version manually in PRs now. The CI pipeline will handle this.

ray6080: @acquamarin Can you also take a look at the KUZU_API changes in this PR?

src/binder/bound_scan_source.cpp
royi-luo: updated

CMakeLists.txt
royi-luo: It probably because the only virtual method before is `copy` which is pure virtual (so the definitions are all in the child classes). The method I added has a default implementation so it actually needs to exported for child classes in the extensions to access.

ray6080 (approved):

PR #4648, opened 2024-12-18
Remove offset from table func

andyfengHKU:
# Description

See title

Fixes # (issue)

# Contributor agreement

### DIFF
--- a/src/include/binder/copy/bound_copy_from.h
+++ b/src/include/binder/copy/bound_copy_from.h
@@ -22,9 +22,9 @@ struct ExtraBoundCopyFromInfo {
 struct BoundCopyFromInfo {
     // Table entry to copy into.
     catalog::TableCatalogEntry* tableEntry;
-    // Data source
+    // Data source.
     std::unique_ptr<BoundBaseScanSource> source;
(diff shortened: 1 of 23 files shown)


src/include/binder/copy/bound_copy_from.h
```diff
@@ -22,9 +22,9 @@ struct ExtraBoundCopyFromInfo {
 struct BoundCopyFromInfo {
     // Table entry to copy into.
     catalog::TableCatalogEntry* tableEntry;
-    // Data source
+    // Data source.
     std::unique_ptr<BoundBaseScanSource> source;
-    // Row offset of input data to generate internal ID.
+    // Row offset.
     std::shared_ptr<Expression> offset;
```
ray6080, line 28: Can you revisit if we still need this expression? Or if there is a way we can also get rid of this one?

ray6080 (approved):

src/include/binder/copy/bound_copy_from.h
andyfengHKU: I'm inclined we still need one. Because the columns are generated by scan operator. And we don't want every scan operator to generate an offset expression by default.

PR #4650, opened 2024-12-18
Remove unnecessary sink in copy from subquery

andyfengHKU:
# Description

Sanity check benchmark. On `wr1` when copying LDBC100 comment. The time goes from 38s to 28s

Fixes # (issue)

# Contributor agreement


ray6080 (approved):
Can you rename the PR title to be more explicit? e.g., "Remove unnecessary sink in COPY from subqueries". I would update the PR description to document that we now remove sink in COPY from subquery but only when the intermediate result is all in a single factorized group. Also it would be great to include some performance numbers on very simple queries. e.g., `COPY person2 FROM (MATCH (p:person) RETURN  p.*);`

PR #4661, opened 2024-12-20
KCore Parallel implementation

WWW0030:
# Description
Parallel Implementation of K Core Decomposition Algorithm, which computes the maximum K Core value of all nodes within a given graph.

Previous PR: https://github.com/kuzudb/kuzu/pull/4624

A node is said to be K Core if:
- It has at least K edges
- It is connected to at least K other K Core nodes

The following is a graph showcasing a graph's K-Core values:
image

# Algorithm Description
The algorithm utilizes the runUntilConvergence Algorithm twice, using different edge Computes to achieve different goals.
The first run uses the `KCoreInitEdgeCompute` to correctly initalize all nodes to their respective vertex degree
The second run uses the `KCoreEdgeCompute` to remove active nodes starting from smallest vertex degree and peeling until no more vertices are active.

# Possible race condition:
In one of the previous implementations, inwhich the smallest degree is updated by 1 everytime, a race condition can occur:
image
In essence, some vertices which forms a chain reaction of removing vertex degrees need the algorithm to stay on the same vertex degree for more than 1 iterations to compute. This is also what is happening within ligra (where the degree doesn't increase until all nodes of current degree are removed).

# Possible optimization
`updateSmallestDegree` can be optimized to run with vertexCompute (in parallel), though that might require a lot of changes, since the code will need to support vertexCompute between iterations of `runUntilConvergence`. 

```
class KCoreUpdateVertexCompute : public VertexCompute {
    // Runs between iterations
    // Finds the smallest degree to compute in the next vertex
    // Sets the nodes which need computation active inside the frontier
public:
    KCoreFrontierPair* kCoreFrontierPair;
public:
    KCoreUpdateVertexCompute(storage::MemoryManager* mm, processor::GDSCallSharedState* sharedState, KCoreFrontierPair* kCoreFrontierPair)
    : mm{mm}, sharedState{sharedState}, kCoreFrontierPair{kCoreFrontierPair} {}

    ~KCoreUpdateVertexCompute() override {}

    bool beginOnTable(common::table_id_t tableID) override {
        kCoreFrontierPair->pinCurrFrontier(tableID);
        kCoreFrontierPair->pinNextFrontier(tableID);
        return true;
    }

    void vertexCompute(common::offset_t startOffset, common::offset_t endOffset, common::table_id_t tableID) override {
        kCoreFrontierPair->updateSmallestDegree(startOffset, endOffset, tableID);
    }

    std::unique_ptr<VertexCompute> copy() override {
        return std::make_unique<KCoreUpdateVertexCompute>(mm, sharedState, kCoreFrontierPair);
    }

private:
    storage::MemoryManager* mm;
    processor::GDSCallSharedState* sharedState;
};
```
# Contributor agreement


andyfengHKU (commented):
hmmm it seems two of us are not on the same page regarding this k-core algorithm. I tried the Neo4j example as shown [here](https://neo4j.com/docs/graph-data-science/current/algorithms/k-core/?utm_source=GSearch&utm_medium=PaidSearch&utm_campaign=Evergreen&utm_content=AMS-Search-SEMCE-DSA-None-SEM-SEM-NonABM&utm_term=&utm_adgroup=DSA&gad_source=1&gbraid=0AAAAADk9OYrX-dwQ_ribFK1kM47oBLC3x&gclid=CjwKCAiA65m7BhAwEiwAAgu4JEeqU34-Z45mVkdhppCgjwZV5hrRI6JGHrN_E78QYqlp-oJbnKsQaxoC8JUQAvD_BwE). And your's is not generating the same output as Neo4j, can u clarify this in the comment and let me know?

PR #4662, opened 2024-12-20
Support skip/limit options for pandas scan

royi-luo:
# Description

Add support for skipping rows + limiting number of rows to scan when scanning from pandas dataframes.
Also reuse `PyarrowScanConfig` for pandas scan config and rename to more appropriate name (since it isn't pyarrow only)

# Contributor agreement

### DIFF
--- a/tools/python_api/src_cpp/pandas/pandas_scan.cpp
+++ b/tools/python_api/src_cpp/pandas/pandas_scan.cpp
@@ -34,7 +34,8 @@ std::unique_ptr<TableFuncBindData> bindFunc(ClientContext* /*context*/,
     auto numRows = py::len(getFunc(columns[0]));
     auto returnColumns = input->binder->createVariables(names, returnTypes);
     auto scanConfig = PyScanConfig{
-        input->extraInput->constPtrCast<ExtraScanTableFuncBindInput>()->fileScanInfo.options};
+        input->extraInput->constPtrCast<ExtraScanTableFuncBindInput>()->fileScanInfo.options,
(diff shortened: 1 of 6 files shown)


tools/python_api/src_cpp/pandas/pandas_scan.cpp
```diff
@@ -32,10 +33,12 @@ std::unique_ptr<TableFuncBindData> bindFunc(ClientContext* /*context*/,
     auto getFunc = df.attr("__getitem__");
     auto numRows = py::len(getFunc(columns[0]));
     auto returnColumns = input->binder->createVariables(names, returnTypes);
-    auto scanConfig =
-        input->extraInput->constPtrCast<ExtraScanTableFuncBindInput>()->fileScanInfo.copy();
-    return std::make_unique<PandasScanFunctionData>(std::move(returnColumns), df, numRows,
-        std::move(columnBindData), std::move(scanConfig));
+    auto scanConfig = PyScanConfig{
+        input->extraInput->constPtrCast<ExtraScanTableFuncBindInput>()->fileScanInfo.options};
+    KU_ASSERT(numRows >= scanConfig.skipNum);
```
ray6080, line 38: hmm Do we have a check somewhere to ensure `scanConfig.skipNum` is always smaller than `numRows`? I'm concerned if this will lead to overflow under the case `numRows < scanConfig.skipNum`. 

I would also add a test case on that if we don't have one already.

ray6080 (approved):
Looks great. Please check my comment below.

PR #4666, opened 2024-12-24
Refactor scalar function constructor

andyfengHKU:
# Description

This is a followed up PR of #4664. The goal is to reduce the number of constructors for scalar function.

Fixes # (issue)

# Contributor agreement


ray6080 (approved):
Can you update the PR description? Let's not use the template as it is.
Why do we choose to move `bindFunc` outside of constructor?

PR #4625, opened 2024-12-11
Allow reordering of joins in `Planner::planRegularMatch`

royi-luo:
# Description
For multi-part queries in cases where there is no dependency between the inner and outer plans we can choose to reorder the join if it produces a plan with lower cost.

# Contributor agreement

- [ ] I have read and agree to the [Contributor Agreement](https://github.com/kuzudb/kuzu/blob/master/CLA.md).

PR #4668, opened 2024-12-28
Hash join flatten fix

andyfengHKU:
# Description

This PR fixes a bug in hash join planning exposed in PR #4625. Below is a description of the bug.

We append flatten to probe side if and only if join key is an internal ID and this ID is guaranteed to be unique on the build side. The check was done based on number of factorization group. This is not true if an aggregation happens, which will always reduce the number of factorization group to 1.

This PR checks by looking at operators along the build side subplan. 

### Testing

The bug cannot be revealed easily on master because we don't reorder multi-part queries. So aggregation does not happen on build side. I have tested by cherry-picking to #4625. 

Fixes # (issue)

# Contributor agreement

### DIFF
--- a/src/planner/operator/logical_hash_join.cpp
+++ b/src/planner/operator/logical_hash_join.cpp
@@ -1,6 +1,5 @@
 #include "planner/operator/logical_hash_join.h"
 
-#include "common/cast.h"
 #include "planner/operator/factorization/flatten_resolver.h"
 #include "planner/operator/factorization/sink_util.h"
 #include "planner/operator/scan/logical_scan_node_table.h"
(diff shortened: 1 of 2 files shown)


src/planner/operator/logical_hash_join.cpp
```diff
@@ -165,6 +165,24 @@ binder::expression_vector LogicalHashJoin::getJoinNodeIDs(
     return result;
 }
 
+class JoinNodeIDUniquenessAnalyzer {
+public:
+    bool isUnique(LogicalOperator* op, const binder::Expression& joinNodeID) {
```
ray6080, line 170: Why not make this function static?

ray6080 (approved):

src/planner/operator/logical_hash_join.cpp
andyfengHKU: Done

PR #4670, opened 2024-12-29
Refactor table func interface with const params

ray6080:
# Description

See title.

### DIFF
--- a/src/include/function/table_functions.h
+++ b/src/include/function/table_functions.h
@@ -73,20 +73,20 @@ struct TableFunctionInitInput {
 };
 
 using table_func_bind_t = std::function<std::unique_ptr<TableFuncBindData>(main::ClientContext*,
-    function::TableFuncBindInput*)>;
-using table_func_t = std::function<common::offset_t(TableFuncInput&, TableFuncOutput&)>;
+    const TableFuncBindInput*)>;
+using table_func_t = std::function<common::offset_t(const TableFuncInput&, TableFuncOutput&)>;
 using table_func_init_shared_t =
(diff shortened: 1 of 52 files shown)


src/include/function/table_functions.h
```diff
@@ -73,20 +73,20 @@ struct TableFunctionInitInput {
...
-    std::function<std::unique_ptr<TableFuncSharedState>(TableFunctionInitInput&)>;
+    std::function<std::unique_ptr<TableFuncSharedState>(const TableFunctionInitInput&)>;
 using table_func_init_local_t = std::function<std::unique_ptr<TableFuncLocalState>(
-    TableFunctionInitInput&, TableFuncSharedState*, storage::MemoryManager*)>;
+    const TableFunctionInitInput&, TableFuncSharedState*, storage::MemoryManager*)>;
 using table_func_can_parallel_t = std::function<bool()>;
 using table_func_progress_t = std::function<double(TableFuncSharedState* sharedState)>;
 using table_func_finalize_t =
-    std::function<void(processor::ExecutionContext*, TableFuncSharedState*)>;
+    std::function<void(const processor::ExecutionContext*, TableFuncSharedState*)>;
```
andyfengHKU, line 85: Mark input as const is fine. I'm slightly worried about marking ExecutionContext & ClientContext as const because some table function might modify it, e.g. if we keep error message in ExecutionContext then `clearWarning` will try to modify it.

No need to revert these changes, we can always change on demand.

andyfengHKU (approved):

src/include/function/table_functions.h
ray6080: hmm maybe, but for now I don't think we do that in `finalize`. Let's see how this evolves later.

PR #4671, opened 2024-12-30
Add more complete order by key types check

ray6080:
# Description

Add more binder type checks for order by key expression.
Follow-up from the conversion in #4669.

### DIFF
--- a/src/binder/bind/bind_projection_clause.cpp
+++ b/src/binder/bind/bind_projection_clause.cpp
@@ -12,7 +12,6 @@ using namespace kuzu::parser;
 
 namespace kuzu {
 namespace binder {
-
 void validateColumnNamesAreUnique(const std::vector<std::string>& columnNames) {
     auto existColumnNames = std::unordered_set<std::string>();
     for (auto& name : columnNames) {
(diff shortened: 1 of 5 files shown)


src/binder/bind/bind_projection_clause.cpp
```diff
@@ -230,16 +229,28 @@ expression_vector Binder::bindOrderByExpressions(
...
+        if (!isOrderByKeyTypeSupported(expr->dataType)) {
+            throw BinderException(stringFormat("Cannot order by {}. Order by {} is not supported.",
+                expr->toString(), expr->dataType.toString()));
         }
         exprs.push_back(std::move(expr));
     }
     return exprs;
 }
 
+bool Binder::isOrderByKeyTypeSupported(const LogicalType& dataType) {
```
andyfengHKU, line 241: Can u also open an issue about supporting more order by key types and assign to ziyi. I believe many of them can be naturally supported since we already support hash & compare

andyfengHKU (approved):

src/binder/bind/bind_projection_clause.cpp
ray6080: I think it's more about how to encode different types of keys, but all should be doable.

ray6080: Updated the existing issue https://github.com/kuzudb/kuzu/issues/1212

PR #4674, opened 2025-01-02
Rework auto begin/commit/rollback transaction during query lifetime

ray6080:
# Description

This PR fixes two problems related to the usage of transactions:
1. Parser relies on transaction to resolve data type for default null value, which should be done in the binder phase.
2. Parser and prepare unnecessarily starts and commits transactions, leading to multiple transactions in a single statement.

For the first problem, the PR moves resolving data type of default null from Parser to Binder, thus, Parser doesn't rely on transactions.
For the second problem, the PR unifies how transactions should auto begin and commit/rollback in `TransactionHelper::runFuncInTransaction()`, and avoids unnecessary begin/commits of transactions.

### DIFF
--- a/src/binder/bind/bind_ddl.cpp
+++ b/src/binder/bind/bind_ddl.cpp
@@ -51,15 +51,36 @@ static void validateSerialNoDefault(const Expression& expr) {
     }
 }
 
+// TODO(Ziyi/Xiyang/Guodong): Ideally this function can be removed, and we can make use of
+// `implicitCastIfNecessary` to handle this. However, the current approach doesn't keep casted
+// boundExpr, instead the ParsedExpression, which is not casted, in PropertyDefinition. We have to
+// change PropertyDefinition to store boundExpr to get rid of this function.
+static void tryResolvingDataTypeForDefaultNull(const ParsedPropertyDefinition& parsedDefinition,
+    const LogicalType& type) {
+    if (parsedDefinition.defaultExpr->getExpressionType() == ExpressionType::LITERAL) {
+        auto& literalVal =
+            common::ku_dynamic_cast<ParsedLiteralExpression*>(parsedDefinition.defaultExpr.get())
+                ->getValueUnsafe();
+        if (literalVal.isNull() &&
+            literalVal.getDataType().getLogicalTypeID() == LogicalTypeID::ANY) {
+            literalVal.setDataType(type);
+        }
+        KU_ASSERT(literalVal.getDataType().getLogicalTypeID() != LogicalTypeID::ANY);
+    }
+}
+
 std::vector<PropertyDefinition> Binder::bindPropertyDefinitions(
     const std::vector<ParsedPropertyDefinition>& parsedDefinitions, const std::string& tableName) {
     std::vector<PropertyDefinition> definitions;
     for (auto& parsedDefinition : parsedDefinitions) {
         auto type = LogicalType::convertFromString(parsedDefinition.getType(), clientContext);
+        // For default null value, we may need to resolve its data type, as its type may not be
+        // resolved during parsing and thus may be ANY.
+        tryResolvingDataTypeForDefaultNull(parsedDefinition, type);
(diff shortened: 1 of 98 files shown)


src/include/main/client_context.h
```diff
@@ -138,26 +138,52 @@ class KUZU_API ClientContext {
...
-    std::unique_ptr<PreparedStatement> preparedStatementWithError(std::string_view errMsg);
+    std::vector<std::shared_ptr<parser::Statement>> parseQuery(std::string_view query);
 
     // when we do prepare, we will start a transaction for the query
     // when we execute after prepare in a same context, we set requireNewTx to false and will not
     // commit the transaction in prepare when we only prepare a query statement, we set requireNewTx
     // to true and will commit the transaction in prepare
     std::unique_ptr<PreparedStatement> prepareNoLock(
-        std::shared_ptr<parser::Statement> parsedStatement, bool requireNewTx = true,
+        std::shared_ptr<parser::Statement> parsedStatement, bool shouldCommitNewTransaction = true,
```
andyfengHKU, line 186: Let's not give default value to `shouldCommitNewTransaction` because it's not true to most function call will set it true. Also the comment is no longer accurate.

src/main/client_context.cpp
```diff
@@ -309,242 +372,202 @@ std::unique_ptr<PreparedStatement> ClientContext::preparedStatementWithError(
...
+}
+
 // If there is an active transaction in the context, we execute the function in current active
 // transaction. If there is no active transaction, we start an auto commit transaction.
-void ClientContext::runFuncInTransaction(const std::function<void()>& fun) {
-    // check if we are on AutoCommit. In this case we should start a transaction
-    bool startNewTrx = !transactionContext->hasActiveTransaction();
-    if (startNewTrx) {
-        transactionContext->beginAutoTransaction(false /* readOnlyStatement */);
+void ClientContext::TransactionHelper::runFuncInTransaction(TransactionContext& context,
```
andyfengHKU, line 548: I'm fine with this design. Though many usage, e.g. add/removeFunction, StandaloneStatementRewriter is trying to runFuncInNewTransaction.

This is up to u. We may want to add a runFuncInNewTransaction where `requireNewTransaction` is always true.

andyfengHKU (approved):

src/binder/bind/bind_ddl.cpp
```diff
@@ -51,15 +51,32 @@ static void validateSerialNoDefault(const Expression& expr) {
...
+}
+
 std::vector<PropertyDefinition> Binder::bindPropertyDefinitions(
     const std::vector<ParsedPropertyDefinition>& parsedDefinitions, const std::string& tableName) {
     std::vector<PropertyDefinition> definitions;
     for (auto& parsedDefinition : parsedDefinitions) {
         auto type = LogicalType::convertFromString(parsedDefinition.getType(), clientContext);
+        // For default null value, we may need to resolve its data type, as its type may not be
+        // resolved during parsing and thus may be ANY.
+        tryResolvingDataTypeForDefaultNull(parsedDefinition, type);
```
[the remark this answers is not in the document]
ray6080: I actually don't know. Let me give a try, if implicitCast can handle this, it would be great.

src/main/client_context.cpp
ray6080: I don't think so. add/removeFunction are not really always running func in new transaction. They can be run in a manual transaction.

src/binder/bind/bind_ddl.cpp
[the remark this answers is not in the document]
ray6080: I cannot directly use implicitCast here because the boundExpr—after binding and implicit casting—is not actually utilized in PropertyDefinition. Instead, PropertyDefinition continues to hold the parsed expression, which is not cast.

I believe the current approach of storing ParsedExpression in PropertyDefinition is not well designed. This method essentially discards the results of binding and casting (which are only used for validation) and re-binds the default expression whenever it is needed.

PR #4675, opened 2025-01-03
Add parsing time to compilation time and remove unnecessary clone of plan

ray6080:
# Description

As stated in the title, this PR takes Parser into consideration when profiling the time spent on query compilation.

Also, removed the unnecessary clone of the best logical plan in `Planner::getBestPlan()`.

### DIFF
--- a/src/include/parser/statement.h
+++ b/src/include/parser/statement.h
@@ -9,13 +9,15 @@ namespace parser {
 class Statement {
 public:
     explicit Statement(common::StatementType statementType)
-        : statementType{statementType}, internal{false} {}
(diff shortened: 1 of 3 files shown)


src/include/parser/statement.h
```diff
@@ -40,6 +42,7 @@ class Statement {
     }
 
 private:
+    double parsingTime;
```
andyfengHKU, line 45: Up to u but I prefer to have `std::vector<std::shared_ptr<Statement>> ClientContext::parseQuery` this function return a struct including Statement & parsing time

andyfengHKU (approved):

PR #4695, opened 2025-01-10
Vector index prepare execute

ray6080:
# Description

This is on top of #4578.

Add support for passing in a parameter expression (e.g., python list) as the query vector, so the input query vector doesn't have to be a float array casted from string. Potentially this can reduce parsing overheads of large query vectors.

### DIFF
--- a/src/function/table/call/hnsw/query_hnsw_index.cpp
+++ b/src/function/table/call/hnsw/query_hnsw_index.cpp
@@ -1,5 +1,6 @@
 #include "binder/binder.h"
 #include "binder/expression/literal_expression.h"
+#include "binder/expression/parameter_expression.h"
 #include "catalog/catalog_entry/hnsw_index_catalog_entry.h"
 #include "catalog/catalog_entry/node_table_catalog_entry.h"
 #include "common/types/value/nested.h"
@@ -35,7 +36,7 @@ static std::unique_ptr<TableFuncBindData> bindFunc(main::ClientContext* context,
     const TableFuncBindInput* input) {
     const auto indexName = input->getLiteralVal<std::string>(0);
     const auto tableName = input->getLiteralVal<std::string>(1);
-    const auto& queryVal = input->params[2]->constCast<binder::LiteralExpression>().getValue();
+    auto inputQueryExpression = input->params[2];
     const auto k = input->getLiteralVal<int64_t>(3);
 
     auto nodeTableEntry = storage::IndexUtils::bindTable(*context, tableName, indexName,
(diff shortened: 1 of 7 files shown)


src/include/function/table/hnsw/hnsw_index_functions.h
```diff
@@ -52,7 +52,7 @@ struct CreateHNSWLocalState final : TableFuncLocalState {
 struct BoundQueryHNSWIndexInput {
     catalog::NodeTableCatalogEntry* nodeTableEntry;
     catalog::HNSWIndexCatalogEntry* indexEntry;
-    std::vector<float> queryVector;
+    std::shared_ptr<binder::Expression> queryExpression;
```
andyfengHKU, line 55: Better call this inputExpr?

src/function/table/call/hnsw/query_hnsw_index.cpp
```diff
@@ -35,39 +36,97 @@ static std::unique_ptr<TableFuncBindData> bindFunc(main::ClientContext* context,
...
+    auto value = common::Value::createDefaultValue(queryExpression.dataType);
+    switch (queryExpression.expressionType) {
+    case common::ExpressionType::LITERAL: {
+        value = queryExpression.constCast<binder::LiteralExpression>().getValue();
+    } break;
+    case common::ExpressionType::PARAMETER: {
+        value = queryExpression.constCast<binder::ParameterExpression>().getValue();
+    } break;
+    default: {
+        KU_UNREACHABLE;
```
andyfengHKU, line 100: What will happen if expression is not of these two types in release mode? I think it's better to throw here.

src/function/table/call/hnsw/query_hnsw_index.cpp
```diff
@@ -78,9 +137,18 @@ static common::offset_t tableFunc(const TableFuncInput& input, TableFuncOutput&
             bindData->boundInput.indexEntry->getConfig().copy());
         index->setUpperEntryPoint(bindData->boundInput.indexEntry->getUpperEntryPoint());
         index->setLowerEntryPoint(bindData->boundInput.indexEntry->getLowerEntryPoint());
+
+        auto& indexColumnType =
```
andyfengHKU, line 141: These access pattern is a bit convoluted.

An index entry should know its type. And dimension should be accessed through ArrayType::getDimention(LogicalType)

src/function/table/call/hnsw/query_hnsw_index.cpp
```diff
@@ -35,39 +36,97 @@ static std::unique_ptr<TableFuncBindData> bindFunc(main::ClientContext* context,
...
+    }
+    }
+    common::PhysicalTypeID childValType = getChildQueryValType(value);
+    std::vector<float> queryVector;
+    auto numElements = value.getChildrenSize();
+    if (numElements != dimension) {
+        throw common::RuntimeException("Query vector dimension does not match index dimension.");
+    }
+    queryVector.resize(numElements);
+    switch (childValType) {
```
andyfengHKU, line 110: I might have a template function that converts a Value to std::vector<T>

andyfengHKU (approved):

src/include/function/table/hnsw/hnsw_index_functions.h
ray6080: Renamed to `inputQueryExpression`.

src/function/table/call/hnsw/query_hnsw_index.cpp
ray6080: > An index entry should know its type.

Hmm, I'm not sure about this. The dependency between index and the table column should be enough to provide such info.

> And dimension should be accessed through ArrayType::getDimention(LogicalType)

Nice catch. Thanks!

PR #4698, opened 2025-01-12
Optimize page rank contention

andyfengHKU:
# Description

Improve page rank contention by using backward adjacency list when propagating ranks.

On the same LDBC100 person know person, performance becomes

1 thread - 5.2s
2 thread  - 2.6s
4 thread - 2.3s

The previous benchmark number was

5s, 2.7s, 4s

Fixes # (issue)

# Contributor agreement

### DIFF
--- a/src/function/gds/degrees.h
+++ b/src/function/gds/degrees.h
@@ -1,0 +1,83 @@
+#include "function/gds/gds_object_manager.h"
+#include "function/gds/gds_utils.h"
+#include "graph/graph.h"
+#include "function/gds/compute.h"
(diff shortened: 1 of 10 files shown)


andyfengHKU:
> Looks great to me! The refactoring is nice.
> 
> I have two todos in mind:
> 
> * Can you add a comment to the PageRank implementation to explain the backward traversing? I don't think that is intuitive without any comments.
> * Can you also show performance numbers before the optimization? It's clearer how much improvement this brings into.

Updated the benchmark number. I'll add comments in my next PR.

PR #4699, opened 2025-01-12
Parallel weighted shortest path

andyfengHKU:
# Description

This PR implements the initial version of bellman-ford parallel weighted shortest path. The current version only tracks total weight and dst nodes. The grammar is `[* WSHORTEST (cost)]`. We only support edge property as weight but not general expression. All numerical types except int128 & int128-backed decimal are supported.

This PR contains a refactor that removes DestinationOutput(Writer) and let SingleShortestPath & AllShortestPath creating their own ones. This creates a bit of code duplication but the previous inheritance is killing me and I don't think it's actually maintainable for the long term. I'll add more PRs to get rid of this Output & OutputWriter design.

There will be following PRs that supports
- tracking path
- all weighted shortest path.

Fixes # (issue)

# Contributor agreement


test/test_files/function/gds/weighted_shortest.test
ray6080, line 1: Do we have test cases on negative and mixed positive and negative weights?

ray6080 (approved):

test/test_files/function/gds/weighted_shortest.test
andyfengHKU: Not yet. I'll handle this in a separate PR.

PR #4701, opened 2025-01-13
Refactor on disk graph interface

andyfengHKU:
# Description

This PR contains the following refactor

1. Remove unnecessary tableID in frontier morsel. Given we first pin to a specific table, frontier morsel should contain pure offset.
2. Pass TableCatalogEntry instead of tableID in many interfaces. Practically speaking they don't make a difference, using TableCatalogEntry will let us access catalog for fewer times. And code looks cleaner.
3. Pass edge property name instead of edge property idx. This is to align with vertex scan which takes vertex property names. Also using string based name is easier to interpret for new users.
4. Rename prepareScan to prepareRelScan because we also have a prepareVertexScan.
5. Remove `std::vector<std::pair<common::table_id_t, OnDiskGraphNbrScanState>> scanStates;` because we no longer need to scan multiple rel table in one `scanFwd/scanBwd`. The infrastructure force scan always happens on exactly one rel table. The design can be further simplified but I would do it in a followed up PR.

@benjaminwinger please review for changes 3, 4 & 5

Fixes # (issue)

# Contributor agreement

### DIFF
--- a/src/include/function/gds/gds_task.h
+++ b/src/include/function/gds/gds_task.h
@@ -1,5 +1,7 @@
 #pragma once
 
+#include <utility>
+
 #include "common/enums/extend_direction.h"
 #include "common/task_system/task.h"
 #include "function/gds/gds_frontier.h"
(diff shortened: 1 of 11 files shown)


src/include/graph/graph.h
```diff
@@ -5,6 +5,7 @@
 #include <memory>
 #include <optional>
 
+#include "catalog/catalog_entry/table_catalog_entry.h"
```
ray6080, line 8: Can you double check if you can turn this include into a forward declaration?

src/include/function/gds/gds_task.h
```diff
@@ -1,5 +1,6 @@
 #pragma once
 
+#include "catalog/catalog_entry/table_catalog_entry.h"
```
ray6080, line 3: Ditto on check if this can be turned into a forward declaration.

ray6080 (commented):
I will leave this to @benjaminwinger . Have two minor comments.

src/include/graph/graph.h
andyfengHKU: Done

PR #4709, opened 2025-01-14
Distinct optimizations

benjaminwinger:
Some of the distinct aggregate queries I've tried have had very poor performance due to them processing a single tuple at a time (which the AggregateHashTable is not really designed for). These two small changes mitigate a significant part of the cost of doing so, without having any adverse effects when processing larger batches of tuples.

- Re-use the temporary ValueVectors in `BaseHashTable::computeAndCombineVecHash`
- Don't copy the entire vector when updating `tmpValueIdxes` in `AggregateHashTable::findHashSlots` (this should be universally better. Even if there is some optimization that can be done by passing a constant to memcpy, any time the loop runs multiple times we'd be better off copying less than the full vector).

### DIFF
--- a/src/include/processor/result/base_hash_table.h
+++ b/src/include/processor/result/base_hash_table.h
@@ -46,6 +46,8 @@ protected:
     std::shared_ptr<common::DataChunkState> hashState;
     std::unique_ptr<common::ValueVector> hashVector;
     common::SelectionVector hashSelVec;
+    std::unique_ptr<common::ValueVector> tmpHashResultVector;
+    std::unique_ptr<common::ValueVector> tmpHashCombineResultVector;
 };
 
 } // namespace processor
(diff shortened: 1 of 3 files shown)


ray6080 (approved):
👍 Can you also update the benchmark number here? Thanks!

benjaminwinger:
Benchmarks adapted from https://github.com/ClickHouse/ClickBench/, run on a Ryzen 5 5500
|Query | Baseline | With Changes | 
| --- | ---| --- | 
| `MATCH (h:hits) RETURN h.RegionID, COUNT(DISTINCT h.UserID) AS u ORDER BY u DESC LIMIT 10;` | 192s | 37s |
| `MATCH (h:hits) RETURN h.RegionID, SUM(h.AdvEngineID), COUNT(*) AS c, AVG(h.ResolutionWidth), COUNT(DISTINCT h.UserID) ORDER BY c DESC LIMIT 10;` | 211s | 47s |

PR #4754, opened 2025-01-21
A bit cleanup of the binder code

ray6080:
# Description

Apply const and final qualifier and clean up C-style castings in binder.

### DIFF
--- a/src/binder/bind/bind_export_database.cpp
+++ b/src/binder/bind/bind_export_database.cpp
@@ -20,26 +20,27 @@ using namespace kuzu::storage;
 namespace kuzu {
 namespace binder {
 
-static std::string getPrimaryKeyName(table_id_t tableId, const Catalog& catalog, Transaction* tx) {
-    auto tableEntry = catalog.getTableCatalogEntry(tx, tableId);
+static std::string getPrimaryKeyName(table_id_t tableId, const Catalog& catalog,
+    const Transaction* transaction) {
+    auto tableEntry = catalog.getTableCatalogEntry(transaction, tableId);
     return tableEntry->constCast<NodeTableCatalogEntry>().getPrimaryKeyName();
 }
 
-static std::vector<ExportedTableData> getExportInfo(const Catalog& catalog, Transaction* tx,
-    Binder* binder) {
+static std::vector<ExportedTableData> getExportInfo(const Catalog& catalog,
+    const Transaction* transaction, Binder* binder) {
     std::vector<ExportedTableData> exportData;
-    for (auto tableEntry : catalog.getTableEntries(tx)) {
+    for (auto tableEntry : catalog.getTableEntries(transaction)) {
         ExportedTableData tableData;
-        if (binder->bindExportTableData(tableData, *tableEntry, catalog, tx)) {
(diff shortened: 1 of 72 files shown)


src/binder/bind/bind_export_database.cpp
```diff
@@ -57,25 +59,29 @@ static void bindExportNodeTableDataQuery(const TableCatalogEntry& entry, std::st
...
 
 static void bindExportRelTableDataQuery(const TableCatalogEntry& entry, std::string& exportQuery,
-    const catalog::Catalog& catalog, transaction::Transaction* tx) {
+    const Catalog& catalog, const Transaction* transaction) {
     auto relTableEntry = entry.constPtrCast<RelTableCatalogEntry>();
-    auto srcPrimaryKeyName = getPrimaryKeyName(relTableEntry->getSrcTableID(), catalog, tx);
-    auto dstPrimaryKeyName = getPrimaryKeyName(relTableEntry->getDstTableID(), catalog, tx);
-    auto srcName = catalog.getTableCatalogEntry(tx, relTableEntry->getSrcTableID())->getName();
-    auto dstName = catalog.getTableCatalogEntry(tx, relTableEntry->getDstTableID())->getName();
+    auto srcPrimaryKeyName =
```
andyfengHKU, line 64: I think you should first get src entry here and then get primary key name and table name from the entry.

src/binder/binder.cpp
```diff
@@ -209,18 +208,21 @@ void Binder::addToScope(const std::vector<std::string>& names, const expression_
 
 void Binder::addToScope(const std::string& name, std::shared_ptr<Expression> expr) {
     // TODO(Xiyang): assert name not in scope.
+    // Note to Xiyang: I don't think the TODO still stands here. I tried adding the assertion, but
```
andyfengHKU, line 211: Alright leave this to me.

src/include/binder/query/query_graph_label_analyzer.h
```diff
@@ -11,10 +11,10 @@ class QueryGraphLabelAnalyzer {
     explicit QueryGraphLabelAnalyzer(const main::ClientContext& clientContext, bool throwOnViolate)
         : throwOnViolate{throwOnViolate}, clientContext{clientContext} {}
 
-    void pruneLabel(QueryGraph& graph);
+    void pruneLabel(const QueryGraph& graph) const;
```
andyfengHKU, line 14: Honestly this graph should not be const because are pruning labels out of it.

It works with const because we are modifying field of its field.

andyfengHKU (approved):

src/binder/bind/bind_export_database.cpp
ray6080: Make sense.

src/include/binder/query/query_graph_label_analyzer.h
ray6080: Yeah. sometimes the clang tidy can be annoying as opposed to the semantics. I reverted the change and suppress it with 
```cpp
// NOLINTNEXTLINE(readability-non-const-parameter): graph is supposed to be modified.
```

Issue #3627, opened 2024-06-11
Feature: Unify REL TABLE GROUP with REL TABLE


andyfengHKU:
### Description

Back in the day, we added `REL TABLE GROUP` because our `REL TABLE` cannot have multiple `FROM,TO` pairs. While in practice, users do want to have one (conceptual) `REL TABLE` that connects multiple `NODE TABLE`s.

The expected usage of `REL TABLE GROUP` is to be a drop-in replacement for `REL TABLE` except for `COPY` (see the note below). We have issues like #3608 and #2929 asking about different behaviours between `REL TABLE GROUP` and `REL TABLE`. We can (and should) implement them. As I plan to get started on them, I think we might just do what #2914 suggested, where we do not keep a difference between `REL TABLE GROUP` and `REL TABLE` on the user's end. Instead, we keep `REL TABLE GROUP` as an internal concept that is generated automatically we see multiple `FROM, TO` pairs when creating a `REL TABLE`.

> [!NOTE]
> The problem with `COPY` is that we need to ask user to provide table name for each row, otherwise it's impossible to resolve the table deterministically.

semihsalihoglu-uw:
As I noted before, I don't think there is a very good solution here until we have a solution to the COPY FROM and CREATE statements into rel table groups. I do see the advantage of making Rel Table Group and REL TABLE uniform. However this is not going to be easy. There are edge cases we probably don't want to think about now. For example, what happens when someone starts with a Rel Table with multiple from and to's, so they are actually constructing a rel group but then through alter schema's the number of from and to's reduces to 1. Do we automatically turn that rel group to a rel table internally? What do we show when people write `call show_tables() return *`? Do we return a mix of rel tables and rel table groups? Overall, as long as we don't have a uniform notion internally at the system, we I think we will not be able to offer a consistent experience to users. There will be edge cases and inconsistencies that confuse people. So I think it is better to use some usability but be very clear about two notions: a Rel Table and a Rel Table Group, which is a syntactic sugar around multiple rel tables.

That said, I think we should support a few things to improve the experience:
1. ALTER REL TABLE GROUP: which should allow people to add and remove FROM and TO pairs and add and drop columns. The syntax for adding and dropping columns should mimic:  ALTER TABLE X ADD/DROP C. So this would be ALTER REL TABLE GROUP RelGroup_Foo Add/Drop C. We can also support ALTER TABLE RelGroup_C ADD/Drop C as a syntactic sugar and give a warning. 

[...]

PR #4755, opened 2025-01-21
Deprecate rel group

andyfengHKU:
# Description

Address issue #3627.

We now support the syntax CREATE REL TABLE (FROM nodeA TO nodeB, FROM nodeA TO nodeC, ...) and should update the documentation to deprecate the rel group syntax, although we will continue to support it for compatibility reasons.

Internally, we maintain the distinction between RelTable and RelTableGroup. A relationship table defined with a single pair of FROM and TO node labels is categorized as a RelTable, while one defined with multiple pairs is categorized as a RelTableGroup.


src/binder/bind/bind_ddl.cpp
```diff
@@ -195,11 +205,6 @@ BoundCreateTableInfo Binder::bindCreateRelTableInfo(const CreateTableInfo* info)
         clientContext->shouldUseInternalCatalogEntry());
 }
 
-static std::string getRelGroupTableName(const std::string& relGroupName,
-    const std::string& srcTableName, const std::string& dstTableName) {
-    return relGroupName + "_" + srcTableName + "_" + dstTableName;
-}
-
 BoundCreateTableInfo Binder::bindCreateRelTableGroupInfo(const CreateTableInfo* info) {
```
ray6080, line 208: We should throw exception when there is storage direction option specified. We are not handling it right now.

src/parser/transform/transform_ddl.cpp
```diff
@@ -86,42 +86,31 @@ std::unique_ptr<Statement> Transformer::transformCreateRelTable(
...
-        relMultiplicity = transformSymbolicName(*ctx.oC_SymbolicName());
-    }
-    std::vector<std::pair<std::string, std::string>> srcDstTablePairs;
-    for (auto& connection : ctx.kU_RelTableConnection()) {
-        auto srcTableName = transformSchemaName(*connection->oC_SchemaName(0));
-        auto dstTableName = transformSchemaName(*connection->oC_SchemaName(1));
-        srcDstTablePairs.emplace_back(srcTableName, dstTableName);
+    std::unique_ptr<ExtraCreateTableInfo> extraInfo;
+    auto entryType = CatalogEntryType::DUMMY_ENTRY;
+    if (fromToPairs.size() == 1) {
```
ray6080, line 98: Abstract this into a function.

src/binder/bind/bind_file_scan.cpp
```diff
@@ -96,17 +96,26 @@ std::unique_ptr<BoundBaseScanSource> Binder::bindScanSource(BaseScanSource* sour
     }
 }
 
+static void removeNonScanOptions(case_insensitive_map_t<Value>& options) {
```
ray6080, line 99: Should we silently remove invalid options or should we throw exceptions?

src/catalog/catalog_entry/table_catalog_entry.cpp
```diff
@@ -91,6 +93,20 @@ void TableCatalogEntry::renameProperty(const std::string& propertyName,
     propertyCollection.rename(propertyName, newName);
 }
 
+std::string TableCatalogEntry::getDisplayLabelName(Catalog* catalog,
```
ray6080, line 96: I wonder if we should keep a `label` field in `TableCatalogEntry`, which is different from `name`.

`label` is defined by users, while `name` is used internally by us. Or maybe we even don't need `name` at all?

src/include/catalog/catalog_entry/table_catalog_entry.h
```diff
@@ -63,6 +64,8 @@ class KUZU_API TableCatalogEntry : public CatalogEntry {
     void dropProperty(const std::string& propertyName);
     void renameProperty(const std::string& propertyName, const std::string& newName);
 
+    std::string getDisplayLabelName(Catalog* catalog, const transaction::Transaction* transaction);
```
ray6080, line 67:
```suggestion
    std::string getLabel(Catalog* catalog, const transaction::Transaction* transaction);
```

ray6080 (commented):
Looks great. I will take over the PR to merge it soon.

Some comments:
1. We should rework LDBC benchmark queries to adapt to new syntax later.
2. I need to double check our export behaviour now as we should maintain the compatibility of import/export from an older version if the database contains rel group.
3. Are we handling single-directed rel storage with multi FROM/TO pairs now? are we testing such cases?
4. I will test to make sure the label pruning still works well with new changes.

ray6080 (approved):
I have double checked that import/export, basic label pruning and single-directed rel storage options work fine.

PR #4764, opened 2025-01-21
Support alter for rel groups

### LINKED ISSUE #4762: Improve error message when trying to drop a property on a table associated with an index

royi-luo:
# Description

- Support `ALTER` on rel groups, propagating the changes to the child tables
- Move most validation functions for alter from binder to alter operator

TODOs:
- Add more tests
  - [ ] WAL replaying alter (there's a single test in `rollback.test`)
  - [ ] Alter conflict handling on rel groups

Fixed #4762 as well.

# Contributor agreement

### DIFF
--- a/src/binder/binder.cpp
+++ b/src/binder/binder.cpp
@@ -151,13 +151,6 @@ void Binder::validateOrderByFollowedBySkipOrLimitInWithClause(
     }
 }
 
-void Binder::validateTableExist(const std::string& tableName) const {
-    if (!clientContext->getCatalog()->containsTable(clientContext->getTransaction(), tableName,
-            clientContext->shouldUseInternalCatalogEntry())) {
-        throw BinderException("Table " + tableName + " does not exist.");
-    }
-}
-
 std::string Binder::getUniqueExpressionName(const std::string& name) {
     return "_" + std::to_string(lastExpressionId++) + "_" + name;
 }
--- a/src/processor/operator/ddl/alter.cpp
+++ b/src/processor/operator/ddl/alter.cpp
@@ -1,6 +1,8 @@
 #include "processor/operator/ddl/alter.h"
 
 #include "catalog/catalog.h"
+#include "catalog/catalog_entry/node_table_catalog_entry.h"
+#include "catalog/catalog_entry/rel_group_catalog_entry.h"
 #include "common/enums/alter_type.h"
 #include "storage/storage_manager.h"
 #include "storage/store/table.h"
@@ -12,7 +14,8 @@ namespace processor {
 
 using skip_alter_on_conflict = std::function<bool()>;
 
-bool skipAlter(common::ConflictAction action, const skip_alter_on_conflict& skipAlterOnConflict) {
+static bool skipAlter(common::ConflictAction action,
+    const skip_alter_on_conflict& skipAlterOnConflict) {
     switch (action) {
     case common::ConflictAction::ON_CONFLICT_THROW:
         return false;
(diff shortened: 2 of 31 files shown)


src/processor/operator/ddl/alter.cpp
```diff
@@ -23,46 +25,155 @@ bool skipAlter(common::ConflictAction action, skip_alter_on_conflict skipAlterOn
...
+        const auto* tableEntry = catalog->getTableCatalogEntry(transaction, tableID);
+        BoundAlterInfo tableAlterInfo = info.copy();
+        tableAlterInfo.tableName = tableEntry->getName();
+        if (tableAlterInfo.alterType == common::AlterType::RENAME_TABLE) {
+            KU_ASSERT(tableAlterInfo.tableName.starts_with(relGroupEntry->getName()));
+            // table name is in format {rel_group_name}{suffix}
+            // rename to {new_rel_group_name}{suffix}
+            auto& renameInfo = tableAlterInfo.extraInfo->cast<BoundExtraRenameTableInfo>();
+            auto tableNameSuffix = tableEntry->getName().substr(relGroupEntry->getName().size());
+            renameInfo.newName.append(tableNameSuffix);
```
royi-luo, line 42: do we want renaming rel group to also rename the child tables?

src/binder/binder.cpp
```diff
@@ -152,9 +152,23 @@ void Binder::validateOrderByFollowedBySkipOrLimitInWithClause(
...
+static bool checkTableExists(main::ClientContext* clientContext, const std::string& tableName) {
+    return clientContext->getCatalog()->containsTable(clientContext->getTransaction(), tableName,
+        clientContext->shouldUseInternalCatalogEntry());
+}
+
+static bool checkRelGroupExists(main::ClientContext* clientContext, const std::string& name) {
+    return clientContext->getCatalog()->containsRelGroup(clientContext->getTransaction(), name);
+}
+
+void Binder::validateTableOrRelGroupExist(const std::string& name) {
```
andyfengHKU, line 164: move this to bind_ddl.cpp and make it static

src/processor/operator/ddl/alter.cpp
andyfengHKU: I'm not sure. I guess it's better to also rename children tables so that the naming convention if preserved after ALTER

src/processor/operator/ddl/alter.cpp
```diff
@@ -23,46 +25,155 @@ bool skipAlter(common::ConflictAction action, skip_alter_on_conflict skipAlterOn
...
-                    ->containsProperty(info.extraInfo->constCast<BoundExtraAddPropertyInfo>()
-                                           .propertyDefinition.getName());
-            })) {
-            return;
+    const auto* relGroupEntry = catalog->getRelGroupEntry(transaction, info.tableName);
+    for (auto tableID : relGroupEntry->getRelTableIDs()) {
+        const auto* tableEntry = catalog->getTableCatalogEntry(transaction, tableID);
+        BoundAlterInfo tableAlterInfo = info.copy();
+        tableAlterInfo.tableName = tableEntry->getName();
+        if (tableAlterInfo.alterType == common::AlterType::RENAME_TABLE) {
```
andyfengHKU, line 36: I guess it's better to do this in bind_ddl.cpp?

andyfengHKU (commented):
Overall LGTM.

@ray6080 should check if the new record can be omitted and if we need to store alterInfo in TableCatalogEntry.

andyfengHKU (approved):

PR #4785, opened 2025-01-24
Statically linking runtime for Windows compilation

mewim:


mewim:
Close in favour of #4796

Issue #4783, opened 2025-01-23
Bug: segfault when creating rel


royi-luo:
### Kùzu version

master

### What operating system are you using?

ubuntu 22

### What happened?

To reproduce you can either run the attached queries or the python script.

[python_script.txt](https://github.com/user-attachments/files/18527760/python_script.txt)
[queries.txt](https://github.com/user-attachments/files/18527761/queries.txt)

An assertion failure is being triggered:
```
Error: Assertion failed in file "/home/r34luo/kuzu/kuzu/src/storage/store/column_chunk_data.cpp" on line 249: vector->dataType.getPhysicalType() == dataType.getPhysicalType()
```

royi-luo:
I've reduced the queries to this to reproduce:
```
CREATE NODE TABLE Person(id INT64, PRIMARY KEY (id));
CREATE NODE TABLE IF NOT EXISTS Movie(id INT64, name STRING, PRIMARY KEY (id));
create rel table knows(from person to person, from person to movie, age int64, comment string);
create (:person {id: 12});
create (:movie {id: 12, name: 'm'});
create (:person {id: 25});
create (:movie {id: 25, name: 'z'});
alter table knows drop age;
match (a:person), (b:person) where a.id = 25 and b.id = 12 create (a)-[:knows {comment: -8696457900372165364}]->(b);
```

The assertion failure also hits if the last query is
` match (a:person), (b:person) where a.id = 25 and b.id = 12 create (a)-[:knows {comment: '-8696457900372165364'}]->(b);`

If the rel table schema is 
`create rel table knows(from person to person, from person to movie, age int64, comment int64)`
then we don't run into the failure which suggests there is some mismatch between user-provided type and expected type that we aren't guarding against but I'm not sure why the assertion failure also triggers when the comment is a string.

PR #4786, opened 2025-01-24
Fix mismatched column IDs during insert and copy after drop and add columns

### LINKED ISSUE #4777: Bug: Failure during insert/copy after drop and add column in a table
### Kùzu version

master

MacOS

### What happened?

```cypher
CREATE NODE TABLE person(id INT64 PRIMARY KEY, name STRING);
CREATE (:person {id: 10, name: 'A'});
ALTER TABLE person DROP name;
ALTER TABLE person ADD age INT64;
CREATE (:person {id: 20, age: 20});
```
The last create statement fails after a DROP column and a ADD column statement.
Same when the last statement is COPY FROM.

The failure is due to that we didn't really remove a dropped column from storage before CHECKPOINT (this is due to that we need to maintain various versions of the table). Thus the column to be inserted / copied into are mismatched internally.

_No response_

ray6080:
# Description

Drop column can cause mismatched column IDs when inserting or copying into tables. The main reason is due to that we don't remove the dropped column immediately until checkpoint. The fix introduces column ids when appending into the table.

Fix #4777 and #4783

### DIFF
--- a/src/storage/store/rel_table_data.cpp
+++ b/src/storage/store/rel_table_data.cpp
@@ -20,20 +20,18 @@ PersistentVersionRecordHandler::PersistentVersionRecordHandler(RelTableData* rel
     : relTableData(relTableData) {}
 
 void PersistentVersionRecordHandler::applyFuncToChunkedGroups(version_record_handler_op_t func,
-    common::node_group_idx_t nodeGroupIdx, common::row_idx_t startRow, common::row_idx_t numRows,
-    common::transaction_t commitTS) const {
+    node_group_idx_t nodeGroupIdx, row_idx_t startRow, row_idx_t numRows,
+    transaction_t commitTS) const {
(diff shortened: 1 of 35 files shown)


src/storage/store/rel_table_data.cpp
```diff
@@ -283,7 +281,7 @@ const VersionRecordHandler* RelTableData::getVersionRecordHandler(CSRNodeGroupSc
     }
 }
 
-void RelTableData::rollbackGroupCollectionInsert(common::row_idx_t numRows_, bool isPersistent) {
+void RelTableData::rollbackGroupCollectionInsert(row_idx_t numRows_, bool isPersistent) {
```
andyfengHKU, line 284: Double check coverage here and l274

andyfengHKU (approved):

Issue #4431, opened 2024-10-28
Bug: API Interrupt test fails on CI


ray6080:
### Kùzu version

master

### What happened?

`ApiTest.Interrupt` fails on the asan test on CI.

PR #4789, opened 2025-01-24
Fix ApiTest.Interrupt

royi-luo:
# Description

It's possible for the connection thread to not start until after `interrupt()` is called, which clears the interrupt flag. Add a new thread that repeatedly calls `interrupt()` until the query terminates.

Fixes https://github.com/kuzudb/kuzu/issues/4431

# Contributor agreement

### DIFF
--- a/test/main/connection_test.cpp
+++ b/test/main/connection_test.cpp
@@ -56,9 +56,6 @@ TEST_F(ApiTest, ParallelConnect) {
 
 // TODO: FIX-ME. See Issue #4431.
 TEST_F(ApiTest, Interrupt) {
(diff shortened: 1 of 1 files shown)


mewim (approved):

test/main/connection_test.cpp
```diff
@@ -55,16 +55,23 @@ TEST_F(ApiTest, ParallelConnect) {
 }
 
 // TODO: FIX-ME. See Issue #4431.
```
ray6080, line 57: @royi-luo can you  also remove the TODO here?

royi-luo: done

Issue #4782, opened 2025-01-23
Bug: Default values when adding new columns doesn't work if casting of a literal is required


royi-luo:
### Kùzu version

master

### What operating system are you using?

Ubuntu 22

### What happened?

Idea is similar to https://github.com/kuzudb/kuzu/issues/4389. Consider the queries:
```
create node table Person(id INT64, PRIMARY KEY (id))
create rel table likes (from person to person);
alter table likes add years int32 default 0
```

The query `alter table likes add years int32 default 0` will fail with the message 
```
Runtime exception: Cannot set a non-constant default value when adding columns on REL tables.
```
This is because the literal `0` needs to be casted from int64 to int32, and we don't consider the casted value a constant.

Similarly, the statement `alter table likes add years float default 5.4` will also fail.

PR #4791, opened 2025-01-24
Support casted literals for new column default value

royi-luo:
# Description

- Avoid throwing unsupported error if default value is a casted literal
- Update expression evaluator so that for default evaluators the correct output sel size is returned

Fixes https://github.com/kuzudb/kuzu/issues/4782

# Contributor agreement

### DIFF
--- a/src/binder/expression/expression_util.cpp
+++ b/src/binder/expression/expression_util.cpp
@@ -5,6 +5,7 @@
 #include "binder/expression/literal_expression.h"
 #include "binder/expression/node_rel_expression.h"
 #include "binder/expression/parameter_expression.h"
+#include "binder/expression/scalar_function_expression.h"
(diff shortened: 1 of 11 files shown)


andyfengHKU:
See #4793 for a more generic fix. Sorry I took it over as the fix is relatively simple so I just wrote it up.

Btw

> Update expression evaluator so that for default evaluators the correct output sel size is returned

was this a bug or something related to your fix?

royi-luo:
> See #4793 for a more generic fix. Sorry I took it over as the fix is relatively simple so I just wrote it up.
> 
> Btw
> 
> > Update expression evaluator so that for default evaluators the correct output sel size is returned
> 
> was this a bug or something related to your fix?

It was related to my changes so no need to worry about it

PR #4792, opened 2025-01-25
Add comment to rel group

ray6080:
# Description

### DIFF
--- a/src/catalog/catalog_set.cpp
+++ b/src/catalog/catalog_set.cpp
@@ -63,15 +63,7 @@ CatalogEntry* CatalogSet::getEntryNoLock(const Transaction* transaction,
     return entry;
 }
 
-static void logCreateDropEntryForTrx(Transaction* transaction, CatalogSet& set, CatalogEntry& entry,
-    bool isInternal, bool skipLoggingToWAL = false) {
-    KU_ASSERT(transaction);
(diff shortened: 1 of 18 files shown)


src/catalog/catalog_set.cpp
```diff
@@ -175,64 +171,68 @@ CatalogEntry* CatalogSet::dropEntryNoLock(const Transaction* transaction, const
...
+        // We treat rename table as drop and create.
+        dropEntryNoLock(transaction, alterInfo.tableName, entry->getOID());
+        auto createdEntry = createEntryNoLock(transaction, std::move(newEntry));
+        if (transaction->shouldAppendToUndoBuffer()) {
+            transaction->pushAlterCatalogEntry(*this, *entry, alterInfo);
+            transaction->pushCreateDropCatalogEntry(*this, *createdEntry, isInternal(),
+                true /* skipLoggingToWAL */);
+        }
+    } break;
+    default: {
```
andyfengHKU, line 194: I would explicitly enumerate type here and make default KU_UNREACHABLE

andyfengHKU (approved):

src/catalog/catalog_set.cpp
ray6080: Sounds good. Let me refactor this in my next PR, which I'm doing refactoring already.

Issue #4814, opened 2025-01-30
Bug: [JAVA-API] Duplicated records being returned


arthur-bigeard-gdotv:
### Kùzu version

v0.7.2-dev-37

### What operating system are you using?

MacOS & Windows

### What happened?

The Java API appears to be returning duplicated records for every query being ran.
For instance, `MATCH (s:Comment) RETURN s LIMIT 50` will return the same comment node 50 times. Each time a new query is run, that behaviour persists but with a different node.
The `FlatTuples` extracted from the `QueryResult` all point to the exact same node when using the `hasNext()`/`getNext()` pattern, so it looks like either `getNext()` is not moving the cursor on the query result or the query result itself contains duplicated information.
Variations of queries with `DISTINCT` lead to the same result, but `COUNT` queries accurately return node counts.

### Are there known steps to reproduce?

This can be reproduced on the Java API using the LDBC test sample from Kuzu Explorer. To reproduce this, you need to create a Database instance in the Java API pointing to a dataset, then run one of the queries suggested above. This appears to happen consistently for all datasets.

royi-luo:
Each call to `getNext()` reuses the same underlying `FlatTuple` object so a flat tuple should either be processed or copied before the next call to `getNext()` is made. Thanks @arthur-bigeard-gdotv for catching this!

PR #4820, opened 2025-01-31
Update API docs: QueryResult getNext() reuses the same underlying flat tuple

royi-luo:
# Description

In the C, C++, and Java APIs, `QueryResult::getNext()` returns references to the output flat tuple. Additionally, the same flat tuple object is reused for each call to `getNext()` meaning that each call to `getNext()` will overwrite the return result of the previous call to `getNext()`. This PR updates the API docs to document this behaviour.

Fixes https://github.com/kuzudb/kuzu/issues/4814

# Contributor agreement


mewim (approved):
LGTM

PR #4825, opened 2025-02-01
Use MemoryManager to allocate undo buffer

ray6080:
# Description

Make use of `MemoryManager` for allocating undo buffers.

Additionally, apply `const` and `static` qualifiers where appropriate.


andyfengHKU (approved):
I assume the uncovered lines are false positive

PR #4838, opened 2025-02-03
[nix] feat: init support for nix shell

sdht0:
# Description

Introduce support for `nix-shell ./tools/nix` or `nix develop ./tools/nix`.

Also supports building Kuzu within nixpkgs using `nix build ./tools/nix`. The plan is to open an [upstream](https://github.com/NixOS/nixpkgs) PR such that users will be able to invoke the kuzu cli using `nix-shell -p kuzu`. See the equivalent [duckdb package](https://github.com/NixOS/nixpkgs/blob/45d92089a65e2e6a6de85503c1dcd015b80e823c/pkgs/development/libraries/duckdb/default.nix).

# Contributor agreement


mewim:
Hmm.. I do not think we should maintain a configuration to a specific Linux distribution in our main repository.

sdht0:
It's fine not to commit this.

`nix` can be used on any Linux distro and even macOS btw and allows easily sharing a consistent developer environment. NixOS is a full distro built on top of nix and nixpkgs but that is optional.

The main reason I wanted to share this is to clarify if the release CLI uses any specific cmake flags. From the Makefile, `make release` doesn't seem to be setting any flags.

sdht0:
Regardless of whether we commit this, I'll open the upstream PR after 0.8.0 is released.

mewim:
> It's fine not to commit this.
> 
> `nix` can be used on any Linux distro and even macOS btw and allows easily sharing a consistent developer environment. NixOS is a full distro built on top of nix and nixpkgs but that is optional.
> 
> The main reason I wanted to share this is to clarify if the release CLI uses any specific cmake flags. From the Makefile, `make release` doesn't seem to be setting any flags.

I don’t think this should go into the main repository. For configuration in nix environment, you can create a separate repository and update the developer’s guide here: https://github.com/kuzudb/kuzu-docs/blob/main/src/content/docs/developer-guide/index.md

PR #4840, opened 2025-02-04
Remove SimpleTableFunction and make TableFunction non-extendable

ray6080:
# Description

1. `SimpleTableFunction` is confusing as I don't think the term is clearly defined, neither there are other kinds of table functions other than "Simple". It should just be `TableFunction`. This PR moved static functions in `SimpleTableFunction` into `TableFunction`.
2. `TableFunction` should not be extended. Table function definition structs should purely serve as utility classes to return a defined function set.


src/include/function/table/bind_data.h
```diff
@@ -17,18 +17,24 @@ struct KUZU_API TableFuncBindData {
     // the last {numWarningDataColumns} columns are for temporary internal use
     common::column_id_t numWarningDataColumns;
     common::cardinality_t cardinality;
+    common::offset_t maxOffset;
```
andyfengHKU, line 20: I would make the default to be INVALID_OFFSET

andyfengHKU (approved):

src/include/function/table/bind_data.h
ray6080: hmm I think 0 is more safe here, as `TableFuncSharedState` would use this to dispatch morsels.

PR #4843, opened 2025-02-04
Fix extend cardinality

royi-luo:
# Description

Remove the code that updates extend cardinality in `appendExtend` as extend doesn't affect factorized cardinality (otherwise we double-count with the factorized group multiplier). This was causing the lsqb benchmark to fail.

# Contributor agreement

### DIFF
--- a/src/planner/plan/append_extend.cpp
+++ b/src/planner/plan/append_extend.cpp
@@ -216,7 +216,8 @@ void Planner::appendRecursiveExtendAsGDS(const std::shared_ptr<NodeExpression>& 
     pathPropertyProbe->computeFactorizedSchema();
     auto extensionRate =
         cardinalityEstimator.getExtensionRate(*rel, *boundNode, clientContext->getTransaction());
-    auto resultCard = cardinalityEstimator.multiply(extensionRate, plan.getLastOperator()->getCardinality());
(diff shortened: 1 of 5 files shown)


src/planner/plan/append_extend.cpp
```diff
@@ -223,9 +221,6 @@ void Planner::appendRecursiveExtendAsGDS(const std::shared_ptr<NodeExpression>&
     // Join with input node
     auto joinConditions = expression_vector{boundNode->getInternalID()};
     appendHashJoin(joinConditions, JoinType::INNER, probePlan, plan, plan);
-    // Hash join above should not change the cardinality of probe plan.
-    // plan.setCardinality(probePlan.getCardinality());
-    // KU_ASSERT(plan.getCardinality() == plan.getLastOperator()->getCardinality());
```
royi-luo, line 228: @andyfengHKU can you double check if this is good to remove or if we should be resetting the cardinality here (this is unrelated to the other change)

royi-luo: Also should the line above also be updated?
```
pathPropertyProbe->setCardinality(
    cardinalityEstimator.estimateExtend(extensionRate, plan.getLastOperatorRef()));
```

PR #4850, opened 2025-02-05
Speed up recompilation when changing compile-time config

royi-luo:
# Description
Added a template file `cmake/templates/system_config.h.in` that uses CMake's `configure_file` that is used to generate the header `${CMAKE_BINARY_DIR}/src/include/common/system_config.h` which contains all of our compile-time configurations. Also added `${CMAKE_BINARY_DIR}/src/include` as an include directory so that we can include the generated header as `common/system_config.h`

This means that when we recompile with a different configuration, only files that include the generated `system_config.h` will be recompiled (with the existing method of using compile definitions the whole project will be recompiled each time a configuration is changed).

### Master
image
### New
image
With the current changes I got the number of recompile steps reduced to 569 from 777

I've gotten the speedup to improve to about 220 steps by making changes to the code (e.g. reorganizing includes, moving stuff from header to cpp file) but I'll put that in [a separate PR](https://github.com/kuzudb/kuzu/pull/4863) since it's a large change.

# Contributor agreement


.github/workflows/ci-workflow.yml
```diff
@@ -283,10 +283,6 @@ jobs:
           IN_MEM_MODE: true
         run: make lcov NUM_THREADS=$(nproc)
 
-      - name: Test with coverage vector capacity of 2
-        run: |
-          make lcov VECTOR_CAPACITY_LOG2=1
-
```
royi-luo, line 289: I realised that this doesn't work since recompiling overwrites the coverage data from previous runs. If we want to include it we'll need to find a way to merge the coverage data from multiple runs from different builds.

tools/rust_api/build.rs
```diff
@@ -131,6 +131,7 @@ fn build_bundled_cmake() -> Result<Vec<PathBuf>, Box<dyn std::error::Error>> {
     Ok(vec![
         kuzu_root.join("src/include"),
         build_dir.join("build/src"),
+        build_dir.join("build/src/include"),
```
royi-luo, line 134: @benjaminwinger can you take a look at the rust build changes here to see if they make sense?

ray6080:
@benjaminwinger can I delegate the review of this PR to you?

tools/rust_api/build.rs
benjaminwinger: Looks good to me.

benjaminwinger (approved):

PR #4863, opened 2025-02-06
Optimize recompile times

royi-luo:
# Description

Follow up to https://github.com/kuzudb/kuzu/pull/4850

Reorganize includes, move stuff from header to cpp files, add forward declares to optimize recompile when system config changes. 

The main changes are:
- Remove the need to include `system_config.h` from headers, especially ones that are included everywhere such as `value_vector.h`, `sel_vector.h`, and `data_chunk_state.h`
- If the need for the constants in the headers are less simple to remove (e.g. needing a system config constant to define an member array size), try to forward-declare those classes so that the headers don't need to be included

This reduces the number of recompile steps to about 220 from 570:
image

# Contributor agreement

### DIFF
--- a/src/include/common/data_chunk/sel_vector.h
+++ b/src/include/common/data_chunk/sel_vector.h
@@ -2,7 +2,6 @@
 
 #include <string.h>
 
-#include <array>
 #include <memory>
 
 #include "common/types/types.h"
--- a/src/include/common/data_chunk/sel_vector.h
+++ b/src/include/common/data_chunk/sel_vector.h
@@ -5,7 +5,6 @@
 #include <array>
 #include <memory>
 
-#include "common/system_config.h"
 #include "common/types/types.h"
 #include <span>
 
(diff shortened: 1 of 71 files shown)


src/include/common/data_chunk/sel_vector.h
```diff
@@ -2,18 +2,15 @@
...
 
-#include "common/system_config.h"
 #include "common/types/types.h"
 #include <span>
 
 namespace kuzu {
 namespace common {
 
 class SelectionVector {
-    KUZU_API static const std::array<sel_t, DEFAULT_VECTOR_CAPACITY> INCREMENTAL_SELECTED_POS;
```
royi-luo, line 16: Did we put this array in the header for performance purposes (maybe so that accesses to it could be inlined)? If so, I can revert this change and try to find a different way to avoid including `system_config.h` everywhere that uses `SelectionVector`

benjaminwinger: It looks like it's been that way since @ray6080 added the class in #1095, so I suspect not.
It still could have a performance impact, but I suspect it will be minimal.

src/include/storage/store/struct_column.h
```diff
@@ -5,6 +5,7 @@
 
 namespace kuzu {
 namespace storage {
+class MemoryManager;
```
benjaminwinger, line 8: Was the intention to also remove the `#include` in this file?

benjaminwinger (approved):

src/include/storage/store/struct_column.h
royi-luo: updated, thanks for catching that

PR #4865, opened 2025-02-06
Add Wasm to main page

prrao87:
# Description

Ardan pointed out that we should mention we support wasm on our GitHub page so we can point users to it.

# Contributor agreement


acquamarin (approved):

prrao87:
Closing in favour of #4866

PR #4868, opened 2025-02-07
Support gds var length label pruning

ted-wq-x:
# Description

Fixes [#4396 ](https://github.com/kuzudb/kuzu/pull/4396)

# Contributor agreement

### DIFF
--- a/src/binder/query/query_graph_label_analyzer.cpp
+++ b/src/binder/query/query_graph_label_analyzer.cpp
@@ -1,6 +1,7 @@
 #include "binder/query/query_graph_label_analyzer.h"
 
 #include "catalog/catalog.h"
+#include "catalog/catalog_entry/node_table_catalog_entry.h"
 #include "catalog/catalog_entry/rel_table_catalog_entry.h"
 #include "common/exception/binder.h"
 #include "common/string_format.h"
@@ -29,7 +30,7 @@ void QueryGraphLabelAnalyzer::pruneNode(const QueryGraph& graph, NodeExpression&
         if (queryRel->isRecursive()) {
             continue;
         }
-        common::table_id_set_t candidates;
+        table_id_set_t candidates;
         std::unordered_set<std::string> candidateNamesSet;
         auto isSrcConnect = *queryRel->getSrcNode() == node;
         auto isDstConnect = *queryRel->getDstNode() == node;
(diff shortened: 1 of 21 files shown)


src/binder/query/query_graph_label_analyzer.cpp
```diff
@@ -94,49 +95,231 @@ void QueryGraphLabelAnalyzer::pruneNode(const QueryGraph& graph, NodeExpression&
...
+        for (auto path : paths[i]) {
+            for (auto j = 0u; j < i; ++j) {
+                auto rels = path[j];
+                stepActiveTableIDs[j].insert(rels.begin(), rels.end());
             }
-            prunedEntries.push_back(entry);
         }
     }
-    rel.setEntries(prunedEntries);
+    return stepActiveTableIDs;
```
andyfengHKU, line 208: Let me try to summarize the logic here and pls correct me if I'm wrong.

Consider topology `Person - knows -> Person` and `Person - livesIn -> City` and query 
```
MATCH (a:Person) - [*2..2] -> (b:City)
``` 
WLOG, we assume going from left to right.

The `dfs` part will first generate all size 2 path
```
<<knows>, <knows>>
<<knows>, <livesIn>>
```
and because `<<knows>, <knows>>` dst is not `b:City` it will be pruned.

So eventually we only compute `<<knows>, <livesIn>>`.

If my understanding is correct, can u also give a benchmark number in the PR description to make sure we actually get performance benefit.

src/binder/query/query_graph_label_analyzer.cpp
```diff
@@ -94,49 +95,231 @@ void QueryGraphLabelAnalyzer::pruneNode(const QueryGraph& graph, NodeExpression&
...
+}
+
+std::vector<table_id_set_t> QueryGraphLabelAnalyzer::pruneRecursiveRel(
+    const std::unordered_map<table_id_t, std::unordered_map<table_id_t, table_id_vector_t>>& graph,
+    const table_id_set_t& startTableIDSet, const table_id_set_t& endTableIDSet, size_t lowerBound,
+    size_t upperBound) const {
+
+    // There may be multiple rel type between src and dst
+    using PATH = std::vector<table_id_vector_t>;
+    std::vector<std::vector<PATH>> paths(upperBound + 1);
```
andyfengHKU, line 175: It would be good to comment 

`std::vector<PATH>` are paths of the same length.

And the outer-most std::vector are paths of different length.

In fact, it might be easier to read if just use `std::vector<PATH>` here.

src/binder/query/query_graph_label_analyzer.cpp
```diff
@@ -94,49 +95,231 @@ void QueryGraphLabelAnalyzer::pruneNode(const QueryGraph& graph, NodeExpression&
...
+    }
+    return nodeIDs;
+}
+
+std::pair<std::vector<table_id_set_t>, std::vector<table_id_set_t>>
+QueryGraphLabelAnalyzer::pruneRecursiveRel(const std::vector<TableCatalogEntry*>& relEntries,
+    const table_id_set_t srcTableIDSet, const table_id_set_t dstTableIDSet, size_t lowerBound,
+    size_t upperBound, RelDirectionType relDirectionType) const {
+    // src-->[dst,[rels]]
+    std::unordered_map<table_id_t, std::unordered_map<table_id_t, table_id_vector_t>>
```
andyfengHKU, line 146: Up to use but I would use `std::unrdered_map<table_id_t, STRUCT>`

and STRUCT contains
```
table_id_t dstTableID
std::vector<tabel_id_t> relTableIDs
```

because we always access by srcTableID instead of `src&dstTableID`

src/binder/query/query_graph_label_analyzer.cpp
```diff
@@ -94,49 +95,231 @@ void QueryGraphLabelAnalyzer::pruneNode(const QueryGraph& graph, NodeExpression&
...
+        auto [stepFromLeftTableIDs, stepFromRightTableIDs] =
+            pruneRecursiveRel(rel.getEntries(), srcTableIDSet, dstTableIDSet, rel.getLowerBound(),
+                rel.getUpperBound(), rel.getDirectionType());
+        auto recursiveInfo = rel.getRecursiveInfo();
+        recursiveInfo->stepFromLeftActivationRelInfos = stepFromLeftTableIDs;
+        recursiveInfo->stepFromRightActivationRelInfos = stepFromRightTableIDs;
+        // todo we need reset rel entries?
+        auto temp = mergeTableIDs(stepFromLeftTableIDs, stepFromRightTableIDs);
+        std::vector<TableCatalogEntry*> newRelEntries{temp.begin(), temp.end()};
+        if (!isSameTableCatalogEntryVector(newRelEntries, rel.getEntries())) {
```
andyfengHKU, line 284: Shall we do l284-l312 as a separate PR because

- it can be viewed as another optimization after pruning recursive rel labels; and
- we will also need to get a benchmark number to make sure there are performance improvements for node pruning.

andyfengHKU (approved):

ted-wq-x:
I found gds join performance worse than before, there are so many memory minor fault. This opt does not provide any performance improvement in current code.

Under recursive join ,this opt can get 900ms -->300ms in sf100,"match (a:person {id:24189255912498})-[f*2]-(b) return count(1)" (Previous test results);

I will provide performance benchmark, when i resolve the current performance problem.

PR #4869, opened 2025-02-07
Add Catalog version

ray6080:
# Description

Introduce a basic version info to our `Catalog`, which can be used later by auto-completion (needs to check catalog if there are changes or not to cache table names, etc. frequently) and `execute` to check if a re-bind is necessary (right now we always re-bind).

The version is only incremented after a transaction that updates Catalog commits.


andyfengHKU (approved):
LGTM.

I have a minor question. If we are able to detect catalog change, does this mean we can do prepare statement knowing catalog has changed?

ray6080:
> LGTM.
> 
> I have a minor question. If we are able to detect catalog change, does this mean we can do prepare statement knowing catalog has changed?

I think so, though we need to double check if the catalog has changed within the same transaction or not, which is also doable now.

PR #4872, opened 2025-02-09
Allow COPY FROM in manual transactions

ray6080:
# Description

We had a strong constraint that a `COPY FROM` statement cannot run in manual transactions. However after #4420, we can allow this. This PR removes this restriction now.

This paves the way to make create vector/fts index running in a single transaction.

### DIFF
--- a/src/processor/operator/partitioner.cpp
+++ b/src/processor/operator/partitioner.cpp
@@ -3,7 +3,6 @@
 #include "binder/expression/expression_util.h"
 #include "processor/execution_context.h"
 #include "processor/operator/persistent/rel_batch_insert.h"
-#include "storage/buffer_manager/memory_manager.h"
 #include "storage/store/node_table.h"
 #include "storage/store/rel_table.h"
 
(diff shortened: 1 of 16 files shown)


src/processor/operator/partitioner.cpp
```diff
@@ -55,7 +54,7 @@ partition_idx_t PartitionerSharedState::getNextPartition(idx_t partitioningIdx,
     if (nextPartitionIdxToReturn >= numPartitions[partitioningIdx]) {
         return INVALID_PARTITION_IDX;
     }
-    progressSharedState.partitionsDone++;
+    ++progressSharedState.partitionsDone;
```
andyfengHKU, line 57: Does this change actually matter?

andyfengHKU (commented):
Overall LGTM though I didn't know why we cannot support COPY FROM in manual transactions.

src/processor/operator/partitioner.cpp
ray6080: Not really. It's more of a code style change (pre-increment vs. post-increment), which I forgot to revert. Will clean it up.

andyfengHKU (approved):

ray6080:
> Overall LGTM though I didn't know why we cannot support COPY FROM in manual transactions.

We didn't support COPY FROM in manual transactions mainly due to that the uncommitted node offsets were unstable before #4420. Since we only support single writer at a time, we can maintain stable node offsets now.

PR #4873, opened 2025-02-09
Rework rewriteFunc of table functions

ray6080:
# Description

Table functions that create/drop FTS and HNSW indexes make use of `rewriteFunc` to prepare tables and macros, etc.
Current implementation of `rewriteFunc` is convoluted that it pre-appends rewritten query to be executed first before the table function itself (e..g, CREATE_FTS_INDEX), instead of generating a rewritten query and directs the execution to the rewritten one.

This PR reworks Create/Drop FTS/HNSW Index functions in that:
1. `CreateFTSIndex`, `DropFTSIndex`, `CreateHNSWIndex`, `DropHNSWIndex` will only be responsible for generating rewritten query.
2. Internal functions `_CreateFTSIndex`, `_DropFTSIndex`, `_CreateHNSWIndex`, `_DropHNSWIndex` are responsible for the actual create/drop logic.


src/binder/bind/bind_standalone_call_function.cpp
```diff
@@ -16,10 +16,11 @@ std::unique_ptr<BoundStatement> Binder::bindStandaloneCallFunction(
     auto& funcExpr =
         callStatement.getFunctionExpression()->constCast<parser::ParsedFunctionExpression>();
     auto funcName = funcExpr.getFunctionName();
-    auto entry =
-        clientContext->getCatalog()->getFunctionEntry(clientContext->getTransaction(), funcName);
+    auto entry = clientContext->getCatalog()->getFunctionEntry(clientContext->getTransaction(),
```
andyfengHKU, line 19: Can u rename this to `useInternalCatalog` or `useInternalCatalogEntry`

src/catalog/catalog.cpp
```diff
@@ -373,16 +374,22 @@ void Catalog::dropIndex(Transaction* transaction, table_id_t tableID,
     indexes->dropEntry(transaction, uniqueName, entry->getOID());
 }
 
-bool Catalog::containsFunction(const Transaction* transaction, const std::string& name) const {
-    return functions->containsEntry(transaction, name);
+bool Catalog::containsFunction(const Transaction* transaction, const std::string& name,
+    bool useInternal) const {
```
andyfengHKU, line 378: I had this question about useInternal for a while.

If useInternal is true, should we still check external entries? The current semantic is to always check external entries.

andyfengHKU (approved):

src/catalog/catalog.cpp
ray6080: Yeah we should still check external. `useInternal` is not restricting the visibility to internal only.

PR #4881, opened 2025-02-10
Parallel distinct hash aggregate

benjaminwinger:
Extends the partitioning done in the hash aggregate operator to apply to distinct hash tables as well as the main hash table. The aggregation is then disabled in the thread-local hash tables for distinct keys, and computed from scratch when combining the data into the global tables.

This still doesn't parallelize the simple distinct aggregate, which I'll do in a later PR.

This also fixes support for nested types in the hash aggregate generally (I realised that I had missed implementing the row data versions of the comparison functions for structs and lists).


src/processor/operator/aggregate/aggregate_hash_table.cpp
```diff
@@ -557,16 +629,24 @@ void AggregateHashTable::updateAggState(const std::vector<ValueVector*>& flatKey
...
+            for (auto& dataChunk : aggregateInputs[i].multiplicityChunks) {
+                multiplicity *= dataChunk->state->getSelVector().getSelSize();
+            }
+            updateAggFuncs[i](this, flatKeyVectors, unFlatKeyVectors, aggregateFunctions[i],
+                aggregateInputs[i].aggregateVector, multiplicity, i, aggregateStateOffset);
+            aggregateStateOffset += aggregateFunctions[i].getAggregateStateSize();
+        } else {
+            // If a function is distinct we still need to insert the value into the distinct hash
+            // table
+            distinctHashTables[i]->isAggregateValueDistinctForGroupByKeys(flatKeyVectors,
```
ray6080, line 647: I don't follow here, don't we need to check the return value here and update the hash table if the key is distinct? Is that done already in `isAggregateValueDistinctForGroupByKeys`? If so, the function name is not accurate.

ray6080 (approved):
Looks great to me! Thanks!
Have few minor comments, you can take a look.

src/processor/operator/aggregate/aggregate_hash_table.cpp
benjaminwinger: Yes, the `isAggregateValueDistinctForGroupByKeys` function also inserts. I can update the function name to make it more clear.

src/include/processor/operator/aggregate/aggregate_hash_table.h
```diff
@@ -166,7 +173,8 @@ class AggregateHashTable : public BaseHashTable {
 
     void updateAggStates(const std::vector<common::ValueVector*>& flatKeyVectors,
         const std::vector<common::ValueVector*>& unFlatKeyVectors,
-        const std::vector<AggregateInput>& aggregateInputs, uint64_t resultSetMultiplicity);
+        const std::vector<AggregateInput>& aggregateInputs, uint64_t resultSetMultiplicity,
+        bool updateDistinct = true);
```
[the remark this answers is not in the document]
benjaminwinger: I suppose so

PR #4882, opened 2025-02-10
Add empty columns to chunked node group if needed during COPY

royi-luo:
# Description

Follow-up to https://github.com/kuzudb/kuzu/pull/4786

After dropping a column (and before checkpointing), if we are directly appending newly-flushed chunked groups to a node group (during copy) the newly inserted node group may not have as many columns as the parent node group, which would cause the next checkpoint to fail. 

This PR adds empty columns to the newly inserted node group so that its number of columns is as expected

Also fixes another bug where warning data was being flushed to disk for rel batch insert.

# Contributor agreement

### DIFF
--- a/src/storage/store/chunked_node_group.cpp
+++ b/src/storage/store/chunked_node_group.cpp
@@ -76,7 +76,7 @@ ChunkedNodeGroup::ChunkedNodeGroup(MemoryManager& mm, ChunkedNodeGroup& base,
 
     for (column_id_t i = 0; i < columnTypes.size(); ++i) {
         if (chunks[i] == nullptr) {
-            chunks[i] = std::make_unique<ColumnChunk>(mm, columnTypes[i].copy(), capacity,
+            chunks[i] = std::make_unique<ColumnChunk>(mm, columnTypes[i].copy(), 0,
                 enableCompression, ResidencyState::IN_MEMORY);
         }
     }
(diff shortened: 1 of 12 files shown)


src/storage/store/chunked_node_group.cpp
```diff
@@ -53,6 +53,35 @@ ChunkedNodeGroup::ChunkedNodeGroup(MemoryManager& mm, const std::vector<LogicalT
...
+        KU_ASSERT(baseColumnID < chunks.size());
+        chunks[baseColumnID] = base.moveColumnChunk(i);
+        enableCompression = chunks[baseColumnID]->isCompressionEnabled();
+        KU_ASSERT(chunks[baseColumnID]->getDataType().getPhysicalType() ==
+                  columnTypes[baseColumnID].getPhysicalType());
+    }
+
+    for (column_id_t i = 0; i < columnTypes.size(); ++i) {
+        if (chunks[i] == nullptr) {
+            chunks[i] = std::make_unique<ColumnChunk>(mm, columnTypes[i].copy(), capacity,
```
ray6080, line 79: Can we not set capacity to 0 here for padding chunks?

ray6080 (approved):
Looks good! Thanks!

src/storage/store/chunked_node_group.cpp
royi-luo: updated

PR #4892, opened 2025-02-12
Add GCS support

royi-luo:
# Description

GCS's interoperability mode supports [using S3-style requests](https://cloud.google.com/storage/docs/aws-simple-migration), so we reuse the S3 filesystem while making a few changes to support scans (COPY/LOAD FROM, attach read-only DB) with GCS. In this case, we create a new S3 filesystem instance for communicating with GCS, with some different init-time configurations to signal that it's communicating with GCS. 

Any URLs with prefixes `gs://` or `gcs://` will have their requests directed to GCS. The path formats will be the same as S3: `${PREFIX}/${BUCKET_NAME}/${PATH_TO_FILE_IN_BUCKET}`.

Additionally, I added separate settings for configuring authentication parameters for GCS (they refer to the same parameters as S3):

```
call gcs_access_key_id=...
call gcs_secret_access_key=...
```

# Contributor agreement


extension/delta/src/connector/delta_connector.cpp
```diff
@@ -15,7 +16,9 @@ void DeltaConnector::connect(const std::string& /*dbPath*/, const std::string& /
     executeQuery("load delta;");
     executeQuery("install httpfs;");
     executeQuery("load httpfs;");
-    executeQuery(duckdb_extension::DuckDBSecretManager::getS3Secret(context));
+    for (auto& fsConfig : httpfs::S3FileSystemConfig::getAvailableConfigs()) {
```
acquamarin, line 19: I think we can refactor this for loop as a separate function in duckdb extension:
E.g. setRemoteFSConfig, which blindly sets hte config for s3/gcs

extension/delta/src/main/delta_extension.cpp
```diff
@@ -3,16 +3,18 @@
...
 
 namespace kuzu {
 namespace delta_extension {
 
 void DeltaExtension::load(main::ClientContext* context) {
     auto& db = *context->getDatabase();
     extension::ExtensionUtils::addTableFunc<DeltaScanFunction>(db);
-    httpfs::S3DownloadOptions::registerExtensionOptions(&db);
-    httpfs::S3DownloadOptions::setEnvValue(context);
+    for (auto& fsConfig : httpfs::S3FileSystemConfig::getAvailableConfigs()) {
```
acquamarin, line 14: Ditto, better to refactor as a function

extension/iceberg/src/iceberg_extension.cpp
```diff
@@ -15,8 +15,10 @@ void IcebergExtension::load(main::ClientContext* context) {
     extension::ExtensionUtils::addTableFunc<IcebergScanFunction>(db);
     extension::ExtensionUtils::addTableFunc<IcebergMetadataFunction>(db);
     extension::ExtensionUtils::addTableFunc<IcebergSnapshotsFunction>(db);
-    httpfs::S3DownloadOptions::registerExtensionOptions(&db);
-    httpfs::S3DownloadOptions::setEnvValue(context);
+    for (auto& fsConfig : httpfs::S3FileSystemConfig::getAvailableConfigs()) {
```
acquamarin, line 18: You can refactor as a common function under duckdb extension

extension/httpfs/src/include/s3fs.h
```diff
@@ -92,7 +85,7 @@ struct ParsedS3URL {
     std::string getHTTPURL(std::string httpQueryString = "") const;
 };
 
-class S3FileSystem final : public HTTPFileSystem {
+class S3FileSystem : public HTTPFileSystem {
```
acquamarin, line 88: Why do we make it not final?

acquamarin (commented):
LGTM, Can you add some tests? I want to make sure everything is covered.

extension/httpfs/src/include/s3fs.h
royi-luo: I originally had something else inherit from this, let me change this back

extension/httpfs/src/include/s3fs_config.h
```diff
@@ -0,0 +1,44 @@
...
+
+namespace kuzu {
+namespace main {
+class Database;
+class ClientContext;
+} // namespace main
+
+namespace httpfs {
+
+static constexpr std::array AUTH_OPTIONS = {"ACCESS_KEY_ID", "SECRET_ACCESS_KEY", "ENDPOINT",
```
acquamarin, line 18: Can you double check whether GCS support endpoint, url_style and region parameters?

extension/httpfs/src/httpfs_extension.cpp
```diff
@@ -24,14 +26,18 @@ static void registerExtensionOptions(main::Database* db) {
 
 static void registerFileSystem(main::Database* db) {
     db->registerFileSystem(std::make_unique<HTTPFileSystem>());
-    db->registerFileSystem(std::make_unique<S3FileSystem>());
+    for (auto& fsConfig : httpfs::S3FileSystemConfig::getAvailableConfigs()) {
+        db->registerFileSystem(std::make_unique<S3FileSystem>(fsConfig));
```
acquamarin, line 30: Can we only use one fileSystem instead of two?
When we open a file, we generate a S3AuthParams class based on the URL.
The S3AuthParams is saved in the fileinfo class.
So we don't need to register two fileSystems.

extension/delta/src/connector/delta_connector.cpp
royi-luo: updated

extension/delta/src/main/delta_extension.cpp
royi-luo: updated

extension/iceberg/src/iceberg_extension.cpp
royi-luo: updated

extension/httpfs/src/include/s3fs.h
royi-luo: updated

extension/httpfs/src/include/s3fs_config.h
royi-luo: removed those options for GCS (although they still exist internally)

extension/httpfs/src/httpfs_extension.cpp
royi-luo: I'd prefer to keep two separate filesystems if possible since I feel like it makes this more understandable + flexible unless we're concerned about performance/memory footprint here

extension/httpfs/test/test_files/gcs_upload.test
```diff
@@ -0,0 +1,38 @@
+-DATASET CSV ldbc-sf01
+
+--
+
+-CASE UploadToGCS
+-SKIP
```
royi-luo, line 6: What was the reason for the `UploadToS3` test being disabled, was it because it required writing to a new bucket for each CI run? I've added the test here with `-SKIP` so that we can run it locally (it works locally for me)

acquamarin (approved):

extension/httpfs/test/test_files/gcs_upload.test
acquamarin: Because it is expensive, we want to reduce the cost

PR #4893, opened 2025-02-12
gds: init scc computation using kosaraju's algorithm

sdht0:
# Description

Single-threaded implementation of SCC using [Kosaraju's_algorithm](https://en.wikipedia.org/wiki/Kosaraju's_algorithm). 

* Does not use `EdgeCompute`, but instead is directly called from `exec()`.
* Added couple of tests. Are the internals IDs assigned in a deterministic manner?

# Contributor agreement

### DIFF
--- a/src/function/gds/strongly_connected_components.cpp
+++ b/src/function/gds/strongly_connected_components.cpp
@@ -1,0 +1,250 @@
+#include <iostream>
+#include <thread>
+#include <tuple>
+#include <vector>
+#include <format>
+
+#include "binder/binder.h"
+#include "binder/expression/expression_util.h"
+#include "common/types/types.h"
+#include "function/gds/auxiliary_state/gds_auxilary_state.h"
+#include "function/gds/gds_frontier.h"
+#include "function/gds/gds_function_collection.h"
+#include "function/gds/gds_object_manager.h"
+#include "function/gds/gds_utils.h"
+#include "function/gds/output_writer.h"
+#include "function/gds_function.h"
+#include "graph/graph.h"
+#include "processor/execution_context.h"
+#include "processor/result/factorized_table.h"
+
+using namespace std;
+using namespace kuzu::binder;
+using namespace kuzu::common;
+using namespace kuzu::processor;
+using namespace kuzu::storage;
+using namespace kuzu::graph;
+
+namespace kuzu {
+namespace function {
+
+offset_t NOT_ASSIGNED = std::numeric_limits<offset_t>::max();
+
+/// State for the SCC algorithm
+class SCCState {
+public:
+    SCCState(const table_id_map_t<offset_t>& numNodesMap, MemoryManager* mm) {
+        for (const auto& [tableID, numNodes] : numNodesMap) {
+            componentIDsMap.allocate(tableID, numNodes, mm);
+            visitsMap.allocate(tableID, numNodes, mm);
+            pinTable(tableID);
+            for (auto i = 0u; i < numNodes; ++i) {
+                visits[i].store(false, std::memory_order_relaxed);
+                componentIDs[i].store(NOT_ASSIGNED, std::memory_order_relaxed);
+            }
+        }
+    }
+
+    void pinTable(table_id_t tableID) { componentIDs = componentIDsMap.getData(tableID); 
+        visits = visitsMap.getData(tableID);}
+
+    bool updateVisit(common::offset_t offset, bool value) {
+        visits[offset].store(value, std::memory_order_relaxed);
+        return true;
+    }
+
+    bool getVisit(offset_t offset) {
+        return visits[offset].load(std::memory_order_relaxed);
+    }
+
+    bool updateComponentID(common::offset_t offset, offset_t value) {
+        componentIDs[offset].store(value, std::memory_order_relaxed);
+        return true;
+    }
+
+    offset_t getComponentID(offset_t offset) {
+        return componentIDs[offset].load(std::memory_order_relaxed);
+    }
+
+private:
+    // Pointers to current tableID's visits and componentIDs
+    std::atomic<bool>* visits = nullptr;
+    std::atomic<offset_t>* componentIDs = nullptr;
+    // State for visits and componentIDs
+    ObjectArraysMap<std::atomic<bool>> visitsMap;
+    ObjectArraysMap<std::atomic<offset_t>> componentIDsMap;
+};
+
+class SCCCompute {
+public:
+    SCCCompute(graph::Graph* graph, const table_id_map_t<offset_t>& numNodesMap,
+        SCCState& sccState)
+        : graph{graph}, sccState{sccState}, numNodesMap{numNodesMap} {};
+    ~SCCCompute() = default;
+
+    // Forward DFS
+    void visit_vertex(nodeID_t node, catalog::TableCatalogEntry* relEntry) {
+        if (sccState.getVisit(node.offset)) {
+            // Already visited
+            return;
+        };
+        sccState.updateVisit(node.offset, true);
+        auto scanState = graph->prepareRelScan(relEntry, "");
+        for (auto chunk : graph->scanFwd(node, *scanState)) {
+            chunk.forEach([&](auto nbrNodeID, auto) {
+                visit_vertex(nbrNodeID, relEntry);
+            });
+        }
+        queue.push_back(node);
+    }
+
+    // Backwards DFS
+    void assign_id(nodeID_t node, offset_t root, catalog::TableCatalogEntry* relEntry) {
+        if (sccState.getComponentID(node.offset) != NOT_ASSIGNED) {
+            // Already assigned
+            return;
+        };
+        sccState.updateComponentID(node.offset, root);
+        auto scanState = graph->prepareRelScan(relEntry, "");
+        for (auto chunk : graph->scanBwd(node, *scanState)) {
+            chunk.forEach([&](auto nbrNodeID, auto) {
+                assign_id(nbrNodeID, root, relEntry);
+            });
+        }
+    }
+
+    void compute() {
+        for (const auto& [tableID, numNodes] : numNodesMap) {
+            sccState.pinTable(tableID);
+            for (auto& [fromEntry, toEntry, relEntry] : graph->getRelFromToEntryInfos()) {
+                for (auto i = 0u; i < numNodes; ++i) {
+                    auto nodeID = nodeID_t{i, tableID};
+                    visit_vertex(nodeID, relEntry);
+                }
+                for (auto it = queue.rbegin(); it != queue.rend(); ++it) {
+                    auto node = *it;
+                    assign_id(node, node.offset, relEntry);
+                }
+            }
+        }
+    }
+
+private:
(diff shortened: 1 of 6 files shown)


src/function/gds/strongly_connected_components.cpp
```diff
@@ -0,0 +1,250 @@
...
+        return componentIDs[offset].load(std::memory_order_relaxed);
+    }
+
+private:
+    // Pointers to current tableID's visits and componentIDs
+    std::atomic<bool>* visits = nullptr;
+    std::atomic<offset_t>* componentIDs = nullptr;
+    // State for visits and componentIDs
+    ObjectArraysMap<std::atomic<bool>> visitsMap;
+    ObjectArraysMap<std::atomic<offset_t>> componentIDsMap;
```
sdht0, line 75: Is this the right way to maintain state for multiple variables?

src/function/gds/strongly_connected_components.cpp
```diff
@@ -0,0 +1,250 @@
...
+            chunk.forEach([&](auto nbrNodeID, auto) {
+                assign_id(nbrNodeID, root, relEntry);
+            });
+        }
+    }
+
+    void compute() {
+        for (const auto& [tableID, numNodes] : numNodesMap) {
+            sccState.pinTable(tableID);
+            for (auto& [fromEntry, toEntry, relEntry] : graph->getRelFromToEntryInfos()) {
```
sdht0, line 114: Is there a better way to loop through the vertices and their neighbors?

src/function/gds/strongly_connected_components.cpp
```diff
@@ -0,0 +1,250 @@
...
+    ~SCCCompute() = default;
+
+    // Forward DFS
+    void visit_vertex(nodeID_t node, catalog::TableCatalogEntry* relEntry) {
+        if (sccState.getVisit(node.offset)) {
+            // Already visited
+            return;
+        };
+        sccState.updateVisit(node.offset, true);
+        auto scanState = graph->prepareRelScan(relEntry, "");
```
sdht0, line 91: This state should be recreated in each recursion, right?

src/function/gds/strongly_connected_components.cpp
```diff
@@ -0,0 +1,245 @@
...
+    }
+
+    offset_t getComponentID(offset_t offset) {
+        return componentIDs[offset].load(std::memory_order_relaxed);
+    }
+
+private:
+    // Pointers to current tableID's visits and componentIDs
+    std::atomic<bool>* visits = nullptr;
+    std::atomic<offset_t>* componentIDs = nullptr;
```
sdht0, line 72: If I remove the `atomic`s, I get a CTE: `error: member reference base type 'bool' is not a structure or union`. How do I use non-atomic vars here?

src/function/gds/strongly_connected_components.cpp
```diff
@@ -0,0 +1,245 @@
...
+        sccState.updateVisit(node.offset, true);
+        auto scanState = graph->prepareRelScan(relEntry, "");
+        for (auto chunk : graph->scanFwd(node, *scanState)) {
+            chunk.forEach([&](auto nbrNodeID, auto) { visit_vertex(nbrNodeID, relEntry); });
+        }
+        queue.push_back(node);
+    }
+
+    // Backwards DFS
+    void assign_id(nodeID_t node, offset_t root, catalog::TableCatalogEntry* relEntry) {
```
andyfengHKU, line 99: ditto I would just call this backwardDFS

src/function/gds/strongly_connected_components.cpp
```diff
@@ -0,0 +1,245 @@
...
+                for (auto it = queue.rbegin(); it != queue.rend(); ++it) {
+                    auto node = *it;
+                    assign_id(node, node.offset, relEntry);
+                }
+            }
+        }
+    }
+
+private:
+    std::vector<nodeID_t> queue;
```
andyfengHKU, line 128: I wonder if this will perform if the size of nodes is 1B. Let's see it's performance first.

andyfengHKU (changes requested):
Overall LGTM. Let's discuss a bit more with @semihsalihoglu-uw

src/function/gds/strongly_connected_components.cpp
sdht0: So I implemented this if I understood you correctly. Is it the case that maintaining a second copy of the graph is faster than scanning the main graph?

src/function/gds/strongly_connected_components.cpp
sdht0: Thanks, yeah I might have missed one of them.

src/function/gds/strongly_connected_components.cpp
```diff
@@ -0,0 +1,245 @@
...
+    // Forward DFS
+    void visit_vertex(nodeID_t node, catalog::TableCatalogEntry* relEntry) {
+        if (sccState.getVisit(node.offset)) {
+            // Already visited
+            return;
+        };
+        sccState.updateVisit(node.offset, true);
+        auto scanState = graph->prepareRelScan(relEntry, "");
+        for (auto chunk : graph->scanFwd(node, *scanState)) {
+            chunk.forEach([&](auto nbrNodeID, auto) { visit_vertex(nbrNodeID, relEntry); });
```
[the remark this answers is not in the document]
sdht0: Yes, this was just to make it work. Let's discuss the proper API I should use here.

src/function/gds/strongly_connected_components.cpp
```diff
@@ -0,0 +1,260 @@
...
+            dfsStack.pop();
+            if (!sccState.componentIDSet(node)) {
+                backwardsDFS(node, node, *scanState);
+            }
+        }
+    }
+
+    void forwardDFS(offset_t node, NbrScanState& scanState) {
+        toProcess.push(node);
+
```
semihsalihoglu-uw, line 100: I'd remove this blank line.

src/function/gds/strongly_connected_components.cpp
```diff
@@ -0,0 +1,260 @@
...
+        while (!toProcess.empty()) {
+            auto nextNode = toProcess.top();
+
```
semihsalihoglu-uw, line 103: Ditto.

src/function/gds/strongly_connected_components.cpp
```diff
@@ -0,0 +1,260 @@
...
+            if (sccState.visited(nextNode)) {
+                toProcess.pop();
+                dfsStack.push(nextNode);
+                continue;
+            }
+
```
semihsalihoglu-uw, line 109: Ditto.

src/function/gds/strongly_connected_components.cpp
```diff
@@ -0,0 +1,260 @@
...
+            sccState.setVisited(nextNode);
+
```
semihsalihoglu-uw, line 111: Ditto.

test/test_files/function/gds/scc.test
```diff
@@ -0,0 +1,132 @@
+-DATASET CSV empty
+
+--
+
+-CASE Scc1
+-STATEMENT CREATE NODE TABLE User(id INT64, name STRING PRIMARY KEY);
+---- ok
+-STATEMENT CREATE REL TABLE FRIEND(FROM User to User);
+---- ok
+-STATEMENT CREATE (alice:User {id: 0, name: 'Alice'}),
```
semihsalihoglu-uw, line 10: Your call but as a general principle, you could probably simplify these test data by giving them just generic and indexed names like U1 U2 etc. Probably would simplify writing down the graph for others who might read the code s you do below.

test/test_files/function/gds/scc.test
```diff
@@ -0,0 +1,132 @@
+-DATASET CSV empty
+
+--
+
+-CASE Scc1
```
semihsalihoglu-uw, line 5: What's Scc1 and Scc2 testing exactly?

Can you make sure your test graph covers the following SCCs: 

- completely isolated node n
- node that only has outgoing edges
- node that has only incoming edges
- a small completely isolated scc
- a small scc that only has outgoing edges
- a small scc that only has incoming edges

semihsalihoglu-uw (approved):
I don't need to see this again but I still have a few small-level comments. @andyfengHKU should look at the binder exception, which does not seem to be the right type of exception or the right place to throw the exception.

src/function/gds/strongly_connected_components.cpp
```diff
@@ -0,0 +1,260 @@
...
+// Use the two largest offset_t values as special boolean values instead of
+// using a separate second array.
+static constexpr offset_t NOT_VISITED = numeric_limits<offset_t>::max();
+static constexpr offset_t VISITED = numeric_limits<offset_t>::max() - 1u;
+
+class SCCState {
+public:
+    SCCState(const table_id_map_t<offset_t>& numNodesMap, MemoryManager* mm) {
+        if (numNodesMap.size() != 1) {
+            throw BinderException("SCC only supports operations on one node table.");
```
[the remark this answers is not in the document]
sdht0: Changed to `RuntimeException`.

Will rename later.

src/function/gds/strongly_connected_components.cpp
```diff
@@ -0,0 +1,260 @@
...
+};
+
+class SCCCompute {
+public:
+    SCCCompute(Graph* graph, SCCState& sccState) : graph{graph}, sccState{sccState} {}
+
+    void compute() {
+        auto nbrTables = graph->getForwardNbrTableInfos(sccState.getTableID());
+        if (nbrTables.size() != 1) {
+            throw BinderException("SCC only supports operations on one edge table.");
```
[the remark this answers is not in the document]
sdht0: Twice for node and edge tables.

test/test_files/function/gds/scc.test
sdht0: Refactored the tests to add all these cases.

src/function/gds/strongly_connected_components_kosaraju.cpp
```diff
@@ -0,0 +1,259 @@
...
+                    sccState.setProcessed(nextNode);
+                }
+                continue;
+            }
+            sccState.setVisited(nextNode);
+            auto nextNodeID = nodeID_t{nextNode, sccState.getTableID()};
+            for (auto chunk : graph->scanFwd(nextNodeID, scanState)) {
+                chunk.forEach([&](auto nbrNodeID, auto) {
+                    if (!sccState.visited(nbrNodeID.offset)) {
+                        toProcess.push_back(nbrNodeID.offset);
```
andyfengHKU, line 121: Well here we are still materializing every nodeID from value vector. I thought the design was to use a `ScanState` pool and not materializing them. Did this change?

sdht0: Copying out the neighbors seem to be the straightforward way.
Deferring this optimization until later when it becomes a bottleneck.

semihsalihoglu-uw: I don't know if we want to actually do this at all. It's not really an optimization. This is the standard way to implement iterative DFS-based algorithms.

PR #4909, opened 2025-02-14
Fix comparison of list null entries in the AggregateHashTable

benjaminwinger:
There was also a bug where a few bytes at the end of the list (just for factorized table to factorized table comparisons) wouldn't be checked since we were omitting to take into account the null mask at the beginning.

E.g. https://github.com/kuzudb/kuzu/actions/runs/13324561162/job/37215093604?pr=4908. Lists containing null entries sometimes compare as non-equal when they should be equal, presumably since we don't guarantee that the data part of the entry is a consistent value if the entry is null.


andyfengHKU (commented):
I would add a test case to cover this.

andyfengHKU (approved):

PR #4913, opened 2025-02-14
Separate hash aggregate finalization into its own operator

### LINKED ISSUE #4903: Bug: Undirected query with DISTINCT fails unpredictably on Windows
### Kùzu version

v0.8.0

Windows

### What happened?

Arthur from G.V() ran the following undirected query with a `DISTINCT` keyword on the LDBC1 and LDBC10 datasets using the Java API. On MacOS, it runs pretty flawlessly, but on Windows, it randomly succeeds (many times it crashes Kùzu)

```cypher
 MATCH ()-[r]-() RETURN DISTINCT r LIMIT 1000
```

The task would be to first reproduce the crash on Windows, and see how the behaviour differs from what's happening on macOS.

Run the above query on a small LDBC1 or LDBC10 dataset on Windows, preferably through the Java API (though the issue is likely at the C++ layer).

benjaminwinger:
This fixes the data race I mentioned in #4906. It's a little difficult to rigorously test since I was having a hard time reproducing it, but moving the finalization into a separate operator should guarantee that all threads running HashAggregate operator tasks finish before finalization starts.

I think this may fix the issue in #4903, but I'm not positive that it's the same issue.

### DIFF
--- a/src/include/processor/operator/aggregate/hash_aggregate.h
+++ b/src/include/processor/operator/aggregate/hash_aggregate.h
@@ -4,6 +4,7 @@
 
 #include <cstdint>
 #include <memory>
+#include <vector>
 
 #include "aggregate_hash_table.h"
 #include "common/copy_constructors.h"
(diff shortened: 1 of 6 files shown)


ray6080 (approved):
Looks good! Thanks!

PR #4914, opened 2025-02-15
Break when error in the middle of multi statements

ray6080:
# Description

This fixes the behaviour for execution of multiple statements in a single query. When there is an error with a statement in the middle of the query, we should error and break the execution.

Also changed how shell collects query results. Previously shell manually splits a query by ";" into multiple statements, however, I don't think this is safe in the first place, it also is not consistent with the behaviour of executing a single query with multiple statements through API.

### DIFF
--- a/extension/fts/test/test_files/stopwords.test
+++ b/extension/fts/test/test_files/stopwords.test
@@ -49,7 +49,11 @@ alice|0.457137
 -STATEMENT EXPORT DATABASE '${KUZU_EXPORT_DB_DIRECTORY}_fts/stopwords_export_test'
 ---- ok
 -RELOADDB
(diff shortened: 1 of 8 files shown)


andyfengHKU (approved):

extension/fts/test/test_files/stopwords.test
```diff
@@ -49,8 +49,12 @@ alice|0.457137
 -STATEMENT EXPORT DATABASE '${KUZU_EXPORT_DB_DIRECTORY}_fts/stopwords_export_test'
 ---- ok
 -RELOADDB
+# TODO(Ziyi): Fix import database here. RELOADDB is not the correct way to test export/import. The query result is incorrect after import.
```
ray6080, line 52: @acquamarin I think there is a bug with import/export fts. Can you double check?

PR #4922, opened 2025-02-17
Add all weighted shortest path

andyfengHKU:
# Description

Add all weighted shortest path.

Split weighted shortest path implementation to different files (pure refactor).


src/function/gds/awsp_paths.cpp
```diff
@@ -0,0 +1,179 @@
...
+class AWSPPathsOutputWriter : public PathsOutputWriter {
+public:
+    AWSPPathsOutputWriter(main::ClientContext* context,
+        processor::NodeOffsetMaskMap* outputNodeMask, common::nodeID_t sourceNodeID,
+        PathsOutputWriterInfo info, BFSGraph& bfsGraph)
+        : PathsOutputWriter{context, outputNodeMask, sourceNodeID, info, bfsGraph} {
+        costVector = createVector(LogicalType::DOUBLE(), context->getMemoryManager());
+    }
+
+    void write(processor::FactorizedTable& fTable, common::nodeID_t dstNodeID,
```
ray6080, line 56: The control logic here looks exactly the same as `PathsOutputWriter::dfsFast` except a few control conditions. Can you refactor them together? Also, I'm curious if the fast and slow path applies to awsp too.

src/function/gds/wsp_destinations.cpp
```diff
@@ -0,0 +1,186 @@
...
+
+    void pinTable(table_id_t tableID) { costs = costsMap.getData(tableID); }
+
+    double getCost(offset_t offset) { return costs[offset].load(std::memory_order_relaxed); }
+
+    void setCost(offset_t offset, double cost) {
+        costs[offset].store(cost, std::memory_order_relaxed);
+    }
+
+    bool update(offset_t boundOffset, offset_t nbrOffset, double val) {
```
ray6080, line 26: Can you add a comment here to clarify that this function updates the cost of `nbrOffset` to the one of `boundOffset` plus val, if the latter one is smaller. The function name is too brief to convey the actual computation happened.

ray6080 (approved):
Looks good to me. Have two minor comments for reference.

src/function/gds/wsp_destinations.cpp
andyfengHKU: Done

src/function/gds/awsp_paths.cpp
andyfengHKU: As we discussed we will solve this gradually.

PR #4928, opened 2025-02-18
Add tests with different node group sizes + fix bugs

royi-luo:
#  #Description

# Contributor agreement


test/test_files/agg/simple.test
```diff
@@ -3,7 +3,8 @@
 --
 
 -CASE AggSimple
-
+# If node group size is too small the order of the outputted list is non-deterministic
+-SKIP_NODE_GROUP_SIZE_TESTS
```
ray6080, line 7: I don't follow here. which list do you mean? How come simple aggregation have nondeterministic output?

test/test_files/dml_node/create/create_empty.test
```diff
@@ -113,6 +113,8 @@ fe38f9dc-f761-4184-8b8b
 foobar-ewe-j323-8nd*-ewew
 
 -CASE CreateInt8MaxWidth
+# rows span multiple node groups if node group size is 2
```
ray6080, line 116: I think you can move storage_info related statements into another test?

test/test_files/function/uuid.test
```diff
@@ -3,7 +3,8 @@
 --
 
 -CASE UUIDfunctions
-
+# messing with node group size can change the output order of collect()
```
ray6080, line 6: Can we change the test to sort over collected list?

test/test_files/match/multi_label_large.test
```diff
@@ -0,0 +1,9 @@
+-DATASET CSV ldbc-sf01
+# This test is intended for testing with smaller node group size
+# Which can cause the BM to run out of memory
```
ray6080, line 3: If so, should we test the bm error?

ray6080 (approved):
Looks great! Thanks!

test/test_files/match/multi_label_large.test
royi-luo: Don't think so since running out of memory is expected for in mem mode (we are storing a lot more metadata since the node group is smaller so the storage footprint is much larger) and the main point of the test is to catch a different bug

test/test_files/agg/simple.test
royi-luo: I was referring to the output of `collect` here, since we're not automating tests with node group size 2 there's no need to skip although if we were we can also sort the output of `collect`.

royi-luo: removed the skip

test/test_files/dml_node/create/create_empty.test
royi-luo: moved to `create_int_max_width.test`

test/test_files/function/uuid.test
royi-luo: updated

test/test_files/transaction/create_node/create_empty_checkpoint.test
```diff
@@ -147,116 +147,6 @@ foobar|
 fe38f9dc-f761-4184-8b8b
 foobar-ewe-j323-8nd*-ewew
 
--CASE CreateInt8MaxWidth
```
royi-luo, line 150: these tests are redundant now that we have a separate test run with `checkpoint_threshold=0`

PR #4932, opened 2025-02-19
Add rel filter to projected graph

andyfengHKU:
# Description

This PR adds relationship table filter to projected graph. An example can be found in the following

```
CALL create_projected_graph('PK2', ['person'], {'knows': {'filter': 'r.date > date("1999-01-01")'}})
```

It's a bit weird that we need to write `r.date > ...` instead of `date > ...` but it's the simplest one. Also Neo4j has a very similar grammar.

This PR also renames project_graph to projected_graph


test/test_files/function/gds/special_projected_graph.test
```diff
@@ -0,0 +1,57 @@
...
+7|0.018750
+8|0.018750
+9|0.018750
+10|0.018750
+-STATEMENT CALL create_projected_graph('err', ['person'], ['studyAt']);
+---- error
+Binder exception: organisation is connected to studyAt but not projected.
+-STATEMENT CALL create_projected_graph('err', ['person'], 'studyAt');
+---- error
+Binder exception: studyAt has data type STRING. STRUCT or LIST was expected.
```
ray6080, line 51: The error message is confusing. You mean the input param's data type is mismatched? Is the following understand correct?
```
Input param 'studyAt' has unexpected data type STRING. We expect STRUCT or LIST as input.
```

test/test_files/function/gds/special_projected_graph.test
```diff
@@ -0,0 +1,57 @@
...
+-STATEMENT CALL create_projected_graph('err', ['person'], {'knows': ['filter']});
+---- error
+Binder exception: [filter] has data type STRING[]. STRUCT was expected.
+-STATEMENT CALL create_projected_graph('err', ['person'], {'knows': {'dummy': 'r.date > date("1999-01-01")'}});
+---- error
+Binder exception: Unrecognized configuration dummy.
```
ray6080, line 57: Can we also provide supported configurations in the error message?

test/test_files/function/gds/special_projected_graph.test
ray6080, line 1: Can you add a few more erroring test cases?
- filter on non-existing column
- incorrect filter in grammar

ray6080 (approved):

PR #4934, opened 2025-02-19
Parallel Distinct SimpleAggregate

benjaminwinger:
Extends the same method used to parallelize the hash aggregate and distinct hash aggregate to the simple, non-grouped aggregates.

~~There is still a reasonable amount of work being done single-threaded since after doing the parallel build + finalize of the distinct hash table we have to then scan it to actually compute the function result and that's currently being done in the single-threaded finalize since it can't be started until the parallel finalization is complete.
It would be possible to add another operator which does that in parallel, which I think could yield up to another 2x improvement to the runtime (skipping this step dropped both of the below benchmarks to \~0.9s, so this single-threaded part is clearly a significant part of the total runtime). But I think that's best left for a different PR.~~
 (*Edit: this can actually be done in the current operator fairly easily, we just need to store a state for each partition and combine them in finalize instead*).

| Query | master | 1 thread | 128 threads |
| --- | --- | --- | --- |
| `MATCH (h:hits) RETURN COUNT(DISTINCT h.UserID);` | 20.5s | 18.6s | 0.77s |
| `MATCH (h:hits) RETURN COUNT(DISTINCT h.SearchPhrase);` | 16.6s | 15.9s | 0.85s |

~Probably needs a bit more cleanup so I'm opening as a draft for now.~ (Should be good now; I updated the benchmarks twice between what was mentioned above as well as a small optimization to make it use fixed-sized hash tables prior to partitioning like in the hash aggregate).

### DIFF
--- a/src/include/processor/operator/aggregate/aggregate_hash_table.h
+++ b/src/include/processor/operator/aggregate/aggregate_hash_table.h
@@ -12,6 +12,9 @@
 #include "processor/result/factorized_table_schema.h"
 
 namespace kuzu {
+namespace common {
+class InMemOverflowBuffer;
+}
 namespace storage {
 class MemoryManager;
 }
@@ -66,7 +69,7 @@ public:
 
     // Returns true if the value was distinct and was inserted
     // otherwise if the value already existed, returns false and the hash table is unchanged
-    bool insertAggregateValueIfDistinctForGroupByKeys(
+    virtual bool insertAggregateValueIfDistinctForGroupByKeys(
         const std::vector<common::ValueVector*>& groupByKeyVectors,
         common::ValueVector* aggregateVector);
 
@@ -271,16 +274,28 @@ struct AggregateHashTableUtils {
         storage::MemoryManager& memoryManager,
         const std::vector<common::LogicalType>& groupByKeyTypes,
         const common::LogicalType& distinctKeyType);
+
+    static FactorizedTableSchema getTableSchemaForKeys(
+        const std::vector<common::LogicalType>& groupByKeyTypes,
+        const common::LogicalType& distinctKeyType);
 };
 
-// NOLINTNEXTLINE(cppcoreguidelines-virtual-class-destructor): This is a final class.
-class HashAggregateSharedState;
+// Separate class since the SimpleAggregate has multiple different top-level destinations for
+// partitioning
+class AggregatePartitioningData {
+public:
+    virtual ~AggregatePartitioningData() = default;
+    virtual void appendTuples(const FactorizedTable& table, ft_col_offset_t hashOffset) = 0;
+    virtual void appendDistinctTuple(size_t /*distinctFuncIndex*/, std::span<uint8_t> /*tuple*/,
+        common::hash_t /*hash*/) = 0;
+    virtual void appendOverflow(common::InMemOverflowBuffer&& overflowBuffer) = 0;
+};
 
 // Fixed-sized Aggregate hash table that flushes tuples into partitions in the
 // HashAggregateSharedState when full
 class PartitioningAggregateHashTable final : public AggregateHashTable {
 public:
-    PartitioningAggregateHashTable(HashAggregateSharedState* sharedState,
+    PartitioningAggregateHashTable(AggregatePartitioningData* partitioningData,
         storage::MemoryManager& memoryManager, std::vector<common::LogicalType> keyTypes,
         std::vector<common::LogicalType> payloadTypes,
         const std::vector<function::AggregateFunction>& aggregateFunctions,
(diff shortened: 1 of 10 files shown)


src/include/processor/operator/aggregate/simple_aggregate.h
```diff
@@ -8,23 +10,68 @@ namespace processor {
...
+    private:
+        SimpleAggregateSharedState* sharedState;
+        size_t functionIdx;
+    };
+
 private:
+    bool readyForFinalization;
+    bool hasDistinct;
+    common::MPSCQueue<std::unique_ptr<common::InMemOverflowBuffer>> overflow;
+    std::vector<Partition> globalPartitions;
```
ray6080, line 73: `readyForFinalization`, `overflow` and `globalPartitions`  can be moved into `BaseAggregateSharedState`.

src/processor/operator/aggregate/simple_aggregate.cpp
```diff
@@ -76,27 +172,69 @@ void SimpleAggregate::executeInternal(ExecutionContext* context) {
...
-    function::AggregateFunction* function, AggregateInput* input, function::AggregateState* state,
-    storage::MemoryManager* memoryManager) {
-    auto multiplicity = 1; // Distinct aggregate should ignore multiplicity.
-    if (distinctHT->insertAggregateValueIfDistinctForGroupByKeys(std::vector<ValueVector*>{},
-            input->aggregateVector)) {
-        auto pos = input->aggregateVector->state->getSelVector()[0];
-        if (!input->aggregateVector->isNull(pos)) {
-            function->updatePosState((uint8_t*)state, input->aggregateVector, multiplicity, pos,
-                memoryManager);
+void SimpleAggregateSharedState::finalizeDistinctHashTables(storage::MemoryManager* memoryManager,
```
ray6080, line 193: Minor comment on code style, can you reorganize the ordering of function implementations based on their class? Easier to track by just scrolling through the file.

src/processor/operator/aggregate/simple_aggregate.cpp
```diff
@@ -76,27 +172,69 @@ void SimpleAggregate::executeInternal(ExecutionContext* context) {
...
+    const std::vector<AggregateInfo>& aggInfos) {
+    if (!hasDistinct) {
+        return;
+    }
+    for (auto& partition : globalPartitions) {
+        if (!partition.finalized && partition.mtx.try_lock()) {
+            if (partition.finalized) {
+                // If there was a data race in the above && a thread may get through after another
```
ray6080, line 201: Instead of duplicate the comment and locking/unlocking logic here with `HashAggregateSharedState::finalizeAggregateHashTable()`, can we abstract the common logic into a base function? might be something long this line in the high level.
```cpp
void finalizePartitions(std::vector<Partition> globalPartitions, function finalizeFunc) {
    for (auto& partition : globalPartitions) {
        if (partition.finalized || !partition.mtx.try_lock()) {
            continue;
        }
        if (partition.finalized) {
            // If there was a data race in the above && a thread may get through after another
            // thread has finalized this partition
            // Ignore coverage since we can't reliably test this data race
            // LCOV_EXCL_START
            partition.mtx.unlock();
            continue;
            // LCOV_EXCL_END
        }
        finalizeFunc(partition);
        partition.finalized = true;
        partition.mtx.unlock();
    }
}
```

ray6080 (approved):
Looks nice to me! Thanks!

Should we start to set up the ClickBenchmark internally?

src/processor/operator/aggregate/simple_aggregate.cpp
benjaminwinger: I think I missed that one; I was trying to keep them ordered. hash_aggregate.cpp is much worse though.

src/processor/operator/aggregate/aggregate_hash_table.cpp
```diff
@@ -59,6 +58,17 @@ uint64_t AggregateHashTable::append(const std::vector<ValueVector*>& flatKeyVect
     return numFlatTuples;
 }
 
+bool PartitioningAggregateHashTable::insertAggregateValueIfDistinctForGroupByKeys(
+    const std::vector<ValueVector*>& groupByFlatKeyVectors, ValueVector* aggregateVector) {
+    auto startingNumTuples = getNumEntries();
+    if (startingNumTuples + 1 > maxNumHashSlots ||
+        (double)startingNumTuples + 1 > (double)maxNumHashSlots / DEFAULT_HT_LOAD_FACTOR) {
```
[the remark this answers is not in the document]
benjaminwinger: Yeah. There's also another spot with almost the same logic, so that could be combined into one function.

src/include/processor/operator/aggregate/aggregate_hash_table.h
```diff
@@ -81,7 +84,7 @@ class AggregateHashTable : public BaseHashTable {
```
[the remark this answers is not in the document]
benjaminwinger: It was removed in another PR as it wasn't being used.

src/include/processor/operator/aggregate/simple_aggregate.h
benjaminwinger: globalPartitions can't because there is a different Partition type for each of the two derived classes, but the others certainly can.

PR #4949, opened 2025-02-21
Skip inserting null keys into distinct hash tables

benjaminwinger:
We always skip them when updating the aggregate state, so there's no point including them in the table at all.

There was a small regression introduced in #4881, which led to null values being included in the distinct output.


acquamarin (approved):

benjaminwinger:
I added @acquamarin's fix from #4952, removed some unused code to appease codecov (as one of the changes from this PR was in a function that wasn't being used), and have added a test to reproduce the issue that de95f3bbc87f16d4ece68bffcd5ed88144d48137 fixed.

PR #4972, opened 2025-02-27
Only merge distinct aggregate hash tables into the global queues when full

benjaminwinger:
Instead of whenever their parent hash table is full. This improves performance when the distinct hash tables fill up significantly more quickly than their parents.

| Query | before | after |
| --- | --- | --- |
| `MATCH (h:hits) RETURN h.RegionID, COUNT(DISTINCT h.UserID) AS u ORDER BY u DESC LIMIT 10;` | 2.5s | 2.2s |
| `MATCH (h:hits) RETURN h.RegionID, SUM(h.AdvEngineID), COUNT(*) AS c, AVG(h.ResolutionWidth), COUNT(DISTINCT h.UserID) ORDER BY c DESC LIMIT 10;` | 2.7s | 2.3s |

In more extreme cases the difference would probably be larger.

### DIFF
--- a/src/processor/operator/aggregate/aggregate_hash_table.cpp
+++ b/src/processor/operator/aggregate/aggregate_hash_table.cpp
@@ -45,18 +45,9 @@ AggregateHashTable::AggregateHashTable(MemoryManager& memoryManager,
     initializeTmpVectors();
 }
 
-bool shouldMergeAll(uint64_t startingNumTuples, uint64_t newNumTuples, uint64_t currentMax) {
-    return startingNumTuples + newNumTuples > currentMax ||
-           static_cast<double>(startingNumTuples) + newNumTuples >
-               static_cast<double>(currentMax) / DEFAULT_HT_LOAD_FACTOR;
-}
-
 bool PartitioningAggregateHashTable::insertAggregateValueIfDistinctForGroupByKeys(
     const std::vector<ValueVector*>& groupByFlatKeyVectors, ValueVector* aggregateVector) {
-    auto startingNumTuples = getNumEntries();
-    if (shouldMergeAll(startingNumTuples, 1, maxNumHashSlots)) {
-        mergeAll();
-    }
+    mergeIfFull(1);
     return AggregateHashTable::insertAggregateValueIfDistinctForGroupByKeys(groupByFlatKeyVectors,
         aggregateVector);
 }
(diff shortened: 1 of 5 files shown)


src/processor/operator/aggregate/aggregate_hash_table.cpp
```diff
@@ -804,9 +792,35 @@ uint64_t PartitioningAggregateHashTable::append(const std::vector<ValueVector*>&
     return numFlatTuples;
 }
 
-void PartitioningAggregateHashTable::mergeAll() {
-    partitioningData->appendTuples(*factorizedTable,
-        tableSchema.getColOffset(tableSchema.getNumColumns() - 1));
+bool hasSpace(uint64_t startingNumTuples, uint64_t newNumTuples, uint64_t currentMax) {
```
ray6080, line 795: Should this be `hasSpace` or `outOfSpace`?

src/processor/operator/aggregate/aggregate_hash_table.cpp
```diff
@@ -804,9 +792,35 @@ uint64_t PartitioningAggregateHashTable::append(const std::vector<ValueVector*>&
...
+    return startingNumTuples + newNumTuples > currentMax ||
+           static_cast<double>(startingNumTuples) + newNumTuples >
+               static_cast<double>(currentMax) / DEFAULT_HT_LOAD_FACTOR;
+}
+
+void PartitioningAggregateHashTable::mergeIfFull(uint64_t tuplesToAdd, bool mergeAll) {
+    if (mergeAll || hasSpace(factorizedTable->getNumTuples(), tuplesToAdd, maxNumHashSlots)) {
```
ray6080, line 802: I would refactor this into a function and reuse it at L846 too.

ray6080 (approved):

src/processor/operator/aggregate/aggregate_hash_table.cpp
benjaminwinger: Yes, I guess I got that inverted somehow

src/processor/operator/aggregate/aggregate_hash_table.cpp
benjaminwinger: I've simplified it a little. I'm not sure if it helps to pass `mergeAll` into that function, but the rest I could simply a bit.

Issue #4943, opened 2025-02-21
Bug: Buffer manager exception for simple queries on large data


NeuralFlux:
### Kùzu version

v0.8.0

### What operating system are you using?

MacOS Sequoia 15.2

### What happened?

I'm using the Python API. I have sample biomedical data with about 150k nodes and 1 million edges. Most queries I've tried fail with buffer manager exception. Is this expected? Am I hitting the limits of KuzuDB (for my machine)?

Sample queries that work -
```Cypher
MATCH (a) RETURN COUNT(a);  // count total nodes

MATCH (g:Gene) WHERE g.name CONTAINS 'kinase' RETURN g;  // property filtering for a node
```

Sample queries that fail - 
```Cypher
MATCH ()-[r]->() RETURN COUNT(r);  // count total edges

MATCH (n)-[r]-(m) WHERE n.id = 'C0429886' AND m.id = 'C0015943' RETURN r;  // get all edges between 2 nodes
```

Some more info - I have around 30 node tables and 1800 rel tables, MacBook Pro 3 with 18GB memory

### Are there known steps to reproduce?

Here's the Kuzu export in CSV format to help reproduce the errors. [semmed_export.zip](https://github.com/user-attachments/files/18901421/semmed_export.zip)

ray6080:
hi @NeuralFlux thanks for reporting this. will take a look asap.

ray6080:
hi @NeuralFlux first, let me clarify that you're not hitting the limit of Kuzu 😄 instead this is a performance bug, and thanks for spotting this for us. The bm exception is due to a non-optimized memory usage when scanning from multiple labelled rel tables. We will optimize this away soon. 😄
As a workaround for now, can you try reduce num of threads, e.g., `CALL threads=1;`, or explicitly specify labels if applicable?

NeuralFlux:
Thanks for the quick response, @ray6080 ! Limiting the number of threads did the job. I was mistaken about Kuzu's performance :)

NeuralFlux:
If it's not proprietary, may I know how you debugged this? Just curious, it might help sort issues by myself in the future, haha.

ray6080:
> If it's not proprietary, may I know how you debugged this? Just curious, it might help sort issues by myself in the future, haha.

[...]

PR #4975, opened 2025-02-28
Reuse scan state for multiple table scans

ray6080:
# Description

This optimization greatly elevates the overheads of scan state allocation when scanning of multi-label node tables or rel tables. A simple case is referred here (https://github.com/kuzudb/kuzu/issues/4943). Now the queries can run with ~**24MB** bm, which were previously reported to OOM with **12.8GB** bm.

The idea is to decouple table specific info from the scan state by introducing a new interface `setToTable`, which needs to be called before `initScanState` and actual reads.
To make the PR minimal, a new constructor of `TableScanState` is introduced to keep previous ones as they were, so they can be refactored in following PRs.


src/include/storage/store/table.h
```diff
@@ -49,7 +62,13 @@ struct TableScanState {
     virtual ~TableScanState() = default;
     DELETE_COPY_DEFAULT_MOVE(TableScanState);
 
-    virtual void initState(transaction::Transaction* transaction, NodeGroup* nodeGroup) {
+    virtual void setToTable(Table* table_, std::vector<common::column_id_t> columnIDs_,
```
andyfengHKU, line 65: Up to u but seeing rel table specific field `direction & resetCachedBoundNodeSelVec` is a bit weird.

andyfengHKU (approved):

src/include/storage/store/table.h
ray6080: Yeah it's a bit annoying, but I will see if I can improve this in the coming refactoring.

ted-wq-x:
@ray6080  This PR greatly increases the number of times the csrHeader is scanned. For algorithms like khop that require a large number of edge traversals, the performance is severely degraded.

Issue #4936, opened 2025-02-19
Bug: Out of Memory for large data


kasiSundar:
### What happened?

I’m facing an out of memory issue when I tried to load a huge data volume.

For example, here is my scenario:

- my source data is 90GB (list of parquet files)
- kuzu runs on AKS pod with 32GB RAM size
- my destination storage size is 150GB ( where i write the kuzudb)

I was having an Out of Memory issue when I tried to read parquet file one by one and write it to a kuzudb. 

I tried adding conn.close() after each parquet file read but still I’m getting OOM issue. 

Can you please help me with this?

Also a question, does Kuzu stores all data in-memory as well before writing it to disk, if so how to flush them?

prrao87:
Hi @kasiSundar, for us to be able to test, it would be helpful if you could provide the version number of Kùzu you're seeing this problem on. Additionally, please post the code you're using to ingest data into Kùzu.

Starting from 0.7.0 of Kùzu, we allow large `COPY FROM` jobs to spill to disk. Our [release blog](https://blog.kuzudb.com/post/kuzu-0.7.0-release/#data-spilling-during-bulk-relationship-ingestion-copy-from) describes some of these features. The spill to disk should be automatic (no user action necessary), but you would have to be doing it via `COPY FROM` on the latest versions (0.7.0 and above).

kasiSundar:
I’m using Kuzu 0.8.0, but still have this issue

kasiSundar:
Here is the more details on the experiment I tried:

- input data size 22GB (parquet files)
- steps included which building graph:

1. Create schema
2. Combine and convert parquet files into pandas dataframe
3. Extract list of nodes from data and compare if nodes already exists in graph, if not create nodes in graph
4. Create relationship directly from dataframe
5. After steps are looped to run 3 chunks of data each approx 7GB data. 

The overall in-memory utilized during this process is approx 235GB out of 264GB in memory

The final graph database created holds 120GB data +  9.9GB index

The overall time took to build the graph is 1 hour 30 minutes. Total nodes created 207 million and 1 billion relationship out of 1 billion records.

We used Python code to build this on AKS pod. 

And this is successful run with 264GB RAM for 22GB input data, where the OOM occurred with 80GB RAM. I believe it would have almost thrown OOM even with 264GB RAM since I utilized almost 235GB RAM.

ray6080:
> Extract list of nodes from data and compare if nodes already exists in graph, if not create nodes in graph

[...]

PR #4980, opened 2025-02-28
Statistics update optimization

benjaminwinger:
As far as I can tell, the issue in #4936 (where speed is concerned; I still need to look into memory usage) is that it's hitting a particularly slow way of updating our statistics. Copying from CSV is significantly better, but was still slower than it could be.

I've optimized this particular part of the code by removing the repeated calls to the single-input `ColumnChunkStats::update` and replacing it with code that calculates the min/max of the vector and passing those to the ColumnChunkStats just once (I also inlined the single-input update to remove the unnecessary optionals).

~~I'm still a little concerned by how much time is spent updating statistics when copying relationship data (even after the optimization, \~86% of the time spent in `ChunkedNodeGroup::append` is spent in `ColumnChunkData::updateStats` (and `ChunkedNodeGroup::append` itself is 76% of the runtime of the copies). I'll try and look further into that to see if it's actually reasonable, but this is a significant improvement by itself.~~ *Update: Figured this out: https://github.com/kuzudb/kuzu/pull/4980#issuecomment-2694911770. Due to the partitioner selecting just one value from a vector at a time, and that the SelVector passed in is not the same as the one on the vector (which I think usually has the full vector selected), we end up doing vastly more work than necessary. Of course, optimising the update still will help.*

Copying the first 10000000 edges from the datagen-9.0-fb dataset:

| | Before | After |
| --- | --- | --- |
| Edges from CSV | 13s | 2.3s |
| Edges from Parquet | 70s | 1.8s |
| Edges from Pandas Dataframe (created from parquet in python) | 70s | 1.8s |

### DIFF
--- a/src/include/common/vector/value_vector.h
+++ b/src/include/common/vector/value_vector.h
@@ -1,5 +1,6 @@
 #pragma once
 
+#include <optional>
 #include <utility>
 
 #include "common/assert.h"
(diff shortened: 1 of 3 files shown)


benjaminwinger:
It looks like the performance discrepancy is because we're partitioning and it is appending a single value at a time. We should probably buffer the appended values, or delay the statistics updates somehow, as it would be much more efficient to calculate the statistics in batches, or even as a whole after the partitioner is done.

ray6080 (approved):
Nice catch on this!

PR #4981, opened 2025-03-01
Refactor table scan state interfaces

ray6080:
# Description

This PR mainly propagates `setToTable` interface and unifies ScanState constructors, which avoids setting ValueVector and chunkState manually.


src/include/storage/store/table.h
```diff
@@ -43,26 +42,21 @@ struct KUZU_API TableScanState {
...
 
-    TableScanState(common::table_id_t tableID, std::vector<common::column_id_t> columnIDs,
-        std::vector<const Column*> columns = {},
-        std::vector<ColumnPredicateSet> columnPredicateSets = {})
-        : tableID{tableID}, nodeIDVector(nullptr), outState{nullptr},
-          columnIDs{std::move(columnIDs)}, semiMask{nullptr}, columns{std::move(columns)},
-          columnPredicateSets{std::move(columnPredicateSets)} {
-        rowIdxVector = std::make_unique<common::ValueVector>(common::LogicalType::INT64());
-    }
+    TableScanState(std::vector<common::column_id_t> columnIDs, std::vector<const Column*> columns)
```
andyfengHKU, line 51: You might wanna double check if we need `columnIDs` & `columns` at the same time

andyfengHKU (approved):

src/include/storage/store/table.h
ray6080: This is indeed redundant semantically, but as I'm not passing in `Table`, so they are necessary for now. But I think I should refactor `Column` to keep its own `columnID`, which makes more sense.

PR #4998, opened 2025-03-05
Selection vector slicing with lightweight SelectionView

benjaminwinger:
The goal was to remove the selection vector copy in `ChunkedNodeGroup::append`, since the partitioner used in rel table copies appends values one at a time and the allocation is relatively expensive.

I ended up creating a parent class for the SelectionVector called SelectionView which is a read-only version of the SelectionVector, and then `SelectionVector::slice` produces a SelectionView over a subsequence of the original selection vector (you can also create SelectionViews over the static `INCREMENTAL_SELECTED_POS` array which don't do any heap allocations using its constructor). I had considered just a variant of the SelectionVector which doesn't allocate anything, but I think this works out to be a cleaner interface, and it wouldn't have been possible to easily guarantee that a sliced SelectionVector wouldn't be able to modify the original data using `SelectionVector::operator[]`.

Copying 100 000 000 edges from the datagen-9_0-fb dataset took ~17s before this change and ~15s after.

### DIFF
--- a/src/common/data_chunk/sel_vector.cpp
+++ b/src/common/data_chunk/sel_vector.cpp
@@ -18,15 +18,19 @@ static const std::array<sel_t, DEFAULT_VECTOR_CAPACITY> INCREMENTAL_SELECTED_POS
         return selectedPos;
     }();
 
+SelectionView::SelectionView(sel_t startPos, sel_t selectedSize)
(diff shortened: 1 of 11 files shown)


ray6080 (approved):
LGTM!

PR #5002, opened 2025-03-06
Add is ready only field to function

### LINKED ISSUE #3502: Add Read-only Transaction check to function calls
At `client_context.cpp:320-331` we check if a statement would be a valid read-only statement for read-only transactions. However, the expressions within the statements could contain functions that could perform a write. We currently don't check this. As well, we don't seem to check if macro functions are read-only either. We perhaps want to handle this in the binding stage instead of the parser stage as above.

andyfengHKU:
# Description

Add readOnly field to function class. Avoid naming-based check.

Fixes #3502 

# Contributor agreement

### DIFF
--- a/src/main/client_context.cpp
+++ b/src/main/client_context.cpp
@@ -462,7 +462,12 @@ std::unique_ptr<PreparedStatement> ClientContext::prepareNoLock(
     prepareTimer.start();
     try {
         preparedStatement->preparedSummary.statementType = parsedStatement->getStatementType();
-        preparedStatement->readOnly = StatementReadWriteAnalyzer().isReadOnly(*parsedStatement);
+        auto readWriteAnalyzer = StatementReadWriteAnalyzer(this);
+        TransactionHelper::runFuncInTransaction(
+            *transactionContext, [&]() -> void { readWriteAnalyzer.visit(*parsedStatement); },
+            true /* readOnly */, false /* */,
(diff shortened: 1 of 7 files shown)


src/main/client_context.cpp
```diff
@@ -462,7 +462,12 @@ std::unique_ptr<PreparedStatement> ClientContext::prepareNoLock(
     prepareTimer.start();
     try {
         preparedStatement->preparedSummary.statementType = parsedStatement->getStatementType();
-        preparedStatement->readOnly = StatementReadWriteAnalyzer().isReadOnly(*parsedStatement);
+        auto readWriteAnalyzer = StatementReadWriteAnalyzer(this);
+        TransactionHelper::runFuncInTransaction(
```
ray6080, line 466: I would use DUMMY_READ here unless we find out later that this doesn't work.

ray6080 (approved):
Here are more functions should be marked as write:
- ClearWarnings
- CreateProjectedGraph
- DropProjectedGraph
- LocalCacheArrayColumn （This is debatable, but I think it makes sense to mark it as write)
- CreateHNSWIndex
- InternalCreateHNSWIndexFunction
- InternalFinalizeHNSWIndexFunction
- DropHNSWIndex
- InternalDropHNSWIndexFunction
- CreateFTSIndex
- InternalCreateFTSFunction
- DropFTSIndex
- InternalDropFTSFunction

PR #5006, opened 2025-03-06
Refactor gds frontier

andyfengHKU:
# Description

Remove unnecessary usage of std::atomic in `PathLengths`.

Remove `GDSFrontier` abstraction level, we don't have a second child class and I don't see we will have one in the foreseeable future.

Rename `getNumNodes` to `getMaxOffset` in graph interface. `NumNodes` is confusing/inaccurate when we introduce filtered graph. What we need is maxOffset that is used to initialize dense data structure.

### DIFF
--- a/src/include/function/gds/gds_frontier.h
+++ b/src/include/function/gds/gds_frontier.h
@@ -128,7 +128,9 @@ public:
         storage::MemoryManager* mm);
     ~PathLengths();
 
-    const common::table_id_map_t<common::offset_t>& getNodeMaxOffsetMap() const { return nodeMaxOffsetMap; }
+    const common::table_id_map_t<common::offset_t>& getNodeMaxOffsetMap() const {
+        return nodeMaxOffsetMap;
+    }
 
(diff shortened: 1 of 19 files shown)


src/include/function/gds/gds_frontier.h
```diff
@@ -155,45 +118,46 @@ class KUZU_API SparseFrontier {
  *
  * However, this is not necessary and the caller can also use this to represent a single frontier.
  */
-class KUZU_API PathLengths : public GDSFrontier {
```
ray6080, line 158: Should this be renamed to `DenseFrontier` now? I don't follow why `PathLengths` is a generic name, as we should have cases where the frontier has nothing to do with path?

ray6080 (approved):

src/include/function/gds/gds_frontier.h
andyfengHKU: Maybe but I'm less motivated to do the renaming at this point.

PR #5007, opened 2025-03-06
Add job for testing simsimd dynamic dispatch to nightly build-and-deploy workflow

royi-luo:
# Description

Add a job to the nightly `build-and-deploy` workflow that tests to see if simsimd's dynamic dispatch is working as expected. The job will download the nightly build of the kuzu shell and run the test on a newly-selected CI runner. Ideally this would over time test out all the combinations of build + execution machines.

# Contributor agreement

### DIFF
--- a/scripts/test-simsimd-dispatch.py
+++ b/scripts/test-simsimd-dispatch.py
@@ -1,0 +1,45 @@
+import gdb
+import subprocess
+
+
+def get_machine_architecture():
+    output = subprocess.check_output(["cat", "/proc/cpuinfo"])
+    flags_str = ""
+    for line in output.decode("utf-8").split("\n"):
+        line = line.strip()
+        if line.startswith("flags"):
+            flags_str = line
+            break
+    assert len(line) > 0
+
+    # Uses the same way as _simsimd_capabilities_x86() in simsimd.h to detect supported features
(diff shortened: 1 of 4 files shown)


.github/workflows/build-and-deploy.yml
```diff
@@ -472,3 +472,25 @@ jobs:
...
+        uses: actions/download-artifact@v4
+        with:
+          name: kuzu_cli-linux-x86_64
+
+      - name: Extract kuzu shell
+        run: |
+          tar xf kuzu_cli-linux-x86_64.tar.gz
+
+      - name: Test
+        run: gdb --batch -x scripts/test-simsimd-dispatch.py --args ./kuzu
```
royi-luo, line 496: Not sure if this is the best place to put this new job but it's here right now because it makes it easier to download the nightly build of the shell. @mewim are there other ways to download the nightly build that would work with this job in a different workflow file?

scripts/test-simsimd-dispatch.py
```diff
@@ -0,0 +1,45 @@
+import gdb
```
royi-luo, line 1: Not sure if this is the best place to put the testing script, let me know if there is a better place

mewim (approved):
LGTM

PR #5008, opened 2025-03-06
Add squared distance function for arrays

royi-luo:
# Description

We offer the `l2_squared` distance function for vector index so we should offer a version for arrays as well. The other distance functions offered by vector index (`l2`, `cos`, `dot`) already exist.

Since we now have the library, I also modified the applicable array functions (distance/cos/inner product) to use simsimd for better performance.

# Contributor agreement

### DIFF
--- a/src/include/function/array/functions/array_distance.h
+++ b/src/include/function/array/functions/array_distance.h
@@ -10,7 +10,7 @@ namespace function {
 
 // Euclidean distance between two arrays.
 struct ArrayDistance {
-    template<typename T>
+    template<std::floating_point T>
     static inline void operation(common::list_entry_t& left, common::list_entry_t& right, T& result,
         common::ValueVector& leftVector, common::ValueVector& rightVector,
         common::ValueVector& resultVector) {
--- a/src/include/function/array/functions/array_distance.h
+++ b/src/include/function/array/functions/array_distance.h
@@ -3,6 +3,7 @@
 #include "math.h"
 
 #include "common/vector/value_vector.h"
+#include "function/array/functions/array_squared_distance.h"
 
 namespace kuzu {
 namespace function {
@@ -12,15 +13,8 @@ struct ArrayDistance {
     template<typename T>
     static inline void operation(common::list_entry_t& left, common::list_entry_t& right, T& result,
         common::ValueVector& leftVector, common::ValueVector& rightVector,
-        common::ValueVector& /*resultVector*/) {
-        auto leftElements = (T*)common::ListVector::getListValues(&leftVector, left);
-        auto rightElements = (T*)common::ListVector::getListValues(&rightVector, right);
-        KU_ASSERT(left.size == right.size);
-        result = 0;
-        for (auto i = 0u; i < left.size; i++) {
-            auto diff = leftElements[i] - rightElements[i];
-            result += diff * diff;
-        }
+        common::ValueVector& resultVector) {
+        ArraySquaredDistance::operation(left, right, result, leftVector, rightVector, resultVector);
         result = std::sqrt(result);
     }
 };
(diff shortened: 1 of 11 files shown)


src/include/function/array/vector_array_functions.h
```diff
@@ -30,6 +30,12 @@ struct ArrayDistanceFunction {
     static function_set getFunctionSet();
 };
 
+struct ArraySquaredDistanceFunction {
+    static constexpr const char* name = "ARRAY_SQUARED_DISTANCE";
```
royi-luo, line 34: Following the previous convention `ARRAY_DISTANCE` for naming. I'm wondering if for consistency is it better to rename these to `ARRAY_L2_DISTANCE` and `ARRAY_L2_SQUARED_DISTANCE`?

third_party/simsimd/lib.c
```diff
@@ -7,12 +7,6 @@
 #define SIMSIMD_NATIVE_F16 0
 #define SIMSIMD_NATIVE_BF16 0
 
-/*  Override the primary serial operations to avoid the LibC dependency.
- */
-#define SIMSIMD_SQRT(x) simsimd_approximate_square_root(x)
-#define SIMSIMD_RSQRT(x) simsimd_approximate_inverse_square_root(x)
-#define SIMSIMD_LOG(x) simsimd_approximate_log(x)
-
```
royi-luo, line 15: We have no need to avoid the libc dependency and these functions are less accurate on wasm

ray6080: Why is this? I thought avoiding libc dep is to: 1) improve compatibility; 2) improve performance?

test/test_files/function/hnsw/small.test
ray6080, line 1: Can we also add tests on array distance computation with datasets used here? Basically I meant that we should get same result for these queries without hnsw index.

src/include/function/array/functions/array_distance.h
```diff
@@ -3,24 +3,18 @@
 #include "math.h"
 
 #include "common/vector/value_vector.h"
+#include "function/array/functions/array_squared_distance.h"
 
 namespace kuzu {
 namespace function {
 
 // Euclidean distance between two arrays.
 struct ArrayDistance {
```
ray6080, line 12: I don't think we document this function in our docs 😂 can you double check?

ray6080 (approved):
Looks nice to me.

I have a comment on disabling approximate operations. we can discuss on that.

Also, can you update our docs accordingly? Thanks!

third_party/simsimd/lib.c
royi-luo: The simsimd library was set up to be built as a shared library (if dynamic dispatch is enabled) so it makes sense for it to improve compatibility. However we link against the library statically and I'm pretty sure we already have many libc dependencies so I don't think the benefit applies to us.

Also if you use the library with dynamic dispatch disabled (as a header-only library) they use the built-in math functions anyways (probably assuming you'll link it into you executable anyways) so I don't think there should be any performance difference.

royi-luo: Hmm from reading their docs it does seem like there is a performance difference, but they mention using the `rsqrt` instruction instead of approximating it using magic numbers. I don't see that anywhere, let me check their master version to see if we removed something when copying it over.

royi-luo: Never mind it's there. So their vectorized versions for each function will use the a different fast version of the function. The approximate functions were only used for the serial implementations. I think it's safe to keep using the standard library implementation of `sqrt()` to keep things simple and accurate since performance-wise it only affects the serial version

src/include/function/array/functions/array_distance.h
royi-luo: Yea it's not there I'll update the docs

test/test_files/function/hnsw/small.test
royi-luo: added

PR #5011, opened 2025-03-07
gds: init parallel scc

sdht0:
Parallel version of SCC using the coloring algorithm.

# Contributor agreement

### DIFF
--- a/src/function/gds/strongly_connected_components.cpp
+++ b/src/function/gds/strongly_connected_components.cpp
@@ -133,16 +133,14 @@ public:
     bool beginOnTable(table_id_t) override { return true; }
 
     void vertexCompute(offset_t startOffset, offset_t endOffset, table_id_t) override {
-        frontierPair.getCurDenseFrontier().ptrCast<PathLengths>()->resetCurIter();
-        frontierPair.getNextDenseFrontier().ptrCast<PathLengths>()->resetCurIter();
+        frontierPair.getCurDenseFrontier().resetCurIter();
+        frontierPair.getNextDenseFrontier().resetCurIter();
         for (auto i = startOffset; i < endOffset; ++i) {
             // If the SCC ID has already been computed, the node should not be activated.
             auto initialState = computationState.isSccIdSet(i) ? PathLengths::INITIAL_VISITED :
                                                                  PathLengths::UNVISITED;
-            // FIXME: avoid dynamic casts?
-            frontierPair.getCurDenseFrontier().ptrCast<PathLengths>()->setCurFrontierValue(i,
-                initialState);
-            frontierPair.getNextDenseFrontier().ptrCast<PathLengths>()->setCurFrontierValue(i,
+            frontierPair.getCurDenseFrontier().setCurFrontierValue(i, initialState);
+            frontierPair.getNextDenseFrontier().setCurFrontierValue(i,
                 PathLengths::INITIAL_VISITED);
         }
     }
@@ -274,8 +272,8 @@ public:
         auto clientContext = context->clientContext;
         auto mm = clientContext->getMemoryManager();
         auto graph = sharedState->graph.get();
-        auto numNodesMap = graph->getNumNodesMap(clientContext->getTransaction());
-        auto it = numNodesMap.begin();
+        auto getMaxOffsetMap = graph->getMaxOffsetMap(clientContext->getTransaction());
+        auto it = getMaxOffsetMap.begin();
         auto tableId = it->first;
         auto numNodes = it->second;
 
@@ -285,8 +283,8 @@ public:
         GDSUtils::runVertexCompute(context, graph, *setInitialSccIds);
 
         // The frontiers will be initialized inside the loop.
-        auto currentFrontier = std::make_shared<PathLengths>(numNodesMap, mm);
-        auto nextFrontier = std::make_shared<PathLengths>(numNodesMap, mm);
+        auto currentFrontier = std::make_shared<PathLengths>(getMaxOffsetMap, mm);
+        auto nextFrontier = std::make_shared<PathLengths>(getMaxOffsetMap, mm);
         currentFrontier->pinCurFrontierTableID(tableId);
         nextFrontier->pinCurFrontierTableID(tableId);
         auto frontierPair =
--- a/src/function/gds/strongly_connected_components.cpp
+++ b/src/function/gds/strongly_connected_components.cpp
@@ -1,5 +1,3 @@
-#include <iostream>
-
 #include "binder/binder.h"
 #include "binder/expression/expression_util.h"
 #include "common/exception/runtime.h"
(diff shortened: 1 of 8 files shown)


src/function/gds/strongly_connected_components.cpp
```diff
@@ -0,0 +1,349 @@
...
+
+    std::unique_ptr<VertexCompute> copy() override {
+        return std::make_unique<SetNewSccIds>(computationState);
+    }
+
+private:
+    SCCComputationState& computationState;
+};
+
+class SetInitialWccIds : public VertexCompute {
```
semihsalihoglu-uw, line 107: WccIds -> just Ids

src/include/function/gds/gds_object_manager.h
```diff
@@ -35,6 +35,56 @@ class ObjectBlock {
...
+
+// Pre-allocated array of atomic objects.
+template<typename T>
+class AtomicObjectArray {
+public:
+    AtomicObjectArray(common::offset_t size, storage::MemoryManager* mm, bool initializeToZero = false)
+        : array{ObjectArray<std::atomic<T>>(size, mm, initializeToZero)} {
+    }
+
+    void set(common::offset_t pos, T& value) {
```
semihsalihoglu-uw, line 70: Let's name these operations setRelaxed so it's clear what level of guarantee the set and get functions give.

src/function/gds/strongly_connected_components.cpp
```diff
@@ -0,0 +1,349 @@
...
+using namespace kuzu::storage;
+using namespace kuzu::graph;
+
+namespace kuzu {
+namespace function {
+
+static constexpr offset_t SCC_UNSET = std::numeric_limits<offset_t>::max();
+
+struct SCCComputationState {
+    bool allSccIdsSet;
```
semihsalihoglu-uw, line 29: Again, follow the principle about atomics and make every value that will be accessed by multiple threads atomics.

src/function/gds/strongly_connected_components.cpp
```diff
@@ -0,0 +1,349 @@
...
+    void vertexCompute(offset_t startOffset, offset_t endOffset, table_id_t) override {
+        for (auto i = startOffset; i < endOffset; ++i) {
+            if (!computationState.isSccIdSet(i)) {
+                auto fwdColor = computationState.fwdWccIDs.get(i);
+                auto bwdColor = computationState.bwdWccIDs.get(i);
+                if (fwdColor == bwdColor) {
+                    computationState.sccIDs.set(i, fwdColor);
+                } else {
+                    // If any of the threads find an unset scc id, they will set the flag to false.
+                    // `allSccIdsSet` is never read and written concurrently, so it does
```
semihsalihoglu-uw, line 89: This is not the right principle. Any value that will be concurrently written to (even if multiple threads will write the same value to it), should be made an atomic. If multiple threads are writing the same value and at the end of this writing phase there is a memory barrier (e.g., a common lock acquisition across all threads), then the writes can be memory_order_relaxed, which is as if you're not using any atomics.

src/function/gds/strongly_connected_components.cpp
```diff
@@ -0,0 +1,349 @@
...
+                    // not need to be atomic.
+                    computationState.allSccIdsSet = false;
+                }
+            }
+        }
+    }
+
+    void reset() { computationState.allSccIdsSet = true; }
```
semihsalihoglu-uw, line 97: For this reset function, you should probably use memory_order_seq_cst, so you guarantee it's visible to every thread when the iterations start.

semihsalihoglu-uw (approved):
I have a few important comments about the use of atomics. I should take another quick look after you address them.

src/function/gds/strongly_connected_components.cpp
```diff
@@ -0,0 +1,346 @@
...
+
+            // Bwd colors.
+            computeState.frontierPair->initState();
+            computeState.frontierPair->initGDS();
+            GDSUtils::runVertexCompute(context, graph, *initializeFrontiers);
+            GDSUtils::runFrontiersUntilConvergence(context, computeState, graph,
+                ExtendDirection::BWD, MAX_ITERATION);
+
+            // Find new SCC IDs and exit if all IDs have been found.
+            setNewSccIds->reset();
```
andyfengHKU, line 319: Instead of reset from a vertex compute function. I would directly reset the SCCComputationState.

src/function/gds/strongly_connected_components.cpp
```diff
@@ -0,0 +1,346 @@
...
+using namespace kuzu::processor;
+using namespace kuzu::storage;
+using namespace kuzu::graph;
+
+namespace kuzu {
+namespace function {
+
+static constexpr offset_t SCC_UNSET = std::numeric_limits<offset_t>::max();
+
+struct SCCComputationState {
```
andyfengHKU, line 28: GDSAuxiliaryState is to hold algorithm specific state. So u might want to make this a subclass of GDSAuxiliaryState. Later when u wanna extend SCC to multi-tabled case, you will find this useful.

andyfengHKU (approved):

src/function/gds/strongly_connected_components.cpp
sdht0: Do you mean just making it a subclass or use it differently too?

src/function/gds/strongly_connected_components.cpp
```diff
@@ -0,0 +1,346 @@
...
+
+class InitializeFrontiers : public VertexCompute {
+public:
+    InitializeFrontiers(FrontierPair& frontierPair, SCCComputationState& computationState)
+        : frontierPair{frontierPair}, computationState{computationState} {}
+
+    bool beginOnTable(table_id_t) override { return true; }
+
+    void vertexCompute(offset_t startOffset, offset_t endOffset, table_id_t) override {
+        frontierPair.getCurDenseFrontier().resetCurIter();
```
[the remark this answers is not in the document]
sdht0: So I tried moving this to `beginOnTable` but turns out this function is [not always called](https://github.com/kuzudb/kuzu/blob/15156bf19cdee694d9db649213b18ea0c224e30c/src/function/gds/gds_task.cpp#L117)?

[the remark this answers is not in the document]
andyfengHKU: It is called in GDSUtils::runVertexCompute. It is ran only once on the main thread but not on each local thread. I haven't encountered a case where we need to pin per thread (because all these data structures are globally accessible).

src/function/gds/strongly_connected_components.cpp
```diff
@@ -0,0 +1,346 @@
...
+        GDSUtils::runVertexCompute(context, graph, *setInitialSccIds);
+
+        // The frontiers will be initialized inside the loop.
+        auto currentFrontier = std::make_shared<PathLengths>(getMaxOffsetMap, mm);
+        auto nextFrontier = std::make_shared<PathLengths>(getMaxOffsetMap, mm);
+        currentFrontier->pinCurFrontierTableID(tableId);
+        nextFrontier->pinCurFrontierTableID(tableId);
+        auto frontierPair =
+            std::make_unique<DoublePathLengthsFrontierPair>(currentFrontier, nextFrontier);
+        frontierPair->getCurSparseFrontier().disable();
```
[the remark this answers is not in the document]
andyfengHKU: Alright leave a TODO in the code and I'll take a look.

src/function/gds/strongly_connected_components.cpp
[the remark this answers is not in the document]
sdht0: Sorry incorrect comment. I moved this to `initState()`

PR #5014, opened 2025-03-07
Support handling null/deleted nodes during vector index creation

royi-luo:
# Description

Skip any nodes whose embeddings vector is marked as null during vector index creation. When caching the embeddings column, mark deleted nodes as null so that they are also skipped.

# Contributor agreement

### DIFF
--- a/src/function/table/cache_column.cpp
+++ b/src/function/table/cache_column.cpp
@@ -85,9 +85,8 @@ static std::unique_ptr<TableFuncSharedState> initSharedState(const TableFunction
     auto& table = input.context.getStorageManager()
                       ->getTable(bindData->tableEntry->getTableID())
                       ->cast<storage::NodeTable>();
-    auto numCommitedNodeGroups = table.getNumCommittedNodeGroups();
-    return std::make_unique<CacheArrayColumnSharedState>(table,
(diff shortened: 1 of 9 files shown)


src/function/table/cache_column.cpp
```diff
@@ -119,6 +118,28 @@ static std::unique_ptr<TableFuncLocalState> initLocalState(const TableFunctionIn
...
+    transaction::Transaction* transaction, storage::NodeTable& table) {
+    scanState.nodeGroupIdx = nodeGroupIdx;
+    table.initScanState(transaction, scanState);
+
+    // We want to ensure that the offsets in the cached column match the offsets in the
+    // table
+    // To do this we write to the same offsets and set any non-selected (e.g. deleted)
+    // rows to null
+    data->getNullData()->resetToAllNull();
+    while (table.scan(transaction, scanState)) {
```
ray6080, line 132: Should we instead add a slow path in `append` to handle nulls? Let's discuss this further.

royi-luo: Discussed offline, keeping this as-is (with an optimization to the `write` interface) as this is the simplest way of preserving the offsets from the deleted values.

ray6080 (approved):

PR #5015, opened 2025-03-07
Use copyNullMask instead of looping during copies of nulls

benjaminwinger:
The performance improvement isn't as good as I was first expecting, I think it was somewhat over-represented in the profiling.

Copying 100000000 edges from datagen-9_0-fb (as nodes in a parquet file) took ~12.8s before and ~12.2s after (~5% speedup).

I also noticed that the way we are calculating the statistics for nulls/bools was bugged in certain cases since for bools we weren't taking into account the offset of the input data in `getMinMaxStorageValue` (which required rewriting NullMask::getMinMax). That deficiency was being offset by `copyFromBuffer` just assuming that there are always both `false` and `true` values (but only if there are at least some true values, so it wasn't hiding the issue always). 
So this might improve read performance and compression of nulls in some cases where it was previously miscalculating/over-estimating the statistics, but I think in the trivial case where we're just appending full vectors at a time without any nulls it won't make a difference.

### DIFF
--- a/src/common/null_mask.cpp
+++ b/src/common/null_mask.cpp
@@ -28,11 +28,7 @@ bool NullMask::copyNullMask(const uint64_t* srcNullEntries, uint64_t srcOffset,
         bool anyNull = false;
         for (size_t i = 0; i < numBitsToCopy; i++) {
             bool isNull = NullMask::isNull(srcNullEntries, srcOffset + i);
-            if (invert) {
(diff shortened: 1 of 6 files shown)


src/common/null_mask.cpp
ray6080, line 1: Looks like we have a few cases not covered by tests. Can you double check?

ray6080 (approved):
👍

PR #5019, opened 2025-03-09
Filtered HNSW search

ray6080:
# Description

Implement the global adaptive filtered search.
The idea is to make use of semi mask to pass the filtering information on node table to the QueryHNSWIndex function.
Inside the function, we follow a heuristic rule based on the selectivity of semi mask:
- If the selectivity is pretty low (<=0.08), we would perform `blindTwoHopFilteredSearch`, which blindly explores the first and second hop neighbours to find those that are selected.
- If the selectivity is somewhat low (<=0.4), we would perform `directedTwoHopFilteredSearch`, which explores the first hop neighbours first to gat a candidate set, and then explore the second hop neighbours in a more directed way that prioritizes nodes in the candidate set that is closest to the query vector. This will hopefully leads to faster converge to the area with closeted nodes.
- If the selectivity is high (>0.4), we would perform `oneHopFilteredSearch`, which finds all selected nodes in the first hop neighbours.

There is a more advanced way, which is based on each node's selectivity to perform adaptive searches. We can move to that in a later PR.
Also, there are two left TODOs:
- add the brute force computation into the adaptive strategy.
- provide options for configuration of the threshold for each method.

### DIFF
--- a/src/function/table/hnsw/query_hnsw_index.cpp
+++ b/src/function/table/hnsw/query_hnsw_index.cpp
@@ -4,6 +4,8 @@
 #include "binder/query/reading_clause/bound_table_function_call.h"
 #include "catalog/catalog_entry/hnsw_index_catalog_entry.h"
 #include "catalog/catalog_entry/node_table_catalog_entry.h"
+#include "common/exception/binder.h"
+#include "common/mask.h"
 #include "common/types/value/nested.h"
 #include "expression_evaluator/expression_evaluator_utils.h"
 #include "function/table/bind_data.h"
@@ -25,10 +27,10 @@ QueryHNSWIndexBindData::QueryHNSWIndexBindData(main::ClientContext* context,
     storage::QueryHNSWConfig config, std::shared_ptr<binder::NodeExpression> outputNode)
     : TableFuncBindData{std::move(columns), 1 /*maxOffset*/}, context{context},
       boundInput{std::move(boundInput)}, config{config}, outputNode{std::move(outputNode)} {
-    auto indexEntry = this->boundInput.indexEntry;
+    const auto indexEntry = this->boundInput.indexEntry;
     auto& indexAuxInfo = indexEntry->getAuxInfo().cast<catalog::HNSWIndexAuxInfo>();
     indexColumnID = this->boundInput.nodeTableEntry->getColumnID(indexEntry->getPropertyIDs()[0]);
-    auto catalog = context->getCatalog();
+    const auto catalog = context->getCatalog();
     upperHNSWRelTableEntry =
         catalog->getTableCatalogEntry(context->getTransaction(), indexAuxInfo.upperRelTableID)
             ->ptrCast<catalog::RelTableCatalogEntry>();
@@ -56,14 +58,33 @@ static std::vector<common::LogicalType> inferInputTypes(const binder::expression
 static std::unique_ptr<TableFuncBindData> bindFunc(main::ClientContext* context,
     const TableFuncBindInput* input) {
     context->setUseInternalCatalogEntry(true /* useInternalCatalogEntry */);
-    const auto tableName = input->getLiteralVal<std::string>(0);
+    const auto tableOrGraphName = input->getLiteralVal<std::string>(0);
     const auto indexName = input->getLiteralVal<std::string>(1);
     auto inputQueryExpression = input->params[2];
     const auto k = input->getLiteralVal<int64_t>(3);
 
-    auto nodeTableEntry = storage::IndexUtils::bindTable(*context, tableName, indexName,
-        storage::IndexOperation::QUERY);
-    auto indexEntry = context->getCatalog()->getIndex(context->getTransaction(),
+    catalog::NodeTableCatalogEntry* nodeTableEntry = nullptr;
+    const graph::GraphEntry* graphEntry = nullptr;
+    if (context->getCatalog()->containsTable(context->getTransaction(), tableOrGraphName)) {
+        nodeTableEntry = storage::IndexUtils::bindNodeTable(*context, tableOrGraphName, indexName,
+            storage::IndexOperation::QUERY);
+    } else if (context->getGraphEntrySetUnsafe().hasGraph(tableOrGraphName)) {
+        graphEntry = &context->getGraphEntrySetUnsafe().getEntry(tableOrGraphName);
+        if (graphEntry->nodeInfos.size() > 1 || !graphEntry->relInfos.empty()) {
+            throw common::BinderException{common::stringFormat(
+                "Projected graph {} either contains relationship tables or "
+                "multiple node tables. Projected graphs passed to QUERY_HNSW_INDEX function must "
+                "contain only nodes from a single node table and no relationship tables.",
+                tableOrGraphName)};
+        }
+        nodeTableEntry = graphEntry->nodeInfos[0].entry->ptrCast<catalog::NodeTableCatalogEntry>();
+        storage::IndexUtils::validateIndexExistence(*context, nodeTableEntry, indexName,
+            storage::IndexOperation::QUERY);
+    } else {
+        throw common::BinderException{common::stringFormat(
+            "Cannot find table or projected graph named as {}.", tableOrGraphName)};
+    }
+    const auto indexEntry = context->getCatalog()->getIndex(context->getTransaction(),
         nodeTableEntry->getTableID(), indexName);
     const auto& auxInfo = indexEntry->getAuxInfo().cast<catalog::HNSWIndexAuxInfo>();
     KU_ASSERT(
@@ -76,7 +97,7 @@ static std::unique_ptr<TableFuncBindData> bindFunc(main::ClientContext* context,
     binder::expression_vector columns;
     columns.push_back(outputNode->getInternalID());
     columns.push_back(input->binder->createVariable("_distance", common::LogicalType::DOUBLE()));
-    auto boundInput = BoundQueryHNSWIndexInput{nodeTableEntry, indexEntry,
+    auto boundInput = BoundQueryHNSWIndexInput{nodeTableEntry, graphEntry, indexEntry,
         std::move(inputQueryExpression), static_cast<uint64_t>(k)};
     auto config = storage::QueryHNSWConfig{input->optionalParams};
     return std::make_unique<QueryHNSWIndexBindData>(context, std::move(columns), boundInput, config,
@@ -94,7 +115,7 @@ static std::vector<float> convertQueryVector(const common::Value& value) {
 }
 
 static std::vector<float> getQueryVector(main::ClientContext* context,
-    const std::shared_ptr<binder::Expression> queryExpression, uint64_t dimension) {
+    std::shared_ptr<binder::Expression> queryExpression, uint64_t dimension) {
     std::shared_ptr<binder::Expression> literalExpression = queryExpression;
     if (queryExpression->expressionType == common::ExpressionType::PARAMETER) {
         literalExpression = std::make_shared<binder::LiteralExpression>(
@@ -111,7 +132,7 @@ static std::vector<float> getQueryVector(main::ClientContext* context,
 }
 
 static const common::LogicalType& getIndexColumnType(const BoundQueryHNSWIndexInput& boundInput) {
-    auto columnName = boundInput.indexEntry->getPropertyIDs()[0];
+    const auto columnName = boundInput.indexEntry->getPropertyIDs()[0];
     return boundInput.nodeTableEntry->getProperty(columnName).getType();
 }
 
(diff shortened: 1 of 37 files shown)


src/include/graph/graph_entry.h
```diff
@@ -61,6 +61,10 @@ class GraphEntrySet {
         KU_ASSERT(hasGraph(name));
         return nameToEntry.at(name).copy();
     }
+    const GraphEntry& getEntryRef(const std::string& name) const {
```
andyfengHKU, line 64: Call this getEntry. We should remove getXXXRef from codebase.

src/include/function/table/table_function.h
```diff
@@ -57,6 +58,11 @@ struct KUZU_API TableFuncSharedState {
     common::offset_t maxOffset;
     common::offset_t curOffset;
     common::offset_t maxMorselSize;
+    // This for now is only used for QueryHNSWIndex.
+    // TODO(Guodong): This is not a good way to pass semiMasks to QueryHNSWIndex function.
+    // However, to avoid function specific logic when we handle semi mask in mapper, so we can move
+    // HNSW into an extension, we have to let semiMasks be owned by a base class.
+    common::table_id_map_t<std::unique_ptr<common::SemiMask>> semiMasks;
```
andyfengHKU, line 65: You don't need to change this but I'll reuse NodeOffsetMaskMap here for you.

src/function/table/hnsw/query_hnsw_index.cpp
```diff
@@ -207,26 +233,60 @@ std::unique_ptr<TableFuncLocalState> initQueryHNSWLocalState(const TableFunction
...
         }
         auto& node = callOp->getBindData()->getNodeOutput()->constCast<binder::NodeExpression>();
+        // Check if there is filter predicate on the base node table. If so, we need to plan the
+        // filtered search.
+        if (bindData->boundInput.graphEntry) {
+            auto& nodeInfo = bindData->boundInput.graphEntry->nodeInfos[0];
+            if (nodeInfo.predicate) {
+                // We have filter predicate on the base node table. Should add a semi mask subplan
+                // and pass the semi mask to QueryHNSWIndexFunction to perform filtered search.
+                std::vector<std::shared_ptr<planner::LogicalOperator>> nodeMaskRoots;
```
andyfengHKU, line 273: Apparently you don't need a hash join here.

src/storage/index/hnsw_index.cpp
```diff
@@ -339,37 +338,272 @@ common::offset_t OnDiskHNSWIndex::searchNNInUpperLayer(transaction::Transaction*
...
+    return neighbors;
+}
+
+common::offset_vec_t OnDiskHNSWIndex::oneHopFilteredSearch(common::SemiMask* semiMask, uint64_t ef,
+    graph::Graph::EdgeIterator& nbrItr, VisitedState& visited) {
+    common::offset_vec_t result;
+    result.reserve(ef);
+    for (const auto& neighborChunk : nbrItr) {
+        neighborChunk.forEach([&](auto neighbor, auto) {
+            if (!visited.contains(neighbor.offset)) {
```
semihsalihoglu-uw, line 492: We could simplify this to if (!visited.contains(neighbor.offset) && semiMask->isMasked(neighbor.offset)) { }.

src/storage/index/hnsw_index.cpp
```diff
@@ -339,37 +338,272 @@ common::offset_t OnDiskHNSWIndex::searchNNInUpperLayer(transaction::Transaction*
...
+                    }
                 }
             });
         }
     }
-    return popTopK(result, k);
+    return filteredNeighbors;
+}
+
+common::offset_vec_t OnDiskHNSWIndex::blindTwoHopFilteredSearch(common::SemiMask* semiMask,
```
semihsalihoglu-uw, line 565: I don't think this is the correct heuristic. What we want in blindTwoHop is this:
First iterate through each 1st degree neighbors and run dist only on the ones in mask (in the directed one you run the distance function on all 1st degree neighbors). 
Then, after all 1st degree neighbor traversal is done, iterate through each 2nd degree neighbors.

Also as above we should be doing this until M many distance calculations.

src/storage/index/hnsw_index.cpp
```diff
@@ -339,37 +338,272 @@ common::offset_t OnDiskHNSWIndex::searchNNInUpperLayer(transaction::Transaction*
...
+        if (numVisitedNbrs >= numNbrsInFirstHop) {
+            break;
+        }
+        if (visited.contains(candidate)) {
+            continue;
+        }
+        visited.add(candidate);
+        auto secondHopNbrItr =
+            lowerGraph->scanFwd(common::nodeID_t{candidate, nodeTableID}, secondHopScanState);
+        for (const auto& secondHopNbrChunk : secondHopNbrItr) {
```
semihsalihoglu-uw, line 550: You seem to be iterating over all 2nd degree neighbors here. Instead, we should stop early when we run distance computations on M many nodes (M is the maximum degree in the lower layer). That is we continue visiting neighbors (1st degree or 2nd degree) until M many unvisited nodes that are also in the mask are seen.

src/storage/index/hnsw_index.cpp
```diff
@@ -339,37 +338,272 @@ common::offset_t OnDiskHNSWIndex::searchNNInUpperLayer(transaction::Transaction*
...
+        secondHopScanState = lowerGraph->prepareRelLookup(lowerRelTableEntry);
+    }
+    while (!candidates.empty()) {
+        auto [candidate, candidateDist] = candidates.top();
+        // Break here if adding closestNode to result will exceed efs or not improve the results.
+        if (results.size() >= ef && candidateDist > results.top().distance) {
+            break;
+        }
+        candidates.pop();
+        auto neighbors = searchFilteredNeighbors(transaction, candidate, queryVector, searchType,
```
semihsalihoglu-uw, line 438: Ideally we would not have this copy here. Instead, ideally each heuristic implementation would be wrapped around an iterator interface (that wraps the iterator interface given by OnDiskGraph). And internally the getNextNode functions would implement the heuristic search. That would also avoid all duplicated code across the heuristics to manage scan states and run distance calculations etc. All that logic would be in this one place.

Is that possible to do?

I also see some code duplication in the unfiltered and filtered search, e.g., some of this code:
```
    min_node_priority_queue_t candidates;
    max_node_priority_queue_t result;
    searchState.visited.reset();
    const auto entryVector =
        embeddings->getEmbedding(transaction, *searchState.embeddingScanState.scanState, entryNode);
    processEntryNodeInKNNSearch(queryVector, entryVector, entryNode, searchState.visited,
        config.distFunc, embeddings.get(), candidates, result);
    const auto ef = std::max(searchState.k, static_cast<uint64_t>(searchState.config.efs));
    auto scanState = lowerGraph->prepareRelLookup(lowerRelTableEntry);
```
It would be good to think of these refactors now than later.

semihsalihoglu-uw (changes requested):
I have a few comments. I mainly want to see the "iterating over 1st and 2nd degree neighbors until M" logic implemented.

src/function/table/hnsw/query_hnsw_index.cpp
```diff
@@ -223,26 +249,54 @@ std::unique_ptr<TableFuncLocalState> initQueryHNSWLocalState(const TableFunction
...
-    return std::make_unique<QueryHNSWLocalState>(input.context.getTransaction(), mm,
-        *hnswSharedState->nodeTable, hnswBindData->indexColumnID, hnswSharedState->numNodes);
+    storage::HNSWSearchState searchState{input.context.getTransaction(), mm,
+        *hnswSharedState->nodeTable, hnswBindData->indexColumnID, hnswSharedState->numNodes};
+    searchState.k = hnswBindData->boundInput.k;
+    searchState.config = hnswBindData->config;
+    const auto tableID = hnswBindData->boundInput.nodeTableEntry->getTableID();
+    auto& semiMasks = hnswSharedState->semiMasks;
+    if (semiMasks.containsTableID(tableID)) {
+        semiMasks.pin(tableID);
```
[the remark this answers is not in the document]
ray6080: I agree with having a different term. but I'd like to ask @andyfengHKU to follow up the discussion here, as he is more familiar with the usage of the terms. (As I remember, GDS has similar interfaces).

src/include/storage/index/hnsw_index.h
```diff
@@ -167,17 +181,53 @@ class OnDiskHNSWIndex final : public HNSWIndex {
...
+        graph::NbrScanState& secondHopScanState) const;
+
+    static bool visitNodeInBlindSearch(common::offset_t offset, common::SemiMask* semiMask,
+        VisitedState& visited, common::offset_vec_t& result, uint64_t ef);
+
+    FilteredSearchType getFilteredSearchType(transaction::Transaction* transaction,
+        const HNSWSearchState& searchState) const;
 
 private:
+    static constexpr double BLIND_SEARCH_UP_SEL_THRESHOLD = 0.08;
```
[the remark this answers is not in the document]
ray6080: Not in this PR. I'd like to do it in a separate one.

src/function/table/hnsw/query_hnsw_index.cpp
```diff
@@ -56,14 +58,32 @@ static std::vector<common::LogicalType> inferInputTypes(const binder::expression
...
+    catalog::NodeTableCatalogEntry* nodeTableEntry = nullptr;
+    const graph::GraphEntry* graphEntry = nullptr;
+    if (context->getCatalog()->containsTable(context->getTransaction(), tableOrGraphName)) {
+        nodeTableEntry = storage::IndexUtils::bindNodeTable(*context, tableOrGraphName, indexName,
+            storage::IndexOperation::QUERY);
+    } else if (context->getGraphEntrySetUnsafe().hasGraph(tableOrGraphName)) {
+        graphEntry = &context->getGraphEntrySetUnsafe().getEntry(tableOrGraphName);
+        if (graphEntry->nodeInfos.size() > 1 || !graphEntry->relInfos.empty()) {
+            throw common::BinderException{common::stringFormat(
+                "Graph {} has multiple node tables or contain rel tables, please limit it to only "
```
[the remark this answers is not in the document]
ray6080: Done.

src/function/table/hnsw/query_hnsw_index.cpp
```diff
@@ -56,14 +58,32 @@ static std::vector<common::LogicalType> inferInputTypes(const binder::expression
...
+                "one node table without any rel tables.",
+                tableOrGraphName)};
+        }
+        nodeTableEntry = graphEntry->nodeInfos[0].entry->ptrCast<catalog::NodeTableCatalogEntry>();
+        storage::IndexUtils::validateIndexExistence(*context, nodeTableEntry, indexName,
+            storage::IndexOperation::QUERY);
+    } else {
+        throw common::BinderException{
+            common::stringFormat("Cannot find table or graph {}.", tableOrGraphName)};
```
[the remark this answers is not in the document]
ray6080: hmm I don't think adding a separate param or use list of params is a good idea. The syntax should be simple as users can input a table or a projected graph in the same way. So I don't think there is a good way we can figure out users' intention.

src/storage/index/hnsw_index.cpp
ray6080: Done.

src/storage/index/hnsw_index.cpp
ray6080: I've changed the implementation to first iterate all 1st degree neighbours, then start with 2nd ones. I realized I was following the implementation of acron's blind search.

src/storage/index/hnsw_index.cpp
ray6080: Changed it to stop early here.

src/storage/index/hnsw_index.cpp
ray6080: I reworked the function a bit to make the logic more clear. I don't find an easy way to handle everything inside the iterator interface, as there is a mixed logic of one-hop search and two-hop search. 

After the change, now unfiltered and one-hop filtered can share the same code for iterating 1-hop neighbour, as they're both one-hop search only. The only difference is checking semiMask or not, which can be hidden by `HNSWSearchState::isMasked`.
The directedTwoHop and blindTwoHop share the same code for iterating 2-hop neighbour node, but have their own code of iterating 1-hop neighbour, which is the part these two differ from each other.

src/storage/index/hnsw_index.cpp
```diff
@@ -340,39 +339,223 @@ common::offset_t OnDiskHNSWIndex::searchNNInUpperLayer(transaction::Transaction*
...
+                        searchState.visited.add(nbrOffset);
+                        results.push({nbrOffset, dist});
+                        candidates.push({nbrOffset, dist});
+                    }
                 }
-            });
+            }
+        });
+    }
+    while (!candidatesForSecHop.empty()) {
```
semihsalihoglu-uw, line 487: Can this be refactored with the end of blindTwoHop function?

semihsalihoglu-uw (approved):

PR #5026, opened 2025-03-11
Fix pin table error in page rank

excaliburwyj:
# Description

In a multi-label graph like SF100, when running page_rank, ther memory error occurs mainly because Degrees & pValues are not correctly pinned with the pinTable operation:

1. The calculation of degrees uses EmptyGDSAuxiliaryState, which does not have pinTable.
2. In the PageRankAuxiliaryState, pNext should use pinTable(fromTableID) because pNext updates with boundNodeID.
3. It is necessary to inherit from FrontierPair to implement PageRankFrontierPair, because the activesNode added to nextDenseFrontier is boundNodeID.

# Contributor agreement

### DIFF
--- a/src/function/gds/page_rank.cpp
+++ b/src/function/gds/page_rank.cpp
@@ -119,16 +119,31 @@ public:
     PageRankAuxiliaryState(Degrees& degrees, PValues& pCurrent, PValues& pNext)
         : degrees{degrees}, pCurrent{pCurrent}, pNext{pNext} {}
 
-    void beginFrontierCompute(table_id_t, table_id_t toTableID) override {
+    void beginFrontierCompute(table_id_t fromTableID, table_id_t toTableID) override {
         degrees.pinTable(toTableID);
         pCurrent.pinTable(toTableID);
-        pNext.pinTable(toTableID);
+        pNext.pinTable(fromTableID);
     }
 
 private:
     Degrees& degrees;
     PValues& pCurrent;
     PValues& pNext;
+};
+
+class KUZU_API PageRankFrontierPair : public FrontierPair {
+public:
+    PageRankFrontierPair(std::shared_ptr<PathLengths> curFrontier,
+        std::shared_ptr<PathLengths> nextFrontier)
+        : FrontierPair{curFrontier, nextFrontier} {}
+
+    void pageRankPinCurrFrontier(common::table_id_t tableID) {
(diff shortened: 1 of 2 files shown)


src/function/gds/page_rank.cpp
```diff
@@ -133,6 +133,37 @@ class PageRankAuxiliaryState : public GDSAuxiliaryState {
...
+    PageRankFrontierPair(std::shared_ptr<PathLengths> curFrontier,
+        std::shared_ptr<PathLengths> nextFrontier)
+        : FrontierPair{curFrontier, nextFrontier} {}
+
+    void pageRankPinCurrFrontier(common::table_id_t tableID) {
+        curSparseFrontier->pinTableID(tableID);
+        curDenseFrontier->pinCurFrontierTableID(tableID);
+        // In PNextUpdateEdgeCompute, addNodesToNextDenseFrontier is boundID
+        // should pin fromTableID in there.
+        nextDenseFrontier->pinNextFrontierTableID(tableID);
```
andyfengHKU, line 147: Hi @excaliburwyj, other changes looks fine to me. But I don't understand this line

If I'm extending from person -> organization. Next dense frontier you are pinning for `person` instead of `organization`. Why is this the case?

andyfengHKU (changes requested):

src/function/gds/page_rank.cpp
andyfengHKU, line 147: Actually you made a good point here.

Since the convergence of PageRank is purely decided by epsilon & maxIteration. I think we could avoid putting `boundNodeID` into `nextDenseFrontier`. Instead we always mark `currentFrontier` as all visited before a new round of edge compute starts. 

This solutions seems to be better than manually pin current tableID for next dense frontier.

excaliburwyj: Hi @andyfengHKU :
I get it. And i think has a more efficient approach. Since `currentFrontier` is set to all visited in every iteration, it would be better to initialize `currentFrontier` once outside the loop. Then, in the `PageRankFrontierPair`'s `beginNewIterationInternalNoLock` method, can remove `swap` and `incrementCurIter` . This way can keep using the same `currentFrontier`.  Is this approach suitable?

andyfengHKU: Yeah this this should work. Feel free to make the change. Also you might wanna rebase to master because #5048  has made a big change to the infrastructure.

PR #5033, opened 2025-03-11
GDS refactor

sdht0:
* Renames the functions to be shorter
* Uses the new encapsulated arrays.

# Contributor agreement

### DIFF
--- a/src/function/gds/k_core_decomposition.cpp
+++ b/src/function/gds/k_core_decomposition.cpp
@@ -16,53 +16,37 @@ using namespace kuzu::graph;
 namespace kuzu {
 namespace function {
 
-class CoreValues {
-public:
-    CoreValues(const table_id_map_t<offset_t>& maxOffsetMap, MemoryManager* mm) {
-        init(maxOffsetMap, mm);
-    }
+struct KcoreComputationState {
+    AtomicObjectArraysMap<degree_t> coreValuesMap;
 
-    void pinTable(table_id_t tableID) { coreValues = coreValuesMap.getData(tableID); }
-
-    bool isValid(offset_t offset) {
-        return coreValues[offset].load(std::memory_order_relaxed) != INVALID_DEGREE;
-    }
-    degree_t getValue(offset_t offset) {
-        return coreValues[offset].load(std::memory_order_relaxed);
-    }
-    void setCoreValue(offset_t offset, degree_t value) {
-        coreValues[offset].store(value, std::memory_order_relaxed);
-    }
-
-private:
-    void init(const table_id_map_t<offset_t>& maxOffsetMap, MemoryManager* mm) {
+    KcoreComputationState(const table_id_map_t<offset_t>& maxOffsetMap, MemoryManager* mm) {
         for (const auto& [tableID, maxOffset] : maxOffsetMap) {
-            coreValuesMap.allocate(tableID, maxOffset, mm);
-            pinTable(tableID);
+            coreValuesMap.allocateArray(tableID, maxOffset, mm);
+            coreValuesMap.pinTable(tableID);
             for (auto i = 0u; i < maxOffset; ++i) {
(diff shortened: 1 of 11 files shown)


src/function/gds/k_core_decomposition.cpp
```diff
@@ -16,53 +16,37 @@ using namespace kuzu::graph;
...
-    bool isValid(offset_t offset) {
-        return coreValues[offset].load(std::memory_order_relaxed) != INVALID_DEGREE;
-    }
-    degree_t getValue(offset_t offset) {
-        return coreValues[offset].load(std::memory_order_relaxed);
-    }
-    void setCoreValue(offset_t offset, degree_t value) {
-        coreValues[offset].store(value, std::memory_order_relaxed);
-    }
+struct KcoreComputationState {
```
andyfengHKU, line 19: Revert the naming to CoreValues. Otherwise I find KcoreComputationState.isValid to be very confusing.

src/function/gds/page_rank.cpp
```diff
@@ -88,63 +88,57 @@ static void addCAS(std::atomic<double>& origin, double valToAdd) {
 }
 
 // Represents PageRank value for all nodes
-class PValues {
-public:
-    PValues(table_id_map_t<offset_t> maxOffsetMap, storage::MemoryManager* mm, double val) {
+struct PrComputationState {
```
andyfengHKU, line 91: Ditto revert renaming.

src/include/function/gds/gds_function_collection.h
```diff
@@ -7,31 +7,31 @@ namespace kuzu {
 namespace function {
 
 struct WeaklyConnectedComponentsFunction {
-    static constexpr const char* name = "WEAKLY_CONNECTED_COMPONENT";
+    static constexpr const char* name = "GDS_WCC";
```
andyfengHKU, line 10: revert all naming changes in this PR. We should add alias naming instead of renaming.

src/include/function/gds/gds_object_manager.h
```diff
@@ -87,6 +90,15 @@ class AtomicObjectArray {
         }
         return false;
     }
+
+    void compare_and_swap(const common::offset_t pos, double valToAdd) {
```
andyfengHKU, line 94: This is not what the function is doing. It's an addition using CAS, so call it addCAS.

Also for `compare_exchange_strong_max` do a similar naming with camel case.

src/function/gds/strongly_connected_components_kosaraju.cpp
```diff
@@ -24,40 +24,34 @@ static constexpr offset_t PROCESSED = numeric_limits<offset_t>::max();
 static constexpr offset_t VISITED = numeric_limits<offset_t>::max() - 1;
 static constexpr offset_t NOT_VISITED = numeric_limits<offset_t>::max() - 2;
 
-class SCCState {
-public:
-    SCCState(const offset_t tableID, const offset_t numNodes, MemoryManager* mm) {
-        componentIDsMap.allocate(tableID, numNodes, mm);
-        componentIDs = componentIDsMap.getData(tableID);
+struct SccComputationState {
```
andyfengHKU, line 27: I would also call this SCCIDs

src/function/gds/weakly_connected_components.cpp
```diff
@@ -15,83 +15,68 @@ using namespace kuzu::graph;
 namespace kuzu {
 namespace function {
 
-class ComponentIDs {
-public:
-    ComponentIDs(const table_id_map_t<offset_t>& maxOffsetMap, MemoryManager* mm) {
+struct WccComputationState {
```
andyfengHKU, line 18: Ditto revert renaming and I would merge this with SCCIDs. Also why we have Wcc, WCC, Scc, SCC at this same time? Let's use one convention and if we decide to change we change in one PR all together.

andyfengHKU (changes requested):

PR #5037, opened 2025-03-12
Expose semi mask sub-plan in the logical plan tree

ray6080:
# Description

Introduced a new logical operator: `LogicalDummySink`, which is used, in this PR, to construct a sub plan for semi masker.
With the help of `LogicalDummySink`, this PR moves the semi mask sub plan into the logical plan tree, which makes the semi mask sub plan be visible to optimizer (such as FilterPushDown).

For now, I'm keeping the semi mask sub plan as child of `LogicalGDSCall` and `LogicalRecursiveExtend`, which may not be the best idea. Open to more refactoring around this.

### DIFF
--- a/src/include/planner/operator/extend/logical_recursive_extend.h
+++ b/src/include/planner/operator/extend/logical_recursive_extend.h
@@ -6,16 +6,16 @@
 namespace kuzu {
 namespace planner {
 
-class LogicalRecursiveExtend : public LogicalOperator {
+class LogicalRecursiveExtend final : public LogicalOperator {
     static constexpr LogicalOperatorType operatorType_ = LogicalOperatorType::RECURSIVE_EXTEND;
 
 public:
     LogicalRecursiveExtend(std::unique_ptr<function::RJAlgorithm> function,
         function::RJBindData bindData, binder::expression_vector resultColumns,
-        common::table_id_set_t nbrTableIDSet, std::shared_ptr<LogicalOperator> nodeMaskRoot)
-        : LogicalOperator{operatorType_}, function{std::move(function)}, bindData{bindData},
-          resultColumns{std::move(resultColumns)}, nbrTableIDSet{std::move(nbrTableIDSet)},
-          limitNum{common::INVALID_LIMIT}, nodeMaskRoot{std::move(nodeMaskRoot)} {}
+        common::table_id_set_t nbrTableIDSet)
+        : LogicalOperator{operatorType_}, function{std::move(function)},
+          bindData{std::move(bindData)}, resultColumns{std::move(resultColumns)},
+          nbrTableIDSet{std::move(nbrTableIDSet)}, limitNum{common::INVALID_LIMIT} {}
 
     void computeFlatSchema() override;
     void computeFactorizedSchema() override;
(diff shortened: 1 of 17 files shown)


src/planner/operator/logical_dummy_sink.cpp
```diff
@@ -0,0 +1,17 @@
+#include "planner/operator/logical_dummy_sink.h"
+
+namespace kuzu {
+namespace planner {
+
+void LogicalDummySink::computeFactorizedSchema() {
+    createEmptySchema();
+    schema = children[0]->getSchema()->copy();
```
andyfengHKU, line 8: use copyChildSchema(0)

src/include/planner/operator/logical_dummy_sink.h
```diff
@@ -0,0 +1,27 @@
...
+public:
+    explicit LogicalDummySink(std::shared_ptr<LogicalOperator> child)
+        : LogicalOperator{type_, {std::move(child)}} {}
+    explicit LogicalDummySink(logical_op_vector_t children)
+        : LogicalOperator{type_, {std::move(children)}} {}
+
+    void computeFactorizedSchema() override;
+    void computeFlatSchema() override;
+
+    std::string getExpressionsForPrinting() const override { return "DUMMY_SINK"; }
```
andyfengHKU, line 20: this should be empty

src/planner/operator/logical_dummy_sink.cpp
```diff
@@ -0,0 +1,17 @@
...
+}
+
+void LogicalDummySink::computeFlatSchema() {
+    createEmptySchema();
+    schema = children[0]->getSchema()->copy();
```
andyfengHKU, line 13: ditto

src/include/planner/operator/logical_gds_call.h
```diff
@@ -10,28 +10,38 @@ class LogicalGDSCall final : public LogicalOperator {
...
-    explicit LogicalGDSCall(binder::BoundGDSCallInfo info,
-        std::vector<std::shared_ptr<LogicalOperator>> nodeMaskRoots)
-        : LogicalOperator{operatorType_}, info{std::move(info)},
-          nodeMaskRoots{std::move(nodeMaskRoots)} {}
+    explicit LogicalGDSCall(binder::BoundGDSCallInfo info)
+        : LogicalOperator{operatorType_}, info{std::move(info)} {}
+
+    // Note that this is a special interface for adding node mask roots as children.
+    // We shouldn't have other cases where this operator has any children.
+    void addDummySinkAsMaskChild(std::shared_ptr<LogicalOperator> child) {
```
andyfengHKU, line 18: Why not use addChild in the base LogicalOperator?

src/include/planner/operator/extend/logical_recursive_extend.h
```diff
@@ -6,20 +6,26 @@
...
+        : LogicalOperator{operatorType_}, function{std::move(function)},
+          bindData{std::move(bindData)}, resultColumns{std::move(resultColumns)},
+          nbrTableIDSet{std::move(nbrTableIDSet)}, limitNum{common::INVALID_LIMIT} {}
 
     void computeFlatSchema() override;
     void computeFactorizedSchema() override;
 
+    // Note that this is a special interface for adding node mask roots as children.
+    // We shouldn't have other cases where this operator has any children.
+    void addDummySinkAsMaskChild(std::shared_ptr<LogicalOperator> child) {
```
andyfengHKU, line 25: ditto

src/processor/map/map_gds.cpp
```diff
@@ -14,6 +15,19 @@ using namespace kuzu::storage;
 namespace kuzu {
 namespace processor {
 
+LogicalSemiMasker* PlanMapper::findSemiMaskerInPlan(LogicalOperator* logicalOperator) {
```
andyfengHKU, line 18: I think we can check the child of LogicalDummySink must be LogicalSemiMasker instead of doing a full top-down traversal?

andyfengHKU (approved):

src/include/planner/operator/logical_gds_call.h
ray6080: hmm I don't think we have that in the base LogicalOperator? Should I add that? I think it's better not to propagate this interface to the base as the usage is very specific to few operators now?

PR #5047, opened 2025-03-12
Vector extension

ray6080:
# Description

Move HNSW related functionality into the `VECTOR` extension.


src/include/processor/execution_context.h
```diff
@@ -1,5 +1,6 @@
 #pragma once
 
+// TODO(Guodong): Change these includes into forward declaration.
```
ray6080, line 3: Separate this in another PR.

acquamarin (approved):
can you also add vss to the list of official_extensions?
```
   static constexpr const char* OFFICIAL_EXTENSION[] = {"HTTPFS", "POSTGRES", "DUCKDB", "JSON",
        "SQLITE", "FTS", "DELTA", "ICEBERG", "UNITY_CATALOG"};
        ```
        So `show_official_extensions` can correctly show vss as an official extension

ray6080:
> can you also add vss to the list of official_extensions?
> 
> ```
>    static constexpr const char* OFFICIAL_EXTENSION[] = {"HTTPFS", "POSTGRES", "DUCKDB", "JSON",
>         "SQLITE", "FTS", "DELTA", "ICEBERG", "UNITY_CATALOG"};
>         ```
>         So `show_official_extensions` can correctly show vss as an official extension
> ```

ah missed it. Will add it in the following PR.

PR #5055, opened 2025-03-15
Remove unused LogicalScanNodeTable's properties

ted-wq-x:
# Description

* Remove unused scanNodeTable properties,while filter be push into scanNodeTable
* Remove unnecessary projection upon scanNodeTable

# Contributor agreement

### DIFF
--- a/src/optimizer/projection_push_down_optimizer.cpp
+++ b/src/optimizer/projection_push_down_optimizer.cpp
@@ -20,6 +20,7 @@
 #include "planner/operator/persistent/logical_insert.h"
 #include "planner/operator/persistent/logical_merge.h"
 #include "planner/operator/persistent/logical_set.h"
+#include <planner/operator/scan/logical_scan_node_table.h>
 
 using namespace kuzu::common;
 using namespace kuzu::planner;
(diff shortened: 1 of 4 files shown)


src/optimizer/projection_push_down_optimizer.cpp
```diff
@@ -20,6 +20,7 @@
 #include "planner/operator/persistent/logical_insert.h"
 #include "planner/operator/persistent/logical_merge.h"
 #include "planner/operator/persistent/logical_set.h"
+#include <planner/operator/scan/logical_scan_node_table.h>
```
andyfengHKU, line 23: change `<>` to `""` for consistency

src/optimizer/projection_push_down_optimizer.cpp
```diff
@@ -361,8 +384,14 @@ void ProjectionPushDownOptimizer::preAppendProjection(LogicalOperator* op, idx_t
         // We don't have a way to handle
         return;
     }
-    auto projection =
-        std::make_shared<LogicalProjection>(std::move(expressions), op->getChild(childIdx));
+    auto child = op->getChild(childIdx);
+    // reset scan node table properties
+    if (child->getOperatorType() == LogicalOperatorType::SCAN_NODE_TABLE) {
```
andyfengHKU, line 389: I would add a new optimizer that remove any unnecessary projection instead of coupling the logic inside projection push down optimizer.

src/optimizer/projection_push_down_optimizer.cpp
```diff
@@ -266,6 +267,28 @@ void ProjectionPushDownOptimizer::visitTableFunctionCall(LogicalOperator* op) {
...
+void ProjectionPushDownOptimizer::visitScanNodeTable(LogicalOperator* op) {
+    clearScanNodeTableProperties(op);
+}
+
+bool ProjectionPushDownOptimizer::clearScanNodeTableProperties(planner::LogicalOperator* op) const {
+    auto& scanNodeTable = op->cast<LogicalScanNodeTable>();
+    if (scanNodeTable.getProperties().empty()) {
+        return false;
+    }
+    bool hasPropertiesInUse = false;
```
andyfengHKU, line 279: First I wanna understand on which query this optimization will take effect so that we can add test & benchmark.

Second, I think we can do better than `clearProperty`. Scan operator can scan a subset of its current properties.

src/processor/operator/scan/primary_key_scan_node_table.cpp
```diff
@@ -36,8 +37,18 @@ void PrimaryKeyScanNodeTable::initLocalStateInternal(ResultSet* resultSet,
     for (auto& pos : info.outVectorsPos) {
         outVectors.push_back(resultSet->getValueVector(pos).get());
     }
-    scanState = std::make_unique<NodeTableScanState>(
-        resultSet->getValueVector(info.nodeIDPos).get(), outVectors, outVectors[0]->state);
+    if (outVectors.empty()) {
```
andyfengHKU, line 40: @ray6080 should review this.

PR #5074, opened 2025-03-19
Track hash index memory usage

benjaminwinger:
TODO: Missing tracking for the overflow file used for strings.

`CopyTest.NodeInsertBMExceptionDuringCheckpointRecovery` is failing for an unknown reason (it appears to be throwing a BufferManagerException despite having a lot of space left).


benjaminwinger:
Increasing the page size (as some of the tests do) also increases the minimum size of the hash index, which evidently becomes too large to fit in the default buffer pool size. We might need to make the default buffer pool size take into account the page size.

PR #5077, opened 2025-03-19
Move OutputNodeMask output GDSComputeState

andyfengHKU:
# Description

See title. OutputNodeMask by definition is not a state and thus should not be stored in GDSComputeState. 

I'm currently differentiating interfaces between GDS and RecursiveJoin because GDS does not have an outputNodeMask. We can unify them alternatively. I don't have a strong opinion at this point.

### DIFF
--- a/src/include/function/gds/gds_state.h
+++ b/src/include/function/gds/gds_state.h
@@ -7,18 +7,14 @@ namespace kuzu {
 namespace function {
 
 struct GDSComputeState {
-    std::unique_ptr<function::FrontierPair> frontierPair = nullptr;
-    std::unique_ptr<function::EdgeCompute> edgeCompute = nullptr;
-    std::unique_ptr<function::GDSAuxiliaryState> auxiliaryState = nullptr;
+    std::unique_ptr<FrontierPair> frontierPair = nullptr;
+    std::unique_ptr<EdgeCompute> edgeCompute = nullptr;
+    std::unique_ptr<GDSAuxiliaryState> auxiliaryState = nullptr;
 
-    common::NodeOffsetMaskMap* outputNodeMask = nullptr;
-
-    GDSComputeState(std::unique_ptr<function::FrontierPair> frontierPair,
-        std::unique_ptr<function::EdgeCompute> edgeCompute,
-        std::unique_ptr<function::GDSAuxiliaryState> auxiliaryState,
-        common::NodeOffsetMaskMap* outputNodeMask)
+    GDSComputeState(std::unique_ptr<FrontierPair> frontierPair,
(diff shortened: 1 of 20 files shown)


src/include/function/gds/gds_utils.h
```diff
@@ -9,9 +9,12 @@ namespace function {
 
 class KUZU_API GDSUtils {
 public:
-    static void runFrontiersUntilConvergence(processor::ExecutionContext* context,
-        GDSComputeState& rjCompState, graph::Graph* graph, common::ExtendDirection extendDirection,
-        uint64_t maxIteration, const std::string& propertyToScan = "");
+    static void runGDS(processor::ExecutionContext* context, GDSComputeState& compState,
+        graph::Graph* graph, common::ExtendDirection extendDirection, uint64_t maxIteration);
+    static void runRecursiveJoin(processor::ExecutionContext* context, GDSComputeState& compState,
```
ray6080, line 14: It looks a bit weird that we have a separate `runRecursiveJoin` under `GDSUtils`. Do you want to separate this into another Utils class?

src/include/function/gds/gds_state.h
```diff
@@ -7,18 +7,16 @@ namespace kuzu {
...
 struct GDSComputeState {
-    std::unique_ptr<function::FrontierPair> frontierPair = nullptr;
-    std::unique_ptr<function::EdgeCompute> edgeCompute = nullptr;
-    std::unique_ptr<function::GDSAuxiliaryState> auxiliaryState = nullptr;
+    std::unique_ptr<FrontierPair> frontierPair = nullptr;
+    std::unique_ptr<EdgeCompute> edgeCompute = nullptr;
+    std::unique_ptr<GDSAuxiliaryState> auxiliaryState = nullptr;
 
-    common::NodeOffsetMaskMap* outputNodeMask = nullptr;
+    // common::NodeOffsetMaskMap* outputNodeMask = nullptr;
```
ray6080, line 14: Remove this line.

src/include/function/gds/gds_utils.h
```diff
@@ -9,9 +9,12 @@ namespace function {
 
 class KUZU_API GDSUtils {
 public:
-    static void runFrontiersUntilConvergence(processor::ExecutionContext* context,
-        GDSComputeState& rjCompState, graph::Graph* graph, common::ExtendDirection extendDirection,
-        uint64_t maxIteration, const std::string& propertyToScan = "");
+    static void runGDS(processor::ExecutionContext* context, GDSComputeState& compState,
```
ray6080, line 12: I think `runFrontiersUntilConvergence` was more accurate than `runGDS`. This function should be part of the execution when we run a GDS algorithm, no?

ray6080 (approved):

Issue #5016, opened 2025-03-08
Feature: Graceful stopping and reloading if connection to database is stopped midquery


DDoyle1066:
### API

Python

### Description

I am not sure if this comes under the heading of feature or bug. But one thing I have noticed (especially when running large queries) is that sometimes if I have written and tried to execute a query that is taking a long time and then I terminate the process, I am unable to connect back to the database and the database just hangs. I don't know what the cause of this is because sometimes when I terminate the process I am able to reconnect just fine; it is not clear how to fix the database to go back to a version that I can query from and I have to reload everything from scratch which is often time consuming.

ray6080:
> if I have written and tried to execute a query that is taking a long time and then I terminate the process, I am unable to connect back to the database and the database just hangs.

hi @DDoyle1066 Are you experiencing this in Python API? "terminate the process, I am unable to connect back to the database and the database just hangs" Do you mean the creation of Kuzu Database object hangs for you if you restart your process to connect to the same db directory?

DDoyle1066:
I have experienced it in both the python API and when running a query in kuzu explorer. The most recent case was that detaching and deleting nodes was taking a lot longer than I anticipated, I terminated the process to figure out how to do it faster and when I tried reconnecting with a fresh process running kuzu.Database("path/to/db") it just hung for a while.
If terminating the process will generally occassionally create a situation where you are unable to reconnect to the database (if it was in the middle of modifying data) then it would be nice to be able to connect to the database from a different process and run a query that halts all queries being run against the database in an orderly fashion.

ray6080:
> I have experienced it in both the python API and when running a query in kuzu explorer. The most recent case was that detaching and deleting nodes was taking a lot longer than I anticipated, I terminated the process to figure out how to do it faster and when I tried reconnecting with a fresh process running kuzu.Database("path/to/db") it just hung for a while. If terminating the process will generally occassionally create a situation where you are unable to reconnect to the database (if it was in the middle of modifying data) then it would be nice to be able to connect to the database from a different process and run a query that halts all queries being run against the database in an orderly fashion.

Yes, I agree. We should be able to stop gracefully and restart. This looks like a bug to me rather than a feature. We should look into this and get it fixed. Are you able to ship a reproducible example to us for debugging? That would help a lot!

DDoyle1066:
[...]

PR #5111, opened 2025-03-24
Allow WAL replaying on some corrupted WAL files (e.g. after DB was killed)

royi-luo:
# Description

After a kuzu process is killed, the WAL may be in an incomplete state. This PR stops an error from being reported if we are replaying an incomplete WAL. Instead the replaying logic will be:
```
for each WAL record:
  record = WAL::deserializeRecord() // if this throws stop replaying
  WAL::replayRecord(record) // if this throws we continue replaying as the next record should be a rollback
```

Fixes https://github.com/kuzudb/kuzu/issues/5016

# Contributor agreement

### DIFF
--- a/src/storage/wal_replayer.cpp
+++ b/src/storage/wal_replayer.cpp
@@ -45,9 +45,20 @@ void WALReplayer::replay() const {
     }
     try {
         Deserializer deserializer(std::make_unique<BufferedFileReader>(std::move(fileInfo)));
+        RUNTIME_CHECK(bool nextRecordShouldBeRollback = false);
         while (!deserializer.finished()) {
+            // If an exception occurs while deserializing we will stop replaying
             auto walRecord = WALRecord::deserialize(deserializer, clientContext);
-            replayWALRecord(*walRecord);
+            KU_ASSERT(
+                !nextRecordShouldBeRollback || walRecord->type == WALRecordType::ROLLBACK_RECORD);
+            try {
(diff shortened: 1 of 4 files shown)


src/storage/wal_replayer.cpp
```diff
@@ -62,8 +62,6 @@ void WALReplayer::replay() const {
             // under this case.
             clientContext.getTransactionContext()->rollback();
```
royi-luo, line 63: Should we also clear the WAL in this case?

ray6080: No. we shouldn't.

tools/python_api/test/test_wal.py
```diff
@@ -0,0 +1,72 @@
...
+
+def run_query_then_kill(tmp_path: Path, build_dir: Path, queries: str):
+    proc = run_query_in_new_process(tmp_path, build_dir, queries)
+    time.sleep(5)
+    proc.kill()
+    proc.wait(5)
+
+
+# Kill the database while it's in the middle of executing a long persistent query
+# When we reload the database it should
```
ray6080, line 34: Finish the sentence?

ray6080 (approved):
Thanks!

tools/python_api/test/test_wal.py
royi-luo: updated

PR #5113, opened 2025-03-25
Fix bug in reloadDB test command

andyfengHKU:
# Description

In `reloadDB` tests, we always reset database before reseting connection. This may lead to unexpected behavior because connection will hold an invalid database ptr.

@ray6080 can u double check if the test in this PR fails on master expected? If so, we can checkin this PR otherwise we should fix in a different way.


ray6080 (commented):
Looks fine to me.

ray6080 (approved):

PR #5138, opened 2025-03-27
Fix rust extensions in testing and document how other binaries can use them

benjaminwinger:
Draft at the moment since I suspect this won't work on all platforms.

It might be a good idea to also add features for bundling the different extensions, as that will probably be more reliable (though the build process may also be complicated, e.g. for the ones which require duckdb (presumably we can make use of the libduckdb-sys crate)).

### DIFF
--- a/tools/rust_api/src/connection.rs
+++ b/tools/rust_api/src/connection.rs
@@ -360,4 +360,30 @@ Invalid input <MATCH (a:Person RETURN>: expected rule oC_SingleQuery (line: 1, o
         temp_dir.close()?;
         Ok(())
     }
+
+    macro_rules! extension_tests {
+        ($($name:ident,)*) => {
+        $(
+            #[test]
+            #[cfg(feature = "extension_tests")]
+            fn $name() -> Result<()> {
+                let temp_dir = tempfile::tempdir()?;
+                let db = Database::new(temp_dir.path(), SystemConfig::default())?;
+                let conn = Connection::new(&db)?;
+                let directory: String = if cfg!(windows) {
+                    std::env::var("KUZU_LOCAL_EXTENSIONS")?.replace("\\", "/")
+                } else {
+                    std::env::var("KUZU_LOCAL_EXTENSIONS")?
+                };
+                let name = stringify!($name);
+                conn.query(&format!("LOAD EXTENSION '{directory}/{name}/build/lib{name}.kuzu_extension'"))?;
+                Ok(())
+            }
+        )*
+        }
+    }
+
+    extension_tests! {
+        fts, duckdb, httpfs, postgres, sqlite, json, delta, iceberg, vector,
+    }
 }
--- a/tools/rust_api/src/lib.rs
+++ b/tools/rust_api/src/lib.rs
@@ -30,6 +30,21 @@
 //! - `KUZU_SHARED`: If set, link dynamically instead of statically
 //! - `KUZU_INCLUDE_DIR`: Directory of kuzu's headers
 //! - `KUZU_LIBRARY_DIR`: Directory containing kuzu's pre-built libraries.
+//!
+//! ## Using Extensions
+//! By default, binaries created using this library will not work with kuzu's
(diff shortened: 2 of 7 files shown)


tools/rust_api/src/lib.rs
```diff
@@ -30,6 +30,16 @@
...
 //! - `KUZU_LIBRARY_DIR`: Directory containing kuzu's pre-built libraries.
+//!
+//! ## Using Extensions
+//!
+//! If you want to use [extensions](https://docs.kuzudb.com/extensions/) in binaries using this
+//! library, you will need to add the following to your build.rs (or create one) so that the binary
+//! produced acts like a library that the extension can link with. Not doing this will produce
+//! undefined symbol errors when the extension is loaded:
+//! ```ignore
+//! println!("cargo:rustc-link-arg=-rdynamic");
```
sdht0, line 41: Does this solve the problem that using `export KUZU_SHARED=1` causes, namely that `cargo install` is not seamless?

benjaminwinger: Yes, this will let the extensions work as normal and there wouldn't be anything a user would have to do. The developers would just have to add that line to their build script.

tools/rust_api/src/connection.rs
```diff
@@ -360,4 +360,29 @@ Invalid input <MATCH (a:Person RETURN>: expected rule oC_SingleQuery (line: 1, o
...
+            fn $name() -> Result<()> {
+                let db = Database::in_memory(SystemConfig::default())?;
+                let conn = Connection::new(&db)?;
+                let directory = if cfg!(windows) {
+                    std::env::var("KUZU_LOCAL_EXTENSIONS")?.replace("\\", "/");
+                } else {
+                    std::env::var("KUZU_LOCAL_EXTENSIONS")?;
+                };
+                let name = stringify!($name);
+                conn.query(&format!("LOAD EXTENSION '{directory}/{name}/build/lib{name}.kuzu_extension'"))?;
```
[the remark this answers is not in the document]
benjaminwinger: Unhelpful; kuzu doesn't parse paths with backslashes so I don't believe it would make any difference

sdht0:
Hi @benjaminwinger. I tested this and doesn't seem to be working for me yet
```bash
$ git fetch origin pull/5138/head:test
$ git checkout test
$ git diff                                       
diff --git i/examples/rust/src/main.rs w/examples/rust/src/main.rs
index a7db8fc1a..5cbc6ffa9 100644
--- i/examples/rust/src/main.rs
+++ w/examples/rust/src/main.rs
@@ -9,6 +9,8 @@ fn main() -> Result<(), Error> {
     )?;
     let connection = Connection::new(&db)?;
 
+    connection.query("INSTALL httpfs;")?;
+    connection.query("LOAD httpfs;")?;
     // Create schema.
     connection.query("CREATE NODE TABLE Person(name STRING, age INT64, PRIMARY KEY(name));")?;
     // Create nodes.
$ cargo run /tmp/db
   Compiling kuzu v0.8.0 <>/kuzu-master/tools/rust_api)
   Compiling kuzu-rust-example v0.1.0 (<>/kuzu-master/examples/rust)
Error: Query execution failed: IO exception: Failed to load library: <>/.kuzu/extension/0.8.0/linux_arm64/httpfs/libhttpfs.kuzu_extension which is needed by extension: httpfs.
Error: <>/.kuzu/extension/0.8.0/linux_arm64/httpfs/libhttpfs.kuzu_extension: undefined symbol: _ZTIN4kuzu6common10FileSystemE.
```
Am I holding this wrong?

Compiling on Aarch64 linux. Will try it on x86_64 as well, although it shouldn't make a difference.

sdht0 (approved):
Alright, builds with
```
diff --git c/examples/rust/build.rs i/examples/rust/build.rs
new file mode 100644
index 000000000..538285b19
--- /dev/null
+++ i/examples/rust/build.rs
@@ -0,0 +1,3 @@
+fn main() {
+    println!("cargo:rustc-link-arg=-rdynamic");
+}
```

@benjaminwinger can you update the example as well?

Thanks for looking into this!

PR #5169, opened 2025-03-31
Fix click benchmark query 29

benjaminwinger:
There were two issues here, one was that I mistyped the regex, leading to it aggregating on 19 million groups instead of ~75 (I think it was failing to match anything, so it produced the original URL instead of replacing it with the domain name), and the second being that the min function (as well as max and presumably other similar aggregate functions) allocates through `MemoryManager::allocateBuffer` which was always increasing the buffermanager's used memory by at least 256KB (19M * 256KB = ~4.6TB).

It might be worthwhile to look at the comparative performance and introduce other size classes (at the moment it would be easy to add 4KB) to the BM-backed allocations supported by the MemoryManager.
And while are other limits on how `BufferManager::pin` works (such as the size of the eviction queue being computed based on the assumption that we cannot ever fill it if all allocations have the smallest page size class (4KB)), it might be possible to have the buffer manager record the actual amount of memory used within the page's region in the PageState so that we can avoid this issue with most sizes that go through the memory manager (to avoid exhausting space in the eviction queue we could use malloc for sizes below 4KB and above 256KB and use this for sizes in-between).

For some reason though, on our 128 thread 512GB machine the query was hanging instead of throwing a BufferManagerException when it ran out of space. It seems like it's just getting stuck just below the buffer manager memory limit and spending a lot of time trying to evict pages to make space for new allocations, and it might just be that it's taking a disproportionately long time to evict the few actually evictable pages in the queue (after which it will throw a BufferManagerException), while the rest are unevictable pages pinned by the MemoryManager (and maybe it's also adding new evictable pages at the same time). I'm not really sure what can be done to improve responsiveness in this case, as I think the issue might be made worse by many threads trying to evict simultaneously and the eviction queue being particularly large due to the default buffer pool size of 410GB, so it might just be an issue with the BufferManager not being very efficient at this scale (using malloc instead of the BM should help with that though, as pages from the MemoryManager that are pinned for long periods of time are a big part of the problem).

For the moment I've just made allocateBuffer use malloc internally for all sizes except 256KB, and merged mallocBuffer with allocateBuffer since there is no reason to distinguish between them any more.


ray6080 (approved):

ray6080:
Let's wait a bit to merge this after the release. Also there seems to be some regression on our benchmark queries, not sure if it's from your changes or not, but can you double check locally?

benjaminwinger:
I can confirm that these changes are causing the regression.

benjaminwinger:
Closing for the moment. The regression was caused by InMemOverflowBuffer relying on allocateBuffer producing 256KB buffers (it would request based on the size of the string being added to the buffer, but is very inefficient if it always allocates one buffer per string). We could change this so that it always requires at least 256KB, but then it doesn't actually change the minimum memory requirements for aggregate functions using it to store strings like `MIN`/`MAX`. Using a 4KB minimum is a decent compromise, but still increased runtime by about 25% in the worst of these benchmarks (var_size_seq_scan/q20). I suspect that introducing a 4KB MemoryManager size class that's allocated through the buffer manager might help, but it would need further work and testing.

PR #5185, opened 2025-04-02
rust: bump min rust version to 1.74

sdht0:
Personal preference is to enable all clippy lints and then selectively disable the unwanted ones. Happy to consider disabling at least the `pedantic` ones by default. 

# Contributor agreement

### DIFF
--- a/tools/rust_api/Cargo.toml
+++ b/tools/rust_api/Cargo.toml
@@ -2,6 +2,7 @@
 name = "kuzu"
 version = "0.9.0"
 description = "An in-process property graph database management system built for query speed and scalability"
+keywords = ["database", "graph", "ffi"]
 # Note: 1.72 required for testing due to latest dependencies of the arrow feature
 rust-version = "1.67"
 readme = "kuzu-src/README.md"
@@ -55,3 +56,34 @@ all-features = true
 [profile.relwithdebinfo]
 inherits = "release"
 debug = true
+
+[lints.clippy]
+correctness = { level = "warn", priority = -1 }
+suspicious = { level = "warn", priority = -1 }
+perf = { level = "warn", priority = -1 }
+cargo = { level = "warn", priority = -1 }
+style = { level = "warn", priority = -1 }
+
+complexity = { level = "warn", priority = -1 }
+too_many_arguments = "allow"
+type_complexity = "allow"
+
+pedantic = { level = "warn", priority = -1 }
+cast_lossless = "allow"
+cast_possible_truncation = "allow"
+cast_possible_wrap = "allow"
+cast_precision_loss = "allow"
+cast_sign_loss = "allow"
+doc_markdown = "allow"
+inline_always = "allow"
+missing_panics_doc = "allow"
+module_name_repetitions = "allow"
+must_use_candidate = "allow"
+needless_pass_by_value = "allow"
+redundant_closure_for_method_calls = "allow"
+return_self_not_must_use = "allow"
+similar_names = "allow"
+struct_excessive_bools = "allow"
+too_many_lines = "allow"
+missing_errors_doc = "allow"
+unreadable_literal = "allow"
--- a/tools/rust_api/Cargo.toml
+++ b/tools/rust_api/Cargo.toml
@@ -24,7 +24,7 @@ include = [
 
 # Note: 1.72 required for testing due to latest dependencies of the arrow feature
 rust-version = "1.67"
-edition = "2018"
+edition = "2021"
 
 [dependencies]
 cxx = "1.0"
--- a/tools/rust_api/Cargo.toml
+++ b/tools/rust_api/Cargo.toml
@@ -3,10 +3,8 @@ name = "kuzu"
 version = "0.9.0"
 description = "An in-process property graph database management system built for query speed and scalability"
 keywords = ["database", "graph", "ffi"]
-# Note: 1.72 required for testing due to latest dependencies of the arrow feature
-rust-version = "1.67"
 readme = "kuzu-src/README.md"
-homepage = "http://kuzudb.com/"
+homepage = "https://kuzudb.com/"
 repository = "https://github.com/kuzudb/kuzu"
 license = "MIT"
 categories = ["database"]
@@ -15,8 +13,8 @@ include = [
     "build.rs",
     "/src",
     "/include",
-# If more files are added to kuzu-src, they should also be added to the rerun-if-changed
-# instructions in build.rs so that rebuilds are detected properly
+    # If more files are added to kuzu-src, they should also be added to the rerun-if-changed
+    # instructions in build.rs so that rebuilds are detected properly
     "/kuzu-src/src",
     "/kuzu-src/cmake",
     "/kuzu-src/third_party",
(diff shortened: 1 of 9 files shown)


examples/rust/Cargo.toml
```diff
@@ -4,8 +4,8 @@ version = "0.1.0"
 edition = "2021"
 
 [dependencies]
-kuzu = {path="../../tools/rust_api"}
-arrow = {version="53", optional=true, default-features=false, features=[]}
+kuzu = { path = "../../tools/rust_api" }
+arrow = { version = "54", optional = true, default-features = false, features = [] }
```
sdht0, line 8: Does enabling the arrow feature mean the internal communication between the rust code and kuzu uses the arrow layout?

benjaminwinger: No. It just adds support for exporting query results as arrow. arrow-rs is a large dependency, so I thought it made sense for it to be optional.

sdht0: Yes of course. I was trying to understand why it's in the example's deps as the code doesn't use it?

benjaminwinger: According to git blame: https://github.com/kuzudb/kuzu/commit/9dedddc1eaa0a66f076414cb701784b88b8256ca. Basically just because the build cache is less efficient without it, or at least, it was at the time. For a while we were using a workspace so that the api tests and the example builds don't duplicate work; I think that might have been removed since I don't see evidence of it (I had been trying to avoid polluting the root of the repo so it was a little hacky).

tools/rust_api/Cargo.toml
```diff
@@ -55,3 +55,34 @@ all-features = true
...
+missing_panics_doc = "allow"
+module_name_repetitions = "allow"
+must_use_candidate = "allow"
+needless_pass_by_value = "allow"
+redundant_closure_for_method_calls = "allow"
+return_self_not_must_use = "allow"
+similar_names = "allow"
+struct_excessive_bools = "allow"
+too_many_lines = "allow"
+unreadable_literal = "allow"
```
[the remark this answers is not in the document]
sdht0: Addressed

tools/rust_api/src/error.rs
```diff
@@ -36,7 +38,7 @@ impl std::fmt::Debug for Error {
 
 impl std::error::Error for Error {
     fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
-        use Error::*;
+        use Error::CxxException;
```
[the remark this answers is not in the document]
sdht0: Inlined

tools/rust_api/src/value.rs
```diff
@@ -298,17 +300,17 @@ impl std::fmt::Display for Value {
...
-            Value::Timestamp(x) => write!(f, "{x}"),
-            Value::TimestampTz(x) => write!(f, "{x}"),
-            Value::TimestampNs(x) => write!(f, "{x}"),
-            Value::TimestampMs(x) => write!(f, "{x}"),
-            Value::TimestampSec(x) => write!(f, "{x}"),
+            Value::Timestamp(x)
+            | Value::TimestampTz(x)
+            | Value::TimestampNs(x)
+            | Value::TimestampMs(x)
+            | Value::TimestampSec(x) => write!(f, "{x}"),
```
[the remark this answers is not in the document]
sdht0: Do I need to address this in the PR?

examples/rust/Cargo.toml
sdht0: Found the workspace: https://github.com/kuzudb/kuzu/actions/runs/14398472589/workflow#L694

benjaminwinger: I think it was also used in the main ci workflow at one point, but I guess that one never got removed.

tools/rust_api/src/value.rs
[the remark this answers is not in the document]
benjaminwinger: No; maybe I should create an issue for it.

tools/rust_api/Cargo.toml
```diff
@@ -2,47 +2,47 @@
...
     "/kuzu-src/CMakeLists.txt",
     "/kuzu-src/tools/CMakeLists.txt",
 ]
 
-edition = "2018"
-
 [dependencies]
-cxx = "1.0"
+arrow = { version = "55", optional = true, default-features = false, features = ["ffi"] }
+cxx = "=1.0.138" # latest version that builds on clang 15
```
benjaminwinger, line 31: I'm concerned that we will probably end up unpinning this relatively soon to handle bugs which newer versions of cxx fix, but I guess we can try it and see when stuff breaks.

tools/rust_api/Cargo.toml
```diff
@@ -2,47 +2,47 @@
...
+rust_decimal = { version = "1.37", default-features = false }
```
benjaminwinger, line 32: Why bump the rust_decimal version? We haven't changed anything we do with rust_decimal since adding it, so it's unlikely that it's no longer working with 1.35, and this keeps things compatible with another crate if they need to use 1.35/1.36 for some particular reason (E.g. the above cxx pin). We could probably even push the requirement back earlier since [the only function we use was introduced in version 1.2.1](https://docs.rs/rust_decimal/1.2.1/rust_decimal/struct.Decimal.html#method.from_i128_with_scale) (I haven't tested it with that version though).

tools/rust_api/build.rs
```diff
@@ -197,12 +197,11 @@ fn main() {
     if let (Ok(kuzu_lib_dir), Ok(kuzu_include)) =
         (env::var("KUZU_LIBRARY_DIR"), env::var("KUZU_INCLUDE_DIR"))
     {
-        println!("cargo:rustc-link-search=native={}", kuzu_lib_dir);
-        // Also add to environment used by cargo run and tests
-        println!("cargo:rustc-env=LD_LIBRARY_PATH={}", kuzu_lib_dir);
+        println!("cargo:rustc-link-search=native={kuzu_lib_dir}");
+        println!("cargo:rustc-link-arg=-Wl,-rpath,{kuzu_lib_dir}");
```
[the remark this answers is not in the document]
benjaminwinger: Seems fine to me. I'm not very familiar with the difference between using `LD_LIBRARY_PATH` and `-rpath`; some of what I could find indicated that `-rpath` has an effect on the library search path embedded in the final executable, but I couldn't actually reproduce a different end result when I tried with both methods (neither of which seem to embed a search path in the executable).

benjaminwinger (approved):
I'm not sure we should be setting the minimum supported rust-version based on our dependencies (particularly optional ones).
Unfortunately, Cargo doesn't handle this very well, but eventually it will default to selecting versions based on their `rust-version` (see [here](https://doc.rust-lang.org/nightly/edition-guide/rust-2024/cargo-resolver.html)), which means that we would be able to do things like indicating that we support a range of arrow versions and it will use the one supported by your current version of the compiler (currently we can also do that, but I think it will pull in the latest version and error saying that it needs a newer rust compiler, so you would have to manually pin the version of arrow that you want to use).

tools/rust_api/build.rs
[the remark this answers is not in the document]
sdht0: `LD_LIBRARY_PATH` didn't work reliably for me even on Linux. Sometimes I had to manually export it. `-rpath` ensures the binary itself knows where to look for the lib.

Standard practice in the Nix world where every lib is stored in separate `/nix/store/hash-lib-version` dir.

tools/rust_api/Cargo.toml
sdht0: Yeah this is to allow the builds to succeed on our mac runners. Eventually we'll have to either move everything to macOS 15 or figure out a way to update clang on the older version (probably through `brew`?).

tools/rust_api/Cargo.toml
sdht0: I was following the "best practice" of updating frequently so that the ecosystem is not stuck on older version, esp through transitive dependencies.

benjaminwinger: This doesn't force 1.35 though; the default is the `^` operator which would mean `>=1.35,<2`

PR #5201, opened 2025-04-06
Add TypeScript definitions for kuzu database API

mewim:


andyfengHKU (approved):
Just one quick question. Do we need to add tests for this?

PR #5205, opened 2025-04-07
Free space management initial PR

royi-luo:
# Description

This PR adds the main infrastructure needed to support free space management. In this PR space is only reclaimed after out-of-place updates of column chunks, reclaiming space after other operations (e.g. dropping columns) will be added in future PRs.

New classes:
- `PageChunkManager`: All data file page allocations are routed through this class. Consumers of this class will request sequential chunks of pages from this class, which will either be allocated by appending new pages to the data file or by reallocating free pages. This class is also used by consumers to free pages when they are no longer needed.
- `FreeSpaceManager`: Keeps track of chunks of pages that are freed. When page allocations are requested, this class can be used to reallocate previously freed pages. This class should not be used directly, instead calls should be made to the `PageChunkManager` which will handle accesses to this class.

Updated benchmarks [here](https://github.com/kuzudb/kuzu/pull/5205#issuecomment-2795154748)

# Contributor agreement


src/include/storage/page_chunk_manager.h
```diff
@@ -0,0 +1,74 @@
...
+
+    common::page_idx_t getStartPageIdx() const { return entry.startPageIdx; }
+    common::page_idx_t getNumPages() const { return entry.numPages; }
+
+    void writePagesToFile(const uint8_t* buffer, uint64_t bufferSize);
+    void writePageToFile(const uint8_t* pageBuffer, common::page_idx_t pageOffset);
+    bool isInMemoryMode() const;
+};
+
+class PageChunkManager {
```
royi-luo, line 36: Not sure if this is the best name for this, let me know if you have better ideas

src/storage/store/column.cpp
```diff
@@ -339,34 +340,33 @@ void Column::write(ColumnChunkData& persistentChunk, ChunkState& state, offset_t
...
     auto& metadata = persistentChunk.getMetadata();
     const auto startOffset = metadata.numValues;
-    const auto numPages = dataFH->getNumPages();
-    // TODO: writeValues should return new pages appended if any.
-    writeValues(state, metadata.numValues, data, nullChunkData, 0 /*dataOffset*/, numValues);
-    const auto newNumPages = dataFH->getNumPages();
-    metadata.numPages += (newNumPages - numPages);
+    const auto numNewPages =
+        writeValues(state, metadata.numValues, data, nullChunkData, 0 /*dataOffset*/, numValues);
+    KU_ASSERT(numNewPages == 0);
```
royi-luo, line 356: Is there actually a case when we append pages here? I only see this function being called when appending to a dictionary chunks for in-place string writes which I don't think should add new pages to the data file (the pages should all have been added when the data/offset chunks were initially allocated). I also haven't hit the assertion I added when running tests.

I feel like if there is a case where we do append pages this way is wrong since `updateShadowedPageWithCursor` just appends to the end of the file which doesn't guarantee that all the pages for a data/offset chunk are contiguous so if we don't use this logic I'll just remove it.

src/include/storage/store/column_chunk.h
```diff
@@ -20,16 +20,16 @@ struct ChunkCheckpointState {
 struct ColumnCheckpointState {
     ColumnChunkData& persistentData;
     std::vector<ChunkCheckpointState> chunkCheckpointStates;
-    common::row_idx_t maxRowIdxToWrite;
+    common::row_idx_t endRowIdxToWrite;
```
royi-luo, line 23: renamed this from `maxRowIdxToWrite` to `endRowIdxToWrite` this value is exclusive and the misleading naming was causing it to be used incorrectly (inclusively)

royi-luo:
# Benchmarks
I tested on three FSM strategies:
1. If a freed chunk contains `n` pages, it is put in the largest level that doesn't exceed size `n`. For example a chunk with `5` pages would go to level 2 (page size 4). When looking for a chunk of size `n` we go to the same level and look for a page that is large enough. If there is none we go up a level and try again.
2. We only allocate blocks sizes that are powers of 2. So if a chunk with 5 pages is requested we allocated a chunk of size 8. This allows us to use the same strategy when searching the FSM which avoid iteration.
3. FSM is disabled
## Results
Scenario|Strategy #|# of used pages|# of free pages|data.kz size|Execution time
----|----|----|----|----|----

test/test_files/function/call/call.test
```diff
@@ -423,3 +423,5 @@ LOWER|SCALAR FUNCTION|(STRING) -> STRING
 -STATEMENT CALL SHOW_FUNCTIONS() WHERE name = 'INTERVAL' RETURN *;
 ---- 1
 INTERVAL|SCALAR FUNCTION|(STRING) -> INTERVAL
+
+
```
royi-luo, line 427: revert

src/storage/page_chunk_manager.cpp
```diff
@@ -0,0 +1,85 @@
...
+}
+
+void PageChunkManager::freeBlock(PageChunkEntry entry) {
+    if constexpr (ENABLE_FSM) {
+        common::UniqLock lck{mtx};
+        for (uint64_t i = 0; i < entry.numPages; ++i) {
+            const auto pageIdx = entry.startPageIdx + i;
+            fileHandle->removePageFromFrameIfNecessary(pageIdx);
+        }
+        freeSpaceManager->addFreeChunk(entry);
```
royi-luo, line 30: Making a freed chunk immediately rewritable may not be fully robust so I may need to change this to only making freed chunks rewritable after a checkpoint completes (the reasoning should be similar to why we have shadow files).

A case like the following might happen otherwise:
begin out of place update -> free existing chunk -> reclaim the same chunk -> start writing -> DB crashes
When the DB is reloaded part of the existing chunk will already be written to so there is no way to recover

src/include/storage/store/column_reader_writer.h
```diff
@@ -9,11 +9,12 @@ class Transaction;
 }
 namespace storage {
 
-class FileHandle;
+class PageManager;
```
royi-luo, line 12: remove

src/include/storage/store/column_reader_writer.h
```diff
@@ -89,7 +92,7 @@ class ColumnReadWriter {
     std::pair<common::offset_t, PageCursor> getOffsetAndCursor(common::offset_t nodeOffset,
         const ChunkState& state) const;
 
-private:
+protected:
```
royi-luo, line 95: this might be able to be reverted

src/storage/store/column.cpp
```diff
@@ -181,8 +182,8 @@ ColumnChunkMetadata Column::flushData(const ColumnChunkData& chunkData, FileHand
     // TODO(Guodong/Ben): We can optimize the flush to write back to same set of pages if new
     // flushed data are not out of the capacity.
```
royi-luo, line 183: Can we remove this TODO @ray6080

src/include/storage/store/column_chunk_data.h
```diff
@@ -234,6 +234,8 @@ class KUZU_API ColumnChunkData {
 
     void updateStats(const common::ValueVector* vector, const common::SelectionView& selVector);
 
+    virtual void reclaimAllocatedPages(FileHandle& dataFH, const ChunkState& state) const;
```
ray6080, line 237: I don't think we actually need this interface. Can we not reclaim just based on ChunkState?

src/include/storage/file_handle.h
```diff
@@ -110,6 +111,13 @@ class FileHandle {
         return isLargePaged() ? common::TEMP_PAGE_SIZE : common::KUZU_PAGE_SIZE;
     }
 
+    PageManager* getPageManager() { return pageManager.get(); }
+    void serializePageManager(common::Serializer& ser) const;
+    void deserializePageManager(common::Deserializer& deSer);
+    void finalizeInit();
```
ray6080, line 117: How about `initPageManager`? Actually why do we need this public function? Why don't we initialize the PageManager when we construct FileHandle?

src/include/storage/free_space_manager.h
```diff
@@ -0,0 +1,66 @@
...
+struct FreeEntryIterator;
+
+class FreeSpaceManager {
+public:
+    static bool entryCmp(const PageChunkEntry& a, const PageChunkEntry& b);
+    using sorted_free_list_t = std::set<PageChunkEntry, decltype(&entryCmp)>;
+
+    FreeSpaceManager();
+
+    void addFreeChunk(PageChunkEntry entry);
```
ray6080, line 19: -> `addFreePages`

src/include/storage/free_space_manager.h
```diff
@@ -0,0 +1,66 @@
...
+    std::optional<PageChunkEntry> popFreeChunk(common::page_idx_t numPages);
```
ray6080, line 20: -> `popFreePages`

src/include/storage/store/column_reader_writer.h
```diff
@@ -72,14 +73,16 @@ class ColumnReadWriter {
...
+    virtual common::page_idx_t writeValuesToPageFromBuffer(ChunkState& state,
+        common::offset_t dstOffset, const uint8_t* data, const common::NullMask* nullChunkData,
+        common::offset_t srcOffset, common::offset_t numValues,
+        const write_values_func_t& writeFunc) = 0;
 
     void readFromPage(const transaction::Transaction* transaction, common::page_idx_t pageIdx,
         const std::function<void(uint8_t*)>& readFunc);
 
-    void updatePageWithCursor(PageCursor cursor,
+    // returns true if a new page was appended to the shadow file
```
ray6080, line 84: Update the comment.

src/include/storage/file_handle.h
```diff
@@ -110,6 +111,13 @@ class FileHandle {
         return isLargePaged() ? common::TEMP_PAGE_SIZE : common::KUZU_PAGE_SIZE;
     }
 
+    PageManager* getPageManager() { return pageManager.get(); }
+    void serializePageManager(common::Serializer& ser) const;
```
ray6080, line 115: I would consider removing `serializePageManager`, `deserializePageManager`, and `finalizeCheckpoint`. Instead just call `getPageManager` first.

src/include/storage/page_manager.h
```diff
@@ -0,0 +1,48 @@
...
+class FileHandle;
+
+class PageManager {
+public:
+    PageManager(FileHandle* fileHandle, std::unique_ptr<FreeSpaceManager> fsm)
+        : freeSpaceManager(std::move(fsm)), fileHandle(fileHandle) {}
+    explicit PageManager(FileHandle* fileHandle)
+        : PageManager(fileHandle, std::make_unique<FreeSpaceManager>()) {}
+
+    PageChunkEntry allocateBlock(common::page_idx_t numPages);
```
ray6080, line 24: Let's not use the term `Block` or `Chunk`. We can just stay with `PageRange` and `pages`. I think they're fine.

-> `allocatePages` or `allocatePageRange`.

src/include/storage/page_manager.h
```diff
@@ -0,0 +1,48 @@
...
+    void freeBlock(PageChunkEntry block);
```
ray6080, line 25: Ditto on not use term `Block`.

src/storage/store/column.cpp
ray6080: Yep. Let's remove this.

src/storage/store/column.cpp
ray6080: @benjaminwinger can you double check here?

src/storage/store/column_reader_writer.cpp
```diff
@@ -437,11 +442,12 @@ void ColumnReadWriter::readFromPage(const Transaction* transaction, page_idx_t p
     fileHandleToPin->optimisticReadPage(pageIdxToPin, readFunc);
 }
 
-void ColumnReadWriter::updatePageWithCursor(PageCursor cursor,
+page_idx_t ColumnReadWriter::updatePageWithCursor(PageCursor cursor,
```
ray6080, line 445: Should the return type be bool?

src/include/storage/page_manager.h
```diff
@@ -0,0 +1,48 @@
...
+
+    // In cases like committing a dropped column chunk we don't want to reclaim pages right away
+    // This way we can undo if the commit is rolled back
+    void addUncheckpointedFreeBlock();
```
ray6080: This doesn't seem to be used or implemented. Remove it.

src/storage/page_manager.cpp
```diff
@@ -0,0 +1,47 @@
...
+
+#include "common/uniq_lock.h"
+#include "storage/file_handle.h"
+#include "storage/shadow_utils.h"
+
+namespace kuzu::storage {
+static constexpr bool ENABLE_FSM = true;
+
+PageChunkEntry PageManager::allocateBlock(common::page_idx_t numPages) {
+    // numPages = std::bit_ceil(numPages);
```
ray6080: Remove L11.

ray6080 (commented):
Thanks! This looks good to me in the high level. I have some initial comments here. Wanna take a closer look after they're addressed.

src/storage/store/column.cpp
royi-luo: ok removed

src/include/storage/page_chunk_entry.h
```diff
@@ -0,0 +1,18 @@
+#pragma once
+
+#include "common/types/types.h"
+namespace kuzu::storage {
+
+struct PageChunkEntry {
```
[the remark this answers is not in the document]
royi-luo: done

src/include/storage/free_space_manager.h
```diff
@@ -0,0 +1,66 @@
...
+    static std::unique_ptr<FreeSpaceManager> deserialize(common::Deserializer& deSer);
+    void finalizeCheckpoint();
+
+    common::row_idx_t getNumEntries() const;
+    std::vector<PageChunkEntry> getEntries(common::row_idx_t startOffset,
+        common::row_idx_t endOffset) const;
+
+private:
+    PageChunkEntry breakUpChunk(PageChunkEntry chunk, common::page_idx_t numRequiredPages);
+    void combineAdjacentChunks();
```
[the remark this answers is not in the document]
royi-luo: updated

src/include/storage/free_space_manager.h
royi-luo: updated

src/include/storage/free_space_manager.h
royi-luo: updated

src/include/storage/store/column_reader_writer.h
royi-luo: changed the return value to `bool`

src/include/storage/page_manager.h
royi-luo: updated

src/include/storage/file_handle.h
royi-luo: updated

src/include/storage/page_manager.h
royi-luo: updated

src/storage/store/column_reader_writer.cpp
royi-luo: updated

src/include/storage/page_manager.h
royi-luo: updated

src/storage/page_manager.cpp
royi-luo: updated

src/include/storage/file_handle.h
royi-luo: had it there since `deserialize()` constructed the page manager but if it was a fresh DB I needed to construct it separately. I updated `deserialize()` to be a void function working on an instance of the page manager which removes the need for this function.

src/include/storage/store/column_chunk_data.h
```diff
@@ -234,6 +235,9 @@ class KUZU_API ColumnChunkData {
 
     void updateStats(const common::ValueVector* vector, const common::SelectionView& selVector);
 
+    virtual void reclaimAllocatedPages(PageChunkManager& pageChunkManager,
```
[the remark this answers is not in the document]
royi-luo: moved logic to `ChunkState`

src/include/storage/store/column_chunk_data.h
royi-luo: updated

src/include/storage/page_chunk_entry.h
ray6080, line 1: rename the file too.

src/include/storage/store/column_chunk_metadata.h
```diff
@@ -2,14 +2,17 @@
 
 #include "common/types/types.h"
 #include "storage/compression/compression.h"
+#include "storage/page_chunk_entry.h"
 
 namespace kuzu::storage {
 struct ColumnChunkMetadata {
-    common::page_idx_t pageIdx;
-    common::page_idx_t numPages;
+    PageRange pageChunk;
```
ray6080, line 9: Can you rename the variable too?

test/test_files/function/call/fsm_info.test
```diff
@@ -0,0 +1,21 @@
+-DATASET CSV empty
+-SKIP_IN_MEM
+--
+
+-CASE FSMInfoRecovery
+# TODO(Royi) enable this test once FSM works on dropped tables
```
ray6080, line 6: Can you use COPY or UNWIND CREATE to trigger FSM so we can enable this test without waiting for reclaiming on drop table?

tools/python_api/test/test_fsm.py
ray6080, line 1: I would add more micro test case here. Also make sure we cover rel tables too.
The basic logic can be 1) run a few insertions or a single copy into a node/rel table, trigger checkpoint, check that free space is empty. 2) insert a bit more tuples or a copy to trigger out-of-place writes, and check we now have some pages in free space. 3) update few tuples from the first insertion to trigger out-of-place updates again, we should be able to use free pages from our last checkpoint.

ray6080 (approved):
Looks good to me. @benjaminwinger can you take another look?

src/include/storage/page_chunk_entry.h
royi-luo: updated

src/include/storage/store/column_chunk_metadata.h
royi-luo: updated

test/test_files/function/call/fsm_info.test
royi-luo: Added more cypher tests

tools/python_api/test/test_fsm.py
royi-luo: Moved the python tests to cypher (they're now a bit different and similar to how you described)

royi-luo:
I rebenchmarked the FSM efficiency after making the change where freed pages are only reusable at the end of the checkpoint. The original benchmarks are [here](https://github.com/kuzudb/kuzu/pull/5205#issuecomment-2787392617)

Note that FSM efficiency is a bit worse now since we only make pages reusable after the end of the checkpoint (so if you free `n` pages then immediately request it again you won't get the same range)

Another new observation that I made is that fragmentation is a bit better with this version, likely because waiting for freed pages to be reusable gives us to opportunity to combine more adjacent free pages.

## Results
Scenario|Strategy|# of used pages|# of free pages|data.kz size|Execution time|Notes
----|----|----|----|----|----|----
Start with 300k values from 0-255, update values to 0-65535 in batches of 10 (2 batches)|1|450|401|15MB|11.51s|
^|2|
^|3|450|0|113MB|11.81s
Start with 300k values from 0-65535, update values to 0-255 in batches of 10k (30 batches)|1|208|224|13MB|1.38s|
^|2|
^|3|208|0|20MB|1.34s
Insert (create) 100K values between 0-65535 in batches of size 1k|1|101|167|7.1MB|4.37s
^|2|
^|3|101|0|16MB|4.42s
Batch insert (copy) 100K values between 0-65535 in batches of size 1k|1|101|167|7.1MB|4.32s
^|2|
^|3|101|0|16MB|4.27s
Batch insert ldbc-sf1 Comment in batches of size 1k|1|2148|1740|25MB|7.11s
^|2|
^|3|2148|0|383MB|7.02s
Batch insert first 1M rows of ldbc-sf100 Comment in batches of size 20k|1|12153|1071|95MB|4.69s|1071 free pages split over 19 entries, largest free entry is 691 pages
^|2|
^|3|12153|0|277MB|4.73s
Batch insert first 2M rows of ldbc-sf100 comment_replyOf_comment in batches of size 10k|1|7572|837|846MB|287.84s
^|2|
^|3|7572|0|1006MB|173.86s

src/include/storage/page_chunk_manager.h
royi-luo: changed to `PageRange`

src/storage/page_chunk_manager.cpp
royi-luo: Updated, freed chunks are now not immediately rewritable

benjaminwinger (approved):

src/include/common/utils.h
```diff
@@ -60,6 +60,12 @@ bool containsValue(const Container& container, const T& value) {
     return std::find(container.begin(), container.end(), value) != container.end();
 }
 
+template<std::integral T>
+T countBits(T) {
```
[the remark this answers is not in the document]
royi-luo: updated

src/include/storage/free_space_manager.h
```diff
@@ -0,0 +1,77 @@
...
+    static bool entryCmp(const PageRange& a, const PageRange& b);
+    using sorted_free_list_t = std::set<PageRange, decltype(&entryCmp)>;
+    using free_list_t = std::vector<PageRange>;
+
+    FreeSpaceManager();
+
+    void addFreePages(PageRange entry);
+    std::optional<PageRange> popFreePages(common::page_idx_t numPages);
+
+    // These pages are not reusuable until the end of the next checkpoint
```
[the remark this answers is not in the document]
royi-luo: updated

src/storage/free_space_manager.cpp
```diff
@@ -0,0 +1,190 @@
...
+    }
+    return freeLists[level];
+}
+
+FreeSpaceManager::FreeSpaceManager() : freeLists{}, numEntries(0){};
+
+common::idx_t FreeSpaceManager::getLevel(common::page_idx_t numPages) {
+    // level is exponent of largest power of 2 that is <= numPages
+    // e.g. 2 -> level 1, 5 -> level 2
+    return common::countBits(numPages) - common::CountZeros<common::page_idx_t>::Leading(numPages) -
```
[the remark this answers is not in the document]
royi-luo: Yep updated to `CountZeros::Trailing(std:;bit_floor(numPages))`

src/include/storage/page_range.h
```diff
@@ -0,0 +1,18 @@
+#pragma once
+
+#include "common/types/types.h"
+namespace kuzu::storage {
+
+struct PageRange {
+    PageRange() = default;
+    PageRange(common::page_idx_t startPageIdx, common::page_idx_t numPages)
+        : startPageIdx(startPageIdx), numPages(numPages) {}
+    PageRange(const PageRange& o, common::page_idx_t startPageInOther)
```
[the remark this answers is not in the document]
royi-luo: Yep I don't think the subrange is used anywhere, I originally used it in a single spot but changed it when fixing a bug. I'll leave it in as a `subrange` function in case we wish to use it elsewhere. I'll hold off on replacing it with `std::ranges` until the use cases for this class get more complex since currently it is essentially a pure data struct

src/storage/store/column.cpp
royi-luo: Sounds good I have the same understanding, I added the assertion just confirm my guess but I'll remove it now that it's confirmed. Moved the assertion to `updatePageWithCursor`.

src/storage/free_space_manager.cpp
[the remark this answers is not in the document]
benjaminwinger: I think I got the semantics mixed up, I was thinking it could replace the whole expression, but it obviously does not. What you had first was good, but making `countBits` `constexpr` does make a difference here (effectively called on the type, not the value), and `CountZeros::Leading` could probably use `std::countl_zero`; there's a `bsr` instruction that does the heavy lifting, and `countl_zero` is defined on all integral types (it appears to just widen values smaller than 32-bits) so it would be more generally useful than the `__builtin_clzl` that is currently only being used on clang (just for asserting? I think the code was copied from duckdb, which targets c++11 and can't make use of `<bit>`).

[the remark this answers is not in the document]
royi-luo: Sure I was wondering why we weren't using the standard library function, I can make this change in my next PR.

PR #5212, opened 2025-04-08
Fix click benchmark q29

benjaminwinger:
Following up #5169; I've fixed the InMemOverflowBuffer performance by making it scale by starting the first block using the minimum size and doubling the size of each subsequent block (up to 256KB, unless values exceed that), which seems to have kept the original performance while still keeping the minimum size of allocations small (I haven't benchmarked it exhaustively; I just tried on `ldbc-100/var_size_seq_scan/q20`, which is perhaps marginally slower than before (~1720ms compared to ~1680ms).

~~I'm still working on updating the queries which had been failing in `benchmark/click/queries`; I'll update those in a later PR.~~ (Added here)


ray6080 (approved):
LGTM. The ground truth doesn't seem to match in CI runs. You may want to double check on it.

benjaminwinger:
> LGTM. The ground truth doesn't seem to match in CI runs. You may want to double check on it.

It's failing I think; the benchmark runner doesn't report errors

PR #5214, opened 2025-04-09
Implement dynamic sparse/dense frontier switch for recursive joins

andyfengHKU:
# Description

Implement sparse frontier. The incentive is to avoid costly allocation of dense data structures when the database size is large. The sparse frontier will adaptively converted itself to dense frontier when the frontier becomes large.

Some preliminary test result.

For LDBC100 comment- reply-of-comment.

```
MATCH (c:Comment)-[]->(d) WHERE c.ID = 8796093023595 RETURN d.*;
```
Takes about  7ms.

While its recursive version on master
```
MATCH (c:Comment)-[*1..1]->(d) WHERE c.ID = 8796093023595 RETURN d.*;
```
Takes about 300ms

This PR brings it back down to 7ms.

Note: I'm using a very small threshold for sparse frontier for now because all test cases are small. Once add more CI tests, the threshold needs to be made larger.

- [ ] Increase frontier threshold and add CI test for different threshold
- [ ] Add read/write interface to GDSSparse/DenseObjectManager

### DIFF
--- a/src/function/gds/asp_destinations.cpp
+++ b/src/function/gds/asp_destinations.cpp
@@ -1,3 +1,5 @@
+#include <unistd.h>
+
 #include "binder/expression/node_expression.h"
 #include "function/gds/gds_function_collection.h"
 #include "function/gds/rec_joins.h"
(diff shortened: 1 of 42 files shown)


src/function/gds/page_rank.cpp
```diff
@@ -125,6 +125,8 @@ class PageRankAuxiliaryState : public GDSAuxiliaryState {
         pNext.pinTable(fromTableID);
     }
 
+    void switchToDense(ExecutionContext*, Graph*) override {}
+
 private:
     Degrees& degrees;
     PValues& pCurrent;
```
ray6080, line 132: Why don't we design pCurrent and pNext as Pair ?

src/include/function/gds/bfs_graph.h
```diff
@@ -41,79 +45,138 @@ class ParentList {
     std::atomic<ParentList*> next;
 };
 
-class BFSGraph {
-    friend class BFSGraphInitVertexCompute;
-    // Data type that is allocated to max num nodes per node table.
-    using parent_entry_t = std::atomic<ParentList*>;
+class BaseBFSGraph {
```
ray6080, line 48: BFSGraph should also make use of `GSDObjects<ParentList*>` to handle sparse and dense data structures, instead of relying on derived `DenseBFSGraph` and `SparseBFSGraph`.

src/function/gds/frontier_morsel.cpp
```diff
@@ -0,0 +1,35 @@
...
+    nextOffset.store(0u);
+    // Frontier size calculation: The ideal scenario is to have k^2 many morsels where k
+    // the number of maximum threads that could be working on this frontier. However, if
+    // that is too small then we default to MIN_FRONTIER_MORSEL_SIZE.
+    auto idealMorselSize =
+        maxOffset / std::max(MIN_NUMBER_OF_FRONTIER_MORSELS, maxThreads * maxThreads);
+    morselSize = std::max(MIN_FRONTIER_MORSEL_SIZE, idealMorselSize);
+}
+
+bool FrontierMorselDispatcher::getNextRangeMorsel(FrontierMorsel& frontierMorsel) {
```
ray6080, line 24: Isn't the returned boolean flag kinda of carrying duplicated information with frontierMorsel? I think you should be able to find out if there is more morsel by checking if the returned morsel is valid or not.

src/processor/operator/recursive_extend.cpp
```diff
@@ -77,38 +80,40 @@ void RecursiveExtend::executeInternal(ExecutionContext* context) {
...
-            candidates.mergeSparseFrontier(frontierPair->getNextSparseFrontier());
-            if (candidates.enabled()) {
-                GDSUtils::runVertexComputeSparse(candidates, graph, *vertexCompute);
-            } else {
+            auto writer = function->getOutputWriter(context, bindData, *computeState, sourceNodeID,
+                sharedState.get());
+            auto vertexCompute = std::make_unique<RJVertexCompute>(
+                clientContext->getMemoryManager(), sharedState.get(), writer->copy());
+            switch (computeState->frontierPair->getState()) {
+            case GDSDensityState::DENSE: {
```
ray6080, line 109: I don't think neither the operator nor `computeState->frontierPair` should be aware of SPARSE and DENSE of internal GDS objects.
1. we should get rid of the switch/case here. All computation should go through the generic interface `GDSUtils::runVertexCompute(context, graph, *vertexCompute);`
2. we should remove the specialized interface for sparse structures `VertexCompute::vertexCompute(common::table_id_t)`, which is not really needed.
3. `VertexCompute` needs expose an interface to indicate if it can run in parallel or not. In the sparse case, we can make sure we only grab one morsel from `FrontierMorselDispatcher`, thus only spawn one thread.

src/include/function/gds/compute.h
```diff
@@ -50,6 +50,9 @@ class VertexCompute {
     // the function on each node in a graph.
     virtual void vertexCompute(const graph::VertexScanState::Chunk&) {}
     virtual void vertexCompute(common::offset_t, common::offset_t, common::table_id_t) {}
+    // This function assumes the number of nodes is small (sparse) and morsel driven parallelism
+    // is not necessary. It should not be used in parallel computations.
+    virtual void vertexCompute(common::table_id_t) {}
```
ray6080, line 55: We should remove this interface.

src/function/gds/output_writer.cpp
```diff
@@ -65,26 +66,40 @@ static ParentList* getTop(const std::vector<ParentList*>& path) {
...
+
+void PathsOutputWriter::write(FactorizedTable& fTable, nodeID_t dstNodeID, LimitCounter* counter) {
+    if (!inOutputNodeMask(dstNodeID.offset)) {
         return;
     }
-    dfsSlow(firstParent, fTable, counter);
+    dstNodeIDVector->setValue<nodeID_t>(0, dstNodeID);
+    writeInternal(fTable, dstNodeID, counter);
+
+    // auto firstParent = findFirstParent(dstNodeID.offset);
```
ray6080, line 86: Ditto on cleanup.

src/function/gds/output_writer.cpp
```diff
@@ -378,5 +393,30 @@ void PathsOutputWriter::addNode(nodeID_t nodeID, sel_t pos) const {
     ListVector::getDataVector(pathNodeIDsVector.get())->setValue(pos, nodeID);
 }
 
+void SPPathsOutputWriter::writeInternal(FactorizedTable& fTable, nodeID_t dstNodeID,
+    LimitCounter* counter) {
+    auto firstParent = findFirstParent(dstNodeID.offset);
+    if (firstParent == nullptr) {
+        // if (sourceNodeID_ == dstNodeID && info.lowerBound == 0) {
```
ray6080, line 400: Ditto on cleanup.

ray6080 (commented):
Finished my first round of reading on the changes. I was mainly trying to understand the core data structures and the sparse/dense changes. Made some comments below. The main change I'd like to see is a better separation of internal details of how we manage sparse/dense GDS objects from more high-level data structures, such as the `FrontierPair`, and also some part in the computation level, such as RecursiveJoin operator and VectexCompute.

We should converge and then I can continue the rest of more detailed changes in each algorithm, most of which should be trivial.

src/function/gds/asp_destinations.cpp
```diff
@@ -12,90 +14,214 @@ using namespace kuzu::graph;
...
+        KU_ASSERT(curData);
+        return curData[offset].load(std::memory_order_relaxed);
+    }
+
+private:
+    GDSDenseObjectManager<std::atomic<multiplicity_t>>& denseObjects;
+    std::atomic<multiplicity_t>* curData = nullptr;
+};
+
+class MultiplicitiesPair {
```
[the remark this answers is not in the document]
andyfengHKU: will do this in a separate PR

src/function/gds/frontier_morsel.cpp
andyfengHKU: I'll refactor this with a different PR.

ray6080 (approved):
Can you update your TODO list in the PR description? so it's more clear.

PR #5221, opened 2025-04-11
Remove enabled check from semi mask

andyfengHKU:
# Description

The convention is to use nullptr to represent a mask is enabled or not. So this PR removes the enable interface from NodeOffsetMask

Fixes # (issue)

# Contributor agreement

### DIFF
--- a/src/processor/operator/recursive_extend.cpp
+++ b/src/processor/operator/recursive_extend.cpp
@@ -6,7 +6,7 @@
 #include "function/gds/compute.h"
 #include "function/gds/gds_utils.h"
 #include "processor/execution_context.h"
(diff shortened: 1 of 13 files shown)


src/processor/operator/recursive_extend.cpp
```diff
@@ -6,6 +6,7 @@
 #include "function/gds/compute.h"
 #include "function/gds/gds_utils.h"
 #include "processor/execution_context.h"
+#include <binder/expression/node_expression.h>
```
ray6080, line 9:
```suggestion
#include "binder/expression/node_expression.h"
```

ray6080 (approved):

PR #5235, opened 2025-04-15
FSM: Reclaim pages when dropping columns/tables

royi-luo:
# Description

When dropping a column is committed (dropping a table is just dropping a bunch of columns), pushes the page ranges for the column chunks into the FSM's `uncheckpointedFreeChunks`. After the end of the next checkpoint, these pages will become reusable.

Detecting if a column/table has been dropped is done in the storage layer. In the storage manager, any tables that don't exist in the catalog entry during checkpoint are considered to have been dropped. The same is done at the node group level for dropped columns.

# Testing

There are two types of tests in this PR:
1. Cypher tests that perform basic sanity checks (e.g. does the FSM have >0 pages when expected). These tests are easier to debug and hopefully catch simpler bugs
2. Python tests that perform more complex checks to see if the free page ranges (start page + num pages) exactly match their expected values.

# Contributor agreement


ray6080 (approved):
Looks good! Thanks!

PR #5239, opened 2025-04-16
Add negative weight check to weighted shortest path

andyfengHKU:
# Description

See title

### DIFF
--- a/src/function/gds/awsp_paths.cpp
+++ b/src/function/gds/awsp_paths.cpp
@@ -26,6 +26,7 @@ public:
         bool fwdEdge) override {
         std::vector<nodeID_t> result;
         chunk.forEach<T>([&](auto nbrNodeID, auto edgeID, auto weight) {
+            checkWeight(weight);
             if (!block->hasSpace()) {
                 block = bfsGraphManager->getCurrentGraph()->addNewBlock();
             }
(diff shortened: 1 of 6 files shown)


src/function/gds/awsp_paths.cpp
```diff
@@ -26,6 +26,10 @@ class AWSPPathsEdgeCompute : public EdgeCompute {
         bool fwdEdge) override {
         std::vector<nodeID_t> result;
         chunk.forEach<T>([&](auto nbrNodeID, auto edgeID, auto weight) {
+            if (weight < 0) {
+                [[unlikely]] throw RuntimeException(
```
ray6080, line 30: Can you say a bit more in the error message?
"Found negative weight {}." -> "Found negative weight {}, which is not supported yet."

src/function/gds/awsp_paths.cpp
```diff
@@ -26,6 +26,10 @@ class AWSPPathsEdgeCompute : public EdgeCompute {
         bool fwdEdge) override {
         std::vector<nodeID_t> result;
         chunk.forEach<T>([&](auto nbrNodeID, auto edgeID, auto weight) {
+            if (weight < 0) {
```
ray6080, line 29: I would abstract this if check as a function. but up to you.

ray6080 (approved):

PR #5241, opened 2025-04-16
Improve scc performance

andyfengHKU:
# Description

Apparently we were doing unnecessary backward scanning. The following code fix it
```
if (visitedState.componentIDSet(nextNode)) {
     continue;
 }
```

This has brought our execution time on `SNAP soc-epinions` from 100ms to 30ms. The improvement highly depends on the topology on graph. For denser/larger graph, the improvement should be more significant.

This PR also includes removes `DFSStack` which used to be a vector implementation. However, since we know the stack will always have the size of `maxOffset`, it make more sense to pre-allocate upfront. Also rename the stack to `VisitedOrder`.

Finally this PR adds support for kosaraju on filtered projected graph.

### DIFF
--- a/src/function/gds/strongly_connected_components_kosaraju.cpp
+++ b/src/function/gds/strongly_connected_components_kosaraju.cpp
@@ -69,9 +69,7 @@ public:
         return curData[size_];
     }
 
-    uint64_t size() const {
-        return size_;
-    }
+    uint64_t size() const { return size_; }
 
 private:
     offset_t* curData = nullptr;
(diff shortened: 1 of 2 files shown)


src/function/gds/strongly_connected_components_kosaraju.cpp
```diff
@@ -1,11 +1,12 @@
+#include <stack>
```
andyfengHKU, line 1: Remove or replace std::vector with std::stack

mewim (approved):

PR #5249, opened 2025-04-19
Make the scan of relID optional for GDS algorithms

ray6080:
# Description

Under quite a few cases, relID is not needed for computation while we still scan from rel tables. This PR makes the scan of relID optional and configured per algorithm.

Performance comparison between this branch and master on WCC and K-Core are as follows. This is done on a Mac mini with m4 pro chip, 64GB RAM and 1TB SSD.
| Algorithm | Master - cold (ms) | Master - warm (ms) | This branch - cold (ms) | This branch - warm (ms) |
| ---------- | ------------------ | -------------------- | ----------------------- | ------------------------- |
| WCC          | 140.5                       | 92.6                           |  87.1                                | 75.6                                    |
| K-CORE     | 9612.7                     | 9241.8                       | 7168.3                            | 6278.9                                |

The performance different on PageRank is not visible.

TODO:
- [ ] Get rid of the hard coding when accessing rel property vectors inside each algorithm.

### DIFF
--- a/src/function/gds/wsp_paths.cpp
+++ b/src/function/gds/wsp_paths.cpp
@@ -25,13 +25,16 @@ public:
     std::vector<nodeID_t> edgeCompute(nodeID_t boundNodeID, graph::NbrScanState::Chunk& chunk,
         bool fwdEdge) override {
         std::vector<nodeID_t> result;
-        chunk.forEach<T>([&](auto nbrNodeID, auto edgeID, auto weight) {
+        chunk.forEach([&](auto neighbors, auto propertyVectors, auto i) {
+            auto nbrNodeID = neighbors[i];
+            auto edgeID = propertyVectors[0]->template getValue<relID_t>(i);
+            auto weight = propertyVectors[1]->template getValue<T>(i);
             checkWeight(weight);
             if (!block->hasSpace()) {
                 block = bfsGraphManager->getCurrentGraph()->addNewBlock();
             }
             if (bfsGraphManager->getCurrentGraph()->tryAddSingleParentWithWeight(boundNodeID,
-                    edgeID, nbrNodeID, fwdEdge, (double)weight, block)) {
+                    edgeID, nbrNodeID, fwdEdge, static_cast<double>(weight), block)) {
                 result.push_back(nbrNodeID);
             }
         });
--- a/src/graph/graph.cpp
+++ b/src/graph/graph.cpp
@@ -1,7 +1,5 @@
 #include "graph/graph.h"
 
-#include <utility>
-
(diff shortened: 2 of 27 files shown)


src/graph/graph.cpp
```diff
@@ -1,14 +1,14 @@
 #include "graph/graph.h"
 
+#include <utility>
```
andyfengHKU, line 3: remove

src/function/gds/wsp_paths.cpp
```diff
@@ -24,12 +24,15 @@ class WSPPathsEdgeCompute : public EdgeCompute {
     std::vector<nodeID_t> edgeCompute(nodeID_t boundNodeID, graph::NbrScanState::Chunk& chunk,
         bool fwdEdge) override {
         std::vector<nodeID_t> result;
-        chunk.forEach<T>([&](auto nbrNodeID, auto edgeID, auto weight) {
+        chunk.forEach([&](auto neighbors, auto propertyVectors, auto i) {
+            auto nbrNodeID = neighbors[i];
+            auto edgeID = propertyVectors[0]->template getValue<relID_t>(i);
```
andyfengHKU, line 29: I guess this template is not needed.

src/include/graph/on_disk_graph.h
```diff
@@ -30,12 +30,15 @@ class OnDiskGraphNbrScanState : public NbrScanState {
...
         bool randomLookup = false);
 
     Chunk getChunk() override {
-        return createChunk(currentIter->getNbrNodes(), currentIter->getEdges(),
-            currentIter->getSelVectorUnsafe(), propertyVector.get());
+        std::vector<common::ValueVector*> vectors;
+        for (auto& propertyVector : propertyVectors) {
+            vectors.push_back(propertyVector.get());
+        }
+        return createChunk(currentIter->getNbrNodes(), currentIter->getSelVectorUnsafe(), vectors);
```
andyfengHKU, line 41: I have a question, are we creating a new chunk per scanFwd call? I have a feeling we might want to mark Chunk as EXPLICIT_COPY.

src/processor/operator/recursive_extend.cpp
```diff
@@ -147,5 +151,16 @@ void RecursiveExtend::executeInternal(ExecutionContext* context) {
     sharedState->factorizedTablePool.mergeLocalTables();
 }
 
+bool RecursiveExtend::requireRelID(const RJAlgorithm& function) {
```
andyfengHKU, line 154: make this static

andyfengHKU (approved):
I would double check with GDS performance, the improvement on WCC should be noticeable.

Issue #1805, opened 2023-07-12
Rust API Ergonomics


benjaminwinger:
It might be helpful to have query/execute be generic functions that return a particular type, rather than returning a Value enum.

I'm not entirely sure how this should work with iterating over multiple results. Two separate functions might be better than trying to do both as a single function, one which will unwrap the iterator and return a single result of the given type, and one which returns an iterator where the type argument represents the iterator's elements.

E.g.
```rust
let result: String = conn.query_one::<String>("RETURN 'foo';")?;

for name in conn.query::<String>>("MATCH (a:Person) RETURN a.name;")? {
    ...
}

for (name, age) in conn.query::<(String, u64)>>("MATCH (a:Person) RETURN a.name, a.age;")? {
    ...
}
```

instead of

```rust
if let Some(Value::String(result)) = conn.query("RETURN 'foo';")?.next() {
    ...
} else {
    panic!("Unexpected result type!"); // or an Error, preferably
}

for result in conn.query("MATCH (a:Person) RETURN a.name;")? {
    if let Value::String(name) = result[0] {
        ...
    } else {
        panic!("Unexpected result type!");
    }
}

for result in conn.query("MATCH (a:Person) RETURN a.name, a.age;")? {
    if let (Value::String(name), Value::Int64(age)),  = (result[0], result[1]) {
        ...
    } else {
        panic!("Unexpected result type!");
    }
}
```

It would require a trait which returned types implement, and could be further extended to allow the trait to be implemented automatically onto structs via a procedural macro.
The trait could be as simple as `From<Vec<Value>>`, but it might make more sense to use a custom trait so that the specific requirements can be more easily changed in future without necessarily being breaking changes.

Given #1804, it might be helpful to use C++ ffi types as the base type for conversion internally so that it would be possible to convert to arrow data structures directly without going through `Value`. On the other hand, RecordBatch would need the entire result, so in that case I think we would want to convert from `ffi::QueryResult` rather than `ffi::FlatTuple` or `ffi::Value`. Maybe the query function that returns a single value could be implemented in a way that supports that, as the impl for simple types like String could just take the first element of the first tuple, while the impl for a complex type like RecordBatch could consume the entire QueryResult.

I feel like something similar might be able to be done for the C++ API, but that probably deserves its own issue.

PR #5251, opened 2025-04-19
Add basic helper 'as_type' functions to rust-api

hjarrell:
# Description

The rust API currently requires chaining if lets or match statements to parse the Values out of the query:

```rust 
let id = if let Value::String(str) = &row[0] {
        str.clone()
    } else {
        "".to_string()
    };
```

This lets you skip that with things like:
```
let id = row[0].as_string().cloned().unwrap_or_default();
```

Starts working on #1805 in a future PR I want to add an optional serde feature to make querying with types that derive deserialize work.

# Contributor agreement

- I am accepting this CLA in my individual capacity. This contribution is being made on my own time, is not related to my employment, and does not represent my employer.

### DIFF
--- a/tools/rust_api/src/value.rs
+++ b/tools/rust_api/src/value.rs
@@ -959,6 +959,178 @@ impl From<&str> for Value {
     }
 }
 
+impl Value {
+    /// Returns the inner String if this Value is a String, otherwise None
+    pub fn as_string(&self) -> Option<&String> {
+        match self {
+            Value::String(s) => Some(s),
+            _ => None,
+        }
+    }
+
+    /// Consumes the Value and returns the inner String if this Value is a String, otherwise None
+    pub fn into_string(self) -> Option<String> {
+        match self {
+            Value::String(s) => Some(s),
+            _ => None,
+        }
+    }
+
+    /// Returns the inner bool if this Value is a Bool, otherwise None
+    pub fn as_bool(&self) -> Option<bool> {
+        match self {
+            Value::Bool(b) => Some(*b),
+            _ => None,
+        }
+    }
+
+    /// Returns the inner i64 if this Value is an Int64, otherwise None
+    pub fn as_i64(&self) -> Option<i64> {
+        match self {
+            Value::Int64(i) => Some(*i),
+            _ => None,
+        }
+    }
+
+    /// Returns the inner i32 if this Value is an Int32, otherwise None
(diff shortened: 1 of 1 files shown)


benjaminwinger:
Was there a particular reason for having explicit "as_type" functions instead of implementing TryInto? I've been working sporadically on #1805 (other things have taken priority and it's not really working at the moment), but I had implemented `TryInto` for most of the types as part of that and could pull those changes out of what I was working on.
One difference I can see is that `TryInto` doesn't really work for `BLOB`, since it's indistinguishable from `Vec<u8>`/`UINT8[]`. But otherwise I think using `TryInto` fits better with #1805.

Nulls admittedly causes some problems, since you can't turn a `STRING[]` into a `Vec<String>` if it contains a null (though if you expect a `Vec<String>` you probably expect all values to be non-null values. or alternatively nulls could get turned into default values instead of producing an error).

I've moved that part of what I had into its own branch: https://github.com/kuzudb/kuzu/compare/rust-convert-types (it needs some more tests, but I think it's usable).
The work I'd done on #1805 is [here](https://github.com/kuzudb/kuzu/compare/master...benjaminwinger:kuzu:rust-ergonomics) (it's a little stale).

hjarrell:
Sorry yeah I was working on something using this library and feeling some of the pain without some helper functions so figured I'd try to contribute back. Just going to close this for now as I decided not to use the library and seems you have a better plan!

andyfengHKU:
> Sorry yeah I was working on something using this library and feeling some of the pain without some helper functions so figured I'd try to contribute back. Just going to close this for now as I decided not to use the library and seems you have a better plan!

Hi @hjarrell, just curious is it because the lib has some usability or performance issues or because it doesn't match what you are looking for. In the former case, it would be helpful if you can share some feedback with us so that we can continue to improve.

hjarrell:
Honestly, I probably didn't evaluate it for enough time but was working on a little side project at the time with Tauri and when I force quit the app one time the database wouldn't load again. I found something on the discord or github at the time that talked about the DB not becoming recoverable if its not gracefully quit. 

Could be completely wrong though, like I said I didn't give it a fair shot and moved on to a different project.

andyfengHKU:
ah I see. That's indeed a deal-breaker. Could be an edge case that we didn't cover (if you share the corrupted database with us we should be able to fix). In the future, if by chance you pick us up again, feel free to open an issue. We will get to this type of bug asap.

PR #5258, opened 2025-04-22
Replace comparisons with toLower/toUpper with caseInsensitiveEquals

benjaminwinger:
For #5256. This improved the single-threaded performance significantly. On my machine the same benchmark used in #5256 took 2.8s initially (in an in-memory database) and improved to 1.7s with this change.

`toLower`/`toUpper` are not very fast since they operate entirely out of place. We could probably also create versions of them that do faster in-place modifications, or which skip changes if they aren't necessary, but that wasn't needed here.
I'd actually implemented a unicode-aware version of `caseInsensitiveEquals` before realising that it doesn't make a difference in this use-case and we already have an ascii-only version (which should generally be faster on ascii strings since it can assume that strings of different lengths are not equal, but unicode upper-case and lower-case characters do not necessarily have the same number of codepoints, so my unicode-aware implementation was a little slower, but still much faster than `toLower`).

### DIFF
--- a/extension/json/src/functions/table_functions/json_scan.cpp
+++ b/extension/json/src/functions/table_functions/json_scan.cpp
@@ -721,9 +721,9 @@ private:
 
 uint64_t JsonScanBindData::getFieldIdx(const std::string& fieldName) const {
     // From and to are case-insensitive for backward compatibility.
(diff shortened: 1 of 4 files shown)


ray6080 (approved):
👍

PR #5268, opened 2025-04-24
Gds function supports the node type as an input parameter

ted-wq-x:
# Description

Support query like `match (a:person {id:933}) ,(b:person {id:899}) call gds_func('projected_graph_name',a,b) return ...`

# Contributor agreement

### DIFF
--- a/src/binder/bind/bind_table_function.cpp
+++ b/src/binder/bind/bind_table_function.cpp
@@ -25,12 +25,12 @@ BoundTableScanInfo Binder::bindTableFunc(const std::string& tableFuncName,
         auto param = expressionBinder.bindExpression(childExpr);
         if (!childExpr.hasAlias()) {
             ExpressionUtil::validateExpressionType(*param,
-                {ExpressionType::LITERAL, ExpressionType::PARAMETER});
+                {ExpressionType::LITERAL, ExpressionType::PARAMETER, ExpressionType::PATTERN});
             positionalParams.push_back(param);
             positionalParamTypes.push_back(param->getDataType().copy());
         } else {
             ExpressionUtil::validateExpressionType(*param,
-                {ExpressionType::LITERAL, ExpressionType::PARAMETER});
+                {ExpressionType::LITERAL, ExpressionType::PARAMETER, ExpressionType::PATTERN});
             if (param->expressionType == ExpressionType::LITERAL) {
                 auto literalExpr = param->constPtrCast<LiteralExpression>();
                 optionalParams.emplace(childExpr.getAlias(), literalExpr->getValue());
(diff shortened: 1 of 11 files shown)


ted-wq-x:
Anyone Can review this pr?

andyfengHKU:
The semantic of this PR seems to conflict with cypher semantic.

In the context of Cypher semantic, `CALL ..` and `MATCH` are both `Reading Clause`. Consider the following abstract cypher statement
```
Reading clause 1
Reading clause 2
```
The semantic should be for every output of `Reading clause 1`, execute `Reading clause 2`. So they form a cartesian product (with possible join conditions).

In your case, as I understand, Reading clause 1 (`MATCH` part) result serves as the input of `gds_func()` and we only want to run `gds_func()` once. If so, I think doing something like `gds_func(..., 'MATCH (a:person {id:933})')` could  be more clear?

Also I wanna understand which algorithm requires additional nodes input?

ted-wq-x:
I referenced the syntax of Neo4j [Dijkstra Source-Target Shortest Path](https://neo4j.com/docs/graph-data-science/current/algorithms/dijkstra-source-target/) .

Refer to neo4j execution plan :
image

In this pr, i implemented the Cartesian product as a child of TableFunctionCall,following Neo4j's approach.

We indeed need the gds funciton to be executable multiple times. This can be achieved by the gds fucntion itself.However,handing input parameters within the gds framework might be a better approach.

Algorithm like `Source-Target Shortest Path` and `K-Nearest Neighbors` need this.

andyfengHKU:
I see. I agree for pair-wise path computation it make sense. But unifying it with variable-length queries and shortest-path queries is a more elegant solution. What we have in master now are
```
MATCH (a)-[e*]->(b)
MATCH (a)-[e* SHORTEST]->(b)
MATCH (a)-[e* ALL SHORTEST]->(b)
MATCH (a)-[e* WSHORTEST(weight)]->(b)
MATCH (a)-[e* ALL WSHORTEST(weight)]->(b)
```
It would be nice if the Dijkstra is added as
```
MATCH (a)-[e* Dijkstra(weight)]->(b)
```

And then there is an optimizer change, if both `a` and `b` has cardinality 1 (or very small cardinality), we should take cartesian product of `a` and `b` and then run the algorithm. If one side has small cardinality and the other side has a large cardinlity we should go in one direction. This optimization is relevant to all path-finding algorithms. 

This is what I have in mind for path computation in the long term. Does this make sense?


For KNN, I think Neo4j doesn't take input, instead it computes all pair wise similarities. So if only path finding algorithms requires this, let's limit the change to path-finding for now.

Screenshot 2025-04-30 at 15 57 16

ted-wq-x:
My eyes are blurry 🙄 , [Breadth First Search](https://neo4j.com/docs/graph-data-science/current/algorithms/bfs/).

Some use cases require a complex macth like `match (a)-[f1]->(b)-[f2]->(c) call gds(a,c) ...` ,this format give use more flexibility.

PR #5292, opened 2025-04-28
QUERY_VECTOR_INDEX + MATCH bind error

ted-wq-x:
# Description

Execution query like `call QUERY_VECTOR_INDEX(...) yield ... match ... return ...` error.

Caused by useInternalCatalogEntry_ not being reset to false due to the vertex index catalog table being found during binding

Fixes # (issue)

# Contributor agreement


ted-wq-x:
Done

ray6080 (approved):
LGTM. Thanks!

PR #5300, opened 2025-04-29
Flush + populate warnings after COPY FROM subquery

royi-luo:
# Description

When we COPY FROM subquery the source isn't a table function so we don't have a `finalizeFunc` that we can use to flush the warning context for the current query so the warnings currently aren't flushed. This PR adds a call at the end of node/rel batch insert's `finalizeInternal` to flush/populate any warnings that weren't handled by the source operator.

# Contributor agreement


ray6080 (approved):
Nice catch!

PR #5301, opened 2025-04-29
Add GDS benchmark

ray6080:
# Description

Introduce benchmark query set for GDS.
Added datasets `soc-LiveJournal1`, `datagen-sf10k`, and `graph500-27`. Queries are borrowed from #4974.

Introduced `-PRERUN` and `-POSTRUN` in benchmark file.

These have been dry runned locally.


andyfengHKU (changes requested):
The benchmark test file needs to be rewritten.

Also I have a high level question, can we allow each test file contains multiple statements instead of introducing pre/post run?

benchmark/queries/soc-livejournal/gds/louvain.benchmark
```diff
@@ -0,0 +1,15 @@
...
+0|937000
+37|556760
+87|507375
+225|206781
+199|199471
+317|187543
+50|139826
+233|130122
+78|99646
+190|92000
```
sdht0, line 14: This result can be different across `COPY FROM` runs from the same csv.

Short term, we'd probably have to exclude Louvain if it gives too much trouble.

Long term, either the CSV importer needs to be made deterministic across N threads, or the GDS implementation needs to use node.id instead of internal id.

The former is desirable I think.

ray6080: We can disable result check for louvain for now.

andyfengHKU (approved):

PR #5353, opened 2025-05-06
Truncate free pages at the end of the file in finalizeCheckpoint

royi-luo:
# Description

Can slightly reduce the size of the data.kz file + prevents cases such as having 5 free pages at the end of the file and requesting a chunk of 7 pages which would previously cause unnecessary fragmentation.

When incrementally copying from epinions split over 50 partitions, this reduces the size of data.kz from 16MB to 14MB.

# Contributor agreement

### DIFF
--- a/src/storage/free_space_manager.cpp
+++ b/src/storage/free_space_manager.cpp
@@ -16,7 +16,7 @@ static FreeSpaceManager::sorted_free_list_t& getFreeList(
     return freeLists[level];
 }
 
-FreeSpaceManager::FreeSpaceManager() : freeLists{}, numEntries(0){};
+FreeSpaceManager::FreeSpaceManager() : freeLists{}, numEntries(0) {};
 
 common::idx_t FreeSpaceManager::getLevel(common::page_idx_t numPages) {
     // level is exponent of largest power of 2 that is <= numPages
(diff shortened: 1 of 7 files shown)


ray6080 (approved):
LGTM. @benjaminwinger can you take another look?

src/storage/free_space_manager.cpp
```diff
@@ -169,7 +169,15 @@ void FreeSpaceManager::mergePageRanges(free_list_t newInitialEntries) {
             prevEntry = entry;
         }
     }
-    addFreePages(prevEntry);
+    handleLastPageRange(prevEntry, fileHandle);
+}
+
+void FreeSpaceManager::handleLastPageRange(PageRange pageRange, FileHandle* fileHandle) {
+    if (pageRange.startPageIdx + pageRange.numPages == fileHandle->getNumPages()) {
+        fileHandle->removePageIdxAndTruncateIfNecessary(pageRange.startPageIdx);
```
benjaminwinger, line 177: It's not really a problem with this code, but `removePageIdxAndTruncateIfNecessary` isn't actually truncating the file, just the FileHandle, so this could still lead to unused pages if the database is closed and re-loaded before the truncated pages are re-added.

Having it truncate the file causes `CopyTest.NodeInsertBMExceptionDuringCheckpointRecovery` to fail (I think it's trying to update a page past the end of the file when recovering).

benjaminwinger (approved):

Issue #5274, opened 2025-04-24
Feature: CREATE NODE TABLE ... AS


ray6080:
### API

Other

### Description

This is a Cypher feature to minify table creation efforts by users.

The syntax would be `CREATE NODE TABLE ... AS subquery`, similar to https://www.postgresql.org/docs/current/sql-createtableas.html. Kuzu can sniff column names and types automatically.

To be discussed:
- this should be limited to node tables only, as we cannot infer FROM and TO table pairs for rel tables.
- for node tables, we're not able to infer primary key columns, thus we can choose to always create a SERIAL column as the primary key.

prrao87:
I think this is a great feature for node tables (for rel tables it's anyway better to be explicit and ask users to create the table properties based on what they are looking for).

From a documentation perspective, we could describe intent behind the feature as follows:

- Sometimes you may want to run a `MATCH` query to isolate a subgraph, and persist select nodes and properties from that subgraph (not all properties may be selected)
- Since it's not possible to know beforehand whether the user asked for the primary key property in the subquery result, we cannot infer the primary key column in the new table
- The new node table will have a `SERIAL` primary key column and only the selected properties from the subquery result

PR #5354, opened 2025-05-06
Add `CREATE NODE TABLE ... AS` syntax

05st:
For https://github.com/kuzudb/kuzu/issues/5274

Note(s):
The primary key column is set to the first column for now, but we probably want to make it configurable.
An empty vector is passed as the parsing options for the BoundCopyFromInfo at the moment.


scripts/antlr4/Cypher.g4
```diff
@@ -327,6 +328,9 @@ kU_IfNotExists
 kU_CreateNodeTable
     : CREATE SP NODE SP TABLE SP (kU_IfNotExists SP)? oC_SchemaName SP? '(' SP? kU_PropertyDefinitions SP? ( ',' SP? kU_CreateNodeConstraint )? SP? ')' ;
 
+kU_CreateNodeTableAs
```
andyfengHKU, line 331: Let's merge this with `kU_CreateNodeTable`, something like

```
CREATE SP NODE SP TABLE SP (kU_IfNotExists SP)? oC_SchemaName SP? ( ... | oC_Query )
```

src/binder/bind/bind_ddl.cpp
```diff
@@ -276,6 +277,54 @@ std::unique_ptr<BoundStatement> Binder::bindCreateTable(const Statement& stateme
...
+        propertyDefinitions.emplace_back(
+            ColumnDefinition(std::string(columnNames[i]), columnTypes[i].copy()));
+    }
+
+    auto parsingOptions = options_t{}; // temp
+    auto boundSource =
+        bindQueryScanSource(*createTable->getSource(), parsingOptions, columnNames, columnTypes);
+    auto warningDataExprs = boundSource->getWarningColumns();
+
+    expression_vector columns;
```
andyfengHKU, line 300: Can we reuse the logic in bindCopyNodeFrom? This logic seems duplicated.

src/planner/plan/append_simple.cpp
```diff
@@ -39,7 +42,37 @@ void Planner::appendCreateTable(const BoundStatement& statement, LogicalPlan& pl
     auto info = createTable.getInfo();
     auto op = make_shared<LogicalCreateTable>(info->copy(),
         statement.getStatementResult()->getSingleColumnExpr());
-    plan.setLastOperator(std::move(op));
+
+    // If it is CREATE NODE TABLE AS, then copy as well
+    if (createTable.hasCopyInfo()) {
```
andyfengHKU, line 47: discuss with @ray6080 on this comment.

I don't think putting a plan together at logic level is easy/elegant enough. I would do the following.

1. generate an inner logical plan for inner query
2. save the inner logical plan inside LogicalCreateTable
3. make sure we optimize the inner plan through optimizer (you can open an issue and leave this to me)
4. in mapper, create inner physical plan and physical create table. Put them together (with whatever strategy that is easier, dummySink or binary operator)

andyfengHKU (changes requested):

src/include/binder/ddl/bound_create_table.h
```diff
@@ -12,11 +12,19 @@ class BoundCreateTable final : public BoundStatement {
...
               BoundStatementResult::createSingleStringColumnResult()},
           info{std::move(info)} {}
+    
+    BoundCreateTable(BoundCreateTableInfo info, BoundCopyFromInfo copyInfo)
+        : BoundStatement{common::StatementType::CREATE_TABLE,
+              BoundStatementResult::createSingleStringColumnResult()},
+          info{std::move(info)}, copyInfo{std::move(copyInfo)} {}
 
     inline const BoundCreateTableInfo* getInfo() const { return &info; }
+    inline bool hasCopyInfo() const { return copyInfo.has_value(); }
```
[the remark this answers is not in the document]
05st: Fixed with https://github.com/kuzudb/kuzu/pull/5354/commits/f6db9139d56e455484f1ade336ed97805ec3013a

src/planner/plan/append_simple.cpp
```diff
@@ -39,7 +42,37 @@ void Planner::appendCreateTable(const BoundStatement& statement, LogicalPlan& pl
...
+        auto dummySink = std::make_shared<LogicalDummySink>(std::move(op));
+        auto& querySource = createTable.getCopyInfo()->source->constCast<BoundQueryScanSource>();
+        auto queryPlan = getBestPlan(planQuery(*querySource.statement));
```
ray6080, line 50: Can you reuse `planCopyFrom`? I think that would make the code simpler.

src/planner/plan/append_simple.cpp
ray6080: @05st can you double check if the subquery plan goes through optimizer? you can open an issue for now if 

Let's discuss this when you're online. @andyfengHKU in my understanding, I think @05st should just reuse what `planCopyFrom`, and whatever we support with `planCopyFrom` should also be supported here in terms of optimizer. I don't feel "save the inner logical plan inside LogicalCreateTable" is easier, isn't that make the subplan invisible from plan tree? How would that help with optimizer?

ray6080 (commented):
I'd like to see some e2e test cases added:
- Test LOAD FROM as subquery: on top of the tinysnb dataset, you can transform all CREATE NODE TABLE and COPY statements into `CREATE NODE TABLE ... AS (LOAD FROM ... RETURN *)`. And run Cypher queries to match all nodes and relationships to make sure the created tables are correct.
- Test MATCH clauses as subquery: test the cases where the subquery are MATCH clauses. `CREATE NODE TABLE ... AS (MATCH ...)`
- Test error on duplicated primary keys: we should error when the first column has duplicated values.
- Test error on empty output from subquery: test the case where the subquery doesn't have returned columns. We should error.

src/include/binder/copy/bound_copy_from.h
```diff
@@ -37,9 +39,18 @@ struct KUZU_API BoundCopyFromInfo {
         std::unique_ptr<BoundBaseScanSource> source, std::shared_ptr<Expression> offset,
         expression_vector columnExprs, std::vector<common::ColumnEvaluateType> columnEvaluateTypes,
         std::unique_ptr<ExtraBoundCopyFromInfo> extraInfo)
-        : tableEntry{tableEntry}, source{std::move(source)}, offset{std::move(offset)},
+        : tableEntry{tableEntry}, tableName{tableEntry->getName()}, tableType{tableEntry->getTableType()}, source{std::move(source)}, offset{std::move(offset)},
+          columnExprs{std::move(columnExprs)}, columnEvaluateTypes{std::move(columnEvaluateTypes)},
+          extraInfo{std::move(extraInfo)} {}
+
+    BoundCopyFromInfo(std::string tableName,
```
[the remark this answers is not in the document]
05st: (from discord)
If we want to keep just one constructor I think it will have to be the one that takes the table entry (since other code like in create_hnsw_index.cpp relies on it).
Then, I'll have to add some logic in the constructor body to handle nullptr being passed as the table entry + set the table name / table type fields manually at the location(s) the object is constructed.

src/planner/plan/append_simple.cpp
05st: Done with https://github.com/kuzudb/kuzu/pull/5354/commits/89494098143e9a31c38af6915244e43d6231c1d5

src/include/binder/binder.h
```diff
@@ -126,6 +128,13 @@ class Binder {
         const std::string& tableName, const std::string& propertyName);
 
     /*** bind copy ***/
+    static std::pair<common::ColumnEvaluateType, std::shared_ptr<Expression>> matchColumnExpression(
```
andyfengHKU, line 131: Try to hide this in cpp file.

src/binder/binder.cpp
```diff
@@ -26,7 +27,11 @@ std::unique_ptr<BoundStatement> Binder::bind(const Statement& statement) {
     std::unique_ptr<BoundStatement> boundStatement;
     switch (statement.getStatementType()) {
     case StatementType::CREATE_TABLE: {
-        boundStatement = bindCreateTable(statement);
+        if (statement.constPtrCast<CreateTable>()->getSource()) {
+            boundStatement = bindCreateTableAs(statement);
+        } else {
+            boundStatement = bindCreateTable(statement);
```
andyfengHKU, line 33: I would move this check in bindCreateTable so inside it check if you have a source and need to trigger bindCreateTableAs

src/include/binder/binder.h
```diff
@@ -6,6 +6,7 @@
 #include "binder/query/query_graph.h"
 #include "catalog/catalog_entry/table_catalog_entry.h"
 #include "common/copier_config/file_scan_info.h"
+#include "common/enums/column_evaluate_type.h"
```
andyfengHKU, line 9: Let's try to maintain smaller header include as it will propagate. Double check if this is needed.

test/test_files/ddl/ddl.test
```diff
@@ -567,3 +567,35 @@ Binder exception: Expression UPPER(abc) has data type STRING but expected INT64.
...
+---- 2
+0|Alice
+2|Bob
+
+-CASE CreateTableAsDuplicateIntIDs
+-STATEMENT CREATE NODE TABLE Foo AS (LOAD FROM "${KUZU_ROOT_DIRECTORY}/dataset/copy-fault-tests/duplicate-ids/vPerson.csv" RETURN *);
+---- error
+Copy exception: Found duplicated primary key value 10, which violates the uniqueness constraint of the primary key column.
+
+-CASE CreateTableAsDuplicateStringIDs
```
andyfengHKU, line 590: This test is not necessary.

test/test_files/ddl/ddl.test
```diff
@@ -567,3 +567,35 @@ Binder exception: Expression UPPER(abc) has data type STRING but expected INT64.
...
+-STATEMENT CREATE NODE TABLE Foo AS (LOAD FROM "${KUZU_ROOT_DIRECTORY}/dataset/copy-fault-tests/duplicate-ids/vOrg.csv" RETURN *);
+---- error
+Copy exception: Found duplicated primary key value 10, which violates the uniqueness constraint of the primary key column.
+
+-CASE CreateTableAsEmptyReturn
```
andyfengHKU, line 595: Ditto test something like

```
CREATE NODE TABLE Bar AS (MATCH (a) WHERE false RETURN a.*);
```

src/planner/plan/append_simple.cpp
```diff
@@ -40,6 +46,26 @@ void Planner::appendCreateTable(const BoundStatement& statement, LogicalPlan& pl
     auto op = make_shared<LogicalCreateTable>(info->copy(),
         statement.getStatementResult()->getSingleColumnExpr());
     plan.setLastOperator(std::move(op));
+    // If it is CREATE NODE TABLE AS, then copy as well
+    if (createTable.hasCopyInfo()) {
+        const auto copyInfo = createTable.getCopyInfo();
+        const auto catalog = clientContext->getCatalog();
+        const auto transaction = clientContext->getTransaction();
+        if (catalog->containsTable(transaction, info->tableName)) {
```
andyfengHKU, line 54: This is not the right place to handle NOT EXISTS. Remove and open an issue about supporting CREATE TABLE AS with IF (NOT) EXISTS

scripts/antlr4/Cypher.g4
```diff
@@ -325,7 +325,7 @@ kU_IfNotExists
     : IF SP NOT SP EXISTS ;
 
 kU_CreateNodeTable
-    : CREATE SP NODE SP TABLE SP (kU_IfNotExists SP)? oC_SchemaName SP? '(' SP? kU_PropertyDefinitions SP? ( ',' SP? kU_CreateNodeConstraint )? SP? ')' ;
+    : CREATE SP NODE SP TABLE SP (kU_IfNotExists SP)? oC_SchemaName ( SP? '(' SP? kU_PropertyDefinitions SP? ( ',' SP? kU_CreateNodeConstraint )? SP? ')' | SP AS SP? '(' SP? oC_Query SP? ')' ) ;
```
andyfengHKU, line 328: I suspect `()` is not needed in PG grammar. Can u double check? Let's stick to PG on the syntax for this.

andyfengHKU (approved):

scripts/antlr4/Cypher.g4
05st: Fixed with https://github.com/kuzudb/kuzu/pull/5354/commits/f989e4ff19b26efe614daadde7dfe645ed52ae44

src/planner/plan/append_simple.cpp
05st: Removed with https://github.com/kuzudb/kuzu/pull/5354/commits/1a58c6f3626e5511b2ea87c3b3afb6c1372c360a and opened the issue https://github.com/kuzudb/kuzu/issues/5406.

test/test_files/ddl/ddl.test
05st: Removed with https://github.com/kuzudb/kuzu/pull/5354/commits/edcd58e7151be06ae8d53481f72fcc47ec513cf4

test/test_files/ddl/ddl.test
05st: It isn't possible to test this at this moment (see https://github.com/kuzudb/kuzu/issues/5408).

src/include/binder/binder.h
05st: This include is needed now due to changes on line 130 of the same file.

src/binder/binder.cpp
05st: Done with https://github.com/kuzudb/kuzu/pull/5354/commits/59208ac94ae481ece5cd523e2005e40246f0fb0a

PR #5357, opened 2025-05-06
Reclaim allocated pages during rollback

royi-luo:
# Description

The `PageManager` now keeps track of which pages were allocated during the current transaction/checkpoint operation. On rollback, the tracked pages are marked as free so that they can be used in future transactions.

# Contributor agreement

### DIFF
--- a/src/include/storage/free_space_manager.h
+++ b/src/include/storage/free_space_manager.h
@@ -29,7 +29,7 @@ public:
 
     // These pages are not reusable until the end of the next checkpoint
     void addUncheckpointedFreePages(PageRange entry);
-    void rollbackCheckpoint();
+    void rollback();
 
     void serialize(common::Serializer& serializer) const;
     void deserialize(common::Deserializer& deSer);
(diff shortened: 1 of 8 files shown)


src/include/storage/store/column_chunk_data.h
```diff
@@ -156,7 +156,7 @@ class KUZU_API ColumnChunkData {
     virtual void append(ColumnChunkData* other, common::offset_t startPosInOtherChunk,
         uint32_t numValuesToAppend);
 
-    virtual void flush(FileHandle& dataFH);
+    virtual void flush(FileHandle& dataFH, bool reclaimOnRollback = true);
```
royi-luo, line 159: A bit of a hacky way to get around the problem that columns that are added + optimistically flushed + rolled back are already reclaimed in `NodeGroup::checkpointInMemAndOnDisk` using the mechanism that is used for reclaiming dropped columns and we don't want to double free. Let me know if you have any better ways to handle this.

ray6080:
Instead of coupling `PageManager` with the rollback of optimistic writes, I think it's better to have a specialized `OptimisticWriter` which tracks of pages optimistically written out, and we can reclaim through that. I need to think a bit more on this, but most likely I will rework this.

royi-luo:
Closing, taken over by https://github.com/kuzudb/kuzu/pull/5601

PR #5375, opened 2025-05-08
More BM exception robustness fixes

royi-luo:
# Description

When we hit + catch BM exceptions storage structure may not be in a valid state (at least until rollback completes). However other threads may still do a bit of work on the structures before they are interrupted which can lead to data corruption due to working on invalid structures.

This PR fixes two of these problems:
1. We now update the capacity after the allocation in `ColumnChunkData::resize` to avoid the case where we update the capacity but the allocation fails
2. If we hit an exception during `ChunkedNodeGroup::append` we reset the `numValues` of each column chunk to match the number of rows in the chunked group.

# Contributor agreement

### DIFF
--- a/src/storage/store/chunked_node_group.cpp
+++ b/src/storage/store/chunked_node_group.cpp
@@ -149,15 +149,15 @@ uint64_t ChunkedNodeGroup::append(const Transaction* transaction,
     KU_ASSERT(residencyState != ResidencyState::ON_DISK);
     KU_ASSERT(columnVectors.size() == chunks.size());
     const auto numRowsToAppendInChunk = std::min(numValuesToAppend, capacity - numRows);
-    for (auto i = 0u; i < columnVectors.size(); i++) {
-        const auto columnVector = columnVectors[i];
-        try {
+    try {
+        for (auto i = 0u; i < columnVectors.size(); i++) {
+            const auto columnVector = columnVectors[i];
             chunks[i]->getData().append(columnVector,
                 columnVector->state->getSelVector().slice(startRowInVectors,
                     numRowsToAppendInChunk));
-        } catch (std::exception& e) {
-            handleAppendException();
         }
+    } catch ([[maybe_unused]] std::exception& e) {
+        handleAppendException();
     }
     if (transaction->getID() != Transaction::DUMMY_TRANSACTION_ID) {
         if (!versionInfo) {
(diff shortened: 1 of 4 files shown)


raiderss:
Fantastic work on this project! This could integrate well with our FiveM resources at Eyes Store Tebex.

src/storage/store/chunked_node_group.cpp
```diff
@@ -570,5 +580,17 @@ uint64_t ChunkedNodeGroup::spillToDisk() {
     return reclaimedSpace;
 }
 
+void ChunkedNodeGroup::handleAppendException() {
+    // After an exception is thrown other threads may continue to work on this chunked group for a
```
benjaminwinger, line 584: Aren't the ChunkedNodeGroups/ColumnChunks supposed to be single-threaded only? When are we writing to them in multiple threads at once (I don't think they are thread-safe)?

royi-luo: yep they are only single threaded but the case is 
- thread 1 locks, hits exception, leaves chunk in undefined state, unlocks
- thread 2 hasn’t been intereupted yet, locks (note it does have exclusive access) but it’s reading a chunk in an undefined state.

src/storage/store/chunked_node_group.cpp
```diff
@@ -151,8 +151,13 @@ uint64_t ChunkedNodeGroup::append(const Transaction* transaction,
     const auto numRowsToAppendInChunk = std::min(numValuesToAppend, capacity - numRows);
     for (auto i = 0u; i < columnVectors.size(); i++) {
         const auto columnVector = columnVectors[i];
-        chunks[i]->getData().append(columnVector,
-            columnVector->state->getSelVector().slice(startRowInVectors, numRowsToAppendInChunk));
+        try {
```
ray6080, line 154: Should we move the `try ... catch ...` to the outer level?

src/storage/store/chunked_node_group.cpp
```diff
@@ -194,8 +199,13 @@ offset_t ChunkedNodeGroup::append(const Transaction* transaction,
     for (auto i = 0u; i < columnIDs.size(); i++) {
         auto columnID = columnIDs[i];
         KU_ASSERT(columnID < chunks.size());
-        chunks[columnID]->getData().append(&other[i]->getData(), offsetInOtherNodeGroup,
-            numToAppendInChunkedGroup);
+
+        try {
```
ray6080, line 203: Ditto here. Can we move `try ... catch ...` out?

ray6080 (approved):
Looks good to me. @benjaminwinger can you take another look?

PR #5380, opened 2025-05-09
shell: add support for ENTER from middle of a query

sdht0:
Currently, when editing in the multiline mode, after editing a query, the cursor needs to be manually navigated to the end of the query before pressing ENTER to execute the query, otherwise the CLI just appends a new line.

This PR adds the ability to execute a query even if the cursor is in the middle of a query by pressing `SHIFT+ENTER` or `CTRL+J`. This is a minor quality of life improvement as compared to `CTRL+E -> ENTER`.

Tested on Linux and macOS. Needs a test on Windows.

# Contributor agreement


ray6080 (approved):
Can you see if we can add a test case to our shell tests?

Issue #5372, opened 2025-05-08
shell: piping into queries without a `;` leads to a silent exit


sdht0:
```
$ echo "return 1" | kuzu`
Opening the database under in-memory mode.
Enter ":help" for usage hints.

$ echo "return 1;" | kuzu`
Opening the database under in-memory mode.
Enter ":help" for usage hints.
┌───────┐
│ 1     │
│ INT64 │
├───────┤
│ 1     │
└───────┘
(1 tuple)
(1 column)
Time: 7.15ms (compiling), 0.66ms (executing)
```

The first command should show an error about missing semicolon OR should run the query as if `EOF` == `;`

PR #5392, opened 2025-05-12
DEV-5372 cli piping into queries leads to silent exit

tgahunia05:
# Description
echo "return 1" | ./kuzu`

Piping the query above leads directly to an exit of the shell instance with no error message as to what went wrong. The only issue was a missing semicolon. i.e 

echo "return 1;" | ./kuzu`

would run as intended. 

The suggested behaviour was to treat EOF as a ';' and run the query. 

This fix adds a semi colon to the end of input if there is not one already there. 

image

Fixes # (issue)
#5372 

# Contributor agreement

### DIFF
--- a/tools/shell/embedded_shell.cpp
+++ b/tools/shell/embedded_shell.cpp
@@ -588,7 +588,7 @@ void EmbeddedShell::run() {
     int numCtrlC = 0;
     continueLine = false;
     currLine = "";
-    bool reQuery = false;
+    std::string lineStr;
 
 #ifndef _WIN32
     termios raw{};
@@ -611,10 +611,17 @@ void EmbeddedShell::run() {
     SetConsoleOutputCP(CP_UTF8);
 #endif
 
-    while (
-        (line = linenoise(continueLine ? ALTPROMPT : PROMPT, CONPROMPT, SCONPROMPT)) != nullptr) {
-        reQuery = false;
-        auto lineStr = std::string(line);
+    while (true)
+    {
+        line = linenoise(continueLine ? ALTPROMPT : PROMPT, CONPROMPT, SCONPROMPT);
+
+        if (line == nullptr && !continueLine)
+            break;
+        else if (line == nullptr)
+            lineStr = ";";
+        else
+            lineStr = std::string(line);
+
         lineStr = lineStr.erase(lineStr.find_last_not_of(" \t\n\r\f\v") + 1);
         if (!lineStr.empty() && lineStr[0] == ctrl_c) {
             if (!continueLine && lineStr[1] == '\0') {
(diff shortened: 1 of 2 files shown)


tools/shell/embedded_shell.cpp
```diff
@@ -652,6 +655,15 @@ void EmbeddedShell::run() {
         linenoiseHistorySave(path_to_history);
         free(line);
     }
+    if (reQuery) {
+        auto queryResults = processInput(";");
+        if (!queryResults.empty())
+            for (auto& queryResult : queryResults)
+                if (queryResult->isSuccess()) {
+                    printInterrupted = false;
+                    printExecutionResult(*queryResult);
```
sdht0, line 664: As we discussed, try to see if we can avoid this repetition. For instance, this is missing the `else` case?
- Detect EOF before line 633 and append `;` there itself
- Refactor out this fragment in a function.

tools/shell/embedded_shell.cpp
```diff
@@ -588,6 +588,7 @@ void EmbeddedShell::run() {
     int numCtrlC = 0;
     continueLine = false;
     currLine = "";
+    bool reQuery = false;
```
sdht0, line 591: reQuery -> `isIncompleteQuery`

sdht0 (changes requested):

tools/shell/embedded_shell.cpp
tgahunia05: Got rid of reQuery

tools/shell/embedded_shell.cpp
```diff
@@ -610,9 +611,16 @@ void EmbeddedShell::run() {
...
-        auto lineStr = std::string(line);
+    while (true) {
+        line = linenoise(continueLine ? ALTPROMPT : PROMPT, CONPROMPT, SCONPROMPT);
+
+        if (line == nullptr && !continueLine)
+            break;
+        else if (line == nullptr)
+            lineStr = ";";
+        else
+            lineStr = std::string(line);
```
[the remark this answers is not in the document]
tgahunia05: {} have been added

tools/shell/embedded_shell.cpp
```diff
@@ -610,9 +611,16 @@ void EmbeddedShell::run() {
     SetConsoleOutputCP(CP_UTF8);
 #endif
 
-    while (
-        (line = linenoise(continueLine ? ALTPROMPT : PROMPT, CONPROMPT, SCONPROMPT)) != nullptr) {
-        auto lineStr = std::string(line);
+    while (true) {
+        line = linenoise(continueLine ? ALTPROMPT : PROMPT, CONPROMPT, SCONPROMPT);
+
+        if (line == nullptr && !continueLine)
```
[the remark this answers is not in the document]
tgahunia05: I've added some comments explaining the new conditionals and the use of the continueLine boolean to mark when an additional ';' is needed.

PR #5403, opened 2025-05-14
Unify semi mask used in GDS, HNSW index & Recursive join

andyfengHKU:
# Description

Since we are going to support input node mask for GDS. This PR first unify the existing semi mask that we use to support filtered vector index, algorithm on filtered graph and filtered recursive joins.

It also refactors HNSW binding logic by removing `BoundQueryHNSWIndexInput`

Fixes # (issue)

# Contributor agreement


ray6080 (approved):
Looks good!

PR #5418, opened 2025-05-15
Update Java API Exception Handling

### LINKED ISSUE #5400: Improve Java API's exception handling
Right now we didn't explicitly `try ... catch ...` exceptions thrown from cpp side in our Java API. This can lead to unexpected crashes if users don't follow steps exactly or manually try catch as expected.

To improve the usability with exception handling, we should explicitly `try ... catch ...` exceptions thrown from cpp side and convert them into Java Exceptions.

tgahunia05:
# Description

Wrap all the functions with try...catch... and throw Java exception when a C++ exception is caught.
Replace ObjectRefDestroyedException with RuntimeException

Fixes #5400 

# Contributor agreement


tools/java_api/src/jni/kuzu_java.cpp
mewim, line 1: The format of this file needs a bit improvement. There is a newline after some of the `try{` block.
For example, line 179 and line 210. These should be removed.

mewim (approved):

PR #5427, opened 2025-05-18
feat(nodejs_api): add ES module support

jasonviipers:
# Description

This PR adds ES module compatibility and TypeScript support to the Node.js API of Kùzu. These enhancements improve module resolution and provide better integration with modern JavaScript and TypeScript environments.

## Changes
- Introduced `.mjs` files to enable ES module support.
- Ensured the build process outputs the `.d.ts` files.
- Updated `package.json` to include:
  - `module` field for ESM entry point.
  - `exports` field for consistent module resolution across environments.
- Updated build and install scripts to handle `.mjs` and `.d.ts` files appropriately.

## Contributor agreement


tools/nodejs_api/build.js
```diff
@@ -12,3 +12,26 @@ execSync(`make nodejs NUM_THREADS=${THREADS}`, {
...
+  return file.endsWith(".js") || file.endsWith(".mjs") || file.endsWith(".d.ts");
+});
+
+console.log("Files to copy:");
+for (const file of jsFiles) {
+  console.log("  " + file);
+  fs.copyFileSync(path.join(jsSourceDir, file), path.join(buildDir, file));
+}
+
+console.log("TypeScript definitions and ES modules copied successfully.");
```
mewim, line 37: I do not think we should add these here. The build script is only for building the native module. I do not think any copying should be done within it. Instead, it should be handled in the install script.

mewim (changes requested):
Thanks for your contribution. Please take a look at my PR comments.

tools/nodejs_api/build.js
mewim: Actually, I see what you were trying to do. The JS files are added by the CMake build system:
https://github.com/kuzudb/kuzu/blob/3cb81d47818a9a90e2e1283b8a0177684acc843a/tools/nodejs_api/CMakeLists.txt#L41-L42

You can edit this file to also include ts and mjs files:
```
file(GLOB JS_SOURCE_FILES 
    "${CMAKE_CURRENT_SOURCE_DIR}/src_js/*.js"
    "${CMAKE_CURRENT_SOURCE_DIR}/src_js/*.ts"
    "${CMAKE_CURRENT_SOURCE_DIR}/src_js/*.mjs"
)
```

jasonviipers: Thank you for your feedback. I understand your point about handling the copying of TypeScript definitions and ES modules in the install script rather than the build script.

I've reviewed the CMake build system section you mentioned. It makes sense to integrate the copying process there. I'll make the necessary adjustments to the CMakeLists.txt file to include .js, .ts, and .mjs files as you suggested.

I'll update the pull request with these changes shortly.

jasonviipers: > I do not think we should add these here. The build script is only for building the native module. I do not think any copying should be done within it. Instead, it should be handled in the install script.

Thank you for your feedback. I understand your point about handling the copying of TypeScript definitions and ES modules in the install script rather than the build script.

I've reviewed the CMake build system section you mentioned. It makes sense to integrate the copying process there. I'll make the necessary adjustments to the CMakeLists.txt file to include .js, .ts, and .mjs files as you suggested.

I'll update the pull request with these changes shortly.

tools/nodejs_api/src_js/index.mjs
```diff
@@ -0,0 +1,9 @@
+import kuzu from "./index.js";
+
+// Re-export everything from the CommonJS module
+export const Database = kuzu.Database;
+export const Connection = kuzu.Connection;
+export const PreparedStatement = kuzu.PreparedStatement;
+export const QueryResult = kuzu.QueryResult;
+
+export default kuzu;
```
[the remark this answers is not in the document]
jasonviipers: > Newline here

Can I rename the type definition from the index.d.ts to a kuzu.d.ts instead?

mewim (approved):

PR #5429, opened 2025-05-18
Add noop operator and fix planning for create table as

### LINKED ISSUE #5408: Rework CREATE TABLE AS as binary plan
Currently, the optimizer seems to only go down the leftmost subtrees (i.e. child at index 0) for logical operators.

For https://github.com/kuzudb/kuzu/issues/5274 we would like to ideally construct a binary logical plan so the subtree responsible for creating the table is independent from the subtree responsible for copying. This will fix the issue demonstrated below:

```
kuzu> create node table if not exists Foo as load from "test.csv" return *;
kuzu> explain logical create node table Bar as match (a:Foo) where false return a.*;
```

Currently, since we construct a linear plan, the above will result in a logical plan with an `EmptyResult` operator, but the optimizer truncates the children, which includes the `CreateTable` operator.

andyfengHKU:
# Description

Fixes #5408

Introduce a LogicalNoop operator as the parent of CreateTable and Copy pipeline. At physical level, it is mapped as an empty TableScan. I don't observe the necessity of a physical Noop yet.

This PR includes some refactor of DummySink (I thought the Noop will be a sink operator but it turns out not after implementation)

The PR also refactors the constructor of BatchInsertInfo. The previous one was way too complicated.

# Contributor agreement


ray6080 (approved):
This should also fix https://github.com/kuzudb/kuzu/issues/5408. Can you add a test case from 5408?

andyfengHKU:
> This should also fix #5408. Can you add a test case from 5408?

I don't think so. To handle IF NOT EXISTS, we need to let Copy pipeline understand it should not be executed if CREATE is not executed successfully.

PR #5431, opened 2025-05-19
Single file database

ray6080:
# Description

Merge `catalog.kz` and `metadata.kz` into `data.kz`. Note that this PR doesn't remove the directory, which will be done in a following PR.

The PR introduced a `MetadataWriter` which caches all pages in memory until `flush`. This simplifies the write of catalog and storage metadata with a little bit memory overheads. A better design might be to chain pages of catalog and metadata so we can better make use of free space in the data file.
Another optimization is we shouldn't rewrite catalog and metadata if there aren't changes to them.

TODOs to be addressed in following PRs:
- [ ] Move files out of the db directory.
- [ ] Reclaim space for metadata and catalog pages.
- [ ] Reclaim space for index pages.
- [ ] Storage info should also show page usage for indexes, catalog, and metadata.


src/include/common/serializer/buffered_file.h
```diff
@@ -41,6 +42,13 @@ class BufferedFileReader final : public Reader {
 public:
     explicit BufferedFileReader(std::unique_ptr<FileInfo> fileInfo);
 
+    // Note: this function is resting the next file offset to read. Make sure the buffer is empty.
```
benjaminwinger, line 45: I think the "Make sure the buffer is empty." message can be omitted here since it's resetting the bufferOffset and bufferSize. The other function wasn't supposed to be called when there's something in the buffer, but this one wouldn't have the same issue.

benjaminwinger (approved):

PR #5432, opened 2025-05-19
Remove unnecessary copy of ResultSetDescriptor

andyfengHKU:
# Description

Move ResultSetDescriptor output sink constructor and copy interface. 

First, it should not be needed in the copied pipeline if we always obtain resultSet from main thread. Second we should try to shorten our operator constructor. Some of them is getting way too long to read.

Fixes # (issue)

# Contributor agreement

### DIFF
--- a/src/processor/operator/sink.cpp
+++ b/src/processor/operator/sink.cpp
@@ -1,0 +1,15 @@
+#include "processor/operator/sink.h"
+
+namespace kuzu {
+namespace processor {
+
+std::unique_ptr<ResultSet> Sink::getResultSet(storage::MemoryManager* memoryManager) {
+    if (resultSetDescriptor == nullptr) {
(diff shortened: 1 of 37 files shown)


src/processor/operator/sink.cpp
```diff
@@ -0,0 +1,15 @@
...
+std::unique_ptr<ResultSet> Sink::getResultSet(storage::MemoryManager* memoryManager) {
+    if (resultSetDescriptor == nullptr) {
+        // Some pipeline does not need a resultSet, e.g. OrderByMerge
+        return std::unique_ptr<ResultSet>();
+    }
+    return std::make_unique<ResultSet>(resultSetDescriptor.get(), memoryManager);
+}
+
+} // namespace processor
+} // namespace kuzu
```
ray6080, line 15: New line here.

ray6080 (approved):

PR #5439, opened 2025-05-20
Remove the early checkpoint for empty primary key indices

benjaminwinger:
It was necessary to refactor the `DiskArrayCollection` so that it doesn't need to know its header page index until checkpointing begins. 

The way the disk array collection handles this is more complicated than is really necessary, since it's designed to store a variable number of disk arrays, when in reality we always just need 512 (256 hash indexes with one disk array for primary slots and one for overflow slots) and they are all created at once. In its current form I think it would ideally be more self-contained and do something more like what the PrimaryKeyIndex does, serializing and deserializing the location of its header page, but I think that would actually be the wrong direction to take things (unless we actually have a use for a variable number of disk arrays). Otherwise it would probably be a better idea to make it unique to the PrimaryKeyIndex, work with a fixed number of disk arrays and incorporate the HashIndexHeaders.


ray6080 (approved):
LGTM. Thanks!

benjaminwinger:
It looks like this has also slightly increased the hash index BM requirements (I'm not sure why checkpointing an empty hash index reduces the amount of memory used though).

Issue #5363, opened 2025-05-07
test: export from prev release and import into a new release


sdht0:
- [ ] Generate a random dataset with multiple Node and Rel tables that covers all the types that Kuzu supports: https://docs.kuzudb.com/cypher/data-types/
- [ ] Write queries that operate on all the different types atleast once, possibly using as many functions as feasible from https://docs.kuzudb.com/cypher/expressions/
- [ ]  Include extensions such as FTS and Vector

- Generate dataset
- `COPY FROM` into vN-1
- `EXPORT DATABASE`
- `IMPORT DATABASE` into vN
- Run test queries

- [ ] In the long term, we can write the queries in each of the client API we support in a single file, which can become a single-file reference test for the APIs

PR #5449, opened 2025-05-23
Export from prev release and import into a new release

tgahunia05:
# Description

This PR introduces support for running end-to-end (E2E) tests using IMPORT DATABASE instead of CSV imports, advancing progress on [issue #5363](https://github.com/kuzudb/kuzu/issues/5363).

Adds logic to the test runner to detect the E2E_IMPORT_TEST_DIR environment variable.
When set (e.g. E2E_IMPORT_TEST_DIR=dataset/tmp/0.9.0_exported_dbs/), the test runner imports a pre-exported Kùzu database via IMPORT DATABASE statements (in this case from version 0.9.0).

Example 

`E2E_IMPORT_DB_DIR =dataset/tmp/0.9.0_exported_dbs/ {e2e executable} function/array.test`

# Contributor agreement


dataset/tmp/0.9.0_exported_dbs/tinysnb/workAt.csv
```diff
@@ -0,0 +1,4 @@
+a.ID,b.ID,r.year,r.grading,r.rating
```
sdht0, line 1: All the above files should not be in the commit.

test/include/test_helper/test_helper.h
```diff
@@ -25,6 +25,7 @@ struct TestQueryConfig {
 class TestHelper {
 public:
     inline static std::string E2E_TEST_FILES_DIRECTORY = "test/test_files";
+    inline static std::string E2E_IMPORT_TEST_DIR = "dataset/";
```
sdht0, line 28: Let's not conflate the two import types into this single variable as it hides the intent.
Instead, let the existing code remain as it is and use `E2E_OVERRIDE_IMPORT_DIR` or equivalent, with a default empty value. That way it is clear that there is an override of the default import type.

test/graph_test/base_graph_test.cpp
```diff
@@ -96,7 +96,23 @@ void BaseGraphTest::createDBAndConn() {
...
 
 void BaseGraphTest::initGraph(const std::string& datasetDir) const {
-    if (conn) { // normal conn
+    // if we are importing from a previous kuzu version
+    // such as dataset/tmp/0.9.0_prev_exported_dbs/tinysnb
+    if (TestHelper::E2E_IMPORT_TEST_DIR != "dataset/") {
+        // falls out of the remaining conditionals
+        // i.e if conn is null we need to use the connMap
+        // as per the conn documentation
+        Connection* connection = conn ? conn.get() : (connMap.begin()->second).get();
```
sdht0, line 105: Refactor this out to above the if condition above, so we're not writing the logic again below.

test/graph_test/base_graph_test.cpp
```diff
@@ -96,7 +96,23 @@ void BaseGraphTest::createDBAndConn() {
...
+        // see migration related documentation
+        std::string query = "IMPORT DATABASE '" + datasetDir + "';";
+        std::cout << "Starting to execute query: " << query << std::endl;
```
sdht0, line 108:
```suggestion
        std::cout << "Loading database as: " << query << std::endl;
```

test/graph_test/base_graph_test.cpp
```diff
@@ -96,7 +96,23 @@ void BaseGraphTest::createDBAndConn() {
...
+        auto result = connection->query(query);
+        std::cout << "Executed query: " << query << std::endl;
```
sdht0, line 110: Can remove this I think. Unless other queries also print something similar.

test/graph_test/base_graph_test.cpp
```diff
@@ -96,7 +96,23 @@ void BaseGraphTest::createDBAndConn() {
...
+        if (!result->isSuccess()) {
+            throw Exception(stringFormat("Failed to execute statement: {}.\nError: {}", query,
```
sdht0, line 112: Check that this error message is consistent with the other error messages format.

test/graph_test/base_graph_test.cpp
```diff
@@ -96,7 +96,23 @@ void BaseGraphTest::createDBAndConn() {
...
+                result->getErrorMessage()));
+        }
+    } else if (conn) { // normal conn
         TestHelper::executeScript(datasetDir + TestHelper::SCHEMA_FILE_NAME, *conn);
         TestHelper::executeScript(datasetDir + TestHelper::COPY_FILE_NAME, *conn);
     } else {
```
sdht0, line 118: While you're at it, an nice optional change here would be to check if datasetDir is "empty" and just skip the load. Don't have to do in this PR though.

sdht0 (approved):
Nice work. LGTM.

dataset/tmp/0.9.0_exported_dbs/tinysnb/workAt.csv
sdht0: Will think about how to best add the exported dbs to CI. Let's just get the infrastructure in for now.

A nice to have would be a Python or Bash script that we can run to automatically create exported dbs for all the test datasets for any given Kuzu version.

test/graph_test/base_graph_test.cpp
tgahunia05: See void TestHelper::executeScript(const std::string& cypherScript, Connection& conn) in base_graph_test.cpp

Lines 62-68

test/graph_test/base_graph_test.cpp
tgahunia05: See void TestHelper::executeScript(const std::string& cypherScript, Connection& conn) in base_graph_test.cpp

Lines 62-68

test/runner/e2e_test.cpp
```diff
@@ -195,13 +196,23 @@ int main(int argc, char** argv) {
...
         char* env_test_dir = std::getenv("E2E_TEST_FILES_DIRECTORY");
+        char* env_import_data_dir = std::getenv("E2E_IMPORT_TEST_DIR");
         if (env_test_dir != nullptr) {
             test_dir = env_test_dir;
         } else {
             test_dir = "test/test_files";
         }
+
+        if (env_import_data_dir != nullptr) {
+            import_data_dir = env_import_data_dir;
```
sdht0, line 216: There should be a check that `env_import_data_dir` ends in `/`.

Even better would to test that directory exists.

test/graph_test/base_graph_test.cpp
```diff
@@ -96,15 +97,26 @@ void BaseGraphTest::createDBAndConn() {
 }
 
 void BaseGraphTest::initGraph(const std::string& datasetDir) const {
-    if (conn) { // normal conn
-        TestHelper::executeScript(datasetDir + TestHelper::SCHEMA_FILE_NAME, *conn);
-        TestHelper::executeScript(datasetDir + TestHelper::COPY_FILE_NAME, *conn);
+    // if conn is null we need to use the connMap
+    // as per the conn documentation
```
sdht0, line 101: Comments should start with a capital letter and end with a `.` Also use the full span before wrapping.

test/graph_test/base_graph_test.cpp
```diff
@@ -96,15 +97,26 @@ void BaseGraphTest::createDBAndConn() {
...
+    Connection* connection = conn ? conn.get() : (connMap.begin()->second).get();
+    KU_ASSERT(connection != nullptr);
```
sdht0, line 103: Use KU_ASSERT_UNCONDITIONAL

sdht0 (commented):
Some more minor feedback and then we can get this in.

PR #5461, opened 2025-05-27
Remove some dead and unnecessary code

benjaminwinger:
This is mostly just stuff that would have made some of the changes I've been working on for segmentation slightly more complicated and I figured should just be removed.

1. Remove most of the functions from NullColumn since they are duplicates of the default Column functions (or near duplicates; scan just calls scanInternal, while the superclass version also scans the null column if it exists).
2. Remove the code for handling adding new pages when writing in-place since it's unused and wasn't well documented (now we just assert that we don't write off the end of the page range).
3. `ColumnChunkData::getNumValues` no longer needs to be virtual (we're synchronizing the numValues in various places so that it should actually be correct and the nested types don't need to use the null chunk's numValues.

### DIFF
--- a/src/storage/table/column_reader_writer.cpp
+++ b/src/storage/table/column_reader_writer.cpp
@@ -119,26 +119,27 @@ public:
             writeFromVectorFunc, &vectorToWriteFrom->getNullMask());
     }
 
-    page_idx_t writeValuesToPageFromBuffer(ChunkState& state, offset_t dstOffset,
-        const uint8_t* data, const NullMask* nullChunkData, offset_t srcOffset, offset_t numValues,
+    void writeValuesToPageFromBuffer(ChunkState& state, offset_t dstOffset, const uint8_t* data,
+        const NullMask* nullChunkData, offset_t srcOffset, offset_t numValues,
         const write_values_func_t& writeFunc) override {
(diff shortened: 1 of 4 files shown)


ray6080 (approved):

src/storage/table/column_reader_writer.cpp
```diff
@@ -434,27 +434,16 @@ void ColumnReadWriter::readFromPage(const Transaction* transaction, page_idx_t p
...
-    bool insertingNewPage = false;
     if (cursor.pageIdx == INVALID_PAGE_IDX) {
         writeOp(nullptr, cursor.elemPosInPage);
-        return false;
+        return;
     }
-
-    // The implemented mechanism for inserting new pages here doesn't work if we do concurrent
-    // writes We currently don't do concurrent writes but should also actually never hit the case
-    // where we need to insert pages either
```
sdht0, line 447: Was this comment removal intentional? Seems like it still applies to the asset?

benjaminwinger: Yes, it was intentional; it is not meaningful any more since I removed the code that might have allowed this to add new pages to the end of the file (with the following code block removed it should now throw an exception when we attempt to read off the end of the file).
It still may be possible with the assert disabled, but it *should* never happen, is currently untested for that reason, and I don't think we will need it in future either (particularly if we want to support concurrent writes).

PR #5464, opened 2025-05-28
Add generic filtered search for vector index.

andyfengHKU:
This PR adds generic HNSW filtered search that allows the filter generated from arbitrary cypher query.

Fixes # (issue)

# Contributor agreement


extension/vector/src/function/query_hnsw_index.cpp
```diff
@@ -124,10 +108,43 @@ static std::unique_ptr<TableFuncBindData> bindFunc(main::ClientContext* context,
...
+        auto resultColumns = boundStatement->getStatementResult()->getColumns();
+        if (resultColumns.size() != 1) {
+            throw BinderException(stringFormat("The return clause of filter query should contain "
+                                               "exact one node expression. Found {}.",
+                ExpressionUtil::toString(resultColumns)));
+        }
+        auto resultColumn = resultColumns[0];
+        if (resultColumn->getDataType().getLogicalTypeID() != LogicalTypeID::NODE) {
+            throw BinderException(
+                stringFormat("The return clause of filter query should contain exact one node "
```
ray6080:
```suggestion
                stringFormat("The returned expression of a filter query must be of type NODE. Found type {} instead."
```

Issue #5395, opened 2025-05-13
Bug: oom when use collect


Anwenya:
### Kuzu version

0.8.2

### What operating system are you using?

windows 10 19044.1586

### What happened?

i think the `collect` func is not work normallly because of OOM

### Are there known steps to reproduce?

here is a simple scenario that can be reproduced

these models for contact between some account  `Account -> ContactWith -> Account`

this query will trigger an OOM, which is just simply grouping contacts by friend (f:Account). and i have
 40GB RAM

```cypher
MATCH (a:Account)-[cw:ContactWith]->(f:Account)
WHERE NOT (f)<-[:HasAccount]-(:Evidence)
WITH f, collect(a) as aList
RETURN count(a)
```

The total number of paths is only 200,033
```cypher
MATCH (a:Account)-[cw:ContactWith]->(f:Account)
RETURN count(a)
// map[COUNT(a._ID):200033]
```

benjaminwinger:
Thanks for raising this issue. The current implementation of the collect function allocates a 256KB buffer of memory for each group, which gives good performance for collecting a large number of values per group, but unfortunately doesn't scale well to a large number of groups with few values in each (~48GB of memory in your example). We plan to re-write it so that it scales better when there are large numbers of groups.

Anwenya:
> Thanks for raising this issue. The current implementation of the collect function allocates a 256KB buffer of memory for each group, which gives good performance for collecting a large number of values per group, but unfortunately doesn't scale well to a large number of groups with few values in each (~48GB of memory in your example). We plan to re-write it so that it scales better when there are large numbers of groups.

thanks for your reply，i got it.  is there some else way to implement grouping in kuzu? because it's necessary for my needs

andyfengHKU:
Hi @Anwenya, 

I can suggest 2 alternatives

1. Chunk your query by adding predicate
```
MATCH (a:Account)-[cw:ContactWith]->(f:Account)
WHERE f.id > 'lowerBound' AND f.id < 'upperBound'
RETURN f, collect(a) as aList
```
and run the query with difference range multiple times

2. Move the grouping logic to the application side.

Also, we should be able to fix this some time next week. By which time it will be available in our nightly build version

Anwenya:
> Hi [@Anwenya](https://github.com/Anwenya),
> 
> I can suggest 2 alternatives
> 
> 1. Chunk your query by adding predicate
> 
> ```
[...]

Issue #5398, opened 2025-05-13
Collect aggregate function scaling


benjaminwinger:
The collect function currently stores a FactorizedTable for each aggregate state (i.e. each group), which allocates at least one 256KB memory buffer to store its data (and an InMemOverflowBuffer, but that was changed recently to start with small blocks and scale upwards).

The FactorizedTable uses a fixed number of tuples per block, so it would slow down access time if we were to increase the size of the blocks as the number of values increase (though it should still be constant time using the same doubling strategy; albeit a more complicated calculation than just dividing by the number of tuples per block), but we could use a smaller block size and have the blocks be backed by the AggregateHashTable's shared InMemOverflowBuffer instead of being individually allocated which should make up for the performance cost of using smaller blocks.

We also don't strictly need the FactorizedTable: this table stores a single value per tuple with no factorization, and the copyToRowData and copyFromRowData functions work on arbitrary memory addresses. We could move the implementations of `getTuple`, `appendEmptyTuple` and the tracking of the number of flat tuples into the DataBlockCollection, and just use a DataBlockCollection in CollectState (and maybe rename DataBlockCollection).

PR #5467, opened 2025-05-29
Collect memory optimization

royi-luo:
# Description

Closes https://github.com/kuzudb/kuzu/issues/5395 and https://github.com/kuzudb/kuzu/issues/5398

Instead of storing a `FactorizedTable` in each `CollectState`, we use the shared overflow buffer to allocate the elements in the collected list. 

The elements will form a linked list, with each element pointing to the memory in the shared overflow buffer containing the next element. This implementation was chosen over a simpler one (e.g. storing a vector of pointers in each `CollectState` to ensure that `CollectState` is trivially destructable, which is desired because the lifetimes of `CollectState` are difficult to manage (as they are stored in a FactorizedTable with their type casted away, which can prevent their destructor from being called).

Also fixes a concurrency bug where an unprotected shared overflow buffer is used in the finalizing the simple hash aggregates. Now, thread-local overflow buffers are used during finalization and the memory is merged into the shared buffer at the end of each thread's finalization.

# Microbenchmarks

First, we check to see the performance impact of the fix to the concurrency bug:

These benchmarks were done on clickbench.

Query | Execution time (master) | Execution time (new) | Peak memory usage (master) | Peak memory usage (new)
----|----|----|----|----
`MATCH (h:hits) RETURN COUNT(DISTINCT h.UserID);`|507ms|540ms|4458MB|4506MB
`MATCH (h:hits) RETURN COUNT(DISTINCT h.SearchPhrase);`|455ms|395ms|5270MB|5207MB
`MATCH (h:hits) RETURN h.RegionID, COUNT(DISTINCT h.UserID) AS u ORDER BY u DESC LIMIT 10;`|1850ms|1840ms|5388MB|5337MB

Next, we check the performance impact on collect:

These benchmarks were done on ldbc-sf01:

Dataset|Query | Execution time (master) | Execution time (new) | Peak memory usage (master) | Peak memory usage (new)
----|----|----|----|----|----
ldbc-sf01 replyOf_Comment (~77K edges) | `MATCH (a:Comment)-[e:replyOf_Comment]->(b:Comment) WITH b, collect(a) as aList RETURN b, aList;`|2.04s|0.38ms|9592MB|1261MB
twitter (~81K nodes, ~2.4M edges) | `MATCH (a:account)-[e:follows]->(b:account) return b.ID, collect(a);`|44s|1.05s|264GB|1.68GB

# Contributor agreement

### DIFF
--- a/src/function/aggregate/collect.cpp
+++ b/src/function/aggregate/collect.cpp
@@ -1,5 +1,4 @@
 #include "function/aggregate_function.h"
-#include "processor/result/factorized_table.h"
 #include "storage/storage_utils.h"
 
 using namespace kuzu::binder;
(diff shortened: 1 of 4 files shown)


src/function/aggregate/collect.cpp
```diff
@@ -10,49 +9,115 @@ using namespace kuzu::processor;
...
     void moveResultToVector(common::ValueVector* outputVector, uint64_t pos) override;
 
-    std::unique_ptr<processor::FactorizedTable> factorizedTable;
+    void appendElement(ValueVector* input, uint32_t pos, InMemOverflowBuffer* overflowBuffer);
+    void resetList();
+    void appendList(const CollectState& o);
+
+    // We store the head + tail of the linked list
+    CollectListElement head;
+    CollectListElement tail;
```
benjaminwinger, line 53: I think the tail can be omitted if you append to the front of the list.

royi-luo: Discussed offline, that would complicate concatenating lists so keeping it as-is.

benjaminwinger: Never mind, I'd forgotten about concatenating lists

PR #5470, opened 2025-05-29
python: code cleanup

sdht0:
- Updated deps
- ruff check: no warnings
- ruff format

### DIFF
--- a/tools/python_api/src_py/async_connection.py
+++ b/tools/python_api/src_py/async_connection.py
@@ -176,6 +176,8 @@ class AsyncConnection:
 
     async def _prepare(self, query: str, parameters: dict[str, Any] | None = None) -> PreparedStatement:
         """
+        Prepare a query.
+
         The only parameters supported during prepare are dataframes.
         Any remaining parameters will be ignored and should be passed to execute().
         """
--- a/tools/python_api/src_py/async_connection.py
+++ b/tools/python_api/src_py/async_connection.py
@@ -176,11 +176,9 @@ class AsyncConnection:
 
     async def _prepare(self, query: str, parameters: dict[str, Any] | None = None) -> PreparedStatement:
         """
-        Prepare a query.
-
         The only parameters supported during prepare are dataframes.
         Any remaining parameters will be ignored and should be passed to execute().
-        """
+        """  # noqa: D401
         loop = asyncio.get_running_loop()
         conn, conn_index = self.__get_connection_with_least_queries()
 
--- a/tools/python_api/src_py/connection.py
+++ b/tools/python_api/src_py/connection.py
@@ -128,21 +128,21 @@ class Connection:
             raise RuntimeError(msg)  # noqa: TRY004
 
         if len(parameters) == 0 and isinstance(query, str):
-            _query_result = self._connection.query(query)
+            query_result = self._connection.query(query)
         else:
             prepared_statement = self._prepare(query, parameters) if isinstance(query, str) else query
-            _query_result = self._connection.execute(prepared_statement._prepared_statement, parameters)
-        if not _query_result.isSuccess():
-            raise RuntimeError(_query_result.getErrorMessage())
-        current_query_result = QueryResult(self, _query_result)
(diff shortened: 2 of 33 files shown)


tools/python_api/src_py/async_connection.py
```diff
@@ -176,6 +176,8 @@ async def execute(
 
     async def _prepare(self, query: str, parameters: dict[str, Any] | None = None) -> PreparedStatement:
         """
+        Prepare a query.
+
```
mewim, line 180: This method is intentionally not commented. It is not considered as an internal method and not intended to be used by the user.

tools/python_api/src_py/connection.py
```diff
@@ -151,6 +151,8 @@ def _prepare(
         parameters: dict[str, Any] | None = None,
     ) -> PreparedStatement:
         """
+        Prepare a query.
```
mewim, line 154: Same here

tools/python_api/src_py/connection.py
```diff
@@ -128,21 +128,21 @@ def execute(
...
-        while _query_result.hasNextQueryResult():
-            _query_result = _query_result.getNextQueryResult()
-            if not _query_result.isSuccess():
-                raise RuntimeError(_query_result.getErrorMessage())
-            all_query_results.append(QueryResult(self, _query_result))
+        while query_result.hasNextQueryResult():
+            query_result = query_result.getNextQueryResult()
+            if not query_result.isSuccess():
+                raise RuntimeError(query_result.getErrorMessage())
+            all_query_results.append(QueryResult(self, query_result))
```
mewim, line 145: I would not change this part. I feef like it is better to keep it `_query_result` to distinguish the native `query_result` exposed from C++ side from the Python query result wrapper.

tools/python_api/src_py/query_result.py
```diff
@@ -326,32 +326,28 @@ def encode_rel_id(rel: dict[str, Any]) -> tuple[int, int]:
                     continue
                 column_type, _ = properties_to_extract[i]
                 if column_type == Type.NODE.value:
-                    _id = row[i]["_id"]
-                    nodes[(_id["table"], _id["offset"])] = row[i]
-                    table_to_label_dict[_id["table"]] = row[i]["_label"]
+                    id = row[i]["_id"]
```
mewim, line 329: I would not remove the unserscores from this function. Specifically not from `_id`. Python has a built-in `id` function, removing `_` will mask it, which is bad.

tools/python_api/src_py/query_result.py
```diff
@@ -326,32 +326,28 @@ def encode_rel_id(rel: dict[str, Any]) -> tuple[int, int]:
...
-                    _src = row[i]["_src"]
-                    _dst = row[i]["_dst"]
                     rels[encode_rel_id(row[i])] = row[i]
 
                 elif column_type == Type.RECURSIVE_REL.value:
                     for node in row[i]["_nodes"]:
-                        _id = node["_id"]
-                        nodes[(_id["table"], _id["offset"])] = node
-                        table_to_label_dict[_id["table"]] = node["_label"]
+                        id = node["_id"]
```
mewim, line 338: Same here

tools/python_api/src_py/query_result.py
```diff
@@ -326,32 +326,28 @@ def encode_rel_id(rel: dict[str, Any]) -> tuple[int, int]:
...
                                 del rel[key]
-                        _src = rel["_src"]
-                        _dst = rel["_dst"]
                         rels[encode_rel_id(rel)] = rel
 
         # Add nodes
         for node in nodes.values():
-            _id = node["_id"]
-            node_id = node["_label"] + "_" + str(_id["offset"])
+            id = node["_id"]
```
mewim, line 349: Also here. In general, just double check to make sure there is not underscore removed from `_id`

mewim (changes requested):
See my comments

mewim (approved):

PR #5471, opened 2025-05-29
chore: remove all code warnings

sdht0:
Currently, these warnings appear on the latest version of gcc.


.github/workflows/ci-workflow.yml
mewim, line 1: I think this can be passed as an environment variable, so you do not need to do it in each step

sdht0: Already set :100:

ray6080 (approved):

PR #5472, opened 2025-05-29
ci: auto format rust & python code

sdht0:
Don't force manually handing formatting failures.

Depends on #5470

### DIFF
--- a/.github/workflows/ci-workflow.yml
+++ b/.github/workflows/ci-workflow.yml
@@ -804,6 +804,12 @@ jobs:
       - name: Check and fix extension format
         run: python3 scripts/run-clang-format.py --in-place --clang-format-executable /usr/bin/clang-format-18 -r extension/
 
+      - name: Format Python API
+        run: |
+          cd tools/python_api
+          make requirements
+          make format
+
       - name: Update Rust
         run: rustup update
 
--- a/.github/workflows/ci-workflow.yml
+++ b/.github/workflows/ci-workflow.yml
@@ -366,7 +366,6 @@ jobs:
 
   rust-build-test:
     name: rust build & test
-    needs: [ rustfmt-check ]
     runs-on: kuzu-self-hosted-testing
     env:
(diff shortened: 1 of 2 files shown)


mewim (approved):

benjaminwinger (approved):
Personally, I've been using pre-commit hooks (via https://pre-commit.com/) to handle formatting locally when commits are made. 
Would it be helpful to add a `.pre-commit-config.yaml` to the repository so that others can use the same hooks? It is admittedly one more thing that will need to be updated periodically so that the version of the formatting tools used locally match the version used in CI, but it wouldn't be very hard to maintain.

sdht0:
I think pre-commit hooks can be _in addition to_ the auto formatter. Since we use auto formatting for the C++ side, might as well extend to the other code, no?

benjaminwinger:
> I'm more open to native git pre-push hooks that we ask devs to link inside the .git dir manually.

So essentially the same as maintaining an uncommitted config for the pre-commit tool and having it generate pre-commit hooks like I'm doing right now (doing it manually I guess also works, but it seems like unnecessary work).

> For now, how about we continue relying on the auto-commit and encourage devs to turn on auto-format-on-save in their IDEs?

I'm fine with that.

mewim:
I think for clang-format, we did encounter different formatting due to different versions of toolchain and that was the reason we decided to do the auto formatting. I think pre-commit hook does not guarantee a consistent formatting since the local configuration can still be different

benjaminwinger:
> EDIT: I guess the tool could be generating these hooks from the yaml file. I'd rather commit the git hook and cut out the middleman.

The generated hook still uses the tool. A large part of the point of the `pre-commit` tool is that it installs a specific version of the different hooks it uses for consistency (it basically includes its own package manager), which means it could be kept synchronised with the version of clang-format, cargo, ruff etc. that's used in CI (though I'll admit there have been some times when the CI version of clang-format still seems to behave slightly differently).

PR #5475, opened 2025-05-30
Add ADLS support to Azure extension

05st:
I believe ADLS support was already working since we use DuckDB under the hood. All this PR really does is add support for the "abfss://" URI scheme, and some new tests.

### DIFF
--- a/extension/azure/test/test_files/azure.test
+++ b/extension/azure/test/test_files/azure.test
@@ -6,9 +6,12 @@
 ---- error
 Binder exception: Cannot load from file type azure. If this file type is part of a kuzu extension please load the extension then try again.
 
--CASE AzureScan
+-CASE AzureScanBlob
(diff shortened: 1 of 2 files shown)


extension/azure/test/test_files/azure.test
```diff
@@ -44,3 +47,21 @@ Alice|Bob
 Bob|Dan
 Bob|Carol
 Bob|Alice
+-STATEMENT COPY Knows FROM (LOAD FROM "abfss://kuzu-test/dirA/dirB/eKnows_2.csv" (file_format='azure') RETURN *);
```
ray6080, line 50: Not related to this PR. but I'm confusing why is azure a file_format? and why do we need to explicitly specify it?

ray6080 (approved):

PR #5476, opened 2025-05-30
Support detach delete from single direction rel table

royi-luo:
# Description

`RelTable::detachDelete()` now supports deleting from rel tables with single-direction (FWD only) storage. This is for internal use only, the `DETACH DELETE` operator still won't work if there are attached rels with single-direction storage as there is currently no easy way to delete incoming edges for a node in this case.

I also fixed the handling of the `DEFAULT_REL_STORAGE_DIRECTION` setting for compilation, previously the incorrect string was passed into `system_config.h` which actually caused it to always compile with `DEFAULT_REL_STORAGE_DIRECTION=BOTH`.

# Contributor agreement


src/include/storage/table/rel_table.h
```diff
@@ -157,6 +157,9 @@ class KUZU_API RelTable final : public Table {
     void update(transaction::Transaction* transaction, TableUpdateState& updateState) override;
     bool delete_(transaction::Transaction* transaction, TableDeleteState& deleteState) override;
 
+    // Deletes all edges attached to the node(s) specified in the deleteState
+    // Currently only supports deleting from a single src node
```
ray6080, line 161: I don't think we should say "currently only". the meaning of this interface should be to delete outgoing rels of given nodes in each storage direction.

ray6080 (approved):
Thanks!

PR #5479, opened 2025-06-02
Refactor card estimator

andyfengHKU:
# Description

This PR removes `addNodeIDDomAndStats` that is being called inside `appendXXOperator`. It's better the enforce the assumption that the cardinality of base should be ready before join order enumeration. Otherwise, we might end up repeatedly calculate cardinality if `appendXXOperator` is called multiple times. 

Also, we remove `perQueryGraphNodeIDName2dom`. For subquery, we can directly modify based table cardinality with outer cardinality.


royi-luo (approved):
LGTM

Issue #5489, opened 2025-06-03
Bug: Zone map assertion failure due to type mismatch


royi-luo:
### Kuzu version

master

### What operating system are you using?

Ubuntu

### What happened?

If you perform the following queries on a fresh DB with runtime checks enabled:

```
create node table Comment (id STRING, creationDate INT64, locationIP STRING, browserUsed STRING, content STRING, length INT32, PRIMARY KEY (id));
COPY Comment FROM "dataset/ldbc-sf01/Comment.csv";
match (c:Comment) where c.id = 687194777181 return c.browserUsed, c.content;
```

You will get the assertion error `Assertion failed in file "/home/r34luo/kuzu/kuzu-master/src/storage/predicate/constant_predicate.cpp" on line 24: mergedStats.stats.min.has_value() && mergedStats.stats.max.has_value()
`.

Note that in the table schema `id` is a string but in the `match` statement we pass in an int. Not sure if an implicit cast is expected here but the cast should either work or we should throw an error in this case.

PR #5493, opened 2025-06-04
Make zone map work with casted string columns

royi-luo:
# Description

Fixes https://github.com/kuzudb/kuzu/issues/5489

For zone maps we previously assumed that if the literal was a storage value type the column will also be. This is not accurate as the column can be casted from a string to a numeric type. This PR adds support for this case.

# Contributor agreement

### DIFF
--- a/test/test_files/filter/node.test
+++ b/test/test_files/filter/node.test
@@ -15,6 +15,12 @@
 ---- 1
 77|123456789.123000
 
+-CASE ZoneMapNotEqualSkip
+-STATEMENT ALTER TABLE person ADD another_col INT64 DEFAULT 1
+---- ok
+-STATEMENT MATCH (a:person) WHERE a.another_col <> 1 RETURN a.*
+---- 0
+
 -CASE ZoneMapUpdateThenQueryWithoutCheckpoint
 -LOG Comparison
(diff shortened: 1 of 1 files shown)


src/storage/predicate/constant_predicate.cpp
```diff
@@ -21,43 +21,46 @@ bool inRange(T min, T max, T val) {
...
+        auto constant = value.getValue<T>();
+        switch (expressionType) {
+        case ExpressionType::EQUALS: {
+            if (!inRange<T>(min, max, constant)) {
+                return ZoneMapCheckResult::SKIP_SCAN;
+            }
+        } break;
+        case ExpressionType::NOT_EQUALS: {
+            if (Equals::operation<T>(constant, min) && Equals::operation<T>(constant, max)) {
+                return ZoneMapCheckResult::SKIP_SCAN;
```
andyfengHKU, line 38: Seems we miss a coverage here. Can u also add a test case.

andyfengHKU (approved):

src/storage/predicate/constant_predicate.cpp
royi-luo: added

PR #5499, opened 2025-06-05
Reduce HNSW index upper layer memory usage

royi-luo:
# Description

# Contributor agreement

### DIFF
--- a/extension/vector/src/include/index/hnsw_graph.h
+++ b/extension/vector/src/include/index/hnsw_graph.h
@@ -129,6 +129,11 @@ private:
     std::unique_ptr<CompressedOffsetsView> view;
 };
 
+struct NodeToGraphOffsetMap {
+    std::unordered_map<common::offset_t, common::offset_t> nodeToGraphMap;
+    std::unique_ptr<common::offset_t[]> graphToNodeMap;
+};
+
 class InMemHNSWGraph {
 public:
     InMemHNSWGraph(storage::MemoryManager* mm, common::offset_t numNodes,
@@ -174,7 +179,8 @@ public:
     }
 
     void finalize(storage::MemoryManager& mm, common::node_group_idx_t nodeGroupIdx,
-        const processor::PartitionerSharedState& partitionerSharedState);
+        const processor::PartitionerSharedState& partitionerSharedState,
+        const NodeToGraphOffsetMap* selectedNodesMap);
 
     // In the current implementation, race conditions can result in dstNode entries being skipped
     // during insertion. Skipped entries will be marked with this value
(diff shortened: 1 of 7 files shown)


extension/vector/src/include/index/hnsw_graph.h
```diff
@@ -129,6 +129,19 @@ class CompressedNodeOffsetBuffer {
...
 };
 
+struct NodeToGraphOffsetMap {
+    explicit NodeToGraphOffsetMap(common::offset_t numNodesInTable) : numNodes(numNodesInTable){};
+    NodeToGraphOffsetMap(common::offset_t numNodesInTable, common::NullMask* selectedNodes);
+
+    common::offset_t getNumNodesInGraph() const { return numNodes; }
+    common::offset_t nodeToGraphOffset(common::offset_t nodeOffset) const;
+    common::offset_t graphToNodeOffset(common::offset_t graphOffset) const;
+    bool isValid() const { return graphToNodeMap != nullptr; }
```
royi-luo, line 139: should probably rename to something like `isTrivialMapping()` + invert return value

extension/vector/src/include/index/hnsw_index.h
```diff
@@ -202,6 +217,10 @@ class InMemHNSWIndex final : public HNSWIndex {
     std::unique_ptr<InMemHNSWLayer> upperLayer;
     std::unique_ptr<InMemHNSWLayer> lowerLayer;
     std::unique_ptr<InMemEmbeddings> embeddings;
+
+    std::unique_ptr<NodeToGraphOffsetMap> lowerGraphSelectionMap; // this mapping is empty
```
royi-luo, line 221:
```suggestion
    std::unique_ptr<NodeToGraphOffsetMap> lowerGraphSelectionMap; // this mapping is trivial
```

extension/vector/src/index/hnsw_graph.cpp
```diff
@@ -225,5 +227,31 @@ void InMemHNSWGraph::resetCSRLengthAndDstNodes() {
...
+    common::offset_t curOffset = 0;
+    for (common::offset_t i = 0; i < numNodesInTable; ++i) {
+        if (selectedNodes->isNull(i)) {
+            graphToNodeMap[curOffset] = i;
+            ++curOffset;
+        }
+    }
+}
+
+common::offset_t NodeToGraphOffsetMap::nodeToGraphOffset(common::offset_t nodeOffset) const {
```
royi-luo, line 243: This naming is a bit misleading since it currently actually returns the next graph offset in increasing order if the node offset isn't directly part of the mapping. We take advantage of this in `finalize` to find the graph offsets that are part of each node group. Regardless this function should probably be renamed. It may also be good to split it into versions that require exact matches (although this version works for all of our use cases so maybe not, maybe I'll add a boolean parameter that triggers an assert for testing purposes).

PR #5507, opened 2025-06-06
(code-maintenance)-Add-Control-Flow-Braces-Check

tgahunia05:
See
https://clang.llvm.org/extra/clang-tidy/checks/readability/braces-around-statements.html

# Contributor agreement

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -21,7 +21,7 @@ Checks:
   cppcoreguidelines-init-variables,
   cppcoreguidelines-pro-type-member-init,
 
-  google-explicit-constructor
+  google-explicit-constructor,
 
   misc-unused-parameters,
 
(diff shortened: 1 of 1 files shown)


src/include/storage/index/in_mem_hash_index.h
```diff
@@ -182,8 +182,9 @@ class InMemHashIndex final {
         if (deletedPos.has_value()) {
             // Find the last valid entry and move it into the deleted position
             auto newIter = iter;
-            while (nextChainedSlot(newIter))
+            while (nextChainedSlot(newIter)) {
                 ;
```
benjaminwinger, line 186: If you're adding braces, the `;` should just be removed here.

sdht0 (approved):
Thanks!

PR #5516, opened 2025-06-09
HNSW finalize memory optimization

royi-luo:
# Description
In master the entire in-memory HNSW layers are materialized in the finalize operator for `CreateHNSWIndex` before being passed to the batch insert operator. This PR changes it so that each partition (node group) is only materialized right before it is needed by a batch insert thread. This reduces the memory overhead of the materialized partitions to one node group (~140MB with ml=60) per executing thread.

This is done by making introducing a `BasePartitionerSharedState` interface. The existing partitioner shared state remains as `PartitionerSharedState` and contains all materialized partition buffers. For the vector index we introduce a `HNSWLayerPartitionerSharedState` which contains the `HNSWLayer` after the create pipeline completes. Partitions are not actually materialized until `getPartitionBuffer()` is called.

**A quick sanity check on performance:**

Dataset: graph-200-angular

Creating index takes 460.1s on master and 455.5s on the new branch.

# Contributor agreement


royi-luo:
Let's hold off on this change until I complete the rel batch insert refactor, this change may end up not being needed

PR #5519, opened 2025-06-10
extension(llm) create embedding function

tgahunia05:
# Description

Implements additional function signatures to support additional params such as dimensions and region.
Changes function signature to return list. This results in simpler logic; we no longer need to hardcode a list of supported models. 
Support configuration where applicable. 

# Contributor agreement

### DIFF
--- a/extension/llm/test/test_files/amazonbedrock.test
+++ b/extension/llm/test/test_files/amazonbedrock.test
@@ -110,7 +110,8 @@ False
 -STATEMENT CALL CREATE_VECTOR_INDEX('Book', 'title_vec_index', 'title_embedding');
 ---- ok
 
--STATEMENT CALL QUERY_VECTOR_INDEX('Book', 'title_vec_index', create_embedding('Quantum Machines', 'amazon-bedrock', 'amazon.titan-embed-text-v1', 'us-east-1'), 2) RETURN node.title ORDER BY distance;
+-STATEMENT CALL QUERY_VECTOR_INDEX('Book', 'title_vec_index', create_embedding('quantum machine learning', 'amazon-bedrock', 'amazon.titan-embed-text-v1', 'us-east-1'), 2) RETURN node.title ORDER BY distance;
+-CHECK_ORDER
 ---- 2
 Learning Machines
 The Quantum World
--- a/extension/llm/test/test_files/googlegemini.test
+++ b/extension/llm/test/test_files/googlegemini.test
@@ -42,6 +42,9 @@ False
 ---- 1
 False
 
+# TODO(TANVIR)I have not been able to validate this result. I am on the free tier of
+# Google Gemini and rate limits do not let this test run to completion. This should
+# be validated when CI is setup.
 -CASE UseWithVectorExtension
 -STATEMENT load extension "${KUZU_ROOT_DIRECTORY}/extension/llm/build/libllm.kuzu_extension"
 ---- ok
@@ -109,7 +112,8 @@ False
 -STATEMENT CALL CREATE_VECTOR_INDEX('Book', 'title_vec_index', 'title_embedding');
 ---- ok
 
--STATEMENT CALL QUERY_VECTOR_INDEX('Book', 'title_vec_index', create_embedding('Quantum Machines', 'google-gemini', 'gemini-embedding-exp-03-07'), 2) RETURN node.title ORDER BY distance;
+-STATEMENT CALL QUERY_VECTOR_INDEX('Book', 'title_vec_index', create_embedding('quantum machine learning', 'google-gemini', 'gemini-embedding-exp-03-07'), 2) RETURN node.title ORDER BY distance;
+-CHECK_ORDER
 ---- 2
 Learning Machines
 The Quantum World
--- a/extension/llm/test/test_files/googlevertex.test
+++ b/extension/llm/test/test_files/googlevertex.test
@@ -117,7 +117,8 @@ False
 -STATEMENT CALL CREATE_VECTOR_INDEX('Book', 'title_vec_index', 'title_embedding');
 ---- ok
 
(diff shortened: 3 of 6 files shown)


extension/llm/src/function/create_embedding.cpp
```diff
@@ -79,26 +81,105 @@ static void execFunc(const std::vector<std::shared_ptr<common::ValueVector>>& pa
...
+    std::optional<std::string> region = std::nullopt;
+    auto& provider = getInstance(StringUtils::getLower(input.arguments[providerIdx]->toString()));
+
+    if (input.arguments.size() == dimensionsAndRegionSpecified) {
+        try {
+            Binder binder{input.context};
+            std::shared_ptr<Expression> dimensionsExpr = input.arguments[3];
+            dimensionsExpr = binder.getExpressionBinder()->implicitCastIfNecessary(dimensionsExpr,
+                LogicalType(LogicalTypeID::INT64));
+            auto toCast = std::stoll(
```
tgahunia05, line 98: This is an ugly work around for checking that the passed parameter is not negative. When passing a parameter like 
`create_embedding('prompt', 'provider', 'model', 50)`
50 is set to be a INT64; it does not and (from I have seen in casting related code) cannot implicitly cast to a UINT64. 

See 
`uint32_t BuiltInFunctionsUtils::getCastCost(LogicalTypeID inputTypeID, LogicalTypeID targetTypeID)`

sdht0 (approved):
Some minor feedback.

- [x] Check and remove unused headers such as `#include "main/database.h"` in `llm_extension.cpp`.
- [x] `accross` -> `across` in `amazon-bedrock.cpp`. 
- [x] Can you write in a short comment in say `llm_extension.cpp` the overall design of this extension. Especially what is needed to add a new provider?
- [x] I've not tested this extension, so please double check that everywhere works as expected.
- [x] Also asking you to create the corresponding docs PR.

sdht0:
Also for the tests, I'm pretty sure the actual vectors will change overtime. Here's an alternative idea:
- Create a 2-3 line csv dataset, with the current embeddings for the strings stored there.
- In the tests, get the embeddings for the same strings from different providers with different settings and check that the cosine similarity of these embeddings is close to the stored embedding.

A little more work, but I think it's better for long term maintainability.

Happy to discuss other alternatives too.

extension/llm/src/function/create_embedding.cpp
```diff
@@ -79,26 +81,105 @@ static void execFunc(const std::vector<std::shared_ptr<common::ValueVector>>& pa
...
-    return std::make_unique<FunctionBindData>(std::move(types),
+    static constexpr size_t providerIdx = 1;
+    static constexpr size_t modelIdx = 2;
+    static constexpr size_t dimensionsOrRegionSpecified = 4;
+    static constexpr size_t dimensionsAndRegionSpecified = 5;
+    std::optional<uint64_t> dimensions = std::nullopt;
+    std::optional<std::string> region = std::nullopt;
+    auto& provider = getInstance(StringUtils::getLower(input.arguments[providerIdx]->toString()));
+
+    if (input.arguments.size() == dimensionsAndRegionSpecified) {
```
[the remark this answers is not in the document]
tgahunia05: I believe I have applied this feedback accordingly. If not reach out and we can discuss.

extension/llm/src/function/create_embedding.cpp
tgahunia05: I believe I have applied this feedback accordingly. I would like to think that there is a better mechanism to do this test. @andyfengHKU?

extension/llm/src/function/create_embedding.cpp
```diff
@@ -78,27 +80,82 @@ static void execFunc(const std::vector<std::shared_ptr<common::ValueVector>>& pa
     }
 }
 
+static uint64_t parseDimensions(std::shared_ptr<Expression> dimensionsExpr,
+    main::ClientContext* context) {
+    Binder binder{context};
+    auto value =
+        evaluator::ExpressionEvaluatorUtils::evaluateConstantExpression(dimensionsExpr, context)
+            .toString();
+    int64_t dimensions = static_cast<int64_t>(std::stoll(value));
```
sdht0, line 89: Isn't there a function to do this casting that was being used before in the PR?

sdht0 (approved):
LGTM, thanks.

extension/llm/src/function/create_embedding.cpp
tgahunia05: Thanks! I had used` dimensions = parameters[3]->getValue<uint64_t>(0);` before.
There is a similar method I can call to avoid std::stoll.

sdht0 (commented):
Some minor comments.

Are there tests that uses the embedding created using this extension in a vector embedding? If not, add at least one for each provider I think. This is to have a tested reference for how to use llm+vector together.

extension/llm/src/providers/google-gemini.cpp
```diff
@@ -42,15 +42,17 @@ std::vector<float> GoogleGeminiEmbedding::parseResponse(const httplib::Result& r
...
-    static const std::unordered_map<std::string, uint64_t> modelDimensionMap = {
-        {"gemini-embedding-exp-03-07", 3072}, {"text-embedding-004", 768}, {"embedding-001", 768}};
-    auto modelDimensionMapIter = modelDimensionMap.find(model);
-    if (modelDimensionMapIter == modelDimensionMap.end()) {
-        throw(common::BinderException(
-            "Invalid Model: " + model + '\n' + std::string(referenceKuzuDocs)));
+void GoogleGeminiEmbedding::configure(const std::optional<uint64_t>& dimensions,
+    const std::optional<std::string>& region) {
+    if (dimensions.has_value()) {
+        throw(BinderException("Google-Gemini does not support the dimensions argument: " +
```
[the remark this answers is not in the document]
tgahunia05: Changed to Google Gemini. This has been applied.

PR #5537, opened 2025-06-13
Fix usage of system gtest in CMake

royi-luo:
# Description

There were some issues with how we were including gtest into our test builds:

- [gtest targets are prefixed with 'GTest::'](https://cmake.org/cmake/help/latest/module/FindGTest.html). CMake doesn't give any errors when you `target_link_libraries` against a target that doesn't exist so I'm assuming that's why we didn't see it complain before but we're not linking our tests against the correct targets regardless.
- Imported targets do not have global visibility so even if we were using the correct target names the CMake file in `test/` we still wouldn't find the correct targets to link against `GTest::GTest`.

I'm assuming that things may have worked before because I think we use gtest as a header-only library (I don't think we use gtest_main) and the `include(GoogleTest)` was correctly resolving the include directory. However on my system `include(GoogleTest)` doesn't work. I think this is because the gtest CMake files are in a different location from the gtest headers, and thus I rely on the gtest libraries being correctly propagated to the tests as they have the correct include directories.

Regardless this PR fixes that problem.

# Contributor agreement


CMakeLists.txt
```diff
@@ -273,6 +273,7 @@ option(BUILD_EXTENSION_TESTS "Build C++ extension tests." FALSE)
 option(BUILD_KUZU "Build Kuzu." TRUE)
 option(ENABLE_BACKTRACES "Enable backtrace printing for exceptions and segfaults" FALSE)
 option(USE_STD_FORMAT "Use std::format instead of a custom formatter." FALSE)
+option(PREFER_SYSTEM_DEPS "Try to find some dependencies on the system and only downloads then if not found" TRUE)
```
sdht0, line 276:
```suggestion
option(PREFER_SYSTEM_DEPS "Only download certain deps if not found on the system" TRUE)
```

CMakeLists.txt
```diff
@@ -281,10 +282,10 @@ endif()
 
 if (ENABLE_BACKTRACES)
     find_package(cpptrace QUIET)
```
sdht0, line 284: Put this inside the `if` then?

test/graph_test/CMakeLists.txt
```diff
@@ -10,7 +10,7 @@ target_include_directories(
         ../include/
 )
 
-target_link_libraries(graph_test PUBLIC GTest::GTest kuzu)
+target_link_libraries(graph_test PUBLIC KUZU_GTEST kuzu)
```
sdht0, line 13: Optional, KUZU_GTEST -> GTEST_LIB? GTEST doesn't really have any connection to KUZU.

test/gtest/CMakeLists.txt
```diff
@@ -2,10 +2,10 @@
 set(CMAKE_COMPILE_WARNING_AS_ERROR FALSE)
 
 find_package(GTest QUIET)
-if(GTest_FOUND)
+if(${PREFER_SYSTEM_DEPS} AND GTest_FOUND)
```
sdht0, line 5: Same, find_package should be inside the `if`

sdht0 (approved):
Thanks, LGTM!

CMakeLists.txt
royi-luo: updated

CMakeLists.txt
royi-luo: updated, now we have ugly control flow :(

test/gtest/CMakeLists.txt
royi-luo: updated

CMakeLists.txt
sdht0: Yeah sorry, but I find the unused `find_package` more troubling :grimacing:

PR #5538, opened 2025-06-13
Fix HNSW checkpoint transaction issue

ray6080:
# Description

Fix the start transaction timestamp during HNSW checkpoint.
Also reset aux buffer before lookup in embedding columns to avoid unnecessary memory growth.


extension/vector/src/index/hnsw_graph.cpp
```diff
@@ -78,6 +78,7 @@ void* OnDiskEmbeddings::getEmbedding(transaction::Transaction* transaction,
     } else {
         nodeTable.initScanState(transaction, scanState);
     }
+    scanState.outputVectors[0]->resetAuxiliaryBuffer();
```
andyfengHKU, line 81: was this a bug?

andyfengHKU (approved):

PR #5550, opened 2025-06-17
Export from prev release import into a new release

tgahunia05:
# Description
This PR provides a Python script to automatically create exports of all test datasets (with schema.cypher files) for any given Kuzu version. 

We take a path to a kuzu shell executable and the path to the dataset directory as arguments
Example `scripts/export-db.py build/debug/tools/shell/kuzu dataset`

The exported dbs are put at `{dataset_directory}/tmp/{kuzu_version}/{name_of_db}/`
Example `dataset/tmp/0.10.0.5/tinysnb/`

Observations:
1. Found a copy.cypher file that is different to other copy files. See `dataset/import_db_legacy` The name seems to imply this is intentional. This copy file uses a path relative to `dataset/import_db_legacy`. The other copy files use a path relative to `dataset`. This dataset is not successfully exported. Update: This dataset is used in `test/test_files/copy/export_import_db.test` as part of some backwards compatibility test so I don't plan on fiddling with that dataset right now.

2. Failing to export `dataset/copy-special-char-test`. But the test that uses this dataset, `test/test_files/copy/copy_special_char.test`, is currently skipped. When I ran it manually it failed with the same error.
`expected 3 values per row, but got 1. Line/record containing the error: '0|-this is |the first line-|2020-02-18'" thrown in SetUp()`

3. Consider in `dataset/ldbc-sf01/Forum.csv` the lines 

```
893353212271|Group for Ferdinand_II,_Holy_Roman_Emperor in Chennai|20120420114423625
481036351856|Group for Ferdinand_II,_Holy_Roman_Emperor in Chennai|20110318075857635
```

This is exported as 
```
893353212271,"Group for Ferdinand_II,_Holy_Roman_Emperor in Chennai",20120420114423625
481036351856,"Group for Ferdinand_II,_Holy_Roman_Emperor in Chennai",20110318075857635
```

The format is not preserved. This results in a copy exception when running tests with the import dir env set. This seems to be a limitation of the export db function since I cannot find any ways to set delimiters. 

`Conversion exception: Cast failed. Could not convert "_Holy_Roman_Emperor in Chennai"" to INT64. Line/record containing the error: '893353212271,"Group for Ferdinand_II,_Holy_Roman_Emperor in Chennai"...'" thrown in SetUp().
`

Progresses #5363 
# Contributor agreement

### DIFF
--- a/test/graph_test/base_graph_test.cpp
+++ b/test/graph_test/base_graph_test.cpp
@@ -108,10 +108,10 @@ void BaseGraphTest::initGraph(const std::string& datasetDir) const {
     }
 
     // Run tests on datasets exported from a previous Kuzu version. Used to verify that exports and
-    // imports across versions work correctly.Note: we skip importing the empty db
+    // imports across versions work correctly. We skip importing the empty db.
     auto dirs = StringUtils::split(StringUtils::getLower(datasetDir), "/");
     if (std::find(dirs.begin(), dirs.end(), "empty") != dirs.end()) {
-        std::cout << "Skipping empty DB" << std::endl;
+        std::cout << "Skipping Empty Dataset: " << datasetDir << std::endl;
         return;
     }
     std::string query = "IMPORT DATABASE '" + datasetDir + "';";
--- a/test/graph_test/base_graph_test.cpp
+++ b/test/graph_test/base_graph_test.cpp
@@ -7,6 +7,7 @@
 #include "common/string_format.h"
 #include "spdlog/spdlog.h"
 #include "test_helper/test_helper.h"
+#include <common/string_utils.h>
 
 using namespace kuzu::common;
 using namespace kuzu::main;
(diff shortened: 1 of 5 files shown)


test/graph_test/base_graph_test.cpp
```diff
@@ -107,7 +108,12 @@ void BaseGraphTest::initGraph(const std::string& datasetDir) const {
     }
 
     // Run tests on datasets exported from a previous Kuzu version. Used to verify that exports and
-    // imports across versions work correctly.
+    // imports across versions work correctly. We skip importing the empty db.
```
sdht0, line 111:
```suggestion
    // imports across versions work correctly. This skips importing the `empty` dataset.
```

test/graph_test/base_graph_test.cpp
```diff
@@ -107,7 +108,12 @@ void BaseGraphTest::initGraph(const std::string& datasetDir) const {
...
+    auto dirs = StringUtils::split(StringUtils::getLower(datasetDir), "/");
+    if (std::find(dirs.begin(), dirs.end(), "empty") != dirs.end()) {
+        std::cout << "Skipping Empty Dataset: " << datasetDir << std::endl;
```
sdht0, line 114: Optional, but I think the standard in this repo is to use `stringFormat`?

sdht0 (approved):
Thanks, LGTM.

test/graph_test/base_graph_test.cpp
tgahunia05: I am indifferent to either form. I have also seen both being used in the repo. In the same file see 
```
void TestHelper::executeScript(const std::string& cypherScript, Connection& conn) 
```
    
```
    if (!std::filesystem::exists(cypherScript)) {
        std::cout << "cypherScript: " << cypherScript << " doesn't exist. Skipping..." << std::endl;
        return;
    }
```

Although I have seen more uses of the stringFormat. This is a trivial change.

PR #5554, opened 2025-06-18
Add option to read embeddings straight from disk during create HNSW index

royi-luo:
# Description

Add the option `cache_embeddings` to `CREATE_HNSW_INDEX`. This setting defaults to true, and if set to false we will skip caching the embedding column and read the embeddings straight from disk. This comes with a performance penalty but reduces the memory requirements.

I repurposed `OnDiskEmbeddings` to implement the interface `HNSWIndexEmbeddings` which is used to read embeddings during the creation of the HNSW index.

We also had some bugs managing the lifetime of scanned on disk embeddings, and our pattern of getting raw embedding ptrs was prone to lifetime issues in general. I reworked this so that `getEmbedding()` returns an `EmbeddingHandle`, which is a unique reference to an offset in the scanned output data chunk for the embeddings. When an `EmbeddingHandle` goes out of scope, the scan state will automatically reclaim space in the output data chunk allocated to that embedding.

**Microbenchmarks**

All benchmarks are on `CREATE_VECTOR_INDEX`.

Dataset: fashion-mnist-784
Config: 4 threads, buffer pool size 16GB

Version|Peak memory usage|Create index time|Recall
--|--|--|--
master|716.7MB|41.80s|97%
new (cached embeddings)|716.7MB|36.98s|97%
new (no cached embeddings)|591.8MB|139.39s|97%

# Contributor agreement


extension/vector/src/include/index/hnsw_graph.h
```diff
@@ -14,50 +16,164 @@ struct NodeTableScanState;
...
+template<typename T>
+struct VectorAllocator {
+    using value_type = T;
+
+    [[nodiscard]] T* allocate(const std::size_t size) { return impl->allocate(size); }
+
+    void deallocate(T* ptr, const std::size_t size) noexcept { impl->deallocate(ptr, size); }
+
+    std::shared_ptr<VectorAllocatorImpl<T>> impl = std::make_shared<VectorAllocatorImpl<T>>();
+};
```
royi-luo, line 58: This is probably an overoptimization so I'll take it out before I merge this PR, it's just here so that my memory profiling doesn't take an hour to run

extension/vector/src/include/index/hnsw_graph.h
```diff
@@ -14,50 +16,165 @@ struct NodeTableScanState;
...
```
royi-luo, line 59: This is probably an overoptimization so I'll take it out before I merge this PR, it's just here so that my memory profiling doesn't take an hour to run

extension/vector/src/include/index/hnsw_graph.h
```diff
@@ -14,50 +16,166 @@ struct NodeTableScanState;
...
```
royi-luo, line 60: This is probably an overoptimization so I'll take it out before I merge this PR, it's just here so that my memory profiling doesn't take an hour to run

src/storage/table/node_table.cpp
```diff
@@ -308,9 +309,19 @@ bool NodeTable::lookup(const Transaction* transaction, const TableScanState& sca
                 StorageUtils::getStartOffsetOfNodeGroup(scanState.nodeGroupIdx) :
             nodeOffset - StorageUtils::getStartOffsetOfNodeGroup(scanState.nodeGroupIdx);
     scanState.rowIdxVector->setValue<row_idx_t>(nodeIDPos, rowIdxInGroup);
-    return scanState.nodeGroup->lookup(transaction, scanState);
+    if constexpr (lock) {
+        return scanState.nodeGroup->lookup(transaction, scanState);
+    } else {
+        return scanState.nodeGroup->lookupNoLock(transaction, scanState);
```
royi-luo, line 315: Outside of the BM stuff (which already has its own locking) I don't think anything in lookup requires locking in this case since we're not writing to the embeddings, I also ran a create index with TSAN and didn't get any errors. @ray6080 can you double-check that the node group lookup is safe here with no lock?

extension/vector/src/index/hnsw_graph.cpp
```diff
@@ -78,62 +142,108 @@ void* OnDiskEmbeddings::getEmbedding(transaction::Transaction* transaction,
...
+void OnDiskEmbeddingScanState::addEmbedding(const EmbeddingHandle& handle) {
+    KU_ASSERT(!handle.isNull());
+    KU_ASSERT(handle.offsetInData % embeddingDim == 0);
+    const auto offsetToAdd = handle.offsetInData / embeddingDim;
+    allocatedOffsets.set(offsetToAdd);
+    KU_ASSERT(usedEmbeddingOffsets.empty() || offsetToAdd > usedEmbeddingOffsets.top());
+    usedEmbeddingOffsets.push(offsetToAdd);
+}
+
+void OnDiskEmbeddingScanState::reclaimEmbedding(const EmbeddingHandle& handle) {
```
ray6080, line 220: I would simplify the reclaiming strategy here:
1. only keep a counter.
2. once the counter is back to 0, reset aux buffer.

extension/vector/src/include/index/hnsw_graph.h
ray6080: Can you get this TODO done? Let's remove the allocator in this PR.

src/storage/table/node_table.cpp
ray6080: It should be for now, as we don't allow multiple writers at the same time. so no other writer besides creating index.

ray6080 (approved):
LGTM. Thanks!

extension/vector/src/include/index/hnsw_graph.h
royi-luo: removed

extension/vector/src/index/hnsw_graph.cpp
royi-luo: Going to change the reclaiming strategy back. This is because we often have a top level embedding (e.g. the query vector that we are inserting) that stays in scope for the whole insert, which means that the scan state will continue growing for the entire insert (which is pretty bad).

PR #5557, opened 2025-06-18
Java API: fix UUID clone bug

05st:
The physical type of a UUID is `int128_t` but the JNI code was fetching the `std::string` field from the value.


tools/java_api/src/jni/kuzu_java.cpp
```diff
@@ -9,6 +9,8 @@
 // This header is generated at build time. See CMakeLists.txt.
 #include <vector>
 
+#include <iostream>
```
ray6080, line 10: Can you remove this include?

Issue #5513, opened 2025-06-09
Bug: delete_ldbc_sf01.DeleteLikeComment1 test sometimes fails with NODE_GROUP_SIZE_LOG2=4 VECTOR_CAPACITY_LOG2=4


royi-luo:
### Kuzu version

master

### What operating system are you using?

Ubuntu

### What happened?

See https://github.com/kuzudb/kuzu/actions/runs/15535607120/job/43735299396?pr=5443 (this is on a PR but I've been able to reproduce locally on master).

Snippet of test log:

```
[2025-06-09 09:58:40.516] [info] DEBUG LOG: 
[2025-06-09 09:58:40.516] [info] QUERY: MATCH (n:Person)-[e:likes_Comment]->(m:Comment) WHERE n.id=6597069767457 RETURN COUNT(*);
[2025-06-09 09:58:40.530] [info] Query execution took 0.679ms.
[2025-06-09 09:58:40.530] [info] QUERY PASSED.
[2025-06-09 09:58:40.530] [info] DEBUG LOG: 
[2025-06-09 09:58:40.530] [info] QUERY: MATCH (n:Person)-[e:likes_Comment]->(m:Comment) WHERE n.id=6597069767457 AND m.id=412317167195 DELETE e;
[2025-06-09 09:58:40.612] [info] DEBUG LOG: 
[2025-06-09 09:58:40.612] [info] QUERY: MATCH (n:Person)-[e:likes_Comment]->(m:Comment) WHERE n.id=6597069767457 RETURN COUNT(*);
[2025-06-09 09:58:40.613] [info] Query execution took 0.597ms.
Error: 6-09 09:58:40.613] [error] QUERY FAILED.
[2025-06-09 09:58:40.614] [info] PLAN: 
PROJECTION[COUNT_STAR()]
AGGREGATE[Group By [], Aggregate [COUNT_STAR(), ]]
PROJECTION[]
EXTEND[n-e->m]
SCAN_NODE_TABLE[n._ID n.id]
[2025-06-09 09:58:40.614] [info] 66
/home/runner/actions-runner/_work/kuzu/kuzu/test/test_runner/test_runner.cpp:171: Failure
Expected equality of these values:
  resultTuples[i]
    Which is: "66"
  testAnswer.expectedResult[i]
    Which is: "65"
Result tuple at index 0 did not match the expected value
```

### Are there known steps to reproduce?

[...]

PR #5564, opened 2025-06-20
Correctly calculate the number of tuples in a CSR node group after checkpoint

royi-luo:
# Description

Fixes https://github.com/kuzudb/kuzu/issues/5513

The loop used to calculate the number of tuples to checkpoint wasn't including the upper bound of a region's CSR offset. This was sometimes causing node groups to be incorrectly detected as all deleted and reclaimed.

# Contributor agreement


ray6080 (approved):
Thanks!

Issue #2289, opened 2023-10-27
Enabling TSAN to detect spurious buffer manager errors


Riolku:
https://github.com/kuzudb/kuzu/actions/runs/6671688917?pr=2288 showed us that when asserts are enabled, we can get spurious buffer manager failures.

Reproducing this locally proved difficult, so we turned to TSAN, which (after shrinking the 8TB mmap call), gave us a flood of diagnostics.

Steps:

- [ ] Fix the source of TSAN errors
- [ ] Test with TSAN in CI, which requires a switch to change how much memory our buffer manager maps
- [ ] Enable asserts in CI

PR #5569, opened 2025-06-23
Add TSAN workflow in CI + fix warnings

royi-luo:
# Description

Closes https://github.com/kuzudb/kuzu/issues/2289

Title should be pretty self-explanatory. Most of the warnings that are fixed are progress-related, for the most part I just changed the variables used during progress calculations to atomics since the performance impact should be minimal.

# Contributor agreement

### DIFF
--- a/src/include/storage/buffer_manager/spiller.h
+++ b/src/include/storage/buffer_manager/spiller.h
@@ -38,7 +38,7 @@ private:
     BufferManager& bufferManager;
     common::VirtualFileSystem* vfs;
     std::unordered_set<ChunkedNodeGroup*> fullPartitionerGroups;
-    FileHandle* dataFH;
+    std::atomic<FileHandle*> dataFH;
     std::mutex partitionerGroupsMtx;
     mutable std::mutex fileCreationMutex;
 };
--- a/src/storage/buffer_manager/spiller.cpp
+++ b/src/storage/buffer_manager/spiller.cpp
@@ -25,12 +25,12 @@ Spiller::Spiller(std::string tmpFilePath, BufferManager& bufferManager,
 }
 
 FileHandle* Spiller::getDataFH() const {
-    if (dataFH) {
+    if (dataFH.load()) {
         return dataFH;
     }
     std::unique_lock lock(fileCreationMutex);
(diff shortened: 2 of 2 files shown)


src/processor/operator/persistent/reader/csv/serial_csv_reader.cpp
```diff
@@ -129,6 +129,9 @@ void SerialCSVScanSharedState::initReader(main::ClientContext* context) {
             std::make_unique<SharedFileErrorHandler>(fileIdx, nullptr, populateErrorFunc);
         localErrorHandler = std::make_unique<LocalFileErrorHandler>(sharedErrorHandler.get(),
             csvOption.ignoreErrors, context);
+        // We only need to protect the access to the reader here are that can cause a data race with
+        // the progressFunc
+        std::lock_guard lck{mtx};
```
royi-luo, line 134: Added a mutex lock to protect the reader being recreated here since we shouldn't care about parallel performance when using the serial CSV reader anyways.

src/processor/operator/persistent/reader/csv/serial_csv_reader.cpp
```diff
@@ -85,6 +85,7 @@ SerialCSVScanSharedState::SerialCSVScanSharedState(FileScanInfo fileScanInfo, ui
     : ScanFileWithProgressSharedState{std::move(fileScanInfo), numRows, context},
       csvOption{std::move(csvOption)}, columnInfo{std::move(columnInfo)}, totalReadSizeByFile{0},
       queryID(queryID), populateErrorFunc(constructPopulateFunc()) {
+    std::lock_guard lck{mtx};
```
royi-luo, line 88: Added a mutex lock here to protect `initReader` which can reset the reader in the shared state (which can then be read by the progressFunc). We shouldn't care about parallel performance when using the serial CSV reader anyways so adding a lock here shouldn't be a problem.

ray6080 (approved):
@acquamarin can you take another look at changes in reader?

acquamarin (approved):
can you check whether there is a performance regression using atomic?

PR #5570, opened 2025-06-23
Add recovery test CI and a bunch of fixes

ray6080:
# Description


src/include/main/database.h
```diff
@@ -71,6 +73,7 @@ struct KUZU_API SystemConfig {
     uint64_t maxDBSize;
     bool autoCheckpoint;
     uint64_t checkpointThreshold;
+    bool forceCheckpointOnClose;
```
andyfengHKU, line 76: I recommend we organize configurations with the same type together.

src/storage/wal/wal_record.cpp
```diff
@@ -307,6 +307,7 @@ std::unique_ptr<TableInsertionRecord> TableInsertionRecord::deserialize(Deserial
     deserializer.validateDebuggingInfo(key, "num_vectors");
     deserializer.deserializeValue(numVectors);
     auto resultChunkState = std::make_shared<DataChunkState>();
+    resultChunkState->setToFlat();
```
andyfengHKU, line 310: use `getSingleValueDataChunkState`

src/storage/storage_manager.cpp
```diff
@@ -43,10 +45,19 @@ void StorageManager::initDataFileHandle(VirtualFileSystem* vfs, main::ClientCont
...
         const auto dataFilePath = StorageUtils::getDataFName(vfs, databasePath);
         dataFH = memoryManager.getBufferManager()->getFileHandle(dataFilePath, flag, vfs, context);
-    }
-    if (dataFH->getNumPages() == 0) {
-        // Reserve the first page for the database header.
-        dataFH->getPageManager()->allocatePage();
+        if (dataFH->getNumPages() == 0) {
+            if (!readOnly) {
+                // Reserve the first page for the database header.
+                dataFH->getPageManager()->allocatePage();
```
andyfengHKU, line 51: Someone else should double check this. I don't why we are writing dummy header here.

src/include/parser/expression/parsed_function_expression.h
```diff
@@ -35,6 +35,11 @@ class ParsedFunctionExpression : public ParsedExpression {
         : ParsedExpression{expressionType_}, isDistinct{isDistinct},
           functionName{std::move(functionName)} {}
 
+    ParsedFunctionExpression(std::string functionName, std::vector<std::string> optionalArguments,
```
andyfengHKU, line 38: Can u use `addOptionalParams`. I kinda wanna reduce the number of constructors for this class.

src/parser/expression/parsed_function_expression.cpp
```diff
@@ -14,12 +14,17 @@ std::unique_ptr<ParsedFunctionExpression> ParsedFunctionExpression::deserialize(
     deserializer.deserializeValue(isDistinct);
     std::string functionName;
     deserializer.deserializeValue(functionName);
-    return std::make_unique<ParsedFunctionExpression>(std::move(functionName), isDistinct);
+    std::vector<std::string> optionalArguments;
```
andyfengHKU, line 17: I assume this relates to a bug where STRUCT is used as default value. Though I don't see a corresponding test.

andyfengHKU (approved):

src/include/parser/expression/parsed_function_expression.h
ray6080: Not really. I'd have to add a new function to set `optionalArguments` only.

src/parser/expression/parsed_function_expression.cpp
ray6080: The new CI jobs have covered it with existing test cases.

src/storage/storage_manager.cpp
ray6080: I think @royi-luo or @benjaminwinger can take a look. This is to handle the case that COPY into an empty database fails during checkpoint, which leaves the header uninitialized on disk, then following checkpoint will fail. An alternative (might be better) way is to keep a `DatabaseHeader` field under `Database` and reset it to a new one after checkpoint. I should get it refactored in a later PR.

royi-luo: Yep the alternative sounds good to me, I think there are a few spots where it's awkward because we don't have an in-memory version of the database header so it would be nice to have one under `Database`

Issue #5565, opened 2025-06-21
Feature: Consistentize array_slice() behavior on STRING vs. on LIST


poiurewq:
### API

Other

### Description

This is an observation of kuzu's implementation. I don't know how others implement it, but kuzu's array_slice() behavior _excludes_ the stop index when slicing a LIST but _includes_ the stop index when slicing a STRING. Is this due to a design choice, something inherited from Neo4j, or a kuzu quirk? For user convenience is it advisable to consistentize the behavior across logical types?

Example commands to demonstrate:

`RETURN array_slice("abcd",1,4);` => `abcd` includes element at end index of the slice

`RETURN array_slice(['a','b','c','d'],1,4);` => `[a,b,c]` excludes element at end index of the slice

This dichotomy, as expected, is also observed for the aliases of the above functions.

`RETURN STRING("abcd")[1:4];` => `abcd`

`RETURN list_slice(['a','b','c','d'],1,4);` => `[a,b,c]`

andyfengHKU:
This should be fixed in #5573

PR #5573, opened 2025-06-23
Fix slice operator inconsistencies

05st:
This PR fixes inconsistencies with the slicing operator by following what DuckDB does.
For https://github.com/kuzudb/kuzu/issues/5565

### DIFF
--- a/src/function/list/list_slice_function.cpp
+++ b/src/function/list/list_slice_function.cpp
@@ -9,7 +9,6 @@ namespace kuzu {
 namespace function {
 
 static void normalizeIndices(int64_t& startIdx, int64_t& endIdx, uint64_t size) {
-    int64_t sSize = size;
     if (startIdx < 0) {
         startIdx = size + startIdx + 1;
     }
@@ -19,10 +18,10 @@ static void normalizeIndices(int64_t& startIdx, int64_t& endIdx, uint64_t size) 
     if (endIdx < 0) {
         endIdx = size + endIdx + 1;
     }
-    if (endIdx > sSize) {
+    if (endIdx > (int64_t)size) {
         endIdx = size;
     }
-    if (endIdx < startIdx || startIdx > sSize) {
+    if (endIdx < startIdx) {
         startIdx = 1;
         endIdx = 0;
     }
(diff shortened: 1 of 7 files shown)


src/function/list/list_slice_function.cpp
```diff
@@ -9,23 +9,22 @@ namespace kuzu {
 namespace function {
 
 static void normalizeIndices(int64_t& startIdx, int64_t& endIdx, uint64_t size) {
+    int64_t sSize = size;
```
andyfengHKU, line 12: Do we actually need this because size is an invariant in this function as I understand.

src/parser/transform/transform_expression.cpp
```diff
@@ -312,15 +312,15 @@ std::unique_ptr<ParsedExpression> Transformer::transformListOperatorExpression(
             left = transformExpression(*ctx.oC_Expression(0));
             right = transformExpression(*ctx.oC_Expression(1));
         } else if (ctx.oC_Expression().size() == 0) { // [:]
-            left = std::make_unique<ParsedLiteralExpression>(Value(0), "0");
-            right = std::make_unique<ParsedLiteralExpression>(Value(0), "0");
+            left = std::make_unique<ParsedLiteralExpression>(Value(1), "1");
```
andyfengHKU, line 315: I thought we only change the logic for handling right value. Do left value needs to change as well?

src/function/list/list_slice_function.cpp
```diff
@@ -9,23 +9,22 @@ namespace kuzu {
...
+    if (endIdx < 0) {
+        endIdx = size + endIdx + 1;
     }
-
-    if (startIdx > endIdx) {
-        endIdx = startIdx;
+    if (endIdx > sSize) {
+        endIdx = size;
+    }
+    if (endIdx < startIdx || startIdx > sSize) {
```
andyfengHKU, line 25: I believe you don't need or condition. L22 already guarantees `endIdx <= size`. If `startIdx > size` then it must also satisfy `startIdx > endIdx`. So just keep the first condition (or we should the original check)

andyfengHKU (changes requested):

src/function/list/list_slice_function.cpp
05st: Removed with https://github.com/kuzudb/kuzu/pull/5573/commits/879b67ab0554dabb39b67f1b5a7fd8c64455eae1

andyfengHKU (approved):

acquamarin (commented):
Can you add one more test on string slice?
```
RETURN "abc"[-1:2]
```
This should return an empty string but return 'c' right now

PR #5576, opened 2025-06-24
Reenable some BM exception handling tests

royi-luo:
# Description

Fixes the reason why `GracefulBMExceptionHandlingManyThreads` was sometimes failing in in-memory mode. There were a few issues:

1. In `FileHandle::addNewPages` it is possible for `addNewPageWithoutLock` to throw a BM exception if we are in in-mem mode (since we pin a page). In that case previously we would never unlock the lock.
2. We slice out the columns containing data for populating errors in node batch insert. However the slice wasn't restored to the original chunked group if a BM exception was thrown which could cause other threads to read invalid data before they detect that an exception is thrown.

# Contributor agreement

### DIFF
--- a/src/include/common/finally_wrapper.h
+++ b/src/include/common/finally_wrapper.h
@@ -1,0 +1,14 @@
+#pragma once
+
+#include <concepts>
+
+namespace kuzu::common {
+// RAII wrapper that calls an enclosed function when this class goes out of scope
+// Should be used for any cleanup code that must be executed even if exceptions occur
+template<std::invocable Func>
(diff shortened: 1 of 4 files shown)


ray6080 (approved):
Cool! I was uncommenting `NodeInsertBMExceptionDuringCheckpointRecovery` in my other PR too 😂

benjaminwinger (approved):
Looks great!
I hadn't realised that std::invocable exists. We could probably make use of that elsewhere.

ray6080 (approved):

PR #5581, opened 2025-06-24
Fix serializing FSM

royi-luo:
# Description

We serialized slightly out-of-date data for the FSM because the order of the serializing is:
1. We serialize the FSM
2. We allocate the pages for the serialized FSM (which may alter the FSM)
3. Checkpoint ends (the state of the FSM may be slightly different from the serialized version)

This PR fixes this by allocating the pages for the serialized FSM before the serializing actually happens (the estimate the number of pages for serializing the FSM beforehand).

# Contributor agreement

### DIFF
--- a/src/include/storage/free_space_manager.h
+++ b/src/include/storage/free_space_manager.h
@@ -31,6 +31,7 @@ public:
     void addUncheckpointedFreePages(PageRange entry);
     void rollbackCheckpoint();
 
+    common::page_idx_t estimateNumPagesNeededForSerialize() const;
     void serialize(common::Serializer& serializer) const;
     void deserialize(common::Deserializer& deSer);
     void finalizeCheckpoint(FileHandle* fileHandle);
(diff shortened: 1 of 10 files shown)


src/include/storage/free_space_manager.h
```diff
@@ -31,6 +31,7 @@ class FreeSpaceManager {
     void addUncheckpointedFreePages(PageRange entry);
     void rollbackCheckpoint();
 
+    common::page_idx_t estimateNumPagesNeededForSerialize() const;
```
ray6080, line 34: I feel `getMaxNumPages` is more accurate than `estimateNumPages`. => `getMaxNumPagesForSerialization`

src/storage/checkpointer.cpp
```diff
@@ -65,7 +65,21 @@ PageRange Checkpointer::serializeMetadata(const catalog::Catalog& catalog,
     auto metadataWriter = std::make_shared<common::MetaWriter>(clientContext.getMemoryManager());
     common::Serializer metadataSerializer(metadataWriter);
     storageManager.serialize(catalog, metadataSerializer);
-    return metadataWriter->flush(storageManager.getDataFH(), storageManager.getShadowFile());
+
+    // We need to preallocate the pages for the page manager before we actually serialize it
+    // This is because the page manager needs to track the pages used for itself
+    // The number of pages needed for the page manager should only decrease after making an
+    // additional allocation so we just calculate the number of pages needed to serialize the
+    // current state of the page manager
```
ray6080, line 73: Can you also comment that there might be one more page beyond the num pages in physical file being marked as free page?

ray6080 (approved):

src/include/storage/free_space_manager.h
royi-luo: updated

src/storage/checkpointer.cpp
royi-luo: updated

PR #5619, opened 2025-06-30
LLM ollama endpoint config

tgahunia05:
# Description

Implements endpoint config for the Ollama provider. Default is now the Ollama default of `http://localhost:11434`. Scalar functions in kuzu do not support optional configuration like GDS functions do. I had to update the region variable to be `regionOrEndpoint` this is ugly but I don't see it changing until scalar functions support` option_name := option_value` syntax. 
Fixes tests with outdated error messages from error test. 

### DIFF
--- a/extension/llm/src/function/create_embedding.cpp
+++ b/extension/llm/src/function/create_embedding.cpp
@@ -94,20 +94,20 @@ static uint64_t parseDimensions(std::shared_ptr<Expression> dimensionsExpr,
 
 static std::unique_ptr<FunctionBindData> bindFunc(const ScalarBindFuncInput& input) {
     std::optional<uint64_t> dimensions = std::nullopt;
-    std::optional<std::string> region = std::nullopt;
+    std::optional<std::string> regionOrEndpoint = std::nullopt;
     auto& provider = getInstance(StringUtils::getLower(input.arguments[1]->toString()));
     if (input.arguments.size() == 5) {
         dimensions = parseDimensions(input.arguments[3], input.context);
-        region = StringUtils::getLower(input.arguments[4]->toString());
+        regionOrEndpoint = StringUtils::getLower(input.arguments[4]->toString());
     } else if (input.arguments.size() == 4) {
         if (input.arguments[3]->dataType == LogicalType(LogicalTypeID::STRING)) {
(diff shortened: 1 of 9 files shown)


extension/llm/src/function/create_embedding.cpp
```diff
@@ -94,20 +94,20 @@ static uint64_t parseDimensions(std::shared_ptr<Expression> dimensionsExpr,
 
 static std::unique_ptr<FunctionBindData> bindFunc(const ScalarBindFuncInput& input) {
     std::optional<uint64_t> dimensions = std::nullopt;
-    std::optional<std::string> region = std::nullopt;
+    std::optional<std::string> regionOrEndpoint = std::nullopt;
```
sdht0, line 97: How about we call this stringConfig/numConfig here?
provider.configure can then use the generic names instead.

extension/llm/src/function/create_embedding.cpp
```diff
@@ -94,20 +94,20 @@ static uint64_t parseDimensions(std::shared_ptr<Expression> dimensionsExpr,
...
     auto& provider = getInstance(StringUtils::getLower(input.arguments[1]->toString()));
     if (input.arguments.size() == 5) {
         dimensions = parseDimensions(input.arguments[3], input.context);
-        region = StringUtils::getLower(input.arguments[4]->toString());
+        regionOrEndpoint = StringUtils::getLower(input.arguments[4]->toString());
     } else if (input.arguments.size() == 4) {
         if (input.arguments[3]->dataType == LogicalType(LogicalTypeID::STRING)) {
```
sdht0, line 103: Make sure the arguments are processed in well determined order. Add tests for all order combinations, including from 0-3 optional configurations of different types.

sdht0 (approved):
Idea seems okay to me.

extension/llm/src/function/create_embedding.cpp
sdht0: For providers, instead of giving an error saying `dimensions not expected`, use the same style as the other PR.
```
Unexpected arguments:
actual:      (string, string, string, bool)
expected: (string, string, string, string)
                   (string, string, string, int, string)
```

PR #5633, opened 2025-07-03
Remove tests on installing extension from local http server

ray6080:
# Description

See title. Remove these tests for now as they run into failures frequently due to CI setup.


acquamarin (approved):

ray6080:
Close this for now as I realized that we have other tests other than extension install tests relying on the local http server.

PR #5652, opened 2025-07-04
Add tests for docs examples for passing dataframes as parameters

royi-luo:
# Description

Add tests for examples added in [this docs PR](https://github.com/kuzudb/kuzu-docs/pull/546)

# Contributor agreement

### DIFF
--- a/tools/python_api/test/test_scan_polars.py
+++ b/tools/python_api/test/test_scan_polars.py
@@ -151,15 +151,16 @@ def test_scan_from_parameterized_df_docs_example_2(conn_db_empty: ConnDB):
     conn.execute("CREATE NODE TABLE Person(name STRING, age INT64, PRIMARY KEY (name))")
 
     def get_df():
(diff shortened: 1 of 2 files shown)


tools/python_api/test/test_scan_polars.py
```diff
@@ -1,5 +1,6 @@
 from datetime import datetime
 
+import kuzu
```
ray6080, line 3: Double check why we need this include now?

ray6080 (approved):
Should we cover this in async connection test as well?

tools/python_api/test/test_scan_polars.py
royi-luo: removed

royi-luo:
> Should we cover this in async connection test as well?

The async python code would look a bit different and the cases should be covered by the other async tests as well, this PR is just to make sure the exact code in the docs will work as expected. So I think I'll skip adding those tests here.

PR #5665, opened 2025-07-07
Implement database lifecycle manager

mewim:


sdht0:
Will this resolve https://github.com/kuzudb/kuzu-docs/issues/479?

src/include/main/query_result.h
```diff
@@ -6,9 +6,9 @@
 #include "common/arrow/arrow.h"
 #include "common/types/types.h"
 #include "kuzu_fwd.h"
+#include "main/database.h"
```
ray6080, line 9: Can you move `DatabaseLifeCycleManager` into its own header file and we can get rid of the include of `main/database.h`?

test/api/api_test.cpp
```diff
@@ -316,3 +316,31 @@ TEST_F(ApiTest, MissingParam) {
     ASSERT_FALSE(prep->isSuccess());
     ASSERT_STREQ("Parameter val1 not found.", prep->getErrorMessage().c_str());
 }
+
+TEST_F(ApiTest, CloseDatabaseBeforeQueryResultAndConnection) {
```
ray6080, line 320: Can we cover more public interfaces of `QueryResult` and `Connection` in tests?

ray6080 (approved):

src/include/main/query_result.h
mewim: Done

test/api/api_test.cpp
mewim: Done

PR #5669, opened 2025-07-07
Export and Import from Prev Release

tgahunia05:
# Description
Provides a script that builds Kuzu from `base-commit` and exports all datasets in `dataset-dir` using `scripts/export-dbs.py`. The datasets are placed in `output-dir/{base-commit-version}`. The script then runs the test suite in `test-commit` using `make test` and the exported datasets. Cleans by removing `output-dir/{base-commit-version}` and checking out to the commit where the script was initially ran. 

```
Export DBs from dataset-dir to output-dir using base-commit and test in test-commit

options:
  -h, --help            show this help message and exit
  --base-commit BASE_COMMIT
                        Git commit to export databases from
  --test-commit TEST_COMMIT
                        Git commit to test against
  --dataset-dir DATASET_DIR
                        Path to the dataset directory
  --output-dir OUTPUT_DIR
                        Path to output the exported databases
  --cleanup CLEANUP     Delete exported DBs after test (default: True)
```
  

I am attaching a log file from a test run of the script. 

`python -u scripts/export-import-test.py --base-commit de79e0b --test-commit master --dataset-dir dataset --output-dir dataset`

[out.txt](https://github.com/user-attachments/files/21109601/out.txt)

# Contributor agreement

### DIFF
--- a/scripts/export-import-test.py
+++ b/scripts/export-import-test.py
@@ -90,8 +90,6 @@ def main():
                 cwd=kuzu_root,
             )
             os.rename(inprogress_path, export_path + os.sep)
-        # Switch back to working branch (to use the latest script)
-        run_command(f"git checkout {current_branch}", cwd=kuzu_root)
 
         # Checkout commit B and run tests
         run_command(f"git checkout {test_commit}", cwd=kuzu_root)
--- a/scripts/export-import-test.py
+++ b/scripts/export-import-test.py
@@ -8,10 +8,7 @@ import shutil
 def get_version(executable_path):
     try:
         result = subprocess.run(
-            [executable_path, "--version"],
-            capture_output=True,
-            text=True,
-            check=True
+            executable_path, " --version", capture_output=True, text=True, check=True
         )
         output = result.stdout.strip()
         if output.startswith("Kuzu "):
@@ -31,7 +28,9 @@ def run_command(cmd, cwd=None, capture_output=False):
     print(f"> Running: {' '.join(cmd)} (cwd={cwd})")
 
     if capture_output:
-        result = subprocess.run(cmd, cwd=cwd, text=True, capture_output=True, check=True)
+        result = subprocess.run(
+            cmd, cwd=cwd, text=True, capture_output=True, check=True
+        )
         return result.stdout.strip()
     else:
         process = subprocess.Popen(
@@ -42,11 +41,11 @@ def run_command(cmd, cwd=None, capture_output=False):
             stderr=subprocess.STDOUT,
             bufsize=1,
             universal_newlines=True,
-            stdin=subprocess.DEVNULL
+            stdin=subprocess.DEVNULL,
         )
 
         for line in process.stdout:
-            print(line, end='')
+            print(line, end="")
 
         process.stdout.close()
         retcode = process.wait()
@@ -59,11 +58,24 @@ def main():
     parser = argparse.ArgumentParser(
         description="Export DBs from dataset-dir to output-dir using base-commit and test in test-commit"
     )
-    parser.add_argument("--base-commit", required=True, help="Git commit to export databases from")
-    parser.add_argument("--test-commit", required=True, help="Git commit to test against")
-    parser.add_argument("--dataset-dir", required=True, help="Path to the dataset directory")
-    parser.add_argument("--output-dir", required=True, help="Path to output the exported databases")
-    parser.add_argument("--cleanup", type=bool, default=True, help="Delete exported DBs after test (default: True)")
+    parser.add_argument(
+        "--base-commit", required=True, help="Git commit to export databases from"
+    )
+    parser.add_argument(
+        "--test-commit", required=True, help="Git commit to test against"
+    )
+    parser.add_argument(
+        "--dataset-dir", required=True, help="Path to the dataset directory"
+    )
+    parser.add_argument(
+        "--output-dir", required=True, help="Path to output the exported databases"
+    )
+    parser.add_argument(
+        "--cleanup",
+        type=bool,
+        default=True,
+        help="Delete exported DBs after test (default: True)",
+    )
     args = parser.parse_args()
 
     base_commit = args.base_commit
@@ -74,7 +86,9 @@ def main():
 
     script_dir = os.path.dirname(os.path.realpath(__file__))
     kuzu_root = os.path.abspath(os.path.join(script_dir, ".."))
-    current_branch = run_command(["git", "rev-parse", "--abbrev-ref", "HEAD"], cwd=kuzu_root, capture_output=True)
+    current_branch = run_command(
+        ["git", "rev-parse", "--abbrev-ref", "HEAD"], cwd=kuzu_root, capture_output=True
+    )
     try:
         # Checkout commit A and build
         run_command("git checkout " + base_commit, cwd=kuzu_root)
@@ -85,7 +99,9 @@ def main():
 
         # Export databases
         export_script_path = os.path.join(kuzu_root, "scripts", "export-dbs.py")
-        exec_path = os.path.join(kuzu_root, "build", "relwithdebinfo", "tools", "shell", "kuzu")
+        exec_path = os.path.join(
+            kuzu_root, "build", "relwithdebinfo", "tools", "shell", "kuzu"
+        )
         # Determine export path based on version (taken from export_script)
         version = get_version(exec_path)
         if version is None:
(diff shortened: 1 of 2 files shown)


scripts/export-import-test.py
```diff
@@ -0,0 +1,121 @@
...
+        if retcode != 0:
+            raise subprocess.CalledProcessError(retcode, cmd)
+        return None
+
+
+def main():
+    parser = argparse.ArgumentParser(
+        description="Export DBs from datasetDir using commitA and test in commitB"
+    )
+    parser.add_argument("commitA", help="Git commit to export databases from")
```
sdht0, line 64: Can you add descriptive flags: `--base-commit` `--test-commit` `--dataset-dir` `--output-dir`

scripts/export-import-test.py
```diff
@@ -0,0 +1,121 @@
...
+        env = os.environ.copy()
+        env["E2E_IMPORT_DB_DIR"] = exportPath
+        runCommand(["make", "test"], cwd=kuzuRoot, env=env)
+
+        # Restore original branch
+        runCommand(["git", "checkout", currentBranch], cwd=kuzuRoot)
+
+        return 0
+    finally:
+        tmpDir = os.path.join(datasetDir, "tmp")
```
sdht0, line 108: The cleanup should be done optionally with a flag I think. Say `--cleanup` , defaults to True.
If a dev wants to run the tests again and again, they shouldn't have to export multiple times.

scripts/export-import-test.py
```diff
@@ -0,0 +1,121 @@
...
+    scriptDir = os.path.dirname(os.path.realpath(__file__))
+    kuzuRoot = os.path.abspath(os.path.join(scriptDir, ".."))
+    currentBranch = runCommand(["git", "rev-parse", "--abbrev-ref", "HEAD"], cwd=kuzuRoot, capture_output=True)
+
+    try:
+        # Checkout commit A and build
+        runCommand(["git", "checkout", commitA], cwd=kuzuRoot)
+        runCommand(["make", "extension-build", "EXTENSION_LIST=json"], cwd=kuzuRoot)
+
+        # Switch back to working branch (to use the following script)
```
sdht0, line 82:
```suggestion
        # Switch back to working branch (to use the latest script)
```

scripts/export-import-test.py
```diff
@@ -0,0 +1,121 @@
...
+        runCommand(["git", "checkout", currentBranch], cwd=kuzuRoot)
+
+        # Export databases
+        exportScriptPath = os.path.join(kuzuRoot, "scripts", "export-dbs.py")
+        execPath = os.path.join(kuzuRoot, "build", "relwithdebinfo", "tools", "shell", "kuzu")
+        runCommand(["python3", exportScriptPath, execPath, datasetDir], cwd=kuzuRoot)
```
sdht0, line 88: Set the export path to `--output-dir / base-commit-version`.
Also check if this directory is already present and skip exporting if exists.

To make it robust, here's the workflow I imagine happening:
1. Check if `--output-dir / base-commit` exists
  a. If it exists, do nothing
  b. If it does not exist, checkout base commit and run export for `--output-dir / base-commit-version_inprogress`. Perhaps also remove any existing path.
c. If export completes successfully, rename `--output-dir / base-commit-version_inprogress` to `--output-dir / base-commit-version`. This is to make the export **atomic**.
2. Now run the tests

scripts/export-import-test.py
```diff
@@ -0,0 +1,121 @@
...
+        cmd = cmd.split()
+
+    print(f"> Running: {' '.join(cmd)} (cwd={cwd})")
+
+    if capture_output:
+        result = subprocess.run(cmd, cwd=cwd, env=env, text=True, capture_output=True, check=True)
+        print(result.stdout)
+        return result.stdout.strip()
+    else:
+        process = subprocess.Popen(
```
sdht0, line 38: Do we really need 2 different commands based on capture_output?

scripts/export-import-test.py
```diff
@@ -0,0 +1,153 @@
...
+            print("Unexpected version format:", output)
+            return None
+    except subprocess.CalledProcessError as e:
+        print(f"Error running '{executable_path} --version': {e}")
+        return None
+
+
+def run_command(cmd, cwd=None, capture_output=False):
+    if isinstance(cmd, str):
+        cmd = cmd.split()
```
sdht0, line 30: I think this is not needed with `text=True`?

scripts/export-import-test.py
```diff
@@ -0,0 +1,153 @@
...
+import argparse
+import subprocess
+import os
+import shutil
+
+
+def get_version(executable_path):
+    try:
+        result = subprocess.run(
+            executable_path + " --version",
```
sdht0, line 11: nit: use `f""` strings everywhere

scripts/export-import-test.py
```diff
@@ -0,0 +1,153 @@
...
+        cwd=cwd,
+        text=True,
+        capture_output=capture_output,
+        check=True,
+        stdin=subprocess.DEVNULL,
+    )
+    if capture_output:
+        return result.stdout.strip()
+    else:
+        print(result.stdout or "")
```
sdht0, line 45: Won't this always be empty as you're redirecting stdin to devnull?

scripts/export-import-test.py
```diff
@@ -0,0 +1,153 @@
...
+    base_commit = args.base_commit
+    test_commit = args.test_commit
+    dataset_dir = args.dataset_dir
+    output_dir = args.output_dir
+    cleanup = args.cleanup
+
+    script_dir = os.path.dirname(os.path.realpath(__file__))
+    kuzu_root = os.path.abspath(os.path.join(script_dir, ".."))
+    current_branch = run_command(
+        ["git", "rev-parse", "--abbrev-ref", "HEAD"], cwd=kuzu_root, capture_output=True
```
sdht0, line 82: Use string instead of array?

scripts/export-import-test.py
```diff
@@ -0,0 +1,153 @@
...
+    )
+    try:
+        # Checkout commit A and build
+        run_command("git checkout " + base_commit, cwd=kuzu_root)
```
sdht0, line 86: I think all this code can be moved under the conditional "Only run export if export_path DNE" `if` below

scripts/export-import-test.py
```diff
@@ -0,0 +1,153 @@
...
+        run_command("make extension-build EXTENSION_LIST=json", cwd=kuzu_root)
```
sdht0, line 87: Why is only json being built? Add a comment.

scripts/export-import-test.py
```diff
@@ -0,0 +1,153 @@
...
+
+        # Switch back to working branch (to use the latest script)
+        run_command("git checkout " + current_branch, cwd=kuzu_root)
+
+        # Export databases
```
sdht0, line 92: This comment belongs inside the if

scripts/export-import-test.py
```diff
@@ -0,0 +1,153 @@
...
+        version = get_version(exec_path)
+        if version is None:
+            raise Exception("Failed to determine version. Aborting.")
+        export_path = os.path.join(output_dir, version) + os.sep
+
+        # Only run export if export_path DNE
+        if not os.path.exists(export_path):
+            inprogress_path = export_path + "_inprogress" + os.sep
+            run_command(
+                "python3 "
```
sdht0, line 107: Same, use multiline `f""" """` here if needed

scripts/export-import-test.py
```diff
@@ -0,0 +1,153 @@
...
+                + export_script_path
+                + " --executable "
+                + exec_path
+                + " --dataset-dir "
+                + dataset_dir
+                + " --output-dir "
+                + inprogress_path,
+                cwd=kuzu_root,
+            )
+            for item in os.listdir(inprogress_path):
```
sdht0, line 117: Don't need to loop. Just rename `inprogress_path` to `export_path`. Do check if `export_path` already exists and either delete or throw an error.

scripts/export-import-test.py
```diff
@@ -0,0 +1,153 @@
...
+
+        print(f"Restoring original git branch: {current_branch}")
+        try:
+            run_command("git checkout " + current_branch, cwd=kuzu_root)
+        except Exception as e:
+            print(f"Warning: Failed to restore branch {current_branch}: {e}")
+
+
+if __name__ == "__main__":
+    sys.exit(main())
```
sdht0, line 153: With the exception based logic, explicit returns and sys.exit here can be removed I'd say.

sdht0 (approved):

scripts/export-import-test.py
sdht0: Sorry, I was thinking about `shell=True`. Use that and avoid arrays based commands.

scripts/export-import-test.py
tgahunia05: Added a comment, leaving an explanation here too. Datasets tinysnb_json and shortest-path-test-json use `load extension {json_path}` in copy.cypher files. The database cannot be properly exported without this extension.

scripts/export-import-test.py
tgahunia05: Export path includes a version number. The version number is determined by checking out the commit, building and running {path to kuzu shell} --version. If there is a better way to determine version number that avoids this build process then I can rework this like you suggested.

scripts/export-import-test.py
tgahunia05: Was addressed in chats. Leaving a comment about why stdin is redirected to dev null.

scripts/export-import-test.py
tgahunia05: I have done a search for these array based commands and removed them.

PR #5728, opened 2025-07-10
Enable extension loading in read-only database

ray6080:
# Description

The previous change to extension transaction accidentally break extension loading under read-only mode. This PR fixes the issue.

### DIFF
--- a/src/include/extension/extension.h
+++ b/src/include/extension/extension.h
@@ -53,14 +53,14 @@ struct ExtensionSourceUtils {
 };
 
 template<typename T>
-void addFunc(transaction::Transaction* transaction, main::Database& database, std::string name,
-    catalog::CatalogEntryType functionType, bool isInternal = false) {
+void addFunc(main::Database& database, std::string name, catalog::CatalogEntryType functionType,
+    bool isInternal = false) {
     auto catalog = database.getCatalog();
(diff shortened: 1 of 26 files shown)


src/include/extension/extension.h
```diff
@@ -53,14 +53,14 @@ struct ExtensionSourceUtils {
 };
 
 template<typename T>
-void addFunc(transaction::Transaction* transaction, main::Database& database, std::string name,
-    catalog::CatalogEntryType functionType, bool isInternal = false) {
+void addFunc(main::Database& database, std::string name, catalog::CatalogEntryType functionType,
+    bool isInternal = false) {
     auto catalog = database.getCatalog();
-    if (catalog->containsFunction(transaction, name, isInternal)) {
+    if (catalog->containsFunction(&transaction::DUMMY_TRANSACTION, name, isInternal)) {
```
sdht0, line 59: Can `containsFunction` and `addFunction` not work without accepting a transaction?

sdht0 (approved):
Seems straightforward.

I don't see any tests?

ray6080:
> Seems straightforward.
> 
> I don't see any tests?

Still fixing the infra for the tests.

src/include/extension/extension.h
ray6080: From the interface design perspective, they should always require a transaction.

PR #5754, opened 2025-07-14
Fix syntax highlighting for quotes

### LINKED ISSUE #5686: Bug: incorrect syntax highlighting in shell
Image

I believe there should be no red highlighting.
It seems like parentheses inside string literals are not handled properly, and neither are single quote string literals containing an escaped single quote.

05st:
It was skipping the `'` or `"` character which prevented the state from being set back to `STANDARD`.

Resolves #5686

### DIFF
--- a/tools/shell/linenoise.cpp
+++ b/tools/shell/linenoise.cpp
@@ -1701,13 +1701,11 @@ void addErrorHighlighting(uint64_t render_start, uint64_t render_end,
                     comment_start.push_back(i);
                     i++;
                     state = ScanState::IN_COMMENT;
-                    break;
                 } else if (i + 1 < l->len && l->buf[i + 1] == '*') {
                     // /* puts us in a multiline comment
                     comment_start.push_back(i);
                     i++;
                     state = ScanState::IN_MULTILINE_COMMENT;
-                    break;
                 }
                 break;
             case '\'':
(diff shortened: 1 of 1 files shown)


tools/shell/linenoise.cpp
```diff
@@ -1764,28 +1762,16 @@ void addErrorHighlighting(uint64_t render_start, uint64_t render_end,
...
-                }
+            if (c == '\'' && (i == 0 || l->buf[i - 1] != '\\')) {
+                // revert to standard scan state if not escaped quote
+                state = ScanState::STANDARD;
             }
             break;
         case ScanState::IN_DOUBLE_QUOTE:
             // double quote - all that will get us out is an unescaped quote
-            if (c == '"') {
-                if (i > 0 && l->buf[i - 1] == '\\') { // escaped quote
```
sdht0, line 1781: What happens for the else case now, when the state is not reset?

sdht0 (approved):
Looks okay, although I haven't looked at the overall logic.

Can you explain in a comment the before and after behavior?

tools/shell/linenoise.cpp
05st: Suppose the buffer looks like this: `\''`. Previously, when we were at the first `'` character, we incremented `i`. This resulted in skipping the next `'` character for the next iteration which was supposed to end the quote and set the state back to `STANDARD`.

In the else case, the character is part of the string (so either a non-quote character or an escaped quote). So we don't do anything.

PR #5757, opened 2025-07-14
test-tools: update rewriter behaviour for error (regex)

### LINKED ISSUE #5739: Bug: Update Test Rewriter
### Kuzu version

0.10.0.6

MacOS

### What happened?

For post 0.11.0
I experienced this using the test rewriter to update the badkeys test in the llm extension.

The test rewriter has an unreasonable way of handling error regex. What it should do is attempt to match the actual error message with the expected error regex. 

If the message does not fix the regex it should update the test with the produced error message (overwriting the regex in the process).

What it currently does is keep the regex so long as the actual output is an error (no matching attempts are made). This is unreasonable and can mask away issues like tests needing updates. 

_No response_

tgahunia05:
# Description
[Do not review; in test]
Update the rewriter to do the following for error(regex)

1. If the query succeeds (no error). Replace the output with 
`---- ok`

2. If the query fails and matches the regex.
Leave the output as is (i.e write the same expected error back).
`---- error(regex)`
`{old error regex}`

3. If the query fails and does not match the regex
Overwrite the output with 
`---- error`
`{query error msg}`

Closes #5739 

# Contributor agreement

### DIFF
--- a/test/runner/e2e_test.cpp
+++ b/test/runner/e2e_test.cpp
@@ -247,8 +247,14 @@ private:
             // ---- error(regex)
             // ^Error: Binder exception: Variable .* is not in scope\.$
             case ResultType::ERROR_REGEX: {
-                newFile += statement->newOutput;
-                skipExistingOutput(file);
+                auto lines = skipExistingOutput(file);
+                // See ERROR_MSG
+                if (lines.find("${") != std::string::npos) {
+                    newFile += currLine + '\n' + lines;
+                } else {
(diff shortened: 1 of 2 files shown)


tgahunia05:
This is failing `E2E_REWRITE_TESTS=1 make test TEST_JOBS=1`
Expected output on regex tests are overwritten with repeated `----ok`.

ray6080:
Close for now as I don't think we will get to it in the near future.

PR #5774, opened 2025-07-16
cli: create history file in home directory

### LINKED ISSUE #5771: Move the CLI history file to Kuzu home directory by default and cap its size to a reasonable number
### Kuzu version

v0.11.0

Ubuntu 24.04

### What happened?

```
mkdir tmp && cd tmp
kuzu foo
kuzu> create node table node(id int primary key);
...
➜  ls
foo  foo.lock  history.txt
```

Prefer to see only `foo`. But see two extra files.

_No response_

sdht0:
There is a default limit of 1000 lines, which is not configurable as of now.

Fixes #5771
Associated docs (issue or PR): TODO

### DIFF
--- a/tools/shell/shell_runner.cpp
+++ b/tools/shell/shell_runner.cpp
@@ -14,7 +14,7 @@ using namespace kuzu::common;
 int setConfigOutputMode(const std::string& mode, ShellConfig& shell) {
     shell.printer = PrinterFactory::getPrinter(PrinterTypeUtils::fromString(mode));
     if (shell.printer == nullptr) {
-        std::cerr << "Cannot parse '" << mode << "' as output mode." << '\n';
+        std::cerr << "Error: cannot parse output mode: " << mode << '\n';
         return 1;
     }
     shell.stats = shell.printer->defaultPrintStats();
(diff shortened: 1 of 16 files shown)


tools/shell/shell_runner.cpp
```diff
@@ -123,17 +123,30 @@ int main(int argc, char* argv[]) {
...
 
+    const std::string historyFile = "history.txt";
     auto pathToHistory = args::get(historyPathFlag);
     if (!pathToHistory.empty() && pathToHistory.back() != '/') {
         pathToHistory += '/';
     }
-    pathToHistory += "history.txt";
+    // If `historyFile` is present in the current directory, use that to preserve backward
+    // compatibility.
+    if (pathToHistory.empty() && !std::filesystem::exists(historyFile)) {
```
ray6080, line 133: Need to double check, but I assume we must have similar piece of code somewhere else that's creating the kuzu home directory?

sdht0: Yes in `ExtensionInstaller::installExtension` using `ClientContext.homeDirectory`.

sdht0: Moved the getUserHomeDir logic to ClientContext. The extensions code creates something like `~/.kuzu/<extension>/...` which is independent from the history logic.

mewim (approved):
CI and Makefile LGTM

PR #5791, opened 2025-07-19
Add scripts to build from source for uv-managed Python

prrao87:
# Description

This PR adds scripts to allow users to build from source using uv-managed Python rather than system, brew, or pyenv-managed Python. **It's non-intrusive** and does not force users to use uv (it's totally opt-in).

uv has become the defacto standard to manage Python installations and virtual environment for many users, and its most recent release 0.8.0 is a landmark one, because it can [now manage Python installs globally](https://pydevtools.com/blog/uv-0-8-release-automatic-python-installation-to-path/) without conflicting with system Python in any way. It's also orders of magnitude faster to install things in uv than with any other method in Python, for both installing new Python versions, and for downloading packages from PyPI.

The shell scripts provided allow users who choose to use uv to build Python from source (includes a `README.md` to help). Users who prefer doing things the old way don't need to use these scripts.

## How did I test?

I completely uninstalled `pyenv` which I used before to manage global Python (which pybind11 and cmake relied upon). The build process broke, as expected. The system Python installation I have on my mac is far too old to be usable by Kuzu, so the build didn't happen. Then, I ran the scripts included in this PR, and cmake and pybind11 now locate the global uv installation correctly, and the entire build process for Kuzu and Python works seamlessly, end-to-end.

Example commands (includes option for parallel builds):
```bash
# Configure and build everything with uv Python
./scripts/uv-make build

# Build only Python API
./scripts/uv-make python

# Build and run tests
./scripts/uv-make pytest

# Configure only (no build)
./scripts/uv-make configure

# Clean build directory
./scripts/uv-make clean

# Show help
./scripts/uv-make help

# Parallel builds (much faster!)
NUM_THREADS=20 ./scripts/uv-make build
NUM_THREADS=8 ./scripts/uv-make python
```

# Contributor agreement

### DIFF
--- a/scripts/README_uv_python.md
+++ b/scripts/README_uv_python.md
@@ -1,0 +1,142 @@
+# Kuzu uv Python Integration
+
+This directory contains scripts for users who wish to use `uv`-managed Python installations for building Kuzu with Python bindings.
+
+## Quick Start
+
+Choose your preferred approach:
+
+1. **Simple script**: `./scripts/uv-make build`
+2. **Shell functions**: Use `source scripts/kuzu_uv_functions.sh` and then `kuzu_build_uv <>`
+3. **Manual**: Set environment variables and run cmake directly
+
+## Problem
+
+By default, pybind11 looks for Python in system locations or pyenv shims, which may not work correctly with `uv`-managed Python installations. This integration automatically detects your current `uv` Python installation and configures cmake accordingly.
+
+## Solutions
+
+### 1. uv-make script (recommended for most users)
+
+```bash
+# Configure and build everything with uv Python
+./scripts/uv-make build
+
+# Build only Python API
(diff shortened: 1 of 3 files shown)


mewim:
I tested on [uv Docker image](https://docs.astral.sh/uv/guides/integration/docker/) and it does not seem to work for me? It seems after I ran each command it immediately returns without doing anything or printing an error? 

Also, I think https://github.com/kuzudb/kuzu/blob/master/scripts/pip-package/package_tar.py should still be able to generate a source distribution without any third-party dependency. After the source dist is created, uv should be able to install from it.

prrao87:
Because I exclusively use uv and need to build Python from source on my machine using uv, I needed these scripts in the other PR I have open (#5792). So I merged these scripts there. We can continue debugging there.

PR #5792, opened 2025-07-19
Fix Python type hints and add uv scripts to build from source

prrao87:
# Description

## Fix type hints
Commonly used functions in the Python API (iterators, Pandas/Polars DataFrames) currently produce a ton of linter errors. This PR fixes the linting issues so that static type checkers like Pylance and mypy won't complain (this affects live demos I do as I like having static type checking turned on to catch bugs in my Python code before runtime).

Only one test was changed for the `rows_as_dict` function, to address the FBT001 ruff linting error [here](https://docs.astral.sh/ruff/rules/boolean-type-hint-positional-argument/).

## uv scripts to build Python API from source

Adds uv scripts to help the user build from source. See #5791 for context. However, we need to address this issue related to using the uv Docker image before we can consider this resolved.

> I tested on [uv Docker image](https://docs.astral.sh/uv/guides/integration/docker/) and it does not seem to work for me? It seems after I ran each command it immediately returns without doing anything or printing an error? 

> Also, I think https://github.com/kuzudb/kuzu/blob/master/scripts/pip-package/package_tar.py should still be able to generate a source distribution without any third-party dependency. After the source dist is created, uv should be able to install from it.

# Contributor agreement

### DIFF
--- a/tools/python_api/src_py/query_result.py
+++ b/tools/python_api/src_py/query_result.py
@@ -32,7 +32,7 @@ if TYPE_CHECKING:
 class QueryResult:
     """QueryResult stores the result of a query execution."""
 
-    def __init__(self, connection: "_kuzu.Connection", query_result: "_kuzu.QueryResult") -> None:  # type: ignore[name-defined]
+    def __init__(self, connection: _kuzu.Connection, query_result: _kuzu.QueryResult) -> None:  # type: ignore[name-defined]
         """
         Parameters
         ----------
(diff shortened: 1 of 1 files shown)


prrao87:
Okay, I realized that all of this is overkill and there are very simple ways to disable the annoying linting issues in the IDE. The `execute` method returns both a `Query | list[QueryResult]` right now which is the cause of the linting problem, which is a whole separate problem and requires an API-level discussion prior to our v1 release. Let's revisit this.

Closing this for now.

PR #5803, opened 2025-07-22
query profiler: segfault trying to profile export database

### LINKED ISSUE #5802: query profiler: segfault when trying to profile export database
### Kuzu version

0.11.0

MacOS

### What happened?

Found while working on https://github.com/kuzudb/kuzu/issues/5727

```
kuzu> profile export database "tmp";
zsh: segmentation fault  ./build/debug/tools/shell/kuzu
tanvirgahunia@Tanvirs-MacBook-Pro kuzu %
```

_No response_

tgahunia05:
# Description

There is no nullptr check on the function `copyChildSchema` which is called by `LogicalExplain::computeSchema`.
The segfault in the issue was caused by a nullptr dereference on the `schema` member. This PR creates an empty schema if the child schema is nullptr. This may not be the right fix; we may want to check how this was ever nullptr. Pinging @andyfengHKU to validate. 

Fixes #5802 

### DIFF
--- a/src/include/planner/operator/logical_operator.h
+++ b/src/include/planner/operator/logical_operator.h
@@ -135,7 +135,14 @@ public:
 
 protected:
     void createEmptySchema() { schema = std::make_unique<Schema>(); }
-    void copyChildSchema(uint32_t idx) { schema = children[idx]->getSchema()->copy(); }
+    void copyChildSchema(uint32_t idx) { 
+    if (auto childSchema = children[idx]->getSchema()) {
+        schema = childSchema->copy(); 
+    }
+    else {
+        createEmptySchema();
+    }
+}
 
 protected:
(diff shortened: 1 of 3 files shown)


src/include/planner/operator/logical_operator.h
```diff
@@ -135,7 +135,14 @@ class KUZU_API LogicalOperator {
 
 protected:
     void createEmptySchema() { schema = std::make_unique<Schema>(); }
-    void copyChildSchema(uint32_t idx) { schema = children[idx]->getSchema()->copy(); }
+    void copyChildSchema(uint32_t idx) {
```
andyfengHKU, line 138: Let's limit the change to `LogicalExplain` only and keep LogicalOperator as it is

andyfengHKU (approved):

PR #5809, opened 2025-07-25
parsing: fix i128 min check

### LINKED ISSUE #5805: Bug: Possible off-by-one range check on int128 values
### Kuzu version

V0.11.0 nightly grabbed July 19

Fedora 40

### What happened?

Guodong Jin requested I post a bug: https://discord.com/channels/1196510116388806837/1397270552745607449/1397277635981082836

For this table:
`"CREATE NODE TABLE types(i int, i128 int128, primary key(i))"`

An attempt to insert the smallest possible int128 fails. 
```
  Query: "CREATE (it:types {i: 1, i128: -170141183460469231731687303715884105728}) RETURN it"
  Error: "Conversion exception: Cast failed. 170141183460469231731687303715884105728 is not within INT128 range."
```

In addition, a  range exception was _also_ received trying to bind this as a kuzu-value to a prepared statement via the C API (the exception terminated my C-API calling process).  I can't rule it out as a possible bug in my code, which is why I was trying a full-on string query in the example above.

_No response_

tgahunia05:
# Description
Parsing only considered magnitude when validating bounds. This is faulty logic. It resulted in an out of bounds error on i128 min. 

This fix pushes the sign down to the number literal subtree of the parse tree so that it can be used for parsing. 
Closes #5805 

# Contributor agreement


src/common/types/int128_t.cpp
```diff
@@ -114,26 +114,43 @@ int128_t Int128_t::divModPositive(int128_t lhs, uint64_t rhs, uint64_t& remainde
 }
 
 std::string Int128_t::ToString(int128_t input) {
-    uint64_t remainder = 0;
-    std::string result;
+    bool isMin = (input.high == INT64_MIN && input.low == 0);
     bool negative = input.high < 0;
+
+    if (isMin) {
+
```
tgahunia05, line 121: Should just use hard code the value of intMin here.

andyfengHKU (approved):

PR #5810, opened 2025-07-25
parsing: experimental grammar changes for signed integer parsing. 

tgahunia05:
# Description

This is an experimental grammar change that is less "hacky" than what is proposed in https://github.com/kuzudb/kuzu/pull/5809. Expect breaking changes [do not review].

# Contributor agreement


andyfengHKU:
I'm a bit concerned this change of grammar will lead to ambiguous in parsing. Let's use the other solution.

PR #5820, opened 2025-07-28
Fix LSQB benchmark runs

ray6080:
# Description

The LSQB benchmark failed due to the change of single file database. Not we need to pass in a file path instead of a directory to construct a Kuzu Database.

### DIFF
--- a/benchmark/lsqb/benchmark_runner.py
+++ b/benchmark/lsqb/benchmark_runner.py
@@ -126,7 +126,7 @@ def run_query(conn, query_spec):
 
 
 def run_kuzu(sf, serialized_graph_path, num_threads):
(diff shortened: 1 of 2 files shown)


benchmark/lsqb/benchmark_runner.py
```diff
@@ -126,7 +126,7 @@ def run_query(conn, query_spec):
 
 
 def run_kuzu(sf, serialized_graph_path, num_threads):
-    db = kuzu.Database(serialized_graph_path)
+    db = kuzu.Database(os.path.join(serialized_graph_path, "kuzu.db"))
```
sdht0, line 129: nit: can we use `db.kuzu` instead?

sdht0 (approved):

PR #5827, opened 2025-07-29
Clean up unused function

ray6080:
# Description

Realized that the interface `getEstimatedMemUsage()` is now no longer used after some previous refactoring. This PR removes these functions.


sdht0 (approved):
Sweet.

PR #5828, opened 2025-07-30
Add locks to local hash index and add basic concurrent insertion tests

ray6080:
# Description

### DIFF
--- a/src/include/storage/local_storage/local_hash_index.h
+++ b/src/include/storage/local_storage/local_hash_index.h
@@ -30,6 +30,7 @@ public:
         : localDeletions{}, localInsertions{memoryManager, handle} {}
     HashIndexLocalLookupState lookup(KeyType key, common::offset_t& result,
         visible_func isVisible) {
+        std::shared_lock sLock{mtx};
         if (localDeletions.contains(key)) {
             return HashIndexLocalLookupState::KEY_DELETED;
         }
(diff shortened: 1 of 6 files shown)


andyfengHKU (approved):
@benjaminwinger should comeback and review the lock

src/include/storage/local_storage/local_hash_index.h
```diff
@@ -56,45 +62,74 @@ class HashIndexLocalStorage final : public BaseHashIndexLocalStorage {
...
         for (auto& key : localDeletions) {
             deleteOp(key);
         }
         insertOp(localInsertions);
     }
 
-    void reserveInserts(uint64_t newEntries) { localInsertions.reserve(newEntries); }
+    void reserveInserts(uint64_t newEntries) {
+        std::unique_lock xLock{mtx};
+        localInsertions.reserve(newEntries);
```
benjaminwinger, line 119: I don't think this is the right behaviour here (though it should be safe). If we have multiple threads bulk reserving we should reserve the combined amount.
This also applies to `reserveSpaceForAppend`, and maybe the solution should be to track capacity explicitly and merge the two functions, and in both situations just add to the capacity (even now, I think `reserveInserts` could be replaced by `reserveSpaceForAppend` since I think `reserveInserts` is only called when the number of existing entries is zero). I think that should be safe, since the local storage gets cleared during rollbacks, so if there is an interruption we won't just keep growing the capacity if the reserve is tried again.

PR #5843, opened 2025-08-02
fsync() the WAL on open

thisismiller:
# Description

It's generally wise to ensure that whatever set of changes are being recovered from the WAL are actually durable before they're processed. By a strict interpretation of the UNIX spec, the fsync() on open doesn't _have_ to actually do anything, but in practice on Linux and the common ext4/xfs it does, and thus avoids weird crash->recover->power loss, and now storage is now ahead of the WAL sorts of weirdness.

# Contributor agreement

---

I'm mostly just using this as the tiny change to work through the build/test/submit PR process.  I see hints of linter or lcov usage, is there something in the makefile I missed for that?  Is there anything else I should be double checking before submitting PRs?

### DIFF
--- a/src/storage/wal/wal.cpp
+++ b/src/storage/wal/wal.cpp
@@ -70,6 +70,9 @@ void WAL::initWriter(main::ClientContext* context) {
     fileInfo = vfs->openFile(walPath,
         FileOpenFlags(FileFlags::CREATE_IF_NOT_EXISTS | FileFlags::READ_ONLY | FileFlags::WRITE),
         context);
+    // A previous unclean exit may have left non-durable contents in the WAL, so make our best
+    // effort at ensuring that we're recovering durable WAL entries.
+    fileInfo->syncFile();
     writer = std::make_shared<BufferedFileWriter>(*fileInfo);
     // WAL should always be APPEND only. We don't want to overwrite the file as it may still
     // contain records not replayed. This can happen if checkpoint is not triggered before the
--- a/src/storage/wal/wal.cpp
+++ b/src/storage/wal/wal.cpp
@@ -70,9 +70,6 @@ void WAL::initWriter(main::ClientContext* context) {
     fileInfo = vfs->openFile(walPath,
         FileOpenFlags(FileFlags::CREATE_IF_NOT_EXISTS | FileFlags::READ_ONLY | FileFlags::WRITE),
         context);
-    // A previous unclean exit may have left non-durable contents in the WAL, so make our best
-    // effort at ensuring that we're recovering durable WAL entries.
-    fileInfo->syncFile();
     writer = std::make_shared<BufferedFileWriter>(*fileInfo);
     // WAL should always be APPEND only. We don't want to overwrite the file as it may still
     // contain records not replayed. This can happen if checkpoint is not triggered before the
--- a/src/storage/wal/wal_replayer.cpp
+++ b/src/storage/wal/wal_replayer.cpp
@@ -53,6 +53,9 @@ void WALReplayer::replay() const {
         checkpointer.readCheckpoint();
         return;
     }
+    // A previous unclean exit may have left non-durable contents in the WAL, so before we start
+    // replaying the WAL records, make a best-effort attempt at ensuring the WAL is fully durable.
+    fileInfo->syncFile();
     // Start replaying the WAL records.
     try {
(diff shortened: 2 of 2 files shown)


ray6080 (commented):
Hi @thisismiller I'm not sure if I follow this correctly.
> It's generally wise to ensure that whatever set of changes are being recovered from the WAL are actually durable before they're processed

Right now, the logic of handling recovering from wal is written in `WalReplayer`, instead of `wal.cpp`. Could you please describe a more detailed steps for a test case in which the newly added `fileInfo->syncFile()` is needed?

> I'm mostly just using this as the tiny change to work through the build/test/submit PR process. I see hints of linter or lcov usage, is there something in the makefile I missed for that? Is there anything else I should be double checking before submitting PRs?

Regarding the development workflow: thanks for your interest and here's what I think you need to be aware of:
- Code coverage: `lcov` is configured to auto run on CI, so no need to care about it locally.
- Code formatting: I would suggest you to run our code format scripts locally before submitting. You can find all code formatting scripts we use here https://github.com/kuzudb/kuzu/blob/0cfe4d54909cf9e621f86cc373b659ec5453e0fd/.github/workflows/ci-workflow.yml#L845
Note that we have auto code format fix on our CI, but it's not yet enabled for outside PRs yet.
- Testing: Please make sure your changes are covered by test cases when possible. You can run tests locally through our testing framework. https://docs.kuzudb.com/developer-guide/testing-framework/
- Major changes: For any major changes, we encourage contributors to join our discord (https://kuzudb.com/chat) and discuss with the team beforehand.

thisismiller:
Ah, I had wrongly just assumed WALReplayer wrapped around a WAL.  WALReplayer::replay() would indeed be the right place to add an extra fsync after the file open.  I'll go shuffle it over.

The longer version of the way to end up in an odd state would be to have something like (and pardon that I'll fall back to a bit more classical of database descriptions here):

1. Undo/redo logs are emitted to the WAL up to LSN 100, and fsync()'d successfully.
2. Another batch of records are written to the WAL up to LSN 200, but the program crashes right before calling fsync().  The pwrite() calls already moved the records to the kernel's page cache.
3. The program restarts, and reads the WAL.  The read will be served using the kernel's page cache, and will return records up to LSN 200.
4. Recovery proceeds, and the storage file is recovered to some degree, let's say LSN 150.
5. The entire system crashes, losing the dirty pages in the kernel's buffer cache.
6. The system reboots, the application starts, and it begins recovery.
7. The WAL was only made durable though LSN 100, and so the read of the WAL only returns records up through LSN 100.  However, the changes made to the storage file could have been persisted, and could have an LSN greater than what the WAL contains, breaking an invariant.

Which also hopefully summarizes why writing a test around it is a bit awkward.  A _real_ test could be written, I think, by using `echo 3 > /proc/sys/vm/drop_caches` to force the kernel to drop the cache after a fake crash has been simulated by closing the WAL and probably tweaking `vm.dirty_*` via sysctl to discourage dirty page writeback.  Unfortunately, that requires sudo, affects the whole system, and would be... like... awkward to write.

The better way to test this is to continue to build around the file and filesystem abstraction to add tracking of un-fsync'd changes in local buffers and only pwrite() them on the sync() call, which then gives you a way to model what a crash and dumping the page cache would look like.  But, then you're testing against a model of the system behavior, so you need to accept that the model is correct in the first place.  One would then also need to test that subsets of pages being written back when dirty is fine, and that crashing before each fsync is recoverable, and then you're on the path to (deterministic) simulation testing.  Which is a great path, and any of the ways you'd try to get to better testing of durability-related code, it'd be some sets of rather involved testing focused changes that wasn't quite in the direction of what I was planning on poking around in.

If you feel like reading around in this space a bit more, [How CouchDB Prevents Data Corruption](https://neighbourhood.ie/blog/2025/02/26/how-couchdb-prevents-data-corruption-fsync) is a longer example of this. [fsync after open is an elaborate no-op](https://despairlabs.com/blog/posts/2025-03-13-fsync-after-open-is-an-elaborate-no-op/) is a reply (and [the lobsters thread](https://lobste.rs/s/t2bmsy/fsync_after_open_is_elaborate_no_op) is probably worth a follow-on read). Admittedly though, a number of things in this area come back circuitously to me as a source.  I learned this from FoundationDB's source code, and they're also the only project I know of who actually set up a power-off testing rig.

PR #5857, opened 2025-08-08
Rework UpdateInfo for concurrent updates

ray6080:
# Description

We use `UpdateInfo` to track multiple versions of data if any were updated. 

Previously, the `UpdateInfo` wasn't designed for concurrent accesses, this PR adds proper locks within UpdateInfo for concurrent accesses. For each vector of data, a `UpdateNode` is kept as the head of the version list.
The locks are placed at two levels, one for the whole `UpdateInfo` structure, and one for each version list, i.e., `UpdateNode`.

The scan of update data is reworked as well, as there can be multiple transactions ongoing concurrently now. Previously we assume that only one writer exists, and the version chain should contain committed updates only (roll backed ones gets removed from the chain immediately). The scan now performs first by scanning the base data, then traversing the version chain and find the latest available update for each row and overwrite the scanned base data accordingly.

Note that the garbage collection is not done within this PR, which will come in a later one, as some further rework is needed separately.

A simple benchmark was performed locally on Mac mini with M4 pro chip. The benchmark is to update each tuple of a node table with 1M tuples with 1/2/4/8 threads under *in-memory mode*. Each update itself is a transaction. Each thread simply updates a sequential range of nodes to set an int64 property value to a new one.

The benchmark result:
| # threads | branch | time (ms) |
| --------- | ------- | ---------- |
| 1 | master | 106825 |
| 1 | new | 105715 |
| 2 | new | 64242 | 
| 4 | new | 44582 | 
| 8 | new | 33626 |

### DIFF
--- a/src/include/storage/table/update_info.h
+++ b/src/include/storage/table/update_info.h
@@ -41,12 +41,10 @@ struct VectorUpdateInfo {
     }
 
     std::unique_ptr<VectorUpdateInfo> movePrev() { return std::move(prev); }
-    std::unique_ptr<VectorUpdateInfo> movePrevNoLock() { return std::move(prev); }
     void setPrev(std::unique_ptr<VectorUpdateInfo> prev_) { this->prev = std::move(prev_); }
     VectorUpdateInfo* getPrev() const { return prev.get(); }
     void setNext(VectorUpdateInfo* next_) { this->next = next_; }
     VectorUpdateInfo* getNext() const { return next; }
-    VectorUpdateInfo* getNextNoLock() const { return next; }
 };
 
 struct UpdateNode {
@@ -56,7 +54,7 @@ struct UpdateNode {
     UpdateNode() : info{nullptr} {}
     UpdateNode(UpdateNode&& other) noexcept : info{std::move(other.info)} {}
 
-    bool isSet() const {
+    bool isEmpty() const {
         std::shared_lock lock{mtx};
         return info != nullptr;
     }
@@ -114,13 +112,15 @@ public:
     }
 
 private:
+    using iterate_read_from_row_func_t =
+        std::function<void(const VectorUpdateInfo&, uint64_t, uint64_t)>;
+
     UpdateNode& getUpdateNode(common::idx_t vectorIdx);
     UpdateNode& getOrCreateUpdateNode(common::idx_t vectorIdx);
 
     void iterateScan(const transaction::Transaction* transaction, uint64_t startOffsetToScan,
         uint64_t numRowsToScan, uint64_t startPosInOutput,
-        const std::function<void(const VectorUpdateInfo&, uint64_t, uint64_t)>& readFromRowFunc)
-        const;
+        const iterate_read_from_row_func_t& readFromRowFunc) const;
 
 private:
     mutable std::shared_mutex mtx;
--- a/src/include/storage/table/update_info.h
+++ b/src/include/storage/table/update_info.h
@@ -1,6 +1,7 @@
 #pragma once
 
 #include <array>
+#include <shared_mutex>
 
 #include "column_chunk_data.h"
 #include "common/types/types.h"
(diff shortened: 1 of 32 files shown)


test/transaction/transaction_test.cpp
```diff
@@ -341,4 +341,461 @@ TEST_F(EmptyDBTransactionTest, ConcurrentComplexRelationshipInsertions) {
...
+    ASSERT_EQ(committedCount, expectedCommitted);
+
+    res = conn->query("MATCH ()-[r:knows]->() WHERE r.weight = 20.0 RETURN COUNT(r) AS COUNT;");
+    ASSERT_TRUE(res->isSuccess());
+    ASSERT_EQ(res->getNumTuples(), 1);
+    auto originalCount = res->getNext()->getValue(0)->getValue<int64_t>();
+    ASSERT_EQ(originalCount, numTotalUpdates / 2);
+}
+
+// TODO(Guodong): FIX-ME.
```
ray6080, line 698: Will remove this test.

src/storage/table/update_info.cpp
```diff
@@ -14,57 +14,261 @@ using namespace kuzu::common;
...
+    }
+    auto [startVectorIdx, startRowInVector] =
+        StorageUtils::getQuotientRemainder(startRowScanned, DEFAULT_VECTOR_CAPACITY);
+    auto [endVectorIdx, endRowInVector] =
+        StorageUtils::getQuotientRemainder(startRowScanned + numRows, DEFAULT_VECTOR_CAPACITY);
+    idx_t vectorIdx = startVectorIdx;
+    while (vectorIdx <= endVectorIdx) {
+        const auto startRow = vectorIdx == startVectorIdx ? startRowInVector : 0;
+        const auto endRow = vectorIdx == endVectorIdx ? endRowInVector : DEFAULT_VECTOR_CAPACITY;
+        // TODO(Guodong): Should also use a bool vector here.
```
ray6080, line 162: Will get this done before the PR gets merged.

src/storage/table/version_info.cpp
```diff
@@ -255,7 +255,7 @@ row_idx_t VectorVersionInfo::getNumDeletions(transaction_t startTS, transaction_
 void VectorVersionInfo::rollbackInsertions(row_idx_t startRowInVector, row_idx_t numRows) {
     if (isSameInsertionVersion()) {
         // This implicitly assumes that all rows are inserted in the same transaction, so regardless
-        // which rows to be rollbacked, we just reset the sameInsertionVersion.
+        // which rows to be roll backed, we just reset the sameInsertionVersion.
```
royi-luo, line 258:
```suggestion
        // which rows to be rolled back, we just reset the sameInsertionVersion.
```

test/transaction/transaction_test.cpp
```diff
@@ -341,4 +341,357 @@ TEST_F(EmptyDBTransactionTest, ConcurrentComplexRelationshipInsertions) {
     auto verifiedCount = res->getNext()->getValue(0)->getValue<int64_t>();
     ASSERT_EQ(verifiedCount, numTotalInsertions / 3);
 }
+
```
royi-luo, line 344: Do we have any tests elsewhere that test the case where multiple transactions update the same offset (Say transaction A commits before transaction B starts so no write-write conflict)? I don't see any here and it's probably good to test that we handle this correctly in the concurrent case (so something like running the updateNodes tests multiple times with some synchronization step in between so we make sure all the transactions are committed before we start again).

src/include/storage/table/update_info.h
```diff
@@ -29,49 +30,101 @@ struct VectorUpdateInfo {
...
+};
+
+struct UpdateNode {
+    mutable std::shared_mutex mtx;
+    std::unique_ptr<VectorUpdateInfo> info;
+
+    UpdateNode() : info{nullptr} {}
+    UpdateNode(UpdateNode&& other) noexcept : info{std::move(other.info)} {}
+
+    bool isSet() const {
```
royi-luo, line 59: I feel like `isSet()` isn't a very intuitive name, maybe something like `isEmpty()` would work? (same thing for `UpdateInfo`)

ray6080: Renamed to `isEmpty`.

test/test_files/transaction/concurrency/dml_empty_serial_execution.test
```diff
@@ -34,6 +34,27 @@
...
+-STATEMENT [conn2] MATCH (p:person) WHERE p.ID = 0 SET p.fName = 'Alphabet' RETURN p.*;
+---- ok
+-ENDLOOP
+-LOOP i 1 100
+-STATEMENT MATCH (p:person) WHERE p.ID = 0 SET p.fName = 'Alphabet100' RETURN p.*;
+---- ok
+-ENDLOOP
+-STATEMENT MATCH (p:person) WHERE p.ID = 0 RETURN p.fName;
+---- 1
+Alphabet100
```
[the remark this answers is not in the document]
royi-luo: looks good now

PR #5858, opened 2025-08-08
Proper error when cast invalid json

### LINKED ISSUE #5856: Bug: Unable to import database via parquet with empty JSON field
### Kuzu version

v0.11.1

MacOS Sequoia 15.5

### What happened?

If you have a database with the JSON extension enabled, where a JSON typed column has an empty value, and then you export it via the parquet format, when you import it you get an error that says:

```
Error: Runtime exception: Import database failed: Conversion exception: Cannot convert  to JSON.
```

Run the following to create the export:
```
INSTALL json;
LOAD json;
CREATE NODE TABLE User(name STRING PRIMARY KEY, metadata JSON);
CREATE (:User {name: 'Joe', metadata: ''});
CREATE (:User {name: 'John'});
EXPORT DATABASE 'kuzu-export';
```

Then in another session run
```
IMPORT DATABASE 'kuzu-export';
```

and you will be greeted with the message `Error: Runtime exception: Import database failed: Conversion exception: Cannot convert  to JSON.`

Doing the exact same thing but using the csv export `EXPORT DATABASE 'dbexport-csv' (format="csv", header=true);` works fine.

acquamarin:
Closes #5856 
We should properly error when the user tries to cast a invalid json string to json.

For example:
```
INSTALL json;
LOAD json;
CREATE NODE TABLE User(name STRING PRIMARY KEY, metadata JSON);
CREATE (:User {name: 'Joe', metadata: ''});
```
We should error since empty string is not a valid jso.

### DIFF
--- a/extension/json/test/error.test
+++ b/extension/json/test/error.test
@@ -54,18 +54,21 @@ Binder exception: Invalid primary key column type json. Primary keys must be eit
 -STATEMENT CREATE NODE TABLE PERSON (ID INT64, NAME JSON, description STRING, PRIMARY KEY(ID));
 ---- ok
 -STATEMENT CREATE (p:PERSON {ID: 52, NAME: 'KUZU', DESCRIPTION: 'good'});
----- ok
+---- error
+Conversion exception: Cannot convert KUZU to JSON.
+-STATEMENT CREATE (p:PERSON {ID: 52, NAME: '', DESCRIPTION: 'good'});
+---- error
+Conversion exception: Cannot convert  to JSON.
 -STATEMENT CREATE (p:PERSON {ID: 22, NAME: 'neo', DESCRIPTION: 'ok'});
----- ok
+---- error
+Conversion exception: Cannot convert neo to JSON.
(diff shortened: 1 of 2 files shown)


andyfengHKU (approved):
Let's also add a test case for empty json?

extension/json/test/error.test
```diff
@@ -54,25 +54,26 @@ Binder exception: Invalid primary key column type json. Primary keys must be eit
 -STATEMENT CREATE NODE TABLE PERSON (ID INT64, NAME JSON, description STRING, PRIMARY KEY(ID));
 ---- ok
 -STATEMENT CREATE (p:PERSON {ID: 52, NAME: 'KUZU', DESCRIPTION: 'good'});
----- ok
+---- error
+Conversion exception: Cannot convert KUZU to JSON.
+-STATEMENT CREATE (p:PERSON {ID: 52, NAME: '', DESCRIPTION: 'good'});
```
acquamarin, line 59: @andyfengHKU This is the test case where we cast empty string to json

PR #5868, opened 2025-08-12
Avoid OnDiskGraphNbrScanState#getChunk() create new propertyVectors

ted-wq-x:
# Description

Optimize gds framework performance.In my gds algorithm , the time taken was reduced from 1703ms to 1563ms.

# Contributor agreement

### DIFF
--- a/src/graph/graph.cpp
+++ b/src/graph/graph.cpp
@@ -4,8 +4,9 @@
 
 namespace kuzu::graph {
 NbrScanState::Chunk::Chunk(std::span<const common::nodeID_t> nbrNodes,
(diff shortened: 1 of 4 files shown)


sdht0:
Thanks @ted-wq-x. Can you share how you tested the changes? Would be helpful to know the input data and algo you used.

ted-wq-x:
I tested in my implement  K-hop algorithm with twitter dataset. Query like 'call graph_khop('PK1' ,'vertex','6894372','FWD',8) return *;' it means start from a node with vertex label and that node's primary key is 6894372,expand direction is FWD,maximum number of hops  is 8,return the number of neighbors .

sdht0:
I'll take a look soon.

PR #5873, opened 2025-08-13
Add abstract arrow query result

andyfengHKU:
# Description

This PR mainly abstract query result interface so that we are prepared to have different type of query results, specifically  

- MaterializedResult: backed by FactorizedTable
- ArrowResult: currently converted from FactorizedTable but we will create this result natively through operator in the next RP.
- StreamingArrowResult: to be added

This PR also refactors arrow converter/rowBatch class. I found it not well designed by passing QueryResult directly into their public interface. As a principle, we should try to avoid include other module in common module.

# Contributor agreement

### DIFF
--- a/src/common/arrow/arrow_array_scan.cpp
+++ b/src/common/arrow/arrow_array_scan.cpp
@@ -1,8 +1,8 @@
 #include "common/arrow/arrow_converter.h"
 #include "common/exception/runtime.h"
+#include "common/types/int128_t.h"
 #include "common/types/interval_t.h"
 #include "common/types/types.h"
-#include "common/types/int128_t.h"
 #include "common/vector/value_vector.h"
 
 namespace kuzu {
--- a/src/common/arrow/arrow_array_scan.cpp
+++ b/src/common/arrow/arrow_array_scan.cpp
@@ -2,6 +2,7 @@
 #include "common/exception/runtime.h"
 #include "common/types/interval_t.h"
 #include "common/types/types.h"
+#include "common/types/int128_t.h"
 #include "common/vector/value_vector.h"
 
 namespace kuzu {
--- a/src/include/processor/result/flat_tuple.h
+++ b/src/include/processor/result/flat_tuple.h
@@ -1,7 +1,6 @@
 #pragma once
 
 #include <cstdint>
-#include <memory>
 #include <string>
 #include <vector>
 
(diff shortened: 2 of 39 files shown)


src/common/arrow/arrow_array_scan.cpp
```diff
@@ -1,5 +1,6 @@
 #include "common/arrow/arrow_converter.h"
 #include "common/exception/runtime.h"
+#include "common/types/int128_t.h"
```
ray6080, line 3: Why do we need to add the include in this PR?

src/main/query_result/arrow_query_result.cpp
```diff
@@ -0,0 +1,73 @@
...
+std::unique_ptr<ArrowArray> ArrowQueryResult::getNextArrowChunk(int64_t chunkSize) {
+    if (chunkSize != chunkSize_) {
+        throw RuntimeException(
+            stringFormat("Chunk size does not match expected value {}.", chunkSize_));
+    }
+    return std::make_unique<ArrowArray>(arrays[cursor++]);
+}
+
+} // namespace main
+} // namespace kuzu
```
ray6080, line 73: New line here.

src/include/processor/result/flat_tuple.h
```diff
@@ -16,20 +15,26 @@ namespace processor {
...
+    KUZU_API common::idx_t len() const;
 
     /**
      * @param idx value index to get.
      * @return the value stored at idx.
      */
-    KUZU_API common::Value* getValue(uint32_t idx) const;
+    KUZU_API common::Value* getValue(common::idx_t idx);
+
+    KUZU_API common::Value& operator[](common::idx_t idx);
```
ray6080, line 33: Add docstring.

src/include/processor/result/flat_tuple.h
```diff
@@ -16,20 +15,26 @@ namespace processor {
...
+
+    KUZU_API const common::Value& operator[](common::idx_t idx) const;
```
ray6080, line 35: Ditto on docstring.

src/include/processor/result/flat_tuple.h
```diff
@@ -16,20 +15,26 @@ namespace processor {
...
 
-    KUZU_API std::string toString();
+    KUZU_API std::string toString() const;
```
ray6080, line 37: Ditto on docstring.

src/processor/processor.cpp
```diff
@@ -32,9 +34,10 @@ std::shared_ptr<FactorizedTable> QueryProcessor::execute(PhysicalPlan* physicalP
         decomposePlanIntoTask(sink->getChild(i), task.get(), context);
     }
     initTask(task.get());
-    context->clientContext->getProgressBar()->startProgress(context->queryID);
+    auto progress = context->clientContext->getProgressBar();
```
ray6080, line 37:
```suggestion
    auto progressBar = context->clientContext->getProgressBar();
```

ray6080 (approved):
Looks good! We need to extend our testing framework to cover `queryAsArrow` as well. I will open an issue on it.

src/common/arrow/arrow_array_scan.cpp
andyfengHKU: Because we remove certain header from arrow_converter.h

src/main/query_result/materialized_query_result.cpp
```diff
@@ -0,0 +1,104 @@
...
+bool MaterializedQueryResult::hasNextArrowChunk() {
+    return hasNext();
+}
+
+std::unique_ptr<ArrowArray> MaterializedQueryResult::getNextArrowChunk(int64_t chunkSize) {
+    checkDatabaseClosedOrThrow();
+    auto rowBatch = std::make_unique<ArrowRowBatch>(copyVector(columnTypes), chunkSize,
+        false /* fallbackExtensionTypes */);
+    auto rowBatchSize = 0u;
+    while (rowBatchSize < chunkSize) {
```
benjaminwinger, line 93: You're never increasing the rowBatchSize here. It's always producing the maximum number of rows since this loop will continue until the iterator is exhausted (which is causing the rust arrow test to fail).

PR #5880, opened 2025-08-13
Remove nullable flag from Arrow map type key

05st:
Resolves the `Map's keys must be non-nullable` error here https://github.com/kuzudb/kuzu-enterprise/issues/21.


ray6080 (approved):
Let's add a test case for the issue.

ray6080 (approved):

PR #5891, opened 2025-08-15
Recursively set database lifecycle manager for multiple query results

mewim:
This is a follow-up of #5665 to handle the closing order issue correctly for query result of multiple statements. Previous PR only sets the database lifecycle manager for the first query result, which did not fix closing order issue when multiple statements are executed and a linked list is returned. This PR sets the database lifecycle manager for all query results in the linked list.


sdht0 (approved):
Short and sweet.

PR #5899, opened 2025-08-18
Add loop and set random in testing framework

ray6080:
# Description

Added the support for `LOOP` and `random` in the testing framework, which can be used to simulate random workloads with a fixed or dynamic seed more easily.
`LOOP` is intended to take in:
 - `<start> <end> [step]`
 - `<[list of values]>`

Only `randInt64` is added for now, which can be expanded to support `randStr` and more types. The best way is actually to remove `random` from the testing framework, but use Cypher functions, which we do not support right now.

After this, the `create_random_int.test` is rewritten in a much easier way as following.

```
-CASE TSSeedRandomInt64Insertions
-SKIP_WASM
-STATEMENT CREATE NODE TABLE test(id SERIAL, prop INT64, PRIMARY KEY(id));
---- ok
-SET seed current_timestamp()
-LOG seed: ${seed}
-SET _ random.set_seed(${seed})
-LOOP i 1 1000
-SET val random.randInt64(1000000)
-STATEMENT CREATE (t:test {prop: ${val}});
---- ok
-ENDLOOP
-STATEMENT MATCH (t:test) RETURN COUNT(*);
---- 1
1000

-CASE FixedSeedRandomInt64Insertions
-SKIP_WASM
-STATEMENT CREATE NODE TABLE test(id SERIAL, prop INT64, PRIMARY KEY(id));
---- ok
-SET seed 1234567890
-LOG seed: ${seed}
-SET _ random.set_seed(${seed})
-LOOP i 1 1000
-SET val random.randInt64(1000000)
-STATEMENT CREATE (t:test {prop: ${val}});
---- ok
-ENDLOOP
-STATEMENT MATCH (t:test) RETURN COUNT(*);
---- 1
1000
-STATEMENT MATCH (t:test) RETURN sum(t.prop);
---- 1
502790265
```

The PR also reworks a bit how test statements are parsed to make it more readable, though more refactoring is needed later to make it more robust.


test/test_runner/test_parser.cpp
```diff
@@ -423,7 +485,71 @@ std::string TestParser::parseCommandArange() const {
...
+
+        try {
+            uint64_t seedValue = static_cast<uint64_t>(stoll(param));
+            randomEngine.setSeed(seedValue);
+            return Value(static_cast<int64_t>(0)); // Return dummy value for _ variable
+        } catch (const std::exception&) {
+            throw TestException("Invalid seed parameter [" + path + ":" + line + "].");
+        }
+    } else if (funcCall.starts_with("random.randInt64(") && funcCall.ends_with(")")) {
+        // Extract parameter from random.randInt64(max)
```
royi-luo, line 536: If this is expected we should probably write somewhere (maybe here as a comment) that our random engine can only generate values in the int32 range and cast to int64

royi-luo (approved):
LGTM

test/test_runner/test_parser.cpp
ray6080: I chose to rename the function in testing framework to randInt32, and remove the casting to make it less confusing in the future.

PR #5905, opened 2025-08-20
Add arrow result collector

andyfengHKU:
# Description

This is the initial version of arrow result collector so that we can bypass FactorizedTable and directly generate arrow result.

- [ ] We need to setup a CI so that all test cases can go through arrow interface.
- [ ] We need an interface that can append value vector directly to arrow row batch.

# Contributor agreement

### DIFF
--- a/src/include/processor/operator/arrow_result_collector.h
+++ b/src/include/processor/operator/arrow_result_collector.h
@@ -1,9 +1,10 @@
 #pragma once
 
+#include <mutex>
+
+#include "common/arrow/arrow.h"
 #include "processor/operator/sink.h"
 #include "processor/result/flat_tuple.h"
-#include "common/arrow/arrow.h"
-#include <mutex>
 
 namespace kuzu {
 namespace processor {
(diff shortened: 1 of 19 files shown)


src/main/query_result/arrow_query_result.cpp
```diff
@@ -25,18 +32,19 @@ uint64_t ArrowQueryResult::getNumTuples() const {
...
-    auto rowBatch = std::make_unique<ArrowRowBatch>(copyVector(columnTypes), chunkSize,
-        false /* fallbackExtensionTypes */);
+    auto rowBatch =
+        std::make_unique<ArrowRowBatch>(columnTypes, chunkSize, false /* fallbackExtensionTypes */);
     auto rowBatchSize = 0u;
     while (rowBatchSize < chunkSize) {
         if (!iterator.hasNext()) {
             break;
         }
         (void)iterator.getNext(*tuple);
```
ray6080, line 42: What's the `(void)` for here?

src/main/query_result/arrow_query_result.cpp
```diff
@@ -25,18 +32,19 @@ uint64_t ArrowQueryResult::getNumTuples() const {
...
         rowBatch->append(*tuple);
+        rowBatchSize++;
```
ray6080, line 44: Was this a bug?

src/processor/operator/arrow_result_collector.cpp
```diff
@@ -0,0 +1,102 @@
...
+
+void ArrowResultCollectorSharedState::merge(std::vector<ArrowArray> localArrays) {
+    std::unique_lock lck{mutex};
+    for (auto i = 0u; i < localArrays.size(); ++i) {
+        arrays.push_back(localArrays[i]);
+    }
+}
+
+void ArrowResultCollector::executeInternal(ExecutionContext* context) {
+    auto rowBatch = std::make_unique<ArrowRowBatch>(info.columnTypes, info.chunkSize,
```
ray6080, line 46: Looks like you don't need unique_ptr here. `auto rowBatch  = ArrowRowBatch(...)` should be fine.

src/processor/operator/arrow_result_collector.cpp
```diff
@@ -0,0 +1,102 @@
...
+#include "common/arrow/arrow_row_batch.h"
+#include "main/query_result/arrow_query_result.h"
+
+using namespace kuzu::common;
+
+namespace kuzu {
+namespace processor {
+
+bool ArrowResultCollectorLocalState::advance() {
+    for (int64_t i = (int64_t)chunks.size() - 1; i >= 0; --i) {
```
ray6080, line 12: Let's avoid c-cstyle casting.

src/processor/operator/arrow_result_collector.cpp
```diff
@@ -0,0 +1,102 @@
...
+    }
+}
+
+void ArrowResultCollectorLocalState::resetCursor() {
+    for (auto i = 0u; i < chunkCursors.size(); ++i) {
+        chunkCursors[i] = 0;
+    }
+}
+
+void ArrowResultCollectorSharedState::merge(std::vector<ArrowArray> localArrays) {
```
ray6080, line 38:
```suggestion
void ArrowResultCollectorSharedState::merge(const std::vector<ArrowArray>& localArrays) {
```

src/include/processor/operator/arrow_result_collector.h
```diff
@@ -0,0 +1,87 @@
...
+    void merge(std::vector<ArrowArray> localArrays);
+
+private:
+    std::mutex mutex;
+};
+
+struct ArrowResultCollectorLocalState {
+    std::vector<ArrowArray> arrays;
+    std::vector<common::ValueVector*> vectors;
+    std::vector<common::sel_t*> vectorsSelPos;
```
ray6080, line 25: I would instead consider `reference_wrapper` here instead of raw pointer.

ray6080 (approved):

PR #5916, opened 2025-08-21
algo: fix weight property materialization

tgahunia05:
In spanning forest, when materializing the weight property we blindly interpret the number as double. This is faulty although I can't seem to create a test that produces an incorrect result. If you attempt the newly added test in master and print the values that are materialized you would see inf and nan. This fix properly materializes numbers to double for comparison. 


extension/algo/src/function/spanning_forest.cpp
```diff
@@ -245,6 +250,59 @@ void KruskalCompute::mergeComponents(const offset_t& srcCompId, const offset_t&
     }
 }
 
+void KruskalCompute::getValueAsDouble(const std::shared_ptr<common::ValueVector>& valueVector,
```
sdht0, line 253: Is this also needed in PR and other algorithms? Then this function should be put in common utils somewhere.

@ray6080 is there no existing function to do this casting? It seems core operators will need something similar when reading values?

sdht0 (approved):
Thanks.

@ray6080 can you take a second look to see if that's a good place for the function?

andyfengHKU (approved):

PR #5922, opened 2025-08-21
Rollback undo buffer before local storage

royi-luo:
# Description

Rolling back local storage will evict any optimistically-allocated pages. However, rolling back the undo buffer may require scanning these pages to rollback the PK index so that needs to be done first.

# Contributor agreement

### DIFF
--- a/src/transaction/transaction.cpp
+++ b/src/transaction/transaction.cpp
@@ -74,8 +74,11 @@ void Transaction::commit(storage::WAL* wal) {
 }
 
 void Transaction::rollback(storage::WAL*) {
-    localStorage->rollback();
+    // Rolling back the local storage will free + evict all optimistically-allocated pages
+    // Since the undo buffer may do some scanning (e.g. to delete inserted keys from the hash index)
+    // this must be rolled back first
     undoBuffer->rollback(clientContext);
+    localStorage->rollback();
     hasCatalogChanges = false;
 }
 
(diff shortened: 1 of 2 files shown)


benjaminwinger:
I think this should fix [this TODO](https://github.com/kuzudb/kuzu/blob/da2357c19e6065f38ebe06f73c03197cbcd33432/test/test_files/transaction/copy/copy_node.test#L5-L6), so those lines could be removed.

src/transaction/transaction.cpp
```diff
@@ -74,8 +74,11 @@ void Transaction::commit(storage::WAL* wal) {
 }
 
 void Transaction::rollback(storage::WAL*) {
-    localStorage->rollback();
+    // Rolling back the local storage will free + evict all optimistically-allocated pages
```
ray6080, line 77: I feel it would be good to remove the dependency between `undoBuffer` and `localStorage`, though not necessarily to flip the ordering here.

Would it work if we move the `clearEvictedBMEntriesIfNeeded()` logic out of `LocalStorage::rollback()`?

royi-luo: We would also need to remove the `optimisticAllocators->rollback()` since that is what is doing the evicting, although I feel that like more properly belongs in the rollback for local storage.

I think the dependency makes sense to a degree since the undo buffer is undoing operations performed on local storage. Maybe in the future if we directly implement rollback for the hash index then we wouldn't need to perform scanning when looping through the undo buffer.

ray6080 (approved):

PR #5944, opened 2025-08-28
Add configuration to disable checksumming

royi-luo:
# Description

Since checksumming has a performance cost, users may want to disable this feature. This PR adds an `enableChecksums` config to the `SystemConfig`, currently only applies to the WAL. Additionally, when loading an existing DB, add a check to see if the `enableChecksums` config matches the config used to serialize the WAL file.

Currently, the default behaviour is for `enableChecksums=true`. This PR also adds a CI workflow that runs for `enableChecksums=false`.

# Contributor agreement

### DIFF
--- a/test/transaction/wal_test.cpp
+++ b/test/transaction/wal_test.cpp
@@ -205,6 +205,22 @@ TEST_F(WalTest, CorruptedWALChecksumMismatchInBodyNoThrow) {
     EXPECT_STREQ("Binder exception: Table test does not exist.", result->getErrorMessage().c_str());
 }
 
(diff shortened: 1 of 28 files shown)


test/transaction/wal_test.cpp
```diff
@@ -3,6 +3,7 @@
 #include "api_test/api_test.h"
 #include "common/exception/runtime.h"
 #include "common/exception/storage.h"
+#include "gmock/gmock.h"
```
ray6080, line 6: Why do we need this include?

ray6080 (approved):

PR #5953, opened 2025-09-03
Implement rand related functions

### LINKED ISSUE #5951: Feature: rand() inside queries
### API

Python

### Description

a `rand()` function that can be used inside queries, similar to [rand() function for neo4j's Cypher](https://neo4j.com/docs/cypher-manual/current/functions/mathematical-numeric/#functions-rand)

Use case: one example is that it can be used to return a random set of `n` rows from a given query where one can order by `rand()` and limit the result to the first `n`.  For example, to get 10 random neighbors of node 12345, one might write a query like below.
```
MATCH (n:nodes {node_id: 12345})-[r]-(neighbor)
    RETURN n.node_id, neighbor.node_id, rand() as r
    ORDER BY r
    LIMIT 10
```

acquamarin:
Closes #5951 
Implemented `rand()` and `setseed()` functions as requested by the user.


src/function/arithmetic/rand_function.cpp
```diff
@@ -0,0 +1,26 @@
...
+
+namespace kuzu {
+namespace function {
+
+using namespace kuzu::common;
+
+struct Rand {
+    static void operation(double& result, void* dataPtr) {
+        auto context = reinterpret_cast<FunctionBindData*>(dataPtr)->clientContext;
+        result = (double)(context->getRandomEngine()->nextRandomInteger()) / (double)UINT32_MAX;
```
andyfengHKU, line 14: use static_cast inside of C style casting

andyfengHKU (approved):

src/function/arithmetic/rand_function.cpp
```diff
@@ -0,0 +1,27 @@
...
+#include "function/arithmetic/vector_arithmetic_functions.h"
+#include "function/scalar_function.h"
+#include "main/client_context.h"
+
+namespace kuzu {
+namespace function {
+
+using namespace kuzu::common;
+
+struct Rand {
```
ray6080, line 11: Would it be a good idea that `rand()` shares the same random engine with `gen_random_uuid()`? and change the seed for one will reset the other as well.

acquamarin: As we discussed, the kuzu should share one random seed

PR #5955, opened 2025-09-03
vector extension optional params + features

### LINKED ISSUE #5808: extensions(fts,vector): support IF NOT EXISTS
- [x] It'd be very convenient to add some support for `IF NOT EXISTS` logic. Something like `CALL CREATE_VECTOR_INDEX(..., skip_if_exists :- true)`.
- [x] Plus the complementary `if EXISTS` for dropping: `CALL DROP_VECTOR_INDEX(, skip_if_not_exists :- TRUE);`
- [ ] Also some form of `CALL DROP_ALL_VECTOR_INDEXES()`

carminite:
# Description
* Implements `skip_if_exists` optional param for `CREATE_VECTOR_INDEX` cypher function
* Implements `skip_if_not_exists` optional param for `DROP_VECTOR_INDEX` cypher function
* Basic test of all above in `error_suppress.test`
* ~~Implemented DROP_ALL_VECTOR_INDEXES(tableName) to erase all (and only) vector indexes from a table.~~FEATURE REMOVED
* ~~Basic test of above in `drop_all.test`~~ ditto

Fixes #5808 
#5808
#5755 (not too sure whether I should cover this as well)

# Contributor agreement

### DIFF
--- a/extension/vector/src/index/hnsw_index_utils.cpp
+++ b/extension/vector/src/index/hnsw_index_utils.cpp
@@ -12,7 +12,7 @@
 namespace kuzu {
 namespace vector_extension {
 
-inline bool HNSWIndexUtils::indexExists(const main::ClientContext& context,
+bool HNSWIndexUtils::indexExists(const main::ClientContext& context,
     const transaction::Transaction* transaction, const catalog::TableCatalogEntry* tableEntry,
     const std::string& indexName) {
     return catalog::Catalog::Get(context)->containsIndex(transaction, tableEntry->getTableID(),


sdht0 (commented):
I'll let @ray6080 do the main review. Just leaving some observations.

ray6080 (changes requested):

carminite:
New solution checks if `indexExists` twice with the same values, might be redundant. In the future, split `bindNodeTable` and move some logic to the operation bindFunc

carminite:
Above comment resolved, should be good to go

extension/vector/test/test_files/drop_all.test
```diff
@@ -0,0 +1,24 @@
+-DATASET CSV empty
```
acquamarin, line 1: ditto, remove this file

acquamarin (changes requested):

extension/vector/src/include/index/hnsw_config.h
```diff
@@ -119,11 +132,20 @@ struct HNSWIndexConfig {
 private:
     HNSWIndexConfig(const HNSWIndexConfig& other)
         : mu{other.mu}, ml{other.ml}, pu{other.pu}, metric{other.metric}, alpha{other.alpha},
-          efc{other.efc}, cacheEmbeddingsColumn(other.cacheEmbeddingsColumn) {}
+          efc{other.efc}, cacheEmbeddingsColumn(other.cacheEmbeddingsColumn),
+          skipIfExists(other.skipIfExists) {}
```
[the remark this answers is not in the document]
carminite: I want to keep the params this way because it's what's proposed in #5808 and storing params in a different form to the input is confusing.

extension/vector/src/include/function/hnsw_index_functions.h
```diff
@@ -17,16 +17,18 @@ struct CreateHNSWIndexBindData final : function::TableFuncBindData {
     catalog::TableCatalogEntry* tableEntry;
     common::property_id_t propertyID;
     HNSWIndexConfig config;
+    bool skipAfterBind;
```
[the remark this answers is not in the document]
carminite: `skipAfterBind` is a flag that switches on when the operation is invalid AND ConflictAction is DO_NOTHING. I don't know if it can be removed without it being replaced by a `tableEntry == nullptr` check

extension/vector/src/include/index/hnsw_config.h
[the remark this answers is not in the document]
carminite: On second thought, the conversion shouldn't be complicated

extension/vector/src/index/hnsw_index_utils.cpp
```diff
@@ -12,41 +12,62 @@
...
         }
+        return false;
     } break;
     case IndexOperation::DROP:
     case IndexOperation::QUERY: {
-        if (!catalog::Catalog::Get(context)->containsIndex(transaction, tableEntry->getTableID(),
-                indexName)) {
-            throw common::BinderException{common::stringFormat(
-                "Table {} doesn't have an index with name {}.", tableEntry->getName(), indexName)};
+        if (!indexExists(context, transaction, tableEntry, indexName)) {
```
acquamarin, line 43: If this is query vector index, we definitely need to throw an exception.

extension/vector/src/index/hnsw_index_utils.cpp
```diff
@@ -12,41 +12,62 @@
...
+catalog::TableCatalogEntry* HNSWIndexUtils::bindTable(const main::ClientContext& context,
+    const std::string& tableName) {
     binder::Binder::validateTableExistence(context, tableName);
     auto transaction = transaction::Transaction::Get(context);
     const auto tableEntry =
         catalog::Catalog::Get(context)->getTableCatalogEntry(transaction, tableName);
     binder::Binder::validateNodeTableType(tableEntry);
-    validateIndexExistence(context, tableEntry, indexName, indexOperation);
-    return tableEntry->ptrCast<catalog::NodeTableCatalogEntry>();
+    return tableEntry;
```
acquamarin, line 70: You can still cast tableEntry as NodeTableCatalogEntry since you have validated the table type in line 69.

extension/vector/src/index/hnsw_config.cpp
```diff
@@ -188,6 +193,21 @@ MetricType HNSWIndexConfig::getMetricType(const std::string& metricName) {
...
+DropHNSWConfig::DropHNSWConfig(const function::optional_params_t& optionalParams) {
+    for (auto& [name, value] : optionalParams) {
+        auto lowerCaseName = common::StringUtils::getLower(name);
+        if (SkipIfNotExists::NAME == lowerCaseName) {
+            value.validateType(SkipIfNotExists::TYPE);
+            conflictAction = value.getValue<bool>() ?
+                                 common::ConflictAction::ON_CONFLICT_DO_NOTHING :
+                                 common::ConflictAction::ON_CONFLICT_THROW;
+        } else {
+            throw common::BinderException{common::stringFormat(
```
acquamarin, line 205: Add a test case for this exception message.

extension/vector/src/index/hnsw_index_utils.cpp
carminite: validateIndexExistence is called in QUERY with the default ConflictAction value of THROW (since there isn't a parameter to change it), but I'll separate the cases for clarity

Issue #5943, opened 2025-08-28
Feature: Add support for viewing, altering, dropping, and exporting/importing macros.


ampexian:
### API

Python

### Description

The current implementation permits creating macros but does not permit dropping, altering, or viewing existing macros in a database. Additionally, the Export Database function does not output a macro.cypher file as the documentation indicates: https://docs.kuzudb.com/migrate/#export-database

PR #5956, opened 2025-09-03
Add the support of using "show_macros()" to view all the macros

Jasonlmx:
# Description

Created a function "show_marcos" in the table function to show all the marcos of the local/attached database 

Fixes #(5943) (first part: macro viewing)
Issue #5943 : Feature: Add support for viewing, altering, dropping, and exporting/importing macros.


src/function/table/show_macros.cpp
```diff
@@ -0,0 +1,101 @@
...
+    const TableFuncBindInput* input) {
+    std::vector<std::string> columnNames;
+    std::vector<LogicalType> columnTypes;
+    columnNames.emplace_back("name");
+    columnTypes.emplace_back(LogicalType::STRING());
+    columnNames.emplace_back("definition");
+    columnTypes.emplace_back(LogicalType::STRING());
+    std::vector<MacroInfo> macroInfos;
+    auto transaction = context->getTransaction();
+    if (!context->hasDefaultDatabase()) {
```
andyfengHKU, line 58: @acquamarin should comment this. I feel the current design will not scale if we keep adding functions that needs to check default database and attached database separately. The ideal case should be getting the current database and check its macro. 

Also, do have ways to test attached database macros?

src/function/table/show_macros.cpp
```diff
@@ -0,0 +1,101 @@
...
+        auto catalog = Catalog::Get(*context);
```
acquamarin, line 59: I don't think we allow users to add macros to remote databases.
You can always assume only local kuzu has macros.
You can get the database by `ClientContext::getDatabase()` interface.

acquamarin (approved):

PR #5957, opened 2025-09-03
Add test_list in root dir

carminite:
# Description

Solely in response to https://github.com/kuzudb/kuzu/pull/5955#discussion_r2319600626

The testing framework example outlined in https://docs.kuzudb.com/developer-guide/testing-framework/#running-a-specific-group-or-test-case has the user generate a `test_list` file in their current directory (likely the root directory). Might as well spare ~~me~~ them the grief of accidental change commits and `.gitignore` it beforehand.

# Contributor agreement

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@ bazel-*
 .bazelrc
 **/test/test_files/test_list
 extension/test_list
+test_list
 mm-256KB
 history.txt
 


.gitignore
```diff
@@ -8,6 +8,7 @@ bazel-*
 .bazelrc
 **/test/test_files/test_list
 extension/test_list
```
sdht0, line 10: We can probably remove line 9 and 10 then. Can you verify and remove them.

sdht0 (approved):
Yes wanted this as well, thanks.

.gitignore
carminite: image

Seems good

Issue #5954, opened 2025-09-03
Bug: Phantom rel left after detach delete


andreasdamm-shure:
### Kuzu version

0.11.2

### What operating system are you using?

Windows 11

### What happened?

After running the queries detailed in the steps to reproduce section using the Windows Kuzu CLI, that same CLI will return a rel that should have been deleted when querying for rels only with

```cypher
match ()-[r]->() return r;
```

Windows output

```
kuzu> match ()-[r]->() return r;
┌───────────────────────────────────────────────┐
│ r                                             │
│ REL                                           │
├───────────────────────────────────────────────┤
│ (2:1)-{_LABEL: TARGET, _ID: 5:1}->(0:1)       │
│ (1:0)-{_LABEL: REQUESTED_BY, _ID: 3:1}->(2:1) │
└───────────────────────────────────────────────┘
```

Whereas the Linux CLI opening the same file outputs

```
kuzu> match ()-[r]->() return r;
┌─────────────────────────────────────────┐
│ r                                       │
│ REL                                     │
├─────────────────────────────────────────┤
│ (2:1)-{_LABEL: TARGET, _ID: 5:1}->(0:1) │
└─────────────────────────────────────────┘
```

The node with id 0:1 has been deleted, yet the rel still exists leading to failures when writing queries the check for the existence of rels into node with id 2:1.

### Are there known steps to reproduce?

Using the Kuzu Windows CLI run the following queries

```cypher
create node table Device(id string primary key);

create node table Execution(id string primary key);  

create node table Request(id string primary key);

create rel table REQUESTED_BY(from Execution to Request);

create rel table TARGET(from Request to Device);

create (:Execution { id: 'E1' })-[:REQUESTED_BY]->(:Request { id: 'R1' })-[:TARGET]->(:Device { id: 'D1' });

match (execution:Execution { id: 'E1' })
create (execution)-[:REQUESTED_BY]->(:Request { id: 'R2' })-[:TARGET]->(:Device { id: 'D2' });

match (request:Request { id: 'R1' })
detach delete request;

match (execution:Execution { id: 'E1' })
detach delete execution;
```

The resulting database file is attached as [test-db.kz.gz](https://github.com/user-attachments/files/22122744/test-db.kz.gz).

When combining the creation queries as in

```cypher
create
(execution:Execution { id: 'E1' })-[:REQUESTED_BY]->(:Request { id: 'R1' })-[:TARGET]->(:Device { id: 'D1' }),
[...]

PR #5968, opened 2025-09-04
Fix flattening of scanned rels before detach delete

royi-luo:
# Description

When flattening scanned rels for detach delete, we were incorrectly assuming that the scanned input is unfiltered (it can be filtered due to deletes made on the scanned rel table). This PR fixes the flattening logic.

Fixes https://github.com/kuzudb/kuzu/issues/5954

# Contributor agreement


sdht0:
Why is this surfacing only on Windows?

PR #5981, opened 2025-09-08
Refactor the macro function catalog entry

acquamarin:
This PR:
1. Added a new catalog entry set: macros
2. Moved `isListLambda` and `isVarlen` to the scalarFunction body since they are only used by scalarFunctions.


andyfengHKU (approved):

src/catalog/catalog.cpp
```diff
@@ -428,13 +429,14 @@ void Catalog::dropFunction(Transaction* transaction, const std::string& name) {
 CatalogEntry* Catalog::getFunctionEntry(const Transaction* transaction, const std::string& name,
```
ray6080, line 429: Should this be renamed to `getFuncOrMacroEntry`?

PR #5982, opened 2025-09-08
api(rust): fix CI builds

sdht0:
- `cxx` and `cxx-bridge` should always be the same version. Updated to the latest releases for both, and disabled tests on clang-15 where linking fails.
- Fixes a warning from latest Rust release.


ray6080 (approved):
thanks!

benjaminwinger (approved):

PR #5984, opened 2025-09-08
Drop macros

Jasonlmx:
Add support for dropping macros.

Fixed:
Issue #5943, dropping macros

### DIFF
--- a/src/catalog/catalog.cpp
+++ b/src/catalog/catalog.cpp
@@ -496,6 +496,17 @@ ScalarMacroCatalogEntry* Catalog::getScalarMacroCatalogEntry(const Transaction* 
     return result->ptrCast<ScalarMacroCatalogEntry>();
 }
 
+ScalarMacroCatalogEntry* Catalog::getScalarMacroCatalogEntry(const Transaction* transaction,
+    kuzu::common::oid_t macroID) const {
+    auto result = functions->getEntryOfOID(transaction, macroID);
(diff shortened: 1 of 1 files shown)


src/processor/operator/ddl/drop.cpp
```diff
@@ -97,6 +100,27 @@ void Drop::dropTable(const main::ClientContext* context) {
     catalog->dropTableEntryAndIndex(transaction, dropInfo.name);
     appendMessage(stringFormat("Table {} has been dropped.", dropInfo.name), memoryManager);
 }
+void Drop::dropMacro(const main::ClientContext* context) {
+    auto catalog = Catalog::Get(*context);
+    auto transaction = transaction::Transaction::Get(*context);
+    auto memoryManager = storage::MemoryManager::Get(*context);
+    if (!catalog->containsMacro(transaction, dropInfo.name)) {
```
acquamarin, line 107: Can we refactor this to a new function?
E.g. handleMacroExistence()

acquamarin (commented):
The operator part looks good to me.
I think @ray6080 is reviewing the transaction/testing part.
You can merge after guodong approves.

src/processor/operator/ddl/drop.cpp
Jasonlmx: Should we put the function into drop.cpp directly, or should we put it somewhere else like catalog.cpp?

PR #5990, opened 2025-09-09
Rework parameter handling

andyfengHKU:
# Description

Please include a summary of the changes and the related issue (if applicable).
Please also include relevant motivation and context.

Fixes #(issue)

# Contributor agreement


test/api/api_test.cpp
```diff
@@ -316,8 +319,13 @@ TEST_F(ApiTest, MissingParam) {
...
-    ASSERT_FALSE(prep->isSuccess());
-    ASSERT_STREQ("Parameter val1 not found.", prep->getErrorMessage().c_str());
+    ASSERT_TRUE(prep->isSuccess());
+    auto result = conn->execute(prep.get(), std::make_pair(std::string("s"), 3));
+    ASSERT_FALSE(result->isSuccess());
+    ASSERT_STREQ("Parameter val2 not found.", result->getErrorMessage().c_str());
+    result = conn->execute(prep.get(), std::make_pair(std::string("val2"), 1.1));
+    ASSERT_TRUE(result->isSuccess());
+    ASSERT_STREQ("4.100000\n", result->getNext()->toString().c_str());
 }
```
royi-luo, line 329: Let's also add a test case for the case where a different param value is supplied for prepare/execute, I don't think I see one here. Something like
```
prep = conn.prepare("return $1", {"1": 1})
result = conn.query(prep, {"1": 2})
// result should be 2
```

royi-luo (approved):

PR #5991, opened 2025-09-09
Add UINT128 type

EnyuPan:
# Description

Add uint128 type and edit int128 type for conversion between the two.
Integrate the UINT128 logical type. Use to_uint128() to cast values into uint128. Using it as a primary key is supported.

# Contributor agreement


src/include/common/types/int128_t.h
```diff
@@ -59,6 +61,9 @@ struct int128_t {
     explicit operator uint8_t() const;
     explicit operator double() const;
     explicit operator float() const;
+
+    // implicit casting from int128 to uint128
+    operator uint128_t() const; // NOLINT: Allow implicit conversion from numeric values
```
royi-luo, line 66: let's keep this explicit if possible

src/include/common/types/uint128_t.h
```diff
@@ -0,0 +1,216 @@
...
+
+// comparison operators
+KUZU_API bool operator==(const uint128_t& lhs, const uint128_t& rhs);
+KUZU_API bool operator!=(const uint128_t& lhs, const uint128_t& rhs);
+KUZU_API bool operator>(const uint128_t& lhs, const uint128_t& rhs);
+KUZU_API bool operator>=(const uint128_t& lhs, const uint128_t& rhs);
+KUZU_API bool operator<(const uint128_t& lhs, const uint128_t& rhs);
+KUZU_API bool operator<=(const uint128_t& lhs, const uint128_t& rhs);
+
+class Uint128_t {
```
royi-luo, line 85: `UInt128_t` might be more correct in terms of naming convention? Not 100% sure so I'll leave this up to you

src/include/common/types/uint128_t.h
```diff
@@ -0,0 +1,216 @@
...
+public:
+    static std::string ToString(uint128_t input);
```
royi-luo, line 87: this should be `toString()`, if `int128_t` is using this naming you can also change that

src/include/common/types/uint128_t.h
```diff
@@ -0,0 +1,216 @@
...
+
+    template<class T>
+    static bool tryCast(uint128_t input, T& result);
+
+    template<class T>
+    static T Cast(uint128_t input) {
```
royi-luo, line 93:
```suggestion
    static T cast(uint128_t input) {
```

src/include/common/types/uint128_t.h
```diff
@@ -0,0 +1,216 @@
...
+    static uint128_t BinaryNot(uint128_t val);
+
+    static uint128_t divMod(uint128_t lhs, uint128_t rhs, uint128_t& remainder);
+    static uint128_t divModPositive(uint128_t lhs, uint64_t rhs, uint64_t& remainder);
+
+    static bool addInPlace(uint128_t& lhs, uint128_t rhs);
+    static bool subInPlace(uint128_t& lhs, uint128_t rhs);
+
+    // comparison operators
+    static bool Equals(uint128_t lhs, uint128_t rhs) {
```
royi-luo, line 132:
```suggestion
    static bool equals(uint128_t lhs, uint128_t rhs) {
```

src/include/common/types/uint128_t.h
```diff
@@ -0,0 +1,216 @@
...
+    }
+
+    static bool lessThan(uint128_t lhs, uint128_t rhs) {
+        return (lhs.high < rhs.high) || (lhs.high == rhs.high && lhs.low < rhs.low);
+    }
+
+    static bool lessThanOrEquals(uint128_t lhs, uint128_t rhs) {
+        return (lhs.high < rhs.high) || (lhs.high == rhs.high && lhs.low <= rhs.low);
+    }
+    static const uint128_t powerOf10[40];
```
royi-luo, line 155: Not sure if we have a test that makes sure that this is large enough, can you check if we do and add one if we don't? Also we probably only need one of these between the signed/unsigned versions (probably keep the unsigned one?) and we can cast

test_casting.cypher
```diff
@@ -0,0 +1,17 @@
+// casting floating point to int
```
royi-luo, line 1: Don't forget to delete this from your branch when you're done with it

test/common/uint128_test.cpp
```diff
@@ -0,0 +1,438 @@
...
+        uint128_t value4 = 20000000000000000000.142857f;
+        uint128_t expected4{1553255926290448384, 1};
+        int error4 = (int)(value4 > expected4 ? value4 - expected4 : expected4 - value4);
+        EXPECT_TRUE(error4 < 500000000);
+        EXPECT_THROW(
+            {
+                float negative_value = -17939777519457.96f;
+                [[maybe_unused]] uint128_t value = negative_value;
+            },
+            std::overflow_error);
```
royi-luo, line 100: Why are some tests throwing `std::overflow_error` while others are throwing `common::OverflowException`?

test/common/uint128_test.cpp
```diff
@@ -0,0 +1,438 @@
...
+    { // modulo by 0
+        uint128_t value1{INT32_MAX, 73};
+        uint128_t value2{0, 0};
+        EXPECT_THROW(value1 % value2, RuntimeException);
+    }
+}
+
+TEST(Uint128Tests, BitwiseTest1) {
+    { // small numbers
+        uint128_t value1{0b10100000, 0b01111001};
```
royi-luo, line 316: can we organize these into separate cases? Something like 

```
uint128_t value=...
uint128_t expected_value=...
EXPECT_EQ(expected_value, op(value))
```

test/common/uint128_test.cpp
```diff
@@ -0,0 +1,438 @@
...
+        EXPECT_EQ(expected_left_shift1, value1 << 55);
+        EXPECT_EQ(expected_left_shift1, Uint128_t::LeftShift(value1, 55));
+        EXPECT_EQ(expected_left_shift2, value2 << 126);
+        EXPECT_EQ(expected_left_shift2, Uint128_t::LeftShift(value2, 126));
+    }
+}
+
+TEST(Uint128Tests, BitwiseTest2) {
+    { // large numbers (with most significant bit set)
+        uint128_t value1{0, (1ULL << 63) + (1ULL << 62)};                     // 11000...000
```
royi-luo, line 345: Same here let's organize this

test/common/uint128_test.cpp
```diff
@@ -0,0 +1,438 @@
...
+        uint128_t value = int128_t((uint64_t)1, (int64_t)1);
+        EXPECT_EQ(value.low, 1);
+        EXPECT_EQ(value.high, 1);
+    }
+
+    { // conversion to int128_t
+        uint128_t value = {35, 165};
+        int128_t int128_value = value;
+        EXPECT_EQ(int128_value.low, 35);
+        EXPECT_EQ(int128_value.high, 165);
```
royi-luo, line 55: As an extra check you can do something `EXPECT_EQ(value, int128_t(value))` which will test how our `==` operator works with casting

src/common/types/uint128_t.cpp
```diff
@@ -0,0 +1,740 @@
...
+    uint128_t((uint64_t)100000000000),
+    uint128_t((uint64_t)1000000000000),
+    uint128_t((uint64_t)10000000000000),
+    uint128_t((uint64_t)100000000000000),
+    uint128_t((uint64_t)1000000000000000),
+    uint128_t((uint64_t)10000000000000000),
+    uint128_t((uint64_t)100000000000000000),
+    uint128_t((uint64_t)1000000000000000000),
+    uint128_t((uint64_t)1000000000000000000) * uint128_t((uint64_t)10),
+    uint128_t((int64_t)1000000000000000000) * uint128_t((int64_t)100),
```
royi-luo, line 37: Repeat for lines below too
```suggestion
    uint128_t((uint64_t)1000000000000000000) * uint128_t((uint64_t)100),
```

src/include/common/types/uint128_t.h
```diff
@@ -0,0 +1,225 @@
...
+template<>
+bool UInt128_t::tryCastTo(uint128_t value, uint128_t& result);
+template<>
+bool UInt128_t::tryCastTo(float value, uint128_t& result);
+template<>
+bool UInt128_t::tryCastTo(double value, uint128_t& result);
+template<>
+bool UInt128_t::tryCastTo(long double value, uint128_t& result);
+
+// TODO: const char to int128
```
royi-luo, line 217: Is this TODO still relevant?

src/include/function/aggregate/avg.h
```diff
@@ -16,11 +17,20 @@ struct AvgState : public AggregateStateWithNull {
...
+                  common::Int128_t::cast<long double>(count);
+        }
+    }
+
+    void finalize()
+        requires common::UnsignedIntegerTypes<T>
+    {
+        if (!isNull) {
+            avg = common::UInt128_t::cast<long double>(sum) /
+                  common::UInt128_t::cast<long double>(count);
```
royi-luo, line 33: You can consolidate the two functions with `std::conditional`

src/include/function/cast/functions/cast_from_string_functions.h
```diff
@@ -18,7 +18,7 @@ struct KUZU_API CastString {
 
     template<typename T>
     static inline bool tryCast(const ku_string_t& input, T& result) {
-        // try cast for signed integer types
+        // try cast for signed integer types (not including int128)
```
royi-luo, line 21: Why not including `int128`? Is it possible to consolidate the logic for the two into just a single function?

src/include/function/cast/functions/cast_string_non_nested_functions.h
```diff
@@ -125,7 +126,7 @@ struct Int128CastData {
             if (digits > DECIMAL_PRECISION_LIMIT) {
                 return false;
             }
-            if (!Int128_t::tryMultiply(result, Int128_t::powerOf10[digits], result)) {
+            if (!Int128_t::tryMultiply(result, (int128_t)UInt128_t::powerOf10[digits], result)) {
```
royi-luo, line 129: Now that I've thought about it some more it might make sense to have a power of 10 array for int128_t as well so we don't need to cast every time we do this. Let's have that power of 10 array constructed from the unsigned one though so there's only a single declaration.

src/include/function/cast/functions/cast_string_non_nested_functions.h
```diff
@@ -185,6 +186,79 @@ inline void simpleInt128Cast(const char* input, uint64_t len, int128_t& result)
...
+        digits = 0;
+        intermediate = 0;
+        return true;
+    }
+};
+
+struct UInt128CastOperation {
+    template<typename T, bool NEGATIVE>
+    static bool handleDigit(T& result, uint8_t digit) {
+        if (NEGATIVE) {
```
royi-luo, line 219:
```suggestion
        if constexpr (NEGATIVE) {
```

src/include/function/cast/functions/numeric_limits.h
```diff
@@ -24,16 +25,33 @@ struct NumericLimits {
...
-        return {std::numeric_limits<int64_t>::max(),
-            static_cast<int64_t>(std::numeric_limits<uint64_t>::max())};
+        return {std::numeric_limits<uint64_t>::max(), std::numeric_limits<int64_t>::max()};
+    }
+    template<typename V>
+    static bool isInBounds(V val) {
+        return minimum() <= val && val <= maximum();
     }
     static constexpr bool isSigned() { return true; }
     static constexpr uint64_t digits() { return 39; }
```
royi-luo, line 38: This could probably be renamed to something better like `maxNumDigits()`

src/include/function/cast/functions/numeric_limits.h
```diff
@@ -139,5 +157,51 @@ constexpr std::array<common::int128_t, NumericLimits<common::int128_t>::digits()
     }; // Couldn't find a clean way to do this
 }
 
+template<>
+constexpr std::array<common::uint128_t, NumericLimits<common::uint128_t>::digits()>
```
royi-luo, line 161: This is duplicated with the pow10 array in `uint128_t` right?

src/processor/result/base_hash_table.cpp
```diff
@@ -227,7 +227,7 @@ static compare_function_t getCompareEntryFunc(const LogicalType& type) {
     default: {
         TypeUtils::visit(
             type.getPhysicalType(), [&]<HashableTypes T>(T) { func = compareEntry<T>; },
-            [](auto) { KU_UNREACHABLE; });
+            [&](uint128_t) { func = compareEntry<uint128_t>; }, [](auto) { KU_UNREACHABLE; });
```
royi-luo, line 230: Let's add a TODO to remove this when we implement hashing, same with below

test/test_files/arithmetic/add.test
```diff
@@ -140,6 +140,26 @@ Overflow exception: INT128 is out of range: cannot add.
 ---- error
 Overflow exception: INT128 is out of range: cannot add.
 
+-CASE AddUINT128
```
royi-luo, line 143: Just double-checking we have a test case here where we overflow the lower 64 bits right?

test/test_files/arithmetic/subtract.test
```diff
@@ -140,6 +140,28 @@ Overflow exception: INT128 is out of range: cannot subtract.
 ---- error
 Overflow exception: INT128 is out of range: cannot subtract.
 
+-CASE SubtractUINT128
```
royi-luo, line 143: Double-checking do we have a case that tests for underflow of the lower 64 bits?

src/common/types/uint128_t.cpp
```diff
@@ -0,0 +1,746 @@
...
+    return NarrowCast<uint16_t>(*this);
+}
+
+uint128_t::operator uint8_t() const {
+    return NarrowCast<uint8_t>(*this);
+}
+
+uint128_t::operator double() const {
+    double result = NAN;
+    if (!UInt128_t::tryCast(*this, result)) { // LCOV_EXCL_START
```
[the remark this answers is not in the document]
EnyuPan: It's removed for casting into int128 (we do have tests for that). But currently I believe uint128 cannot fail to be cast into double, and when it is out of range for float, it simply returns `inf`. Should we keep that behavior?

src/function/vector_hash_functions.cpp
```diff
@@ -218,6 +218,7 @@ void VectorHashFunction::computeHash(const ValueVector& operand,
         [&](list_entry_t) {
             computeListVectorHash(operand, operandSelectVec, result, resultSelectVec);
         },
+        [&](uint128_t) {},
```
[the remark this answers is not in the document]
EnyuPan: Some of the tests (`e2e_test_dml_rel~delete~delete_empty.DeleteSecondNodeGroup` and `e2e_test_function~call~fsm_info.FSMReclaimRelNewTable`) would fail if we added KU_UNREACHABLE to this. It probably has something to do with hash functions being set up for certain types including UINT128.

test/test_files/arithmetic/add.test
EnyuPan: We do (same as underflow for subtraction), but I added one more test case with overflow just in case.

src/include/function/cast/functions/numeric_limits.h
EnyuPan: As you suggested, we ended up using this one wherever the pow10 array from uint128_t is used.

src/common/types/uint128_t.cpp
```diff
@@ -0,0 +1,746 @@
...
+}
+
+template<>
+bool UInt128_t::tryCastTo(float value, uint128_t& result) {
+    return tryCastTo(double(value), result);
+}
+
+template<class REAL_T>
+bool castFloatingToUint128(REAL_T value, uint128_t& result) {
+    // TODO: Maybe need to add func isFinite in value.h to see if every type is finite.
```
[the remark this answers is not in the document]
royi-luo: Let's remove it then (and maybe add a note to the int128 todo)

src/common/types/uint128_t.cpp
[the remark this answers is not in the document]
royi-luo: That behaviour is fine, although if that's the case we shouldn't have this branch. Instead we should assert that the cast succeeds.

src/include/common/types/uint128_t.h
EnyuPan: I'm not sure (this was also copied over from int128_t.h). Does it mean return the ASCII value? I don't see many uses for casting char into int128, though.

src/include/function/cast/functions/cast_from_string_functions.h
royi-luo: I think the tryCast for int128/uint128 should be specializations of this instead of being a completely different function. This makes things simpler for whatever code is using these functions.

src/include/common/types/uint128_t.h
royi-luo: I think it's a separate overload for converting const char* instead of std::string or std::string_view. We can probably remove this TODO as I don't see a need for it right now.

src/function/vector_arithmetic_functions.cpp
```diff
@@ -740,7 +740,7 @@ static std::unique_ptr<FunctionBindData> genericBinaryArithmeticFunc(
     if constexpr (FUNC::matchToOutputLogicalType) {
         common::TypeUtils::visit(
             resultingType.getPhysicalType(),
-            [&]<IntegerTypes T>(T) {
+            [&]<SignedIntegerTypes T>(T) {
                 asScalar->execFunc = ScalarFunction::BinaryStringExecFunction<T, T, T, FUNC>;
             },
             [](auto) { KU_UNREACHABLE; });
```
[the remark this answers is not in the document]
EnyuPan: I changed it from IntegerTypes to SignedIntegerTypes sometime ago, but I don't remember why. I changed it back to IntegerTypes.

src/include/function/cast/functions/cast_from_string_functions.h
EnyuPan: Right now tryCast, trySimpleInt128Cast, and trySimpleUInt128Cast are all using the tryIntegerCast template on `cast_string_non_nested_functions.h:88`, but the ones for int128 and uint128 have different "cast data structs" passed in as template arguments than the ones for simple integers. tryCast wraps around trySimpleIntegerCast with the IS_SIGNED argument set to true, and it also takes in a ku_string_t which it reinterpret-casts into `const char*`. I think things would be less confusing if we renamed tryCast to show that it is meant for simple signed integers. Or we could use trySimpleIntegerCast directly whenever tryCast is used.

src/function/vector_hash_functions.cpp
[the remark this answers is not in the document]
EnyuPan: Shouldn't we wait until hashing for uint128 is implemented? Then we could remove this empty overload because uint128_t would be one of the HashableNonNestedTypes.

royi-luo (approved):

PR #5996, opened 2025-09-09
Support SET with dictionary

carminite:
# Description

Supports the following syntax:

`MATCH (n) SET n = $data` where `$data` is a dict containing property keys and values.

NOTE that only property keys are allowed, not property expressions (like `a.*`). Nor are multiple assignments allowed (like `SET n = { ... }, m = { ... }` or `SET n.var = val, m = { ... }`)

# Contributor agreement


src/parser/transform/transform_updating_clause.cpp
```diff
@@ -44,8 +44,20 @@ std::unique_ptr<UpdatingClause> Transformer::transformMerge(CypherParser::OC_Mer
 
 std::unique_ptr<UpdatingClause> Transformer::transformSet(CypherParser::OC_SetContext& ctx) {
     auto setClause = std::make_unique<SetClause>();
-    for (auto& setItem : ctx.oC_SetItem()) {
-        setClause->addSetItem(transformSetItem(*setItem));
+    if (ctx.kU_Properties()) {
+        KU_ASSERT(ctx.kU_Properties()->oC_PropertyKeyName().size() ==
```
andyfengHKU, line 48: No need to assert. This is guaranteed by the grammar.

test/test_files/dml_node/set/set_read.test
```diff
@@ -63,3 +63,40 @@ Alice2021-06-30
 Bob1950-05-14
 Carol2000-01-01
 Dan
+
+-CASE SetMultipleReadTest
```
andyfengHKU, line 67: You probably don't need this test because it's not related to the change you made in this PR.

test/test_files/dml_node/set/set_read.test
```diff
@@ -63,3 +63,40 @@ Alice2021-06-30
...
+1|X|42
+4|X|42
+6|X|42
+
+-CASE SetByDictReadTest
+-STATEMENT CREATE NODE TABLE play(ID INT64, name STRING, score INT64, PRIMARY KEY(ID));
+---- ok
+-STATEMENT CREATE (a:play {ID: 0, name: 'AA', score: 8});
+---- ok
+-STATEMENT MATCH (a:organisation:play) SET a = { name: 'X', score: 42 } RETURN a.ID, a.name, a.score;
```
andyfengHKU, line 90: I want more testing for this feature which should include

- Set edge property.
- Data type coverage. Let's set/update all properties appears in `tinysnb` dataset `person` table. 
- Mismatched data type, e.g. try to set a int property with string value
- Incorrect variable name, e.g. `MATCH (a:organisation:play) SET x={}`
- Set with empty dictionary.
- Set when value is NULL.

Ideally these cases should have already been covered. Since you are just rewriting, this feature should automatically inherits existing implementation and no change is needed. I'm requesting these tests to ensure this.

andyfengHKU (approved):
Overall LGTM just need more testing.

PR #5999, opened 2025-09-10
Drop macros new

Jasonlmx:
Add support for dropping macros.
Add test: commit, rollback, recovery on drop macro
Fixed:
Issue https://github.com/kuzudb/kuzu/issues/5943, dropping macros


src/processor/operator/ddl/drop.cpp
```diff
@@ -97,6 +100,34 @@ void Drop::dropTable(const main::ClientContext* context) {
     catalog->dropTableEntryAndIndex(transaction, dropInfo.name);
     appendMessage(stringFormat("Table {} has been dropped.", dropInfo.name), memoryManager);
 }
```
ray6080, line 102: Add a line in between here.

Issue #5841, opened 2025-08-01
[tracking] cypher conformance


sdht0:
Noting down areas where our Cypher support does not conform with openCypher or Neo4j.

- [ ] `MATCH (n) SET n = $data` where `$data` is a dict containing property keys and values.
- [ ] `MATCH (n:LABEL1|LABEL1)`/`MATCH ()-[:LABEL1|LABEL2]->()` https://github.com/kuzudb/kuzu/issues/5807
- [ ] `list[1..3]` instead of `list[1:3]`
- [ ] list comprehension https://github.com/opencypher/openCypher/blob/main/tck/features/expressions/list/List12.feature https://neo4j.com/docs/cypher-manual/current/expressions/list-expressions/#list-comprehension

prrao87:
We should add this to the [docs](https://docs.kuzudb.com/cypher/difference/) if we don't plan to support any of them (for e.g., the first one is hard to do in Kuzu because we are strictly typed).

sdht0:
I think the first one can be done by casting `$data` to a struct and translating each `key` to SET n.key_i = value_i.

sdht0:
Status as of TCK repo commit 
```
total scenarios: 1615
passing: 959
failing: 449
skipped: 207
```

PR #6009, opened 2025-09-12
Implement PROPERTIES(NODE|REL|STRUCT)

carminite:
# Description

Implemented PROPERTIES(NODE|REL|STRUCT) for returning the (non-internal) properties of a variable.

Usage:
```cypher
RETURN PROPERTIES({name: 'alice', age: 9, nullvar: null});
{name: alice, age: 9, nullvar: }

MATCH (p:person) RETURN properties(p);
{ID: 0, fName: Alice, age: 9, ...}

MATCH (p:person)-[e:knows]->(:person) RETURN properties(e);
{date: 2021-06-30, meetTime: 1986-10-21 21:08:31.521, ...}
```

# Contributor agreement


sdht0:
Is having a special `STRUCT_PROPERTIES(x)` helpful when returning `x` has (almost) same outcome (with internal keys present). `properties()` was in listed in https://github.com/kuzudb/kuzu/issues/5841 for openCypher conformance. If we're changing the name of the function, then I'm not sure in which scenario we expect to use it as compared to simply returning the argument of the function itself.

carminite:
@sdht0 you're definitely right,  i'll probably find a way to alias/group both `properties()` functions together. i just wanted this pr open right now for code reviews because raii is hard ;-;

PR #6016, opened 2025-09-16
Optimize checkpointing updates

royi-luo:
# Description

This PR changes the checkpointing of updates so that segments are only scanned + rewritten if updates are applied to them. Previously, this happened at the column chunk level.

This is done by introducing a `ColumnChunkScanner` interface which is used by `ColumnChunk::scanCommitted()` and determines which `ChunkCheckpointStates` to write out. A `LazySegmentScanner` implementation of the interface is then used when scanning + writing updates.

For normal node groups, the `ChunkCheckpointStates` to write mirror the structure (start row, num rows) of the existing segments on disk. However, for CSR node groups we use a different strategy. Here, we skip writing to rows if we can (generally when old/new CSR offsets for an unchanged value are the same) and each new `ChunkCheckpointState` is just a contiguous block of rows that we actually write to. It's then up to `ColumnChunk::checkpoint()` to split these into segments.

### Quick microbenchmark

Configuration: AMD Zen 3 CPU, `max_num_threads=10` (although checkpoint is single-threaded so shouldn't affect execution time by much)

**Node Group Checkpoint:**

Dataset: ldbc-sf100 Comment (first 39060410 rows)

Query:
Config: `auto_checkpoint=true`, `checkpoint_threshold=0`
```sql
match (c:comment) where offset(id(c)) % pow(2, 19) = 0 set c.content = concat(c.content, 'a') return c.id, c.content order by c.id
```

branch|execution time
----|----
master|30s
new|5s

**CSR Node Group Checkpoint:**

Dataset: ldbc-sf100 Comment (first 39060410 rows) + custom rel table (10M rows):

```sql
create rel table replyOf_Comment (FROM Comment TO Comment, str STRING);
copy replyOf_Comment from (match (c:comment) where c.id <> 39582418599937 and offset(id(c)) < 10000000 return 39582418599937, c.id, c.content)
```

Query:
Config: `auto_checkpoint=true`, `checkpoint_threshold=0`
```sql
match (:comment)-[r:replyOf_Comment]->(:comment) where offset(id(r)) % pow(2, 19) = 0 set r.str = concat(r.str, 'a') return r.str order by r.str
```

branch|execution time (1st time running query)|execution time (2nd time running query)
----|----|----
master|8.6s|7.9s
new|2.7s|2.4s

~~The reason why there is no performance improvement the 1st time we run the query (which is the first time we checkpoint after the COPY) is that currently we do not calculate in-memory stats when performing COPY (at least in some cases). During the 1st checkpoint, this causes us the underestimate the sizes of string segments (not sure about other segments) which results in us under-splitting which undermines the benefits of this optimization. It looks like after the 1st checkpoint, the in-memory statistics are accurate (enough?) and thus we can see the benefits of this optimization.~~

~~Since the fix for this is unrelated to this PR and the benefits of this PR are still visible through 2+ checkpoints, I won't consider that an issue here.~~

# Contributor agreement

### DIFF
--- a/src/storage/table/csr_node_group.cpp
+++ b/src/storage/table/csr_node_group.cpp
@@ -4,7 +4,6 @@
 #include "storage/buffer_manager/memory_manager.h"
 #include "storage/storage_utils.h"
 #include "storage/table/column_chunk_data.h"
-#include "storage/table/combined_chunk_scanner.h"
 #include "storage/table/csr_chunked_node_group.h"
 #include "storage/table/lazy_segment_scanner.h"
 #include "storage/table/rel_table.h"
--- a/src/storage/table/chunked_node_group.cpp
+++ b/src/storage/table/chunked_node_group.cpp
@@ -13,7 +13,6 @@
 #include "storage/table/column_chunk.h"
 #include "storage/table/column_chunk_data.h"
 #include "storage/table/column_chunk_scanner.h"
-#include "storage/table/combined_chunk_scanner.h"
 #include "storage/table/node_table.h"
 
 using namespace kuzu::common;


src/storage/table/csr_node_group.cpp
```diff
@@ -587,62 +601,248 @@ void CSRNodeGroup::checkpointColumn(const UniqLock& lock, column_id_t columnID,
...
-        auto regionCheckpointState = checkpointColumnInRegion(lock, columnID, csrState, region);
-        if (regionCheckpointState.numRows == 0) {
-            // Skip the case when we have no rows to write for the region. This can happen when all
-            // rows are deleted within the region. We don't aggressively reclaim the space in the
-            // region, but keep deleted rows aa gaps.
-            continue;
+        auto regionCheckpointStates = checkpointColumnInRegion(lock, columnID, csrState, region);
+        // Skip the case when we have no rows to write for the region. This can happen when all
+        // rows are deleted within the region. We don't aggressively reclaim the space in the
+        // region, but keep deleted rows as gaps.
```
royi-luo, line 607: update comment

PR #6052, opened 2025-10-08
Add index hashing for uint128

EnyuPan:
# Description

Add hash index for the UINT128 type and support using it as a primary key.

# Contributor agreement


src/processor/result/base_hash_table.cpp
```diff
@@ -227,9 +227,6 @@ static compare_function_t getCompareEntryFunc(const LogicalType& type) {
     default: {
         TypeUtils::visit(
             type.getPhysicalType(), [&]<HashableTypes T>(T) { func = compareEntry<T>; },
-            [&](uint128_t) {
-                func = compareEntry<uint128_t>;
-            }, // TODO: remove when uint128_t hashing is implemented
```
royi-luo, line 232: Can you double-check if there were any other TODOs like this from your last PR?

test/test_files/uint128/uint128.test
```diff
@@ -30,6 +30,34 @@
...
+-STATEMENT CREATE (n6: tab {uint128key: 170141183460469231731687303715884105737, value: 1869}) return n6.*;
+---- error
+Runtime exception: Found duplicated primary key value 170141183460469231731687303715884105737, which violates the uniqueness constraint of the primary key column.
+-STATEMENT CREATE (n7: tab {uint128key: 2025, value: NULL}) return n7.*;
+---- 1
+2025|
+-STATEMENT CREATE (n8: tab {uint128key: NULL, value: 1}) return n8.*;
+---- error
+Runtime exception: Found NULL, which violates the non-null constraint of the primary key column.
+
```
royi-luo, line 60: Can you add some more statements to check that matching by primary key works?

royi-luo (approved):

src/processor/result/base_hash_table.cpp
EnyuPan: I took a look a didn't find any left.

EnyuPan: (Except for the comment `TODO: Maybe need to add func isFinite in value.h to see if every type is finite.` in int128_t.cpp, which was left over from before. There isn't a line like this in uint128_t.cpp.)

